PR #405, opened 2018-03-17
Fixing AttributeError for Kanban

safwanrahman:
While loading Kanban with fresh install without any plugin, and load the Kanban Page, it raises Error.
Finding out more, it seems like `AttributeError` is not cached.


ajrbyers:
Before this is merged, can you please post the entire stack of this error so I can work out what is causing it rather than just catching this error.

safwanrahman:
@ajrbyers `WORKFLOW_PLUGINS` is not available always if you do not have any plugin installed.

PR #419, opened 2018-04-05
Feature: nicer config

lsh-0:
* single definitive settings.py file instead of per-environment and example settings.py files
* settings that actually vary between environment and users live in app.cfg
* app.cfg is created from a template `example.cfg` file if it doesn't exist upon installation
* installation happens with `install.sh`, which assumes host environment has any necessary non-python dependencies installed


ajrbyers:
@lsh-0 what will happen if someone who already has a settings.py file pulls this update?

install.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+set -e # everything must succeed.
+echo "[-] install.sh"
+
+python=/usr/bin/python3.6
```
ajrbyers: Does this assume python 3.6 is installed?

src/core/settings.py
```diff
@@ -240,61 +251,70 @@ def ugettext(s):
...
 SILENCED_SYSTEM_CHECKS = (
     'fields.W340',
 )
 
-'''
-# This section should only be enabled if you intend to use Sentry for error reporting.
 LOGGING = {
     'version': 1,
     'disable_existing_loggers': True,
+    # TODO: do we really need to be affecting the root logger here?
```
ajrbyers: The Sentry documentation suggest, yes. https://docs.sentry.io/clients/python/integrations/django/

install.sh
lsh-0: yup. I've opened another PR with a test script and travis integration but your test suite has a few failing cases. Travis will test against 3.4, 3.5 and 3.6, which roughly correspond to the LTS Ubuntu releases since 12.04 to 18.04 - I think that's reasonable.

To make the python version here more flexible we can just do a `$(which python3)` and then issue a warning if the `x` in `python --version` (3.x.y) is not supported.

I didn't know if you target a specific Python3 or not

src/core/settings.py
lsh-0: yeah, dead comment. will remove

lsh-0:
> what will happen if someone who already has a settings.py file pulls this update?

pretty sure the pull will fail with an error like: `error: Untracked working tree file 'src/core/settings.py' would be overwritten by merge.`

this *is* a large change, it would affect deployment and require migrating the various `settings.py` files around to app.cfg

lsh-0:
typically breaking changes get a major version bump (1.0.0 -> 2.0.0)

I'm not sure how you're managing versioning at the moment as I can see the `master` branch is behind the last several releases. Do you have a develop or integration branch?

lsh-0:
better yet, we can dogfood this so that if the doc is updated I'll know how to best open a PR:
https://github.com/BirkbeckCTP/janeway#contributions

ajrbyers:
There are three main branches at the moment:

- Master
- 1.2.4 (Stable Branch)
- 1.3-working (Development of 1.3)

I suppose we could look to integrate this change into the 1.3 release, there are some already large changes in the way the Workflow system works.

install.sh
ajrbyers: 3.5+ is required, post merge I can look at integrating this install script with the Django installation script we have so that it can do the full install. We would also need to update the base requirements to include venv and dropp virtualenv/virtualenvwrapper.

lsh-0: yes, I have an unpushed change around here for that ...

PR #420, opened 2018-04-08
adds migrations for two database fields dyanmicsites app generates

lsh-0:
removes them from the auto-loading fixture src/core/fixtures/dev/dummy_sites.sql


ajrbyers:
I've merged this, though it is covered in the [install_janeway](https://github.com/BirkbeckCTP/janeway/blob/master/src/utils/management/commands/install_janeway.py) management command when the following is called 

```
call_command('makemigrations', 'sites')
```

lsh-0:
good to know about that command. I'm approaching this from the POV of another dev evaluating how quickly I can get up and running with this project, code quality, features, etc.

One thing that gave me a slight pause was once I finally had all the bits working, it took me to example.org. I'm guessing this is default behaviour in the absence of a site, but a friendlier option when `debug=True` might be a "congratulations! the next step is to install a journal" type page.

PR #596, opened 2018-11-12
Added social links to the Editorial team for Material.

ajrbyers:


src/themes/material/templates/elements/journal/editorial_social_content.html
```diff
@@ -0,0 +1,5 @@
+{% if user.website %}<a href="{{ user.website }}"><i class="fa fa-globe"></i></a>{% endif %}
+{% if user.twitter %}<a href="https://www.twitter.com/{{ user.twitter }}"><i class="fa fa-twitter-square"></i></a>{% endif %}
+{% if user.facebook %}<a href="https://www.facebook.com/{{ user.facebook }}"><i class="fa fa-facebook-square"></i></a>{% endif %}
+{% if user.github %}<a href="https://www.github.com/{{ user.github}}"><i class="fa fa-github-square"></i></a>{% endif %}
+{% if user.linkedin %}<a href="https://www.linkedin.com/in/{{ user.linkedin }}"><i class="fa fa-linkedin-square"></i></a>{% endif %}
```
mauromsl: We could add properties to the Account model to generate the urls for us, as I imagine we will need to use these often

PR #634, opened 2018-11-19
Feature dev settings improvements

mauromsl:


src/core/example_settings.py
```diff
@@ -395,3 +395,26 @@ def filter(self, record):
...
 SILENT_IMPORT_CACHE = False
+
+# Development Overrides
+if DEBUG:
+    EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
+    for t in TEMPLATES:
+        t["OPTIONS"]["string_if_invalid"] = "Invalid variable: %s!!"
+
+# Testing Overrides
+if IN_TEST_RUNNER:
```
ajrbyers: Can you clarify what this does? If running `manage.py test` skip migrations?

mauromsl: Yes, Having to wait for all the migrations on every test run is quite a pain

ajrbyers:
You can use —keepdb to maintain a db across multiple test runs.

mauromsl:
Good point, handled that case

PR #656, opened 2018-11-23
Adds logger config for development

mauromsl:


src/core/example_settings.py
```diff
@@ -95,6 +95,7 @@
 INSTALLED_APPS += plugin_installed_apps.load_homepage_element_apps()
 
 MIDDLEWARE_CLASSES = (
+    'utils.middleware.TimeMonitoring',
```
ajrbyers: I don't think this should be enabled by default... example settings could be used in production and this seems like a development thing.

mauromsl: Done

PR #666, opened 2018-11-26
#665: Stop sending email notifications to users with dummy email adress

### LINKED ISSUE #665: Don't send notification emails to users with dummy emails
During imports, we create users with a dummy email address if one is not provided.
The email notifier should ignore those users.

mauromsl:
closes #665

### DIFF
--- a/src/utils/notify_plugins/notify_email.py
+++ b/src/utils/notify_plugins/notify_email.py
@@ -24,7 +24,7 @@ def send_email(subject, to, html, journal, request, bcc=None, cc=None, attachmen
         else:
             to = [to]
     elif isinstance(to, Iterable)
-        to = [email for email in to if not DUMMY_EMAIL_DOMAIN in to]
+        to = [email for email in to if not DUMMY_EMAIL_DOMAIN in email]
 
     if request and request.user and not request.user.is_anonymous():
         reply_to = [request.user.email]
--- a/src/utils/notify_plugins/notify_email.py
+++ b/src/utils/notify_plugins/notify_email.py
@@ -1,9 +1,12 @@
+from collections import Iterable
+
 from django.core.mail import EmailMultiAlternatives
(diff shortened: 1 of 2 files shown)


src/utils/notify_plugins/notify_email.py
```diff
@@ -1,9 +1,12 @@
+from collections import Iterable
+
 from django.core.mail import EmailMultiAlternatives
 from django.utils.html import strip_tags
 
 from utils import setting_handler
 from utils import notify
 
+DUMMY_EMAIL_DOMAIN = "@journal.com"
```
ajrbyers: Do you think we should set this in settings.py so it can be altered?

src/utils/notify_plugins/notify_email.py
```diff
@@ -15,8 +18,13 @@ def send_email(subject, to, html, journal, request, bcc=None, cc=None, attachmen
     else:
         from_email = request.press.main_contact
 
-    if not type(to) in [list, tuple]:
-        to = [to]
+    if isinstance(to, str):
+        if DUMMY_EMAIL_DOMAIN in to:
+            to = []
```
ajrbyers: I suspect Mailgun will give a 400 if `to` is empty.

src/utils/notify_plugins/notify_email.py
mauromsl: Alright

src/utils/notify_plugins/notify_email.py
mauromsl: I'll handle that  case then

mauromsl: Actually, seems like Django has thought of this for us :)
https://github.com/django/django/blob/58f02c498b659a906e9c30d946bd89bedc4717e5/django/core/mail/message.py#L336

PR #677, opened 2018-11-28
Closes #676, adds a signal to create a congifuration for new journals…

### LINKED ISSUE #676: New Journal Error
RelatedObjectDoesNotExist at /submit/start/
Journal has no submissionconfiguration.

ajrbyers:
… and handles those that are missing by creating defaults


src/submission/forms.py
```diff
@@ -33,6 +33,11 @@ def __init__(self, *args, **kwargs):
         self.fields['competing_interests'].label = ''
         self.fields['comments_editor'].label = ''
 
+        try:
+            journal.submissionconfiguration
+        except models.SubmissionConfiguration.DoesNotExist:
+            models.SubmissionConfiguration.objects.create(journal=journal)
+
```
mauromsl: When do we expect this to happen?

ajrbyers: It happened to me yesterday. Basically when this model was introduced a Signal should have been added to generate them automatically but it wasn't so there may be journals out there without a configuration.

mauromsl: But this code will live here forever, shouldn't we have it on some sort of upgrade script/migration instead?

ajrbyers:
Reverted forms and views code and added a migration to ensure all journals have one. Retained the if created signal.

Issue #671, opened 2018-11-27
User doesn't get any feedback when registering DOIs via Crossref


mauromsl:
When clicking on the "Issue DOI" button, the `identifiers.logic.register_crossref_component` is called and a number of Log entries will be recorded.

The user, however, doesn't receive any feedback from the `submission.views.edit_identifier` view

PR #679, opened 2018-11-29
#671 Adds feedback for the Issue DOI button.

ajrbyers:


ajrbyers:
Okay, can you check this over @mauromsl

src/journal/views.py
```diff
@@ -706,7 +706,8 @@ def publish_article(request, article_id):
             # Attempt to register xref DOI
             for identifier in article.identifier_set.all():
                 if identifier.id_type == 'doi':
-                    identifier.register()
+                    status = identifier.register()
+                    messages.add_message(request, messages.INFO, status)
```
mauromsl: I suppose this should really be messages.ERROR when it fails..

PR #681, opened 2018-11-29
#672 view pdf galley feature

### LINKED ISSUE #672: Allow PDF galleys to be opened in browser
At the moment, we server the galleys via request, we could also offer them statically so that they can be opened in the browser without the user needing to download it and then open it.

This was proposed by Megan:

> I have the following query from a journal editor: when the ‘download’ button is clicked – can the pdf be viewed in browser in a new window instead of asking for a download location? This helps with reading materials over smart devices and computers where someone is not intending to download a copy but just view it only?

ajrbyers:
Closes #672


src/core/files.py
```diff
@@ -358,6 +358,33 @@ def serve_file_to_browser(file_path, file_to_serve, public=False):
...
+        settings.BASE_DIR,
+        'files',
+        'articles',
+        str(article.id),
+        str(file.uuid_filename)
+    )
+
+    try:
+        response = HttpResponse(
+            FileWrapper(open(file_path, 'rb'), 8192),
```
mauromsl: Why hardcode he buffer size to 8192 here? Its the default behaviour of `FileWrapper`. if anythong we should ask the os via `os.stat()` or use the default `io.DEFAULT_BUFFER_SIZE` (which is what we are doing on the `open(file_path, 'rb')` call)

PR #691, opened 2018-12-06
#582: two-layered settings

### LINKED ISSUE #582: Improve settings handling
When new a new feature requires a setting to be added to the django settings.py, there should be a way to include a default case so that users are not required to immediately change their own settings files.

Janeway needs a way of supporting default settings, similar to how django's django.conf.global_settings behave. https://docs.djangoproject.com/en/1.11/topics/settings/#using-settings-without-setting-django-settings-module

mauromsl:
Closes #582

Summary:
 - Current example settings moved to `janeway_global_settings.py`
 - `example_settings.py` now contains those settings that MUST be set by the user
 - janeway will load settings from `core.settings.py` and override values from `janeway_global_settings.py`
 - Due to the above, users won't need to make any changes to their current settings.
 - The custom settings module can be overriden by setting the env variable `JANEWAY_SETTINGS_MODULE` (similar to how django uses `DJANGO_SETTINGS_MODULE`)


src/core/example_settings.py
```diff
@@ -1,464 +1,24 @@
...
 # You should change this key before you go live!
 SECRET_KEY = 'uxprsdhk^gzd-r=_287byolxn)$k6tsd8_cepl^s^tms2w1qrv'
-
-# SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = True
-COMMAND = sys.argv[1:]
-IN_TEST_RUNNER = COMMAND[:1] == ['test']
-ALLOWED_HOSTS = ['*']
-
-ENABLE_TEXTURE = False
```
ajrbyers: This should be a non-global setting.

src/core/example_settings.py
```diff
@@ -1,464 +1,24 @@
...
+        'HOST': '',
+        'PORT': '',
+        'OPTIONS': {'init_command': 'SET default_storage_engine=INNODB'},
     }
 }
 
-# Internationalization
-# https://docs.djangoproject.com/en/1.8/topics/i18n/
-
-LANGUAGE_CODE = 'en'
```
ajrbyers: Should we move this one over as well?

src/core/example_settings.py
```diff
@@ -1,464 +1,24 @@
...
-URL_CONFIG = 'domain'  # path or domain
-
-# Captcha
-# You can get reCaptcha keys for your domain here: https://developers.google.com/recaptcha/intro
-# You can set either to use Google's reCaptcha or a basic math field with no external requirements
-
-INSTALLED_APPS.append('snowpenguin.django.recaptcha2')
-
-CAPTCHA_TYPE = 'recaptcha'  # should be either simple_math or recaptcha to enable captcha fields otherwise disabled
-RECAPTCHA_PRIVATE_KEY = ''
```
ajrbyers: Probably a good idea to have the captcha settings over in example as they need configuring.

src/core/example_settings.py
```diff
@@ -1,464 +1,24 @@
...
-
-PLUGIN_HOOKS = {}
-
-NOTIFY_FUNCS = []
-
-ENABLE_ORCID = True
-ORCID_API_URL = 'http://pub.orcid.org/v1.2_rc7/'
-ORCID_URL = 'https://orcid.org/oauth/authorize'
-ORCID_TOKEN_URL = 'https://pub.orcid.org/oauth/token'
-ORCID_CLIENT_SECRET = ''
```
ajrbyers: ORCID settings need configuring so should probably be in example.

src/core/example_settings.py
mauromsl: I don't see this used anywhere, what is its purpose?

ajrbyers: Its used in settings.py and will be used to enable a feature that is half finished... maybe leave it for now.

PR #702, opened 2018-12-11
Added test and tweaked migration.

### LINKED ISSUE #700: journal has no submissionconfiguration
navigating to `/submit/manager/configurator/` throws an error that the journal has no `submissionconfiguration`

ajrbyers:
- Tweaked the migration to make it reversible
- Added a test for signals
- Closes #700

### DIFF
--- a/src/submission/tests.py
+++ b/src/submission/tests.py
@@ -1,0 +1,38 @@
+__copyright__ = "Copyright 2017 Birkbeck, University of London"
+__author__ = "Martin Paul Eve & Andy Byers"
+__license__ = "AGPL v3"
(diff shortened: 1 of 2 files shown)


src/submission/tests.py
```diff
@@ -0,0 +1,38 @@
...
+from django.test import TestCase
+
+from journal import models as journal_models
+
+
+# Create your tests here.
+class SubmissionTests(TestCase):
+
+    def test_new_journals_has_submission_configuration(self):
+        if not self.journal_one.submissionconfiguration:
```
mauromsl: Could check `jorunal_two` as well, otherwise why would we create it?

PR #703, opened 2018-12-11
Added handler for odd edge case. Made field read only to avoid in future

### LINKED ISSUE #653: TypeError: invalid file: None
It looks like `self_article_path()` can return `None`:
https://github.com/BirkbeckCTP/janeway/blob/c80dba661a50bf3220fa63b5e2eaad2c6678821d/src/core/models.py#L628-L630
Which triggers a TypeError being raised here:
https://github.com/BirkbeckCTP/janeway/blob/c80dba661a50bf3220fa63b5e2eaad2c6678821d/src/core/models.py#L659

ajrbyers:
- Returns an error where the article_id is missing but doesn't break the page
- article_pk is now a read-only field in admin to stop this happening in future
- Closes #653

### DIFF
--- a/src/core/admin.py
+++ b/src/core/admin.py
@@ -85,6 +85,7 @@ class FileAdmin(admin.ModelAdmin):
     list_filter = ('mime_type', 'article_id')
     raw_id_fields = ('owner',)
     filter_horizontal = ('history',)
+    readonly_fields = ['article_id']
 
     def article_pk(self, obj):
         if obj.article:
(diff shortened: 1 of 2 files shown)


ajrbyers:
@mauromsl yes, there are some cases where core.File is used for non-article objects, News and Logos etc. I think this is a decent enough solution for the time being.

mauromsl:
> @mauromsl yes, there are some cases where core.File is used for non-article objects, News and Logos etc. I think this is a decent enough solution for the time being.

But the fix is for galley files on the article page, which should have always an `article_id` right?

PR #708, opened 2018-12-12
654 UUID regex

### LINKED ISSUE #654: ValidationError/profile/{uuid}/
The profile/{uuid} url regex its too flexible and it is letting invalid UUIDs through which raise an exception in the view:
https://github.com/BirkbeckCTP/janeway/blob/c80dba661a50bf3220fa63b5e2eaad2c6678821d/src/core/include_urls.py#L180-L181

`["'291' is not a valid UUID."]`

ajrbyers:
Closes #654

### DIFF
--- a/src/core/include_urls.py
+++ b/src/core/include_urls.py
@@ -178,7 +178,7 @@ urlpatterns = [
         name='edit_identifiers_with_event'),
 
     # Public Profiles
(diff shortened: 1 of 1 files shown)


src/utils/middleware.py
```diff
@@ -23,13 +23,16 @@ def process_response(self, request, response):
             del _local.request
         return response
 
+
 class TimeMonitoring(object):
     """Monitors the resource usage of a request/response cycle """
+
```
mauromsl: Why this newline?

PR #711, opened 2018-12-14
Sidebars now loop over all header tags in the article.

### LINKED ISSUE #687: Pickup All H tags for the sidebar

ajrbyers:
Closes #687

### DIFF
--- a/src/themes/OLH/assets/js/table_of_contents.js
+++ b/src/themes/OLH/assets/js/table_of_contents.js
@@ -9,7 +9,7 @@ $( document ).ready(function() {
     iter = 0;
     toc = '';
 
-    $("article h2").each(function () {
+    $("#main_article :header").each(function () {
 
         link = $(this);
         title = link.html();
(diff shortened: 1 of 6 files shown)


src/themes/OLH/assets/js/table_of_contents.js
```diff
@@ -9,7 +9,7 @@ $( document ).ready(function() {
     iter = 0;
     toc = '';
 
-    $("article h2").each(function () {
+    $("#main_article :header").each(function () {
```
mauromsl: Why do we need to target by ID in this tag and by `article` in material?

ajrbyers: Because there are other <hx> inside <article> on the OLH theme that aren't in the Material one that we don't want in the sidebar.

mauromsl: Perhaps we should get rid of the parent `<article>` tag then? https://github.com/BirkbeckCTP/janeway/pull/711/files#diff-fe4cba43336c6287b7612e37085e1406L68

Or alternatively, use the `#main_article` id on all themes?

PR #720, opened 2018-12-20
Added handler for empty author searches.

ajrbyers:
This function is too long, need to move to forms/logic but a fix is in place for the empty search issue.

https://github.com/BirkbeckCTP/janeway/issues/721


ajrbyers:
@mauromsl yes, done!

PR #724, opened 2018-12-20
723 bugfix author records

### LINKED ISSUE #723: Author Records
- [x] Add a signal to delete an author-article record when a frozen author is deleted
- [x] Update submission.views.delete_author to delete also the ArticleAuthorOrder record
- [x] Move the point of freezing authors to submission

ajrbyers:
Closes #723

### DIFF
--- a/src/submission/models.py
+++ b/src/submission/models.py
@@ -1255,4 +1255,12 @@ def remove_author_from_article(sender, instance, **kwargs):
     :param instance: FrozenAuthor instance
     :return: None
     """
+    try:
+        ArticleAuthorOrder.objects.get(
+            author=instance.author,
+            article=instance.article,
+        ).delete()
+    except ArticleAuthorOrder.DoesNotExist:
+        pass
+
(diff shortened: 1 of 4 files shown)


src/submission/models.py
```diff
@@ -1240,3 +1242,17 @@ def handle_defaults(self, article):
...
+@receiver(pre_delete, sender=FrozenAuthor)
+def remove_author_from_article(sender, instance, **kwargs):
+    """
+    This signal will remove an author from a paper if the user deletes the
+    frozen author record to ensure they are in sync.
+    :param sender: FrozenAuthor class
+    :param instance: FrozenAuthor instance
+    :return: None
+    """
+    instance.article.authors.remove(instance.author)
```
mauromsl: do we not need to remove the ArticleAuthorOrder record here as well? (shame that there is no signal for removing m2m)

ajrbyers: We do, commit added.

Issue #728, opened 2018-12-21
Log attempts to acess a resource over which a user has no permission


mauromsl:
Attempts to access a resource behind a view for which a user doesn't have permission should be logged

ajrbyers:
Adding this to Trello, please ignore.

PR #729, opened 2018-12-21
#728: Log attempts to access a view without correct permission

mauromsl:


ajrbyers:
Adding this to Trello, please ignore.

PR #753, opened 2019-01-14
747 migration fix

ajrbyers:


src/core/migrations/0019_set_default_news_items.py
```diff
@@ -1,34 +1,43 @@
...
+            )
+
+            plugin_setting_value, c = PluginSettingValue.objects.get_or_create(
+                setting=plugin_setting,
+                journal=journal
+            )
+
+            SQL = """
+            UPDATE utils_pluginsettingvalue_translation SET
+            value = 5 WHERE master_id = {master_id} AND value IS NULL;
```
mauromsl: This is not handling the case where the value is set to "" or " " which I've come across on the live db.

src/core/migrations/0019_set_default_news_items.py
```diff
@@ -1,34 +1,43 @@
...
-                        value=5,
-                        journal=journal,
-                )
+
+            plugin_setting, c = PluginSetting.objects.get_or_create(
+                plugin=plugin,
+                name="number_of_articles"
+            )
+
+            plugin_setting_value, c = PluginSettingValue.objects.get_or_create(
```
mauromsl: do we need to create the `utils_pluginsettingvalue_translation` rows when `c` is `True`? or does the library do that automagically?

PR #775, opened 2019-01-21
Changed default review visibility to double-blind

ajrbyers:


ajrbyers:
> Do we need a migration to handle current installations?

We don't want a migration to over-write the other choices, I think this will be okay as it its.

PR #798, opened 2019-01-25
Adds Dev settings and functionality to merge settings from 2 sources

### LINKED ISSUE #787: Remove DEBUG options from janeway_global_settings.py
#582 added support for having 2 layers of settings (global and local) but some DEBUG specific settings were left in the `src/core/janeway_global_settings.py`. Whilst this is not a problem for current installations, new installations would need to explicitly override this debug options on their custom local `settings.py`

Any debug specific settings should be moved to a dev_settings.py file that the docker container can then choose as the default.

mauromsl:
Closes #787

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -48,7 +48,7 @@ services:
       - DB_USER
       - DB_NAME
       - PYTHONDONTWRITEBYTECODE=yes
-      - JANEWAY_SETTINGS_MODULE=core.janeway_global_settings
+      - JANEWAY_SETTINGS_MODULE=core.dev_settings
     depends_on:
       - "start_dependencies"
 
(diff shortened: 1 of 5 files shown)


ajrbyers:
@mauromsl I think you may need to talk me through this PR before I approve it.

mauromsl:
The problem was that the global settings were making a check on the value of `settings.DEBUG` before evaluating the user-defined `settings.py` (It was always evaluated as `DEBUG = True`). Instead, I've moved those development options to a `dev_settings.py` file which will be loaded by default with `docker-compose` commands.

Additionally, cbc62e4 is a commit I've borrowed from my Jenkins PR that allows for merging some setting values across the two settings files (`janeway_global_settings.py` and `settings.py`) which enables the `settings.py` file to expand a number of whitelisted settings (at this point `MIDDLEWARE_CLASSES` and `INSTALLED_APPS`)

Issue #795, opened 2019-01-25
1.3.2 Bugs


ajrbyers:
The following bugs were identified during a 1.3.2 test run, and should be fixed:

- The Submitted Articles list should show all articles a user is an author for
- The Author Copyedit Decision field should be required
- Prodcution Dashboard numbers aren't filtered by user
- Authors don't see Author Proofing requests on the dashboard
- Proofing Task page has a typo "compelte" should read "complete"
- On the Prepub page the title for setting the Pub Date is incorrect
- MathCaptcha should be the default captcha type

PR #799, opened 2019-01-28
795 bugfixes

ajrbyers:
Trello card: https://trello.com/c/W7ju84Xk/494-795-132-bugs#
Issue: #795

### DIFF
--- a/src/copyediting/forms.py
+++ b/src/copyediting/forms.py
@@ -27,7 +27,5 @@ class AuthorCopyeditForm(forms.ModelForm):
         fields = ('decision', 'author_note')
 
     def __init__(self, *args, **kwargs):
-        # first call parent's constructor
         super(AuthorCopyeditForm, self).__init__(*args, **kwargs)
-        # there's a `fields` property now
         self.fields['decision'].required = True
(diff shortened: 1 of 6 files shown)


ajrbyers:
[#795: 1.3.2 Bugs](https://trello.com/c/W7ju84Xk/494-795-132-bugs)

src/copyediting/forms.py
```diff
@@ -25,3 +25,9 @@ class AuthorCopyeditForm(forms.ModelForm):
     class Meta:
         model = models.AuthorReview
         fields = ('decision', 'author_note')
+
+    def __init__(self, *args, **kwargs):
+        # first call parent's constructor
```
ajrbyers: Comments have been dropped in future commit fyi.

src/core/views.py
```diff
@@ -866,7 +866,7 @@ def add_user(request):
     form = forms.EditAccountForm()
     registration_form = forms.AdminUserForm(active='add', request=request)
     return_url = request.GET.get('return', None)
-    role = request.GET.get('role', None)
```
mauromsl: We are only assigning the role if another role is not defined in `request.GET["role"]`. This seems to contradict the commit title. Is this what we intended?

ajrbyers: Indeed! Commit incoming.

PR #808, opened 2019-01-29
require last_name and first_name in Account model

hachacha:
require first name and last name to be set in account model.


mauromsl:
Thanks @hachacha

Issue #819, opened 2019-01-30
Manage Galleys in Proofing Stage deny_access


hachacha:
Manage galleys button is available in the proofing stage and directs to `url/production/article_id/`

![image](https://user-images.githubusercontent.com/1020384/52002596-7185de80-2490-11e9-9b3d-60fae867e704.png)

that page has a decorator: `@article_stage_production_required` which requires that the article is on typesetting stage resulting in `deny_access`

ajrbyers:
Yup I think we can drop this decorator we only care that the user is a pm or editor.

PR #820, opened 2019-01-31
Removed production stage decorator as it is no longer required. Wrapp…

ajrbyers:
…ed link in check for production_manager role.

Fix for #819


ajrbyers:
[#819: Manage Galleys in Proofing Stage deny_access](https://trello.com/c/EIAvjeB8/514-819-manage-galleys-in-proofing-stage-denyaccess)

PR #831, opened 2019-02-01
Closes #457, adds option to send pre and post acceptance reminder ema…

### LINKED ISSUE #457: Resend Review Emails
- Option to send a manual reminder email
- If a review has not been accepted, add an option to gently remind the user of the request making the option to decline obvious as it’s better the editor know the person will not review the paper than leave the request open.

ajrbyers:
…ils manually

### DIFF
--- a/src/review/views.py
+++ b/src/review/views.py
@@ -19,9 +19,11 @@ from django.conf import settings
 from core import models as core_models, files, forms as core_forms
 from events import logic as event_logic
 from review import models, logic, forms, hypothesis
-from security.decorators import editor_user_required, reviewer_user_required, reviewer_user_for_assignment_required, \
-    article_edit_user_required, file_user_required, article_decision_not_made, article_author_required, \
-    editor_is_not_author, senior_editor_user_required, section_editor_draft_decisions
+from security.decorators import editor_user_required, reviewer_user_required, \
+    reviewer_user_for_assignment_required,  article_edit_user_required, \
+    file_user_required, article_decision_not_made, article_author_required, \
+    editor_is_not_author, senior_editor_user_required, \
+    section_editor_draft_decisions, article_stage_review_required
 from submission import models as submission_models
 from utils import models as util_models, ithenticate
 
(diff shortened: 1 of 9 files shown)


ajrbyers:
This is not ready for review, I need to add a test for the new decorator first.

src/review/views.py
```diff
@@ -19,9 +19,11 @@
...
 from events import logic as event_logic
 from review import models, logic, forms, hypothesis
-from security.decorators import editor_user_required, reviewer_user_required, reviewer_user_for_assignment_required, \
-    article_edit_user_required, file_user_required, article_decision_not_made, article_author_required, \
-    editor_is_not_author, senior_editor_user_required, section_editor_draft_decisions
+from security.decorators import editor_user_required, reviewer_user_required, \
+    reviewer_user_for_assignment_required,  article_edit_user_required, \
+    file_user_required, article_decision_not_made, article_author_required, \
+    editor_is_not_author, senior_editor_user_required, \
+    section_editor_draft_decisions, article_stage_review_required
```
mauromsl: I can't see where "article_stage_review_required" lives
```
$ git grep "def article_stage_"
src/security/decorators.py:def article_stage_production_required(func):
src/security/decorators.py:def article_stage_accepted_or_later_required(func):
src/security/decorators.py:def article_stage_accepted_or_later_or_staff_required(func):
```

I suspect this is a new decorator that you forgot commit :P

ajrbyers: https://github.com/BirkbeckCTP/janeway/pull/831/files

I can see it in the commit...

mauromsl: Where?

mauromsl: Ah, GitHub decided to no show me the diff by default

src/security/decorators.py
```diff
@@ -931,6 +1174,27 @@ def wrapper(request, *args, **kwargs):
...
+def article_stage_review_required(func):
+    """
+    Checks that the article is in one of the review stages
+    :param func: func
+    :return: PermissionDenied or func
+    """
+    def review_required_wrapper(request, *args, **kwargs):
+        article = get_object_or_404(
+            models.Article,
+            pk=kwargs['article_id']
```
mauromsl: Could throw a KeyError.

src/security/decorators.py
```diff
@@ -697,13 +866,22 @@ def wrapper(request, *args, **kwargs):
...
             return func(request, *args, **kwargs)
 
-        typeset_assignment = get_object_or_404(production_models.TypesetTask, pk=typeset_id)
-        if request.user == typeset_assignment.typesetter and not typeset_assignment.completed and \
+        typeset_assignment = get_object_or_404(
+            production_models.TypesetTask,
+            pk=typeset_id
+        )
+        if request.user == typeset_assignment.typesetter and\
+                not typeset_assignment.completed and \
```
mauromsl: Backslashes are evil and operator should be at the beginning of the line, is autopep8 doing this?

I think we should leave this kind of task for its own issue instead.

src/security/decorators.py
ajrbyers: it would 404 before a key error.

mauromsl: Definately not, `__getitem__` will be invoked on kwargs before `get_object_or_404` is called

ajrbyers: handler and test added

src/security/decorators.py
ajrbyers: Dropped autopep8 crap from this module.

ajrbyers:
Requested changes done.

src/security/decorators.py
```diff
@@ -952,3 +953,29 @@ def deny_access(request, *args, **kwargs):
...
+
+def article_stage_review_required(func):
+    """
+    Checks that the article is in one of the review stages
+    :param func: func
+    :return: PermissionDenied or func
+    """
+
+    def review_required_wrapper(request, *args, **kwargs):
+        article_id = kwargs.get('article_id', None)
```
mauromsl: This means that the article_id MUST be a kwarg.
If that is the case, python3 introduced a cool feature to express this in the function signature:

```
def review_required_wrapper(request, *args, article_id=None,**kwargs):
     if article_id is None:
          raise Http404
     [...]
```

ajrbyers: Further changes made.

PR #838, opened 2019-02-04
add OLH and material themes templates for author list 

hachacha:
I was missing these in my initial PR. remedying that now with styles that match the editorial_team.html for both those themes. Also, made the default template file a little more legible with white space.

### DIFF
--- a/src/themes/OLH/templates/journal/authors.html
+++ b/src/themes/OLH/templates/journal/authors.html
@@ -10,12 +10,15 @@
     <div class="row" data-equalizer data-equalize-on="medium">
        
         {% for author in author_list %}
+
             <div class="medium-3 columns {% if forloop.last %}end{% endif %}">
                 <div class="{% if journal_settings.general.enable_editorial_images %}editorial-team{% endif %} callout" data-equalizer-watch>
                    
                     {% if journal_settings.general.enable_editorial_images %}
+
                         <img class="thumbnail editorial-image" src="{% if author.profile_image %}{{ author.profile_image.url }}{% else %}{% static "common/img/icons/users.png" %}{% endif %}"
                              alt="Photo of Uranus.">
+
                     {% endif %}
                    
                     <strong>{{ author.full_name }}</strong><br/>
@@ -24,11 +27,13 @@
                     </i>
                     <br/>
                     
-                    {% if user.enable_public_profile %}
+                    {% if author.enable_public_profile %}
+
                         <p>
                             <small><a href="{% url 'core_public_profile' author.uuid %}">View
                                 Profile</a></small>
                         </p>
+                        
                     {% endif %}
 
                     {% include "elements/journal/editorial_social_content.html" with user=author %}
--- a/src/themes/OLH/templates/journal/authors.html
+++ b/src/themes/OLH/templates/journal/authors.html
@@ -1,0 +1,43 @@
+{% extends "core/base.html" %}
+{% load static from staticfiles %}
+
+{% block title %}Authors{% endblock title %}
+{% block page_title %}Authors{% endblock page_title %}
+
+{% block body %}
+
+    <h3>Authors</h3>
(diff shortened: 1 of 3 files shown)


Bbkctp:

src/themes/OLH/templates/journal/authors.html
```diff
@@ -0,0 +1,43 @@
...
+                             alt="Photo of Uranus.">
+                    {% endif %}
+                   
+                    <strong>{{ author.full_name }}</strong><br/>
+                    <i>
+                        <small>{{ author.affiliation }}</small>
+                    </i>
+                    <br/>
+                    
+                    {% if user.enable_public_profile %}
```
mauromsl: shouldn't it be `author.enable_public_profile`?

src/themes/material/templates/journal/authors.html
```diff
@@ -0,0 +1,43 @@
...
+
+                    </div>
+                </div>
+            </div>
+                       
+        {% endfor %}
+
+    </div>
+
+{% endblock body %}
```
mauromsl: Missing new line here (They are missing throughout most of the code base but we are trying to add them for new files and whenever we touch current files)

src/themes/OLH/templates/journal/authors.html
```diff
@@ -0,0 +1,43 @@
...
+                    {% include "elements/journal/editorial_social_content.html" with user=author %}
+                   
+                </div>
+            </div>
+            
+        {% endfor %}
+
+    </div>
+
+{% endblock body %}
```
mauromsl: Missing new line

src/themes/OLH/templates/journal/authors.html
```diff
@@ -0,0 +1,43 @@
...
+    <h3>Authors</h3>
+    <div class="row" data-equalizer data-equalize-on="medium">
+       
+        {% for author in author_list %}
+            <div class="medium-3 columns {% if forloop.last %}end{% endif %}">
+                <div class="{% if journal_settings.general.enable_editorial_images %}editorial-team{% endif %} callout" data-equalizer-watch>
+                   
+                    {% if journal_settings.general.enable_editorial_images %}
+                        <img class="thumbnail editorial-image" src="{% if author.profile_image %}{{ author.profile_image.url }}{% else %}{% static "common/img/icons/users.png" %}{% endif %}"
+                             alt="Photo of Uranus.">
```
mauromsl: And the meme goes on and on!

src/themes/OLH/templates/journal/authors.html
```diff
@@ -0,0 +1,43 @@
...
+
+    <h3>Authors</h3>
+    <div class="row" data-equalizer data-equalize-on="medium">
+       
+        {% for author in author_list %}
+            <div class="medium-3 columns {% if forloop.last %}end{% endif %}">
+                <div class="{% if journal_settings.general.enable_editorial_images %}editorial-team{% endif %} callout" data-equalizer-watch>
+                   
+                    {% if journal_settings.general.enable_editorial_images %}
+                        <img class="thumbnail editorial-image" src="{% if author.profile_image %}{{ author.profile_image.url }}{% else %}{% static "common/img/icons/users.png" %}{% endif %}"
```
mauromsl: Could use `default_if_none` here: https://docs.djangoproject.com/en/1.11/ref/templates/builtins/#default-if-none
Although I think you are trying to be consistent with other templates

hachacha: Yeah I was trying to be consistent.
on
`{{ author.profile_image.url|default_if_none:'just/to/test.png' }}`
I get a valueerror when updating to
`The 'profile_image' attribute has no file associated with it.`
should I just stick with what's currently there?

mauromsl: Ah fair enough, the lack of a `profile_image` is the problem, not the lack of a `url`. Yes, leave it as it is

PR #839, opened 2019-02-05
Closes #812, makes managing typeset tasks fit description given.

### LINKED ISSUE #812: Possible to Edit Accepted Typesetting task which is contray to stated instructions (?)
Logged in as Production Manager and assign a typesetting task,
Which is accepted by another account with typesetting role:
screenshot 102
screenshot 104

As Production Manager it show as accepted:
screenshot 103
If I click on Edit, I get the next screen with the following instructions:
"You can edit the details of the typesetting task below. Note, if the assignment has been accepted you cannot edit it, if it has been rejected you may reset it."

screenshot 105

When I go back to the typesetter account, the text has changed and the manuscript file is no longer associated as per edited task:

As it has been accepted, I assume this means I can't edit the task (?) However, I can.
I add some text and unclick the associated file:

screenshot 106

screenshot 107

If this ability to edit is by design, and I am misinterpreting the instructions or functionality here,  I would say the instructional text needs changing . . . ?

ajrbyers:
#812

### DIFF
--- a/src/core/dev_settings.py
+++ b/src/core/dev_settings.py
@@ -10,4 +10,17 @@ EMAIL_BACKEND = 'django.core.mail.backends.smtp.EmailBackend'
 URL_CONFIG = 'path'  # path or domain
 EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 
-MIDDLEWARE_CLASSES = ('utils.middleware.TimeMonitoring',)
+MIDDLEWARE_CLASSES = (
+    'utils.middleware.TimeMonitoring',
+    'debug_toolbar.middleware.DebugToolbarMiddleware'
+)
+INSTALLED_APPS = ['debug_toolbar']
+
+
+def show_toolbar(request):
+    return True
+
+
+DEBUG_TOOLBAR_CONFIG = {
+    "SHOW_TOOLBAR_CALLBACK": show_toolbar,
+}
--- a/src/production/models.py
+++ b/src/production/models.py
@@ -66,3 +66,13 @@ class TypesetTask(models.Model):
                 self.completed.strftime("%Y-%m-%d %H:%M"))}
         elif self.completed and self.editor_reviewed:
             return {'slug': 'closed', 'friendly': 'Task closed'}
+
+    @property
+    def get_status_code(self):
+        return self.status().get('slug')
+
+    def reset_task_dates(self):
+        self.accepted = None
+        self.completed = None
+        self.editor_reviewed = False
+        self.save()
--- a/src/production/models.py
+++ b/src/production/models.py
@@ -54,22 +54,30 @@ class TypesetTask(models.Model):
         else:
             return False
 
+    @property
     def status(self):
         if self.assigned and not self.accepted and not self.completed:
-            return {'slug': 'assigned', 'friendly': 'Awaiting response'}
+            return "assigned"
         elif self.assigned and self.accepted and not self.completed:
-            return {'slug': 'accepted', 'friendly': 'Task accepted, underway'}
+            return "accepted"
         elif self.assigned and not self.accepted and self.completed:
-            return {'slug': 'declined', 'friendly': 'Task declined'}
+            return "declined"
         elif self.completed and not self.editor_reviewed:
-            return {'slug': 'completed', 'friendly': 'Completed on {0}<br/>Awaiting manager review'.format(
-                self.completed.strftime("%Y-%m-%d %H:%M"))}
+            return "completed"
         elif self.completed and self.editor_reviewed:
-            return {'slug': 'closed', 'friendly': 'Task closed'}
+            return "closed"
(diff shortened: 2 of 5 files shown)


src/production/models.py
```diff
@@ -66,3 +66,13 @@ def status(self):
                 self.completed.strftime("%Y-%m-%d %H:%M"))}
         elif self.completed and self.editor_reviewed:
             return {'slug': 'closed', 'friendly': 'Task closed'}
+
+    @property
+    def get_status_code(self):
```
mauromsl: If it is a property it should probably be named `status_code` instead

src/production/models.py
```diff
@@ -66,3 +66,13 @@ def status(self):
...
+        return self.status().get('slug')
```
mauromsl: The way in which `TypesetTask.status` is coded it could return `None`, which would throw an `AttributeError` here. I think a cleaner implementation would be for `TypesetTask.status` to return the status code and some other `TypesetTask.slug` which calls `TypesetTask.status` and returns a slug structure.

src/templates/admin/production/edit_typesetter_assignment.html
```diff
@@ -73,6 +74,14 @@ <h2>Edit Assignment</h2>
...
                               data-origin="textarea">{{ typeset.typeset_task }}</textarea>
                     <button class="button success" type="submit" name="update"><i class="fa fa-check">&nbsp;</i>Update Task</button>
+                    {% elif typeset.status.slug == 'declined' %}
+                    <p>This typesetting task has been declined. If this was done in error you can
+                        reset it here.</p>
+                        <button class="button" name="reset"><i class="fa fa-recycle"></i> Reset</button>
+
+                    {% else %}
+                    <p>This typesetting task has been accepted so cannot be edited.</p>
+                    {% endif %}
```
mauromsl: I think this is equivalent and easier to read:
```
{% if typeset.accepted %}
[...]
{% elif typeset.declined %}
[...]
{% else %}
[...]
{% endif %}
```
I'm assuming that a task can't be both accepted and declined at the same time.

src/core/dev_settings.py
```diff
@@ -10,4 +10,17 @@
 URL_CONFIG = 'path'  # path or domain
 EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 
-MIDDLEWARE_CLASSES = ('utils.middleware.TimeMonitoring',)
+MIDDLEWARE_CLASSES = (
+    'utils.middleware.TimeMonitoring',
+    'debug_toolbar.middleware.DebugToolbarMiddleware'
+)
+INSTALLED_APPS = ['debug_toolbar']
+
```
mauromsl: spacing looks inconsistent

ajrbyers: Not sure why you mean here clarification?

mauromsl: There is one space between constants but two between a constant and a function

src/production/models.py
ajrbyers: Done.

src/production/models.py
ajrbyers: Fixed.

src/production/models.py
```diff
@@ -54,15 +54,33 @@ def is_active(self):
...
         elif self.completed and self.editor_reviewed:
-            return {'slug': 'closed', 'friendly': 'Task closed'}
+            return "closed"
+
+    FRIENDLY_STATUSES = {
+            "assigned": "Awaiting response",
+            "accepted": "Task accepted",
+            "declined": "Task declined",
+            "completed": "Task completed",
+            "closed": "Task closed",
```
mauromsl: might be desirable to have an `unassigned` status as well? (its the only hole I can find the `status` property

ajrbyers: I'm not so sure about an `unassigned` status, they are by default assigned, but wonder if we should add an `unknown` status encase someone messes with the data outside the worklflow?

src/production/views.py
```diff
@@ -312,19 +312,49 @@ def edit_typesetter_assignment(request, typeset_id):
...
+            )
         elif 'update' in request.POST:
             logic.update_typesetter_task(typeset, request)
+        elif 'reset' in request.POST and typeset.status == 'declined':
+            typeset.reset_task_dates()
+        else:
+            messages.add_message(
+                request,
+                messages.WARNING,
+                'An invalid operation has been attempted for this task.'
```
mauromsl: Might be useful to include `task.status` in this message

src/production/models.py
mauromsl: If it is always assigned why are we checking `if self.assigned...` over and over in the `status` property?

ajrbyers: It can be deleted in admin, its not an immutable field. Safest to check its not null.

mauromsl:
test this please

PR #841, opened 2019-02-05
Closes #807, removes article ordering records when an article is remo…

### LINKED ISSUE #807: Remove Article from Issue Still Returns Article
To recreate:
- create issue id=2
- add article id=8, id=4
- remove 4 using /manage/issues/2/remove/?article=4
- on /manage/issues/2/  the number of articles = 1

- on the manage page `table_of_contents.html` TOC shows both 8 and 4
- on the front end `/issue/2/info/` shows both 8 and 4

- only 8 is showing as selected in url `/admin/journal/issue/2/change/`

the journal view's `def issue` uses 
	[Issue.structure](https://github.com/BirkbeckCTP/janeway/blob/master/src/journal/models.py#L487) which returns: `<QuerySet [<ArticleOrdering: 0: Article, #4>, <ArticleOrdering: 1: Article, #8>]>`
as does the table of contents template  {% for section, articles in issue.structure.items %}

4 is still showing in /admin/journal/articleordering/
so the articles are being found and displayed through the `ArticleOrdering` model and not really through `Issue` 

deleting 4 from ArticleOrdering.article.pk=4 removes it from all the places it should not be.

the remove function should probably also delete the article ordering entry as well as it relates to Issue and/or views could gather the ordering integer and not the article from articleordering with `issue_articles` and `get_article_order`?
I don't fully understand what sections are or do but there seems to be a portion in that structure function that loops through and adds articles to the final returned list based on ArticleOrdering [here](https://github.com/BirkbeckCTP/janeway/blob/master/src/journal/models.py#L493):
``` 
    for section in sections:
            article_with_order = ArticleOrdering.objects.filter(issue=self, section=section)

            article_list = list()
            for order in article_with_order:
                article_list.append(order.article)

            for article in articles.filter(section=section):
                if not article in article_list:
                    article_list.append(article)
            structure[section] = article_list
```

ajrbyers:
…ved from an issue.

### DIFF
--- a/src/journal/models.py
+++ b/src/journal/models.py
@@ -551,6 +551,18 @@ class Issue(models.Model):
                 issue = latest_issue.issue + 1
                 return (latest_issue.volume, issue)
 
+    def remove_article(self, article):
+        """Removes an article from an issue and deletes its ordering."""
+        self.articles.remove(article)
+        try:
+            ordering = ArticleOrdering.objects.get(
+                issue=self,
+                article=article,
+            )
+            ordering.delete()
+        except ArticleOrdering.DoesNotExist:
+            pass
+
     def __str__(self):
         return u'{0}: {1} {2} ({3})'.format(self.volume, self.issue, self.issue_title, self.date.year)
 
--- a/src/journal/models.py
+++ b/src/journal/models.py
@@ -12,7 +12,8 @@ import os
 
 from django.conf import settings
 from django.db import models
-from django.db.models.signals import post_save
+from django.db.models.signals import post_save, m2m_changed
+
 from django.dispatch import receiver
 from django.urls import reverse
 from django.utils import timezone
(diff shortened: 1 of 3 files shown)


ajrbyers:
A signal hadn’t even occurred to me. There is a m2m_changed signal that could be used I supposed.

src/journal/models.py
```diff
@@ -683,3 +684,29 @@ def setup_default_form(sender, instance, created, **kwargs):
...
+def article_remove_from_issue(sender, **kwargs):
+    """
+    When an article is removed from an issue this signal will delete any
+    orderings for that article in the issue.
+    """
+    issue = kwargs.get('instance')
+    action = kwargs.get('action')
+
+    if issue and action == 'post_remove':
+        article_ids = kwargs.get('pk_set')
```
mauromsl: Maybe `kwargs.get('pk_set', [])` ?

src/journal/models.py
```diff
@@ -683,3 +684,29 @@ def setup_default_form(sender, instance, created, **kwargs):
...
+
+def article_remove_from_issue(sender, **kwargs):
+    """
+    When an article is removed from an issue this signal will delete any
+    orderings for that article in the issue.
+    """
+    issue = kwargs.get('instance')
+    action = kwargs.get('action')
+
+    if issue and action == 'post_remove':
```
mauromsl: Looks like we need to check the `reverse` argument as well: 
https://docs.djangoproject.com/en/2.1/ref/signals/#m2m-changed
Otherwise, if `ArticleOrdering` with `pk=1` is removed from `Article` with `pk=2` this signal will be invoked and the ArticleOrdering with `pk=2` will be accidentally deleted

mauromsl: So basically the instance could be either a `submission.models.Article` or a `journal.models.Issue` depending on which direction the removal has hapenned (issue removes article or article removes issue)

ajrbyers:
Need to run some tests for the `reverse` stuff, but works well forwards.

PR #853, opened 2019-02-06
#762: Journals can be conferences

### LINKED ISSUE #762: Seperating Journals from Conferences
- [ ] Add a bool to press `publishes_conferences`
- [ ] Add a bool to journal `is conference`
- [ ] If press publishes conferences display these in a separate listing page

mauromsl:
closes #762


mauromsl:
The reason why I didn't do it is because there is no journals page support for either default nor material themes as far as I can see.

PR #858, opened 2019-02-07
693 login form improvements

### LINKED ISSUE #693: LoginAttempts not clearing
We need to investigate the loginattempt system, it appears to not clear.

mauromsl:
Closes #693

### DIFF
--- a/src/core/forms.py
+++ b/src/core/forms.py
@@ -3,7 +3,7 @@ __author__ = "Martin Paul Eve & Andy Byers"
 __license__ = "AGPL v3"
 __maintainer__ = "Birkbeck Centre for Technology and Publishing"
 
-
+import logging
 import uuid
 
 from django import forms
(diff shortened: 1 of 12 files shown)


mauromsl:
test this please

src/core/forms.py
```diff
@@ -377,13 +379,38 @@ def clean_email(self):
 
 
 class LoginForm(forms.Form):
-    user_name = forms.CharField(max_length=255)
-    user_pass = forms.CharField(max_length=255, widget=forms.PasswordInput)
+    user_name = forms.CharField(max_length=255, label="Email")
```
ajrbyers: Can we make the label a translatable with ugettext?

mauromsl: I left the translation logic in the templates as that is what was the current behaviour, but happy to change that as well if you want

ajrbyers: Woops, missed that. Its fine as is I think.

PR #880, opened 2019-02-12
Closes #732, adds editor checks to sections of summary and metadata m…

### LINKED ISSUE #732: Possible Functionality Issue Depending on Requirements

ajrbyers:
closes #732

### DIFF
--- a/src/templates/admin/elements/metadata.html
+++ b/src/templates/admin/elements/metadata.html
@@ -2,8 +2,9 @@
 {% load roles %}
 
 {% user_has_role request 'editor' as user_is_editor %}
+{% user_has_role request 'production' as user_is_production %}
 
-{% if not view_type == "author" and user_is_editor %}
+{% if not view_type == "author" and user_is_editor or user_is_production %}
     <a class="float-right" href="{% url 'edit_metadata' article.pk %}?return={{ request.path|urlencode }}"><i
             class="fa fa-edit">&nbsp;</i>Edit</a>{% endif %}
 <h4>{{ article.title|safe }}</h4>
--- a/src/templates/admin/elements/metadata.html
+++ b/src/templates/admin/elements/metadata.html
@@ -1,6 +1,9 @@
 {% load settings %}
+{% load roles %}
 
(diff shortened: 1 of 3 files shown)


src/templates/admin/elements/metadata.html
```diff
@@ -1,6 +1,9 @@
 {% load settings %}
+{% load roles %}
 
-{% if not view_type == "author" %}
+{% user_has_role request 'editor' as user_is_editor %}
+
+{% if not view_type == "author" and user_is_editor %}
```
mauromsl: I'm a bit confused here, from the original issue it looks like you agreed with David that the typesetter should have access to this view and to add notes(below) but the code restricts them from doing so

ajrbyers: Yes the wording was confusing but the outcome is what I always intended except instead of making a special model for typesetters I simply hid the bits they can’t use...

ajrbyers: For avoidance of doubt: typesetters should not be allowed to edit metadata or add notes to articles.

mauromsl: Ok cool, the view in the backend is still decorated with `@production_user_or_editor_required` which seems to contradict this logic.

ajrbyers: Good catch, fix committed.

src/templates/admin/elements/metadata.html
```diff
@@ -2,8 +2,9 @@
 {% load roles %}
 
 {% user_has_role request 'editor' as user_is_editor %}
+{% user_has_role request 'production' as user_is_production %}
 
-{% if not view_type == "author" and user_is_editor %}
+{% if not view_type == "author" and user_is_editor or user_is_production %}
```
mauromsl: Okay I;m a bit confused here:
Why do we still have the `if not view_type == "author"` check?

ajrbyers: Commit added to remove it, with the new check it is vestigial.

PR #884, opened 2019-02-12
Adds support for PDF issue Galleys

### LINKED ISSUE #864: Add ability to download issue file
Further to the work done in #511 it would be great if we could have the option to upload a custom issue file. If this is present, we should serve that. If it isn't, then we should fall back to the traditional zip method.

mauromsl:
closes #864


src/journal/models.py
```diff
@@ -560,6 +564,27 @@ class Meta:
         ordering = ("order", "-date")
 
 
+class IssueGalley(models.Model):
+    FILES_PATH = 'issues'
+
+    file = models.ForeignKey('core.File')
+    #An Issue can only have one galley at this time (PDF)
```
ajrbyers: For PEP8 should be `# comment`

src/journal/urls.py
```diff
@@ -37,6 +37,10 @@
     url(r'^issue/current/$', views.current_issue, name='current_issue'),
     url(r'^issue/(?P<issue_id>\d+)/info/$', views.issue, name='journal_issue'),
     url(r'^issue/(?P<issue_id>\d+)/download/$', views.download_issue, name='journal_issue_download'),
+    url(r'^issue/(?P<issue_id>\d+)/download/(?P<galley_id>\d+)$', views.download_issue_galley, name='journal_issue_download_galley'),
```
ajrbyers: Overhang here seems inconsistent.

src/journal/views.py
```diff
@@ -555,6 +560,23 @@ def file_history(request, article_id, file_id):
 
     return render(request, template, context)
 
+@editor_user_required
```
ajrbyers: PEP8 - two new lines between functions required

src/journal/views.py
```diff
@@ -555,6 +560,23 @@ def file_history(request, article_id, file_id):
...
+
+    template = "journal/file_history.html"
+    context = {
+        'article': None,
+        'file': file_object,
+    }
+
+    return render(request, template, context)
 
 @login_required
```
ajrbyers: PEP8 - two new lines before next function required here also.

src/journal/views.py
```diff
@@ -555,6 +560,23 @@ def file_history(request, article_id, file_id):
 
     return render(request, template, context)
 
+@editor_user_required
+def issue_file_history(request, issue_id):
+    """ Returns the file history of a given Issue Galley file
+
+    """
+    #TODO: Combine with `file_history` above, disabled until GH #865
```
ajrbyers: PEP8 - Comment should start with a hash then a space

src/journal/views.py
```diff
@@ -889,10 +912,42 @@ def manage_issues(request, issue_id=None, event=None):
         'issue': issue,
         'form': form,
         'modal': modal,
+        'galley_form': galley_form,
     }
 
     return render(request, template, context)
 
+@editor_user_required
```
ajrbyers: PEP8 - two new lines between functions

src/journal/views.py
```diff
@@ -889,10 +912,42 @@ def manage_issues(request, issue_id=None, event=None):
...
+                        *(models.IssueGalley.FILES_PATH, issue.pk)
+                )
+                models.IssueGalley.objects.create(
+                    file=file_obj,
+                    issue=issue,
+                )
+
+            messages.info(request, "Issue Galley Uploaded")
+
+    return redirect(reverse('manage_issues_id', kwargs={'issue_id': issue.pk}))
```
ajrbyers: PEP8 - Two new lines between functions required.

src/core/dev_settings.py
```diff
@@ -14,7 +14,7 @@
     'utils.middleware.TimeMonitoring',
     'debug_toolbar.middleware.DebugToolbarMiddleware'
 )
-INSTALLED_APPS = ['debug_toolbar']
+INSTALLED_APPS = ['debug_toolbar', 'django_nose']
```
ajrbyers: Added nose here but it is missing from requirements.txt

src/journal/views.py
```diff
@@ -889,10 +912,42 @@ def manage_issues(request, issue_id=None, event=None):
...
+                        public=False
+                        *(models.IssueGalley.FILES_PATH, issue.pk)
+                )
+                models.IssueGalley.objects.create(
+                    file=file_obj,
+                    issue=issue,
+                )
+
+            messages.info(request, "Issue Galley Uploaded")
+
```
ajrbyers: If I attempt to upload a non-pdf it is duly rejected, however, as there is no handler for where `form.is_valid()` is `False` I don't get told why it was rejected, may be useful to add a handler here for that?

mauromsl: Ah good point, the form does render the errors, but it is not immediately shown due to it being a modal. I'll add a message as well

src/core/dev_settings.py
mauromsl: It is present in `dev_requirements.txt`

ajrbyers: Ok - we'll need to document this then because my build error'd out and other will too if they're not also installing dev-requirements.

mauromsl: `dev_requirements.txt` are installed by default in the docker image and the dev_settings are used by the dockerised Janeway. I imagine the problem is that you haven't rebuilt your image in a while, so would have errored if it was added to `requirements.txt` as well (i.e. it shouldn't error on others if they rebuild)

ajrbyers: And if someone not using docker also decides to use the dev_settings file?

mauromsl: That worflow is not documented so I don't expect that to happen. In any case, once we move the wiki to the docs page, yes we should explain the `dev_requirements.txt` file

ajrbyers: Okay, just playing devils advocate. TBH, if I wasn't going to use Docker and was looking for a settings file for use locally and saw dev_settings and example_settings I'd go with dev_settings, but maybe that's just me :+1:

mauromsl: In that case you would raise an issue and I would tell you to follow the manual :nerd_face:

PR #899, opened 2019-02-13
Minor stage rewordings

### LINKED ISSUE #877: Confusing boolean form for settings
your-journal-name/manager/settings/group/Identifiers/setting/use_crossref/

image

Confusing design/ux
The checkbox must be checked and then you have an option of submit or delete. Unchecking the checkbox and clicking submit is what one would expect would undo the value and turn this to False. However, you must check the box (with no description) and then press delete to set to false.

mauromsl:
closes #877


src/submission/models.py
```diff
@@ -228,9 +228,9 @@ def article_media_upload(instance, filename):
 STAGE_CHOICES = [
     (STAGE_UNSUBMITTED, 'Unsubmitted'),
     (STAGE_UNASSIGNED, 'Unassigned'),
-    (STAGE_ASSIGNED, 'Assigned'),
-    (STAGE_UNDER_REVIEW, 'Under Review'),
-    (STAGE_UNDER_REVISION, 'Under Revision'),
+    (STAGE_ASSIGNED, 'Assigned to Editor'),
```
ajrbyers: A migration will be required to update article.stage from the old to the new values

ajrbyers: Rightly so, talking utter crap here! Looks fine :+1:

PR #912, opened 2019-02-15
Pagination and show filter on author list view

hachacha:
for olh, default, and material themes.


mauromsl:
@hachacha, @ajrbyers: this was autoclosed by #916 due to comments. I suspect that was a mistake

PR #928, opened 2019-02-27
#919 adds a view for previewing article galley images in production

### LINKED ISSUE #919: Images in Production Preview
**Describe the bug**
As noted on https://github.com/ajrbyers/pandoc_plugin/pull/10#issuecomment-465766460 the Production Preview page doesn't render out images correctly when an article is generated from Pandoc.

**Janeway version**
Master

ajrbyers:
Closes #919 

@mdlincoln this is the PR for image previewing in production.

### DIFF
--- a/src/production/urls.py
+++ b/src/production/urls.py
@@ -19,7 +19,7 @@ urlpatterns = [
         ),
     url(r'^(?P<article_id>\d+)/preview/(?P<galley_id>\d+)/(?P<file_name>.*)$',
         views.preview_figure,
-        name='preview_figure'
+        name='production_preview_figure'
         ),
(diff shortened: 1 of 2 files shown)


src/production/urls.py
```diff
@@ -17,6 +17,10 @@
         views.preview_galley,
         name='production_preview_galley'
         ),
+    url(r'^(?P<article_id>\d+)/preview/(?P<galley_id>\d+)/(?P<file_name>.*)$',
+        views.preview_figure,
+        name='preview_figure'
```
mauromsl: Name clashes with a url in `proofing.urls`

Also, should we add a link to this in the templates?

ajrbyers: I've updated that name, this url is for the production_preview_galley page to use when rendering HTML/XML. There is already a download link for galley images.

PR #937, opened 2019-02-27
#934 fixes multiple objects returned for correction tasks

### LINKED ISSUE #934: MultipleObjectsReturned: get() returned more than one TypesetterProofingTask -- it returned 2!
**Describe the bug**
When calling can_view_file on proofing correction task page, two objects are found and an error is rendered.

**Janeway version**
1.3.1

**To Reproduce**
1. Add two Correction Task to a Proofing Request
2. As the Typesetter try to do Correction Task

**Expected behavior**
Page will be renderd

Error is on security/logic.py in can_view_file at line 83

ajrbyers:
Closes #934

### DIFF
--- a/src/security/logic.py
+++ b/src/security/logic.py
@@ -77,12 +77,12 @@ def can_view_file(request, user, file_object):
 
     try:
         if file_object.article_id:
-            proofing_models.TypesetterProofingTask.objects.get(
-                proofing_task__round__assignment__article__pk=file_object.article_id,
-                typesetter=request.user,
-                completed__isnull=True
-            )
-            return True
+            if proofing_models.TypesetterProofingTask.objects.filter(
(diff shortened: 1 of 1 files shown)


src/security/logic.py
```diff
@@ -77,12 +77,12 @@ def can_view_file(request, user, file_object):
...
-                typesetter=request.user,
-                completed__isnull=True
-            )
-            return True
+            if proofing_models.TypesetterProofingTask.objects.filter(
+                    proofing_task__round__assignment__article__pk=file_object.article_id,
+                    typesetter=request.user,
+                    completed__isnull=True
+            ).exists():
+                return True
```
mauromsl: I don't think we need the try/except block here anymore instead we could just:
`return proofing_models.TypesetterProofingTask.objects.filter([...]).exists()`

PR #960, opened 2019-03-05
Activate template timezone based on user/browser configuration

### LINKED ISSUE #312: Time zone awareness in templates
- [ ] Add new timezone attribute to the user object, default to UTC(?)
- [ ] Add a new middleware (https://docs.djangoproject.com/en/1.11/topics/i18n/timezones/#selecting-the-current-time-zone) to set the timezone to the current user's if they have it and UTC if not.

mauromsl:
closes #312 

Adds `preferred_timezone` to Account model

New timezone logic:
Janeway will attempt to set the timezone for the session to the first of the following scenarios to be met

 1. User has chosen a timezone for his account
 2. A timezone has been successfully read from the browser
 3. USE_TZ has been set in settings.py (previous behaviour)


src/core/middleware.py
```diff
@@ -200,3 +204,21 @@ def get_current_request(cls):
     @staticmethod
     def process_request(request):
         _threadlocal.request = request
+
```
ajrbyers: PEP8 - two new lines required here

src/core/models.py
```diff
@@ -129,6 +131,7 @@ def profile_images_upload_path(instance, filename):
                    (u'VG', u'Virgin Islands, British'), (u'VI', u'Virgin Islands, U.S.'), (u'WF', u'Wallis and Futuna'),
                    (u'EH', u'Western Sahara'), (u'YE', u'Yemen'), (u'ZM', u'Zambia'), (u'ZW', u'Zimbabwe')]
 
+TIMEZONE_CHOICES = tuple(zip(pytz.all_timezones, pytz.all_timezones))
```
ajrbyers: PEP8 - Two new lines here

src/core/views.py
```diff
@@ -1767,3 +1770,24 @@ def order_workflow_elements(request):
             element.save()
 
     return HttpResponse('Thanks')
+
```
ajrbyers: PEP8 - Two newlines.

PR #982, opened 2019-03-05
855: Adds email page to editor unassign view

### LINKED ISSUE #855: Inconsistent Functionality When Removing Editor Assignment?
The following may well be intentional, but I will just draw your attention to an inconsistency in removing assignments.
If I assign a REVIEW assignment an email can be sent to inform the reviewer:
screenshot 113
If the assignment is canceled for whatever reason, and email can also be sent:
screenshot 112
However, this is not true for an EDITOR assignment: An email can be sent when a task if assigned:
screenshot 110
But it is possible to simply delete the assignment by clicking on the delete button, and no opportunity is presented to send the ex-editor an email in a similar fashion to canceling a review assignment.
screenshot 111

mauromsl:
closes #855

### DIFF
--- a/src/utils/transactional_emails.py
+++ b/src/utils/transactional_emails.py
@@ -31,6 +31,30 @@ def send_reviewer_withdrawl_notice(**kwargs):
         notify_helpers.send_slack(request, description, ['slack_editors'])
 
 
+
+def send_editor_unassigned_notice(request, message, assignment, skip=False):
+    description = "{a.editor} unassigned from {a.article} by {r.user}".format(
+            a=assignment,
+            r=request,
+    )
+
+    if not skip:
+
+        log_dict = {
+                'level': 'Info', 'action_text': description,
+                'types': 'Editor Unassigned',
+                'target': assignment.article
+        }
+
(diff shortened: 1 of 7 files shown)


src/utils/transactional_emails.py
```diff
@@ -31,6 +31,30 @@ def send_reviewer_withdrawl_notice(**kwargs):
         notify_helpers.send_slack(request, description, ['slack_editors'])
```
ajrbyers: Looks like three newlines here?

src/utils/transactional_emails.py
```diff
@@ -31,6 +31,30 @@ def send_reviewer_withdrawl_notice(**kwargs):
...
+
+        notify_helpers.send_email_with_body_from_user(
+                request,
+                'subject_review_withdrawl',
+                assignment.editor.email,
+                message,
+                log_dict=log_dict,
+        )
+    notify_helpers.send_slack(request, description, ['slack_editors'])
+
```
ajrbyers: Move the newline from above down here! :+1:

PR #984, opened 2019-03-06
Allow authors to access their proofreading tasks

### LINKED ISSUE #792: Decorator: proofreader_for_article_required blocks authors
Authors assigned to proofing tasks should pass this check.

mauromsl:
closes #792

### DIFF
--- a/src/security/decorators.py
+++ b/src/security/decorators.py
@@ -795,29 +795,41 @@ def proofreader_for_article_required(func):
     """
 
     def wrapper(request, *args, **kwargs):
+        if 'article_id' in kwargs:
+            article = get_object_or_404(
+                    models.Article,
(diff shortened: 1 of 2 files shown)


src/security/decorators.py
```diff
@@ -795,30 +795,42 @@ def proofreader_for_article_required(func):
...
+
         if not base_check(request):
             return redirect('{0}?next={1}'.format(reverse('core_login'), request.path_info))
 
-        if request.user.is_editor(request) or request.user.is_staff:
+        elif request.user.is_editor(request) or request.user.is_staff:
             return func(request, *args, **kwargs)
 
-        if not request.user.is_proofreader(request):
-            deny_access(request)
```
mauromsl: I've removed this check as I think it is redundant with the new else

PR #990, opened 2019-03-11
Added URLs to specifically handle DOIs.

### LINKED ISSUE #958: Article Galley Downloads
**Describe the bug**
If a user attempts to download a galley file but misses the trailing `/` an attribute error is raised `AttributeError: 'NoneType' object has no attribute 'galley_set'`

**Janeway version**
1.3.3

**To Reproduce**
1. Download a galley but remove the trailing slash

**Expected behavior**
Galley will download

### LINKED ISSUE #988: Images not rendering in article when using upload option in proofing/typesetting
**Describe the bug**
Images are not rendering in the article after publication when they were uploaded during production.

**Janeway version**
39acd7a83eefa4b23076b3f36deb0600aea3c88c

**To Reproduce**
1. submit article and get to the proofing page
2. convert docx to html
3. add images to the converted html via upload
4. continue and publish the article
5. go to the article view page
6. be very sad that there is no image.

**Expected behavior**
You would see an image in the published article just as you saw in the proofing stage.

**Screenshots**
In proofing preview and with attached file (note URL):
image
that image src is pointing @ http://localhost:8000/cmesh/production/18/preview/30/image1.jpeg

after publication:
image
src is pointing @ http://localhost:8000/cmesh/article/id/18/image1.jpeg

**Additional context**
I'm assuming this is some url that was not set up to serve this?

ajrbyers:
Closes #958 and Closes #988

### DIFF
--- a/src/identifiers/models.py
+++ b/src/identifiers/models.py
@@ -21,11 +21,16 @@ identifier_choices = (
     ('pubid', 'Publisher ID'),
 )
 
-IDENTIFIER_TYPES = [
+IDENTIFIER_TYPES = {
     'uri',
     'pubid',
     'id',
-]
+    'doi'
+}
+
+NON_DOI_IDENTIFIER_TYPES = IDENTIFIER_TYPES - {"doi"}
+
+DOI_REGEX_PATTERN = '10.\d{4,9}/[-._;()/:A-Za-z0-9]+'
 
 
 class Identifier(models.Model):
--- a/src/identifiers/models.py
+++ b/src/identifiers/models.py
@@ -21,6 +21,12 @@ identifier_choices = (
     ('pubid', 'Publisher ID'),
 )
 
+IDENTIFIER_TYPES = [
+    'uri',
+    'pubid',
+    'id',
+]
+
 
 class Identifier(models.Model):
     id_type = models.CharField(max_length=300, choices=identifier_choices)
--- a/src/journal/urls.py
+++ b/src/journal/urls.py
@@ -6,18 +6,22 @@ __maintainer__ = "Birkbeck Centre for Technology and Publishing"
 from django.conf.urls import url
 
 from journal import views
+from identifiers.models import NON_DOI_IDENTIFIER_TYPES, DOI_REGEX_PATTERN
 
-from identifiers.models import IDENTIFIER_TYPES
+NON_DOI_PIPE_SEPARATED_IDENTIFIERS = "|".join(NON_DOI_IDENTIFIER_TYPES)
 
-idents = "|".join(IDENTIFIER_TYPES)
+# Various url patterns in this module are have duplicated names
+# This is so we can handle DOI patterns using a more restrictive
+# Regex pattern an example is this is 'article_view'
 
 urlpatterns = [
     # Figures and download patterns
     url(r'^article/(?P<identifier_type>{0})/(?P<identifier>.+)/print/$'
-        ''.format("|".join(IDENTIFIER_TYPES)),
+        ''.format(NON_DOI_PIPE_SEPARATED_IDENTIFIERS),
         views.print_article,
         name='article_print_article'),
-    url(r'^article/(?P<identifier_type>doi)/(?P<identifier>10.\d{4,9}/[-._;()/:A-Za-z0-9]+)/print/$',
+    url(r'^article/(?P<identifier_type>doi)/(?P<identifier>{0})/print/$'
+        ''.format(DOI_REGEX_PATTERN),
         views.print_article,
         name='article_print_article'),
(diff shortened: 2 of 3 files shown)


src/journal/urls.py
```diff
@@ -7,15 +7,26 @@
...
     url(r'^article/(?P<article_id>\d+)/galley/(?P<galley_id>\d+)/figure/(?P<file_name>.*)/$',
         views.article_figure,
-        name='article_figure'),
-    url(r'^article/(?P<identifier_type>.+?)/(?P<identifier>.+)/file/(?P<file_id>\d+)/replace$',
+        name='article_galley_figure'),
+    url(r'^article/(?P<identifier_type>id)/(?P<identifier>.+)/file/(?P<file_id>\d+)/replace$',
         views.replace_article_file,
         name='article_file_replace'),
-    url(r'^article/(?P<identifier_type>.+?)/(?P<identifier>.+)/file/(?P<file_id>\d+|None)/$',
+    url(r'^article/(?P<identifier_type>id)/(?P<identifier>.+)/file/(?P<file_id>\d+|None)/$',
```
mauromsl: Why do we need to support the file id `None` here but not in `article_file_replace` above?

src/identifiers/models.py
```diff
@@ -21,6 +21,12 @@
     ('pubid', 'Publisher ID'),
 )
 
+IDENTIFIER_TYPES = [
+    'uri',
+    'pubid',
+    'id',
```
mauromsl: This might look confusing as it doesn't include `'doi'`, why not include it here and change the order of the urls below?
[ ...]
url(<doi_url>) # more restrictive goes first.
url(identifier_url)

src/journal/urls.py
```diff
@@ -24,15 +35,23 @@
     url(r'^article/(?P<article_id>\d+)/galley/(?P<galley_id>\d+)/view/',
         views.view_galley,
         name='article_view_galley'),
-    url(r'^article/(?P<identifier_type>.+?)/(?P<identifier>.+?/.+?)/table/(?P<table_name>.+)$',
+    url(r'^article/(?P<identifier_type>{0})/(?P<identifier>\d+)/table/(?P<table_name>.+)$'
+        ''.format("|".join(IDENTIFIER_TYPES)),
+        views.download_table,
+        name='article_table'),
```
mauromsl: Could use a comment explaining why we duplicate this urls with the same name

src/journal/urls.py
```diff
@@ -24,15 +35,23 @@
...
+    url(r'^article/(?P<identifier_type>doi)/(?P<identifier>10.\d{4,9}/[-._;()/:A-Za-z0-9]+)/table/(?P<table_name>.+)$',
         views.download_table,
+        name='article_table'),
+    url(r'^article/(?P<identifier_type>{0})/(?P<identifier>\d+)/(?P<file_name>.+)$'
+        ''.format("|".join(IDENTIFIER_TYPES)),
+        views.identifier_figure,
         name='article_figure'),
-    url(r'^article/(?P<identifier_type>.+?)/(?P<identifier>.+?/.+?)/(?P<file_name>.+)$',
+    url(r'^article/(?P<identifier_type>doi)/(?P<identifier>10.\d{4,9}/[-._;()/:A-Za-z0-9]+)/(?P<file_name>.+)$',
```
mauromsl: Could store the DOI regex in a constant to avoid duplication and minor typos

src/journal/urls.py
mauromsl: I believe I found why :grimacing: 
https://github.com/BirkbeckCTP/janeway/blob/39acd7a83eefa4b23076b3f36deb0600aea3c88c/src/journal/views.py#L426

ajrbyers: mauro.sanchez Today at 11:07
if something works, don't touch it, or so they say

mauromsl: haha I believe it was "if ain't broke don't fix-it"

src/journal/urls.py
ajrbyers: Added.

src/journal/urls.py
```diff
@@ -6,16 +6,31 @@
 from django.conf.urls import url
 
 from journal import views
+from identifiers.models import NON_DOI_IDENTIFIER_TYPES, DOI_REGEX_PATTERN
+
+NON_DOI_PIPE_SEPARATED_IDENTIFIERS = "|".join(NON_DOI_IDENTIFIER_TYPES)
+
+# Various url patterns in this module are have duplicated names
```
mauromsl: "are have"

ajrbyers: Yeah. They are have.

PR #1013, opened 2019-03-18
#712 Reworked this setting with help from Helen Saunders

### LINKED ISSUE #712: Typos/Character Mismatch in Review Request Guidelines - Minor Issue

ajrbyers:
Closes #712 

Changes made by Dr. Helen Saunders


mauromsl:
We need a Migration for current installations as well?

ajrbyers:
> We need a Migration for current installations as well?

I think not. Most will have changed this text as its OLH specific. We could add a check to see if the setting value = the old text but I can't imagine it actually hitting for anyone.

PR #1023, opened 2019-03-19
1016 edit press form updates

### LINKED ISSUE #1016: Ability to modify press logo/cover image via manager
The default, material, and OLH templates display a `press_cover` (as an SVG?) but there doesn't appear to be an option to upload a new image for this in the press manager.

An upload form in the press manager to set this image. Also, ideally, this would not be limited to SVG images but could accept other image formats.

ajrbyers:
Closes #1016


src/core/models.py
```diff
@@ -651,6 +650,27 @@ def self_article_path(self):
...
             return os.path.join(settings.BASE_DIR, 'files', 'articles', str(self.article_id), str(self.uuid_filename))
 
+    def url(self):
+        from core.middleware import GlobalRequestMiddleware
+        request = GlobalRequestMiddleware.get_current_request()
+        url_kwargs = {'file_id': self.pk}
+
+        if request.journal and self.article_id:
+            return reverse(
+                'download_journal_file',
```
mauromsl: Does this work? The docs suggest the argument "viewname can be a URL pattern name or the callable view object"
So `journal_views.download_journal_file` or `"journal_file"`?

src/core/models.py
```diff
@@ -651,6 +650,27 @@ def self_article_path(self):
...
+                kwargs=url_kwargs,
+            )
+        elif request.journal:
+            return reverse(
+                'download_journal_file',
```
mauromsl: As above. Also, I think this branch is a duplicate of the above

src/press/forms.py
```diff
@@ -5,29 +5,46 @@
...
-    def __init__(self, *args, **kwargs):
-        super(PressForm, self).__init__(*args, **kwargs)
-        press = models.Press.objects.all()[0]
-        content_type = ContentType.objects.get_for_model(press)
-
-        self.fields['featured_journals'].queryset = journal_models.Journal.objects.filter(hide_from_press=False)
-        self.fields['carousel_news_items'].queryset = comms_models.NewsItem.objects.filter(
-            content_type=content_type,
-            object_id=press.pk
+        exclude = (
```
mauromsl: Might be simpler and safer to write an include instead of an exclude

src/press/forms.py
```diff
@@ -5,29 +5,46 @@
...
+        widgets = {
+            'featured_journals': forms.CheckboxSelectMultiple,
+        }
+
+    def save(self, commit=True):
+        press = super(PressForm, self).save(commit=False)
+        request = GlobalRequestMiddleware.get_current_request()
+
+        if press.thumbnail_image:
+            press.thumbnail_image.delete()
```
mauromsl: Should defer the delete until the end, in case the below fails

src/press/forms.py
```diff
@@ -5,29 +5,46 @@
...
+            'carousel_items', 'homepage_news_items', 'carousel',
+            'carousel_news_items', 'thumbnail_image',
         )
+        widgets = {
+            'featured_journals': forms.CheckboxSelectMultiple,
+        }
+
+    def save(self, commit=True):
+        press = super(PressForm, self).save(commit=False)
+        request = GlobalRequestMiddleware.get_current_request()
```
mauromsl: This feels really clunky, we really need to refactor the files module.

src/core/homepage_elements/journals/templates/featured_journals.html
```diff
@@ -0,0 +1,28 @@
...
+    </div>
+    <div class="content">
+        <form method="POST">
+            {% csrf_token %}
+            {{ form|foundation }}
+            <button class="button success">Save</button>
+        </form>
+    </div>
+</div>
+{% endblock %}
```
mauromsl: No new line at EOF

src/core/models.py
```diff
@@ -651,6 +650,27 @@ def self_article_path(self):
         if self.article_id:
             return os.path.join(settings.BASE_DIR, 'files', 'articles', str(self.article_id), str(self.uuid_filename))
 
+    def url(self):
```
mauromsl: Maybe `download_url` would be more descriptive?

ajrbyers: It has to be URL for the Widget to work, but I do agree.

src/core/models.py
ajrbyers: Raising NotImplementedError for these for the time being,

src/press/forms.py
ajrbyers: It does, and we should!

ajrbyers:
I've addressed the comments as best I can :)

PR #1039, opened 2019-03-25
Account Institution made optional

mdlincoln:
Is there a strong reason why `Institution` is the only required field for user accounts besides email, pw, and name? We've had a couple of our testers remark (correctly, I think!) that this is not a desirable message to send in the modern academic world.


Bbkctp:

mauromsl:
add to whitelist

ajrbyers:
@mdlincoln we’re discussing this at the moment. That said, if the premise of the change is accepted there are a bunch of instances that will need to be handled for an empty Institution field throughout the codebase.

ajrbyers:
We're all on board with making institution an optional field, however, as mentioned above we need to add handlers for this throughout the codebase especially for the front end so that Institution doesn't show up as `None`.

PR #1043, opened 2019-03-25
#996 added html block to OLH theme with styling.

### LINKED ISSUE #996: Custom HTML blocks take whole width in the journal homepage
**Describe the bug**
When adding a custom HTML block to the homepage, it is rendered as-is, which leads to the content not being justified (screenshot added)

We might want to wrap the user content within our own tags or indicate the necessary tags on the edit page.

**Janeway version**
1.3.3.1

**To Reproduce**
1. Go to Journal home settings
2. Add an HTML block
3. type some text
4. See the homepage

**Expected behavior**

**Screenshots**

image

## Front-end Issues

**Additional context**

ajrbyers:
Closes #996

### DIFF
--- a/src/themes/OLH/templates/journal/homepage_elements/html_block.html
+++ b/src/themes/OLH/templates/journal/homepage_elements/html_block.html
@@ -1,0 +1,7 @@
+<section>
+    <div class="row">
+        <div class="large-12 columns">
+            {{ html_content|safe }}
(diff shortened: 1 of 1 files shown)


ajrbyers:
> Should we add the same for the material theme as well?

The material theme doesn't need this as its inside a container already :)

PR #1045, opened 2019-03-26
751 review form fixes

### LINKED ISSUE #751: Inconsistent Instructions vs Behaviour When Submitting A Review, Plus Erronous Outcome?
https://demojournal.janeway.systems/review/requests/131/
As a normal user I attempt to submit my review. The interface suggests that rather than filling in the review text box,  that "Alternatively, you may upload a file in lieu of completing the review form." If I attach a review file, and try to submit, I am told that the filling in the review text box is still a required field, and I can't submit:

screenshot 74

Moreover, if I do add some text, and choose the blank decision option, it is accepted:
screenshot 75
So my recommendation for the article is 'None' which I assume is not very useful (?)

screenshot 76

ajrbyers:
Closes #751

### DIFF
--- a/src/review/views.py
+++ b/src/review/views.py
@@ -832,17 +832,17 @@ def upload_review_file(request, assignment_id):
 
     if access_code:
         assignment = models.ReviewAssignment.objects.get(
-            Q(pk=assignment_id) &
-            Q(is_complete=False) &
-            Q(article__stage=submission_models.STAGE_UNDER_REVIEW) &
-            Q(access_code=access_code)
+            Q(pk=assignment_id)
+            & Q(is_complete=False)
+            & Q(article__stage=submission_models.STAGE_UNDER_REVIEW)
+            & Q(access_code=access_code)
         )
     else:
         assignment = models.ReviewAssignment.objects.get(
-            Q(pk=assignment_id) &
-            Q(is_complete=False) &
-            Q(article__stage=submission_models.STAGE_UNDER_REVIEW) &
-            Q(reviewer=request.user)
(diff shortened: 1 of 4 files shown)


src/review/views.py
```diff
@@ -799,6 +825,66 @@ def do_review(request, assignment_id):
...
 
 
+@require_POST
+@reviewer_user_for_assignment_required
+def upload_review_file(request, assignment_id):
+    access_code = logic.get_access_code(request)
+
+    if access_code:
+        assignment = models.ReviewAssignment.objects.get(
+            Q(pk=assignment_id) &
```
mauromsl: PEP 8 recommends operators to be leading the line rather than trailing for improved readability:

https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

src/review/views.py
```diff
@@ -799,6 +825,66 @@ def do_review(request, assignment_id):
     return render(request, template, context)
 
 
+@require_POST
+@reviewer_user_for_assignment_required
+def upload_review_file(request, assignment_id):
+    access_code = logic.get_access_code(request)
```
mauromsl: It'd make sense for the `reviewer_user_for_assignment_required` to pass the `access_code` as an argument instead of having to do this again here, but this is not an problem with this PR so I'll shut up :zipper_mouth_face:

ajrbyers: SHUSH but yes.

src/review/views.py
ajrbyers: Done.

PR #1046, opened 2019-03-26
1017 issue editors update

### LINKED ISSUE #1017: Add Editorial Board info to Issues
Currently, the editorial board only displays current editors. Add option to list editors responsible for the issue on issue page to capture history of journal editors and acknowledge their work.

See bepress issue view, editorial board displays under issue cover image: https://lib.dr.iastate.edu/macnewsletter/vol45/iss3/ 

--Feature requested by JCTP editorial board.

ajrbyers:
Closes #1017 

Requires some updating to bring wording in line with models. "Guest Editor" will be "Issue Editor".


src/journal/views.py
```diff
@@ -1060,38 +1076,100 @@ def add_guest_editor(request, issue_id):
...
         'users': users,
-        'guest_editors': guest_editors,
+        'editors': models.IssueEditor.objects.filter(issue=issue),
     }
 
     return render(request, template, context)
 
 
+@require_POST
+def remove_issue_editor(request, issue_id):
```
mauromsl: This view should probably be authenticated

src/templates/admin/journal/manage/add_guest_editor.html
```diff
@@ -15,111 +15,123 @@
...
                                 <td>
-                                    <button type="submit" name="user_remove" value="{{ user.pk }}"><i class="fa fa-chain-broken">
-                                        &nbsp;</i>Remove
+                                    {{ editor.role }}
+                                </td>
+                                <td>
+                                    <button type="submit"
+                                            name="user_remove"
+                                            value="{{ editor.pk }}"
+                                            formaction="{% url 'manage_remove_issue_editor' issue.pk %}"
```
mauromsl: I found this is a bit confusing, why do we use two different forms for adding/removing if we are overriding the action on the submit anyways? I think it would be clearer to either the action to the <form> element

src/journal/migrations/0027_auto_20190321_1611.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('journal', '0026_auto_20190321_1542'),
+    ]
+
+    operations = [
+        migrations.RenameField(
```
mauromsl: The changes in these 3 migrations are bit weird, I suppose you changed your mind a couple of times. Should probably have all this changes in journal 0025 instead.

Also, wouldn't this migrations wipe current guest editors data? Shouldn't we migrate those over as IssueEditors?

src/journal/logic.py
```diff
@@ -413,3 +413,14 @@ def parse_html_table_to_csv(table, table_name):
...
+
+def potential_issue_editors(journal, current_editors):
+    users = [role.user for role in
+             core_models.AccountRole.objects.filter(
+                 journal=journal,
+                 user__is_active=True,
+             ).select_related('user').exclude(
+                 user__in=current_editors,
+             )]
+    return set(users)
```
mauromsl: We don't need the intermediate list here, you can pass the generator expression directly into the set literal:
```
return {role.user for role in
             core_models.AccountRole.objects.filter(
                 journal=journal,
                 user__is_active=True,
             ).select_related('user').exclude(
                 user__in=current_editors,
)}

File: src/journal/models.py [hunk 5]
```diff
@@ -239,9 +239,17 @@ def editor_pks(self):
...
         if objects:
-            users = [role.user for role in core_models.AccountRole.objects.filter(journal=self, user__is_active=True)]
+            users = [role.user for role in
+                     core_models.AccountRole.objects.filter(
+                         journal=self,
+                         user__is_active=True,
+                     ).select_related('user')]
         else:
-            users = [role.user.pk for role in core_models.AccountRole.objects.filter(journal=self, user__is_active=True)]
+            users = [role.user.pk for role in
```
mauromsl: could do as above and avoid constructing intermediate list as it is not used for anything

src/journal/models.py
```diff
@@ -586,6 +600,18 @@ def path_parts(self):
         return self.FILES_PATH, self.issue.pk
 
 
+class IssueEditor(models.Model):
+    account = models.ForeignKey('core.Account')
+    issue = models.ForeignKey(Issue)
+    role = models.CharField(max_length=255, default='Guest Editor')
```
mauromsl: This can be confused easily with account roles, but I can't think of a better name

src/journal/admin.py
```diff
@@ -12,16 +12,17 @@ class IssueAdmin(admin.ModelAdmin):
     list_display = ('volume', 'issue', 'issue_title', 'journal', 'issue_type')
     list_filter = ('journal', 'issue_type')
     search_fields = ('issue_title',)
-    filter_horizontal = ('articles', 'guest_editors')
+    filter_horizontal = ('articles', 'editors')
```
mauromsl: So the admin app doesn't use `related_name`s?

src/journal/models.py
ajrbyers: Do you mean return a set rather than a list?

src/journal/models.py
ajrbyers: Me either.

src/journal/admin.py
ajrbyers: I've removed this because filter_horizontal doesn't work with m2m's that have in intermediate table anyway.

src/journal/models.py
mauromsl: It is returning a set already, but constructing an intermediate list that serves no purpose, same as https://github.com/BirkbeckCTP/janeway/pull/1046#discussion_r269079711

src/journal/admin.py
mauromsl: :+1:

src/journal/migrations/0027_auto_20190321_1611.py
ajrbyers: I've added a forwards migration, just need to add backwards.

src/templates/admin/journal/manage/add_guest_editor.html
ajrbyers: Separate forms are to avoid posting the contents of all of the text input when deleting. I've moved the formaction to the <form> action tag.

ajrbyers:
Changes made.

PR #1056, opened 2019-03-28
#1001 added link to production, updated production template

### LINKED ISSUE #1001: Can't add supplementary files to articles loaded through back content plugin
**Describe the bug**
When an article is loaded via the back content plugin, it is not possible to add supplementary files to it. Instead, one must manually move the article back into production through the admin.

**Janeway version**
1.3.3.1

ajrbyers:
updated view to assign user as PM.

Closes #1001

### DIFF
--- a/src/production/views.py
+++ b/src/production/views.py
@@ -158,7 +158,29 @@ def production_article(request, article_id):
     :return: HttpResponse object
     """
     article = get_object_or_404(submission_models.Article, pk=article_id)
-    production_assignment = models.ProductionAssignment.objects.get(article=article)
+
+    try:
+        production_assignment = models.ProductionAssignment.objects.get(
+            article=article
+        )
+    except models.ProductionAssignment.DoesNotExist:
+        # It is assumed that if a user enters this URL by hand and passes
+        # the decorator that they are fine being assigned as the PM
+        production_assignment = models.ProductionAssignment.objects.create(
+            article=article,
+            editor=request.user,
+            production_manager=request.user,
+            assigned=timezone.now(),
+            notified=True,
+        )
+
+        messages.add_message(
+            request,
+            messages.WARNING,
+            'You have been assigned as the '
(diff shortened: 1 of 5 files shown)


src/production/views.py
```diff
@@ -158,7 +158,29 @@ def production_article(request, article_id):
...
-    production_assignment = models.ProductionAssignment.objects.get(article=article)
+
+    try:
+        production_assignment = models.ProductionAssignment.objects.get(
+            article=article
+        )
+    except models.ProductionAssignment.DoesNotExist:
+        # It is assumed that if a user enters this URL by hand and passes
+        # the decorator that they are fine being assigned as the PM
+        production_assignment = models.ProductionAssignment.objects.create(
```
mauromsl: Why is this needed? Seems a weird thing to create this objects on a GET request

ajrbyers: I agree it is a little odd, but if an article doesn't directly go through the Production process (ie. through the back content plugin) folks still want access to some of the controls from here. Its a one-to-one as well so only a single entity can exist so I think its okay. Open to alternative suggestions though.

mauromsl: If that is the problem, what about creating the production assignment at the time of loading the article from the back content plugin? It seems seem to be what we are implicitly doing here.

ajrbyers: Yeah that is a _specific_ use case, but i meant _any_ article that doesn't go through the production workflow element, so I decided it was best to simple handle it here.

mauromsl: Yeah but this doesn't explicitly target "any article that doesn't go through the production workflow element", but rather any article for which we haven't found a production assignment.

ajrbyers: These are one and the same. An article can't go through production without a Production Assignment.

mauromsl: Then we should at least render a little form for the user to confirm this.

PR #1058, opened 2019-03-28
Updates OLH base templates.

### LINKED ISSUE #1057: OLH template logo
On the press site we should display the janeway logo in the footer of the OLH template, on the journal site we should display the press one if there is an SVG and if not, display the Janeway one.

ajrbyers:
Closes #1057

### DIFF
--- a/src/press/templatetags/press_url.py
+++ b/src/press/templatetags/press_url.py
@@ -16,7 +16,7 @@ def press_url(request):
 
 
 @register.simple_tag
-def svg(filename, partial=False):
+def svg(filename):
     path = filename
 
     if not path:
@@ -30,7 +30,7 @@ def svg(filename, partial=False):
     if isinstance(path, (list, tuple)):
         path = path[0]
 
-    if partial:
+    if not path.startswith(settings.BASE_DIR):
         path = os.path.join(settings.BASE_DIR, path)
 
     try:
--- a/src/press/templatetags/press_url.py
+++ b/src/press/templatetags/press_url.py
@@ -1,9 +1,10 @@
 import mimetypes
+import os
 
 from django import template
 from django.utils.safestring import mark_safe
 from django.urls import reverse
-from django.http import Http404
+from django.conf import settings
 
 from press import models as press_models
 register = template.Library()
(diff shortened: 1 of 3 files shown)


src/press/templatetags/press_url.py
```diff
@@ -15,19 +16,24 @@ def press_url(request):
...
+def svg(filename, partial=False):
     path = filename
 
     if not path:
         return None
 
     mimetype = mimetypes.guess_type(path, strict=True)
+
     if not mimetype or mimetype[0] != 'image/svg+xml':
+        print(1)
```
mauromsl: needs deleting

src/press/templatetags/press_url.py
```diff
@@ -15,19 +16,24 @@ def press_url(request):
...
         return mark_safe('<img src="{url}" class="top-bar-image">'.format(url=reverse('press_cover_download')))
 
     if isinstance(path, (list, tuple)):
         path = path[0]
 
+    if partial:
```
mauromsl: Perhaps `relative_path` would make more sense than `partial`.
In fact, why not remove the extra argument and just:
```
if not path.startswith(settings.BASE_DIR):
    path = os.path.join(settings.BASE_DIR, path)
```
It has the same result as using passing the `partial` argument.

src/themes/OLH/templates/core/base.html
```diff
@@ -149,7 +149,7 @@
             {% elif 'svg' in request.press_cover %}
                 <a href="{% url 'website_index' %}">{% svg request.press_cover %}</a>
             {% else %}
-                <a href="{% url 'website_index' %}">{% svg "static/common/img/sample/janeway.svg" %}</a>
+                <a href="https://janeway.systems">{% svg "static/common/img/sample/janeway.svg" "partial" %}</a>
```
mauromsl: Ah, this is why we need the `partial` argument. Maybe the simplest thing is for the `svg` tag to return the actual path required for rendering (like django's `static` template tag)

ajrbyers: Because our external CSS manipulate the SVGs we can't just use a  tag we need the SVG in the HTML.

src/press/templatetags/press_url.py
ajrbyers: Cool, latest commit should resolve this.

PR #1059, opened 2019-03-29
Adds support for configuring journal default settings.

### LINKED ISSUE #828: Allow press to set its own journal defaults

mauromsl:
Closes #828

This is the approach I've taken:
 - `settings.json` are loaded into the db when installing janeway as "default" SettingValue rows.
 - A default Settingvalue has its journal` foreign key set to NULL
 - The `utils.setting_handler` functions will return the default SettingValue when one is not found for the journal
 - The UI for editing settings remains largely the same. There are two new views to allow editing the default journal settings from the press site. This are just proxy views to the journal ones with a check for the `is_staff` role.
 - As a result, newly introduced settings don't require running `sync_settings_with_journals`, just `load_setting_defaults`

An alternative approach would be to split the journal settings and the default settings into their own models, which makes the `edit_setting` view more readable (no more checking `if setting_value.journal is None`) But would result in settings stored in two different sets of tables. The interface of `utils.setting_handler` would remain the same though.


src/core/models.py
```diff
@@ -545,17 +546,30 @@ def __str__(self):
     def __repr__(self):
         return u'%s' % self.name
 
+    @property
+    def default_setting_value(self):
+        return SettingValue.objects.language("en").get(setting=self, journal=None)
```
ajrbyers: This line is too long.

src/core/models.py
```diff
@@ -584,6 +598,32 @@ def process_value(self):
...
+
+        :return: a value
+        """
+        if self.setting.types == 'boolean' and not self.value:
+            return "off"
+        elif self.setting.types == 'boolean':
+            return "on"
+        elif self.setting.types == 'file':
+            if self.journal:
+                return self.journal.site_url(reverse("journal_file", value))
```
ajrbyers: `value` doesn't exist here should this be `self.value`?

src/core/views.py
```diff
@@ -639,6 +653,8 @@ def edit_setting(request, setting_group, setting_name):
         if request.FILES:
             value = logic.handle_file(request, setting_value, request.FILES['value'])
 
+        setting_value = setting_handler.get_setting(
```
ajrbyers: Push these onto multiple lines?

src/utils/setting_handler.py
```diff
@@ -27,11 +27,33 @@ def create_setting(setting_group, setting_name, type, pretty_name, description,
...
+    :fallback: If True, it will attempt to return the value for the base
+        language when no value is available for the current language
+    :default: If True, returns the default SettingValue when no journal specific
+        value is present
+    """
     setting = core_models.Setting.objects.get(name=setting_name)
     lang = get_language() if setting.is_translatable else 'en'
 
-    return _get_setting(setting_group, setting, journal, lang, create, fallback)
+    return _get_setting(
```
ajrbyers: as above, push these onto multiple lines?

src/utils/setting_handler.py
```diff
@@ -46,32 +68,37 @@ def get_requestless_setting(setting_group, setting, journal):
...
+        _fallback = settings.LANGUAGE_CODE
+    else:
+        _fallback = None # deactivates fallback
+
     try:
-        setting = core_models.SettingValue.objects.language(lang).get(
-            setting__group__name=setting_group,
-            setting=setting,
-            journal=journal
+        return core_models.SettingValue.objects \
```
ajrbyers: Aren't backslashes the devil!? :P

ajrbyers:
There is now also a minor conflict in `utils/testing/helpers`

src/utils/setting_handler.py
mauromsl: They are! but they are also a necessary devil in this case :hurtrealbad:

PR #1062, opened 2019-03-29
Adds missing migration

mauromsl:
Seems we missed this migration when adding the plural name for "press"


ajrbyers:
Why are these things even tracked?

mauromsl:
For migrations to be consistent regardless of code changes

PR #1067, opened 2019-04-01
1063 production views checks

### LINKED ISSUE #1063: Production Views Multi Journal
**Describe the bug**
Production views were written before Janeway supported multiple journals. As a result many of its views do not check that a given object belongs to the current journal. Fortunately the decorators will protect from unathorised access but a `staff` user could call up a Production article from one journal inside another if they knew its primary key. 

**Janeway version**
1.3.3.1

ajrbyers:
Partial fix for #1063


ajrbyers:
I can take a look at them, preprint is press based so shouldn't be checking for journal anyway.

ajrbyers:
Perhaps we should open an issue for 1.4 to run over all views and and test for this?

PR #1091, opened 2019-04-24
1089: Removes noisy log line from middleware

### LINKED ISSUE #1089: Noisy middleware warning in log
**Describe the bug**
Unnecessary noisy log warning looks like this:
[Wed Apr 17 15:10:00.850189 2019] [wsgi:error] [pid 11388] [remote 128.48.204.65:65258] \x1b[33mWARNING 2019-04-17 15:10:00,850 middleware P:11388 T:139705087657728 None\x1b[0m

From this line:
https://github.com/BirkbeckCTP/janeway/blob/master/src/core/middleware.py#L90

**Janeway version**
1.3.4

mauromsl:
Closes #1089

### DIFF
--- a/src/core/middleware.py
+++ b/src/core/middleware.py
@@ -87,7 +87,6 @@ class SiteSettingsMiddleware(object):
         """
 
         journal, press, redirect_obj = get_site_resources(request)
-        logging.warning(journal)
         if redirect_obj is not None:
             return redirect_obj
 


ajrbyers:
I find this quite useful... but we should probably add a debug middleware that outputs this sort of debug info rather than add it in here right?

PR #1114, opened 2019-05-08
596 django 2x

ajrbyers:
Upgrade to Django 2.2.


src/templates/admin/review/unassign_editor.html
```diff
@@ -35,7 +35,7 @@ <h5>From {{ request.user.full_name }}</h5>
                     &nbsp;&nbsp;
                     <button type="submit" class="button alert" name="skip"><i class="fa fa-step-forward">&nbsp;</i>Skip
                     </button>
-                    <a href="{% url 'review_unassign_editor' article.pk assignment.editor.pk %}" class="button warning" type="submit">Cancel</a>
```
mauromsl: I think this doesn't belong here, `review.views.unassign_editor` does require both an `article_id` and an `editor_id`

requirements.txt
```diff
@@ -38,10 +38,10 @@ python-crontab==2.2.3
...
 python-docx==0.8.6
-raven==6.0.0 # TODO: Optional for sentry
+raven==6.10.0 # TODO: Optional for sentry
 requests==2.20.0
 six==1.10.0
 sqlparse==0.2.3
 ua-parser==0.7.2
 user-agents==1.0.1
--e git+https://github.com/BirkbeckCTP/django-materialize.git@bcfbfd9150887328dddce965674c84b2b6e6b785#egg=django_materialize
+-e git+https://github.com/BirkbeckCTP/django-materialize.git@baace20177e8b0c9816afa52e07260aca1ebb5f7#egg=django_materialize
```
mauromsl: Can we have a couple of tags on this repo to change between the versions more easily?

PR #1121, opened 2019-05-09
#1119 Tweaked the way core.files.save_file handles path_parts, I thin…

### LINKED ISSUE #1119: issue galleys uploaded to wrong place
**Describe the bug**
When I try to upload an issue galley the (pdf)-files seem to end up in the wrong place,, i.e .../janeway/src/files instead of .. /janeway/src/files/issues/&lt;issueid&gt;/

**Janeway version**
origin/master

**To Reproduce**
1. Create an issue
2. Upload a galley file
3. Try to download the file

**Additional context**
I'm quite sure it's not a file permission issue.
The problem also happens after chown -R apache:apache and chmod -R 775 .../janeway/src/files

ajrbyers:
…k this is more readable and works for the only call from journal.views.issue_galley

Closes #1119

### DIFF
--- a/src/journal/views.py
+++ b/src/journal/views.py
@@ -66,12 +66,18 @@ def home(request):
 
     # call all registered plugin block hooks to get relevant contexts
     for hook in settings.PLUGIN_HOOKS.get('yield_homepage_element_context', []):
-        hook_module = plugin_loader.import_module(hook.get('module'))
-        function = getattr(hook_module, hook.get('function'))
-        element_context = function(request, homepage_elements)
+        try:
+            hook_module = plugin_loader.import_module(hook.get('module'))
+            function = getattr(hook_module, hook.get('function'))
+            element_context = function(request, homepage_elements)
 
-        for k, v in element_context.items():
-            context[k] = v
+            for k, v in element_context.items():
+                context[k] = v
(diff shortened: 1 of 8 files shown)


src/journal/views.py
```diff
@@ -998,11 +998,11 @@ def issue_galley(request, issue_id, delete=False):
...
-                        request,
-                        uploaded_file,
-                        label=issue.issue_title,
-                        public=False
-                        *(models.IssueGalley.FILES_PATH, issue.pk)
+                    request,
+                    uploaded_file,
+                    label=issue.issue_title,
+                    public=False,
+                    path_parts=(models.IssueGalley.FILES_PATH, issue.pk)
```
mauromsl: All the other file functions are using the `*(path_parts)` pattern. I think we should either change all of them or none of them for consistency

ajrbyers: I can change them all, or if you prefer it as it is can you work out why it wasn't working?

mauromsl: It is not working because of the missing comma on the previous line (classic python gotcha). I think  you are right, it is easier to read with the kwarg than with the unpacking

ajrbyers:
Code updated, will look at adding tests tomorrow AM.

PR #1166, opened 2019-06-10
fix articles links and search box when using non standard ports

Beethro:
when using a non standard port the article view links in the default and material themes are missing the port number resulting in a 404
same for the search box in the default theme


Bbkctp:

mauromsl:
add to whitelist

mauromsl:
Thanks @Beethro !

PR #1190, opened 2019-06-20
`ImportCacheEntry.nuke` to handle FileNotFoundError

mauromsl:
Currently, If the import cache files have been cleared from an install, all the import commands stop working as they try to clear the cache themselves.


mauromsl:
test this please

PR #1211, opened 2019-07-15
Fix bug with saving JPGs with alpha channel

MartinPaulEve:
Fixes the JPG bug of not being able to save alpha channels


Bbkctp:

mauromsl:
add to whitelist

PR #1219, opened 2019-07-17
1190 review metadata

ajrbyers:
Closes #1190


src/templates/admin/review/review_requests.html
```diff
@@ -48,4 +48,8 @@ <h2>Completed Reviews</h2>
         </div>
     </div>
 
+    {% for request in new_requests %}
```
mauromsl: I think we should avoid overriding the name "request" here

PR #1242, opened 2019-08-07
931 workflow manager

ajrbyers:


src/workflow/logic.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def drop_workflow_log_entries(article, stages_to_process):
+    """
+    Finds and deletes workflow log entries for stages being processed.
+    :param article: Article object
+    :param stages_to_process: A List of stage names
+    :return: None
+    """
+    del stages_to_process[-1]
```
mauromsl: I don't think we want to mutate the list we've been passed.

mauromsl: > I don't think we want to mutate the list we've been passed.

Maybe something like this would be succinct and readable?:

`*stages_to_delete, new_current_stage = stages_to_process`

PR #1245, opened 2019-08-08
Analytics update

ajrbyers:


ajrbyers:
Geoip2 has been added to requirements but Jenkins failed to build it... (@mauromsl is there a way to get it to rebuild using the latest requirements?)

mauromsl:
@ajrbyers It already does that, and I can see geoip2 was installed during the build:
`10:36:13 Installing collected packages: django-bootstrap4, foundationform, django-mailgun, django-simple-math-captcha, django-summernote, djangorestframework, dynamicsites, ebooklib, django-materialize, django-compat, django-hijack, maxminddb, geoip2`

ajrbyers:
I can't explain what causes this error then, I don't see it locally after installing geoip2 with pip...

```
10:36:40   File "/vol/janeway/src/metrics/logic.py", line 10, in <module>
10:36:40     import geoip2.database
10:36:40 ImportError: No module named 'geoip2'
```

?

mauromsl:
Test this

mauromsl:
test this please

src/metrics/migrations/0006_auto_20190627_1412.py
```diff
@@ -0,0 +1,47 @@
...
+            access.country = country
+            access.save()
+
+            print(access.country)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('metrics', '0005_articleaccess_country'),
```
mauromsl: Should have a dependancy on `core` as well since it is using `core.Country`

src/metrics/migrations/0006_auto_20190627_1412.py
```diff
@@ -0,0 +1,47 @@
...
+
+    for access in accesses:
+        print(access.pk)
+        check = re.match(
+            r"^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}$",
+            access.identifier,
+        )
+
+        if check:
+            print('ip match')
```
mauromsl: I suppose we don't need these print statements

src/metrics/migrations/0006_auto_20190627_1412.py
ajrbyers: Core dependency is in 0005.

src/metrics/migrations/0005_articleaccess_country.py

src/metrics/migrations/0006_auto_20190627_1412.py
ajrbyers: Indeed, removed.

PR #1247, opened 2019-08-08
#578 completing the registration process now requires clicking a button.

### LINKED ISSUE #578: Confusing Error Message When Activating New Account
I create a new account, check using Admin rights that it has been created and is NOT active.
As the new user I click on the link in the registration email and get the following error message:
Error
There was no inactive account with this activation code found. It is possible that your account is already active, you can check by attempting to login.

https://demojournal.janeway.systems/register/step/2/fb66a174-65a4-41b7-b1e7-24e9ace8a773/

However, Admin rights confirm that the account is active, and as the user I can login successfully. This has happened on a number of occasions when registering a new account, although not every time. I assume that clicking on the link does actually activate the account but gives a confusing message to the end user,

ajrbyers:
Closes #578.

This update prevents browser pre-loading from triggering the completion of registration.

### DIFF
--- a/src/core/views.py
+++ b/src/core/views.py
@@ -307,18 +307,29 @@ def orcid_registration(request, token):
 
 def activate_account(request, token):
     """
-    Activates a user account if an Account object with the matching token is found and is not already active.
+    Activates a user account if an Account object with the
+    matching token is found and is not already active.
     :param request: HttpRequest object
     :param token: string, Account.confirmation_token
     :return: HttpResponse object
     """
     try:
         account = models.Account.objects.get(confirmation_code=token, is_active=False)
+    except models.Account.DoesNotExist:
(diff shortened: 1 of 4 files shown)


src/core/views.py
```diff
@@ -307,18 +307,29 @@ def orcid_registration(request, token):
...
     :param request: HttpRequest object
     :param token: string, Account.confirmation_token
     :return: HttpResponse object
     """
     try:
         account = models.Account.objects.get(confirmation_code=token, is_active=False)
+    except models.Account.DoesNotExist:
+        account = None
+
+    if account and request.POST:
```
mauromsl: Now that we are doing this, we could also add a LogEntry for this action, but I'll leave that up to you

Issue #576, opened 2018-10-29
The 'Request Revisions' button does work unless somethings is entered into the editor note box


ccaadeh:
https://demojournal.janeway.systems/review/article/582/revisions/request/
When you make request a revision:
The 'Request Revisions' button does not do anything unless some text is entered into the editor note box, although this does not appear to be a mandatory requirement like the setting of a date(?), which tells you that must enter some input.

ajrbyers:
Adding this to Trello, please ignore.

PR #1250, opened 2019-08-08
#576 marks the form as not requiring HTML5 validation and adds an err…

ajrbyers:
…ors block.

Closs #576


ajrbyers:
> Is there a reason not to make the form element optional instead of not validating the form at all?

Yes, we definitely want input from the editor here.

ajrbyers:
> Is there a reason not to make the form element optional instead of not validating the form at all?

Also, were not removing all validation, just the browser vidation. Django handles it perfectly.

PR #1253, opened 2019-08-09
#1127 when creating a correction task we now display a checkbox.

### LINKED ISSUE #1127: Import comments link should be a checkbox
There is a link on the request correction page `proofing/article_id/$proofing_task/$task_id/corrections/` that modifies the form on the backend to include the original comments from the proofreader task into the new typesetting task. This is done by setting a GET parameter so when the production manager clicks this link they will only see the page being refreshed.

We should change this for a checkbox on the form (and perhaps render a preview of the comments)

image

ajrbyers:
Also tidies up the view a little.

Closes #1127

### DIFF
--- a/src/proofing/views.py
+++ b/src/proofing/views.py
@@ -495,12 +495,8 @@ def request_typesetting_changes(request, article_id, proofing_task_id):
     """
     article = get_object_or_404(submission_models.Article, pk=article_id)
     proofing_task = get_object_or_404(models.ProofingTask, pk=proofing_task_id)
-
-    if request.GET.get('comments') == 'true':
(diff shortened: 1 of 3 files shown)


src/proofing/views.py
```diff
@@ -534,6 +537,7 @@ def request_typesetting_changes(request, article_id, proofing_task_id):
         'user': user if request.POST else None,
         'galleys': galleys if request.POST else None,
         'form': form,
+        'comments': comments,
```
mauromsl: Where is this used?

ajrbyers: It should be used in the template instead of calling it again, will update.

PR #1304, opened 2019-10-16
Added some validation to install_janeway.

### LINKED ISSUE #593: Journal Code Field
During install we shouldn’t accept a code longer than the field limit.

ajrbyers:
Closes #593 

Specifically chose to validate this manually as trying to validate through `save()` causes problems with the transaction.

### DIFF
--- a/src/utils/management/commands/install_janeway.py
+++ b/src/utils/management/commands/install_janeway.py
@@ -9,9 +9,6 @@ from django.utils import translation
 from press import models as press_models
 from journal import models as journal_models
 from utils.install import update_settings, update_license, update_issue_types
-from utils.logger import get_logger
-
-logger = get_logger(__name__)
 
 ROLES_RELATIVE_PATH = 'utils/install/roles.json'
 
@@ -40,7 +37,7 @@ class Command(BaseCommand):
         :return: None
         """
         call_command('migrate')
-        logger.info("Please answer the following questions.\n")
+        print("Please answer the following questions.\n")
         translation.activate('en')
         with transaction.atomic():
             test_one = press_models.Press.objects.all()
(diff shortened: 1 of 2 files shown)


src/utils/management/commands/install_janeway.py
```diff
@@ -38,7 +40,7 @@ def handle(self, *args, **options):
         :return: None
         """
         call_command('migrate')
-        print("Please answer the following questions.\n")
+        logger.info("Please answer the following questions.\n")
```
mauromsl: I think using print here (and in the ones below) was the right thing to do since this is a CLI program and we expect the user to read and react to these

src/utils/management/commands/install_janeway.py
```diff
@@ -49,9 +51,19 @@ def handle(self, *args, **options):
...
+
+            while True:
+                code = input('Journal #1 code (15 characters max): ')
+
+                if len(code) <= 15:
+                    journal.code = code
+                    journal.save()
+                    break
+                else:
+                    logger.warning('Jouranl code must be 15 characters or less.')
```
mauromsl: I see your point about the validation through the model within the transaction. Perhaps we should use a form instead and validate all the fields instead of hardcoding the logic here?

mauromsl: Typo -> Jouranl

ajrbyers: It is possible to fetch the field length with `Journal._meta.get_field('code').max_length`

Thoughts?

mauromsl: Its another option but only covers one form of validation for a single field, so we might have to add all the others. If we use a form then we can use the standard validation methods from django

PR #1333, opened 2019-11-07
Adds zip upload for galley files

### LINKED ISSUE #375: Allow upload of XML image files from zip
It would be really helpful for articles where we have masses of inline images to be able to upload these using a zip file.

ajrbyers:
- Adds a new view for uploading zip files of images
- View is accessible to Editors and Typesetters
- Closes #375


src/production/views.py
```diff
@@ -951,6 +951,87 @@ def edit_galley(request, galley_id, typeset_id=None, article_id=None):
...
+    galley = get_object_or_404(
+        core_models.Galley,
+        pk=galley_id,
+        article=article,
+    )
+
+    if request.POST and 'zip_file' in request.POST:
+        file = request.FILES.get('file')
+        try:
+            errors = logic.process_zip_file(file, galley, request)
```
mauromsl: This is a bit strange to see in python (not so much in go :P )

maybe have `process_zip_file` raise an exception at the end of processing the file and then:
```
Try:
    logic.process_zip_file(file, galley, request)
except BadZipFile:
   [...]
except FoundAdditionalFiles: # or some other name
   [...]
```

src/production/logic.py
```diff
@@ -278,3 +278,58 @@ def get_production_assign_content(user, request, article, url):
...
+            kwargs={'article_id': article.pk, 'galley_id': galley.pk},
+        )
+        redirect_url = '{url}{return_path}'.format(
+            url=url,
+            return_path=return_path,
+        )
+        return redirect(redirect_url)
+
+
+def process_zip_file(file, galley, request):
```
mauromsl: `def handle_zipped_galley_images:`?

src/production/logic.py
```diff
@@ -278,3 +278,58 @@ def get_production_assign_content(user, request, article, url):
...
+                galley.images.add(new_file)
+            else:
+                messages.add_message(
+                    request,
+                    messages.WARNING,
+                    'File in zip not in XML {}'.format(
+                        zipped_file.name
+                    )
+                )
+                return True
```
mauromsl: When this happens, the user will receive a warning and might try to re-upload an ammended Zip which would in turn create a lot of duplicates, perhaps we should update the galley images that are matched by name?

src/core/files.py
```diff
@@ -185,6 +185,46 @@ def save_file_to_article(file_to_handle, article, owner, label=None, description
     return new_file
 
 
+def save_zipped_file_to_article(zipped_file, article, user):
```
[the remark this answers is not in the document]
ajrbyers: I tried this and the way zipfiles work wouldn't allow it to work without refactoring the save_file_to_article into a confusing mess.

src/core/models.py
```diff
@@ -863,19 +863,25 @@ def has_missing_image_files(self):
...
-                        self.images.filter(original_filename=url).first()
-                except File.DoesNotExist:
+                if show_all:
                     missing_elements.append(url)
+                else:
+                    try:
+                        try:
+                            self.images.get(original_filename=url)
+                        except File.MultipleObjectsReturned:
+                            self.images.filter(original_filename=url).first()
```
mauromsl: This looks confusing, should we just call `self.images.filter(original_filename=url).first()` since it already does what `self.images.get(original_filename=url)` would do?

Maybe worth adding a comment explaining why there might be more than one and why we are fine getting the first result

src/core/models.py
```diff
@@ -863,19 +863,25 @@ def has_missing_image_files(self):
...
+                    except File.DoesNotExist:
```
mauromsl: This should be at the same level as the previous `except File.MultipleObjectsReturned:` instead of wrapping it

src/core/models.py
```diff
@@ -863,19 +863,25 @@ def has_missing_image_files(self):
...
+                        missing_elements.append(url)
 
         if not missing_elements:
             return []
         else:
             return missing_elements
 
+    def all_missing_images(self):
+        return self.has_missing_image_files(show_all=True)
```
mauromsl: I don't understand this bit, if I'm reading this correctly, it returns all images declared in the XML, not just the missing ones. Is that right? perhaps a comment would help.

src/core/models.py
ajrbyers: Dropping this as suggested above.

src/core/models.py
ajrbyers: Its a poorly worded name. It should probably be `all_images` as it returns all rather than thos missing.

PR #1336, opened 2019-11-11
Removes all CSS from uploaded HTML Galleys.

### LINKED ISSUE #473: HTML Galleys
We should strip any CSS links or <style> tags from HTML galleys.

mauromsl:
Closes #473 

@ajrbyers I've also removed inline css, although it wasn't explicitly mentioned on the original issue. Does that sound sensible?


ajrbyers:
@mauromsl yes I think that's sensible. I'll review the rest later tonight.

PR #1344, opened 2019-11-14
1322 most downloaded article

ajrbyers:


ajrbyers:
Closes #1344 1322

src/core/homepage_elements/featured/forms.py
```diff
@@ -0,0 +1,66 @@
...
+        most_downloaded, num_most_downloaded, most_downloaded_time = logic.get_popular_article_settings(
+            journal
+        )
+
+        self.fields['num_most_downloaded'].initial = num_most_downloaded
+        self.fields['most_downloaded'].initial = most_downloaded
+        self.fields['most_downloaded_time'].initial = most_downloaded_time
+
+    most_downloaded = forms.BooleanField(
+        label='Display Most Downloaded Articles',
```
mauromsl: Should this say, most popular or most visited since we are not filtering downloads but general accesses?
(same applies to the references to "downloads" below

src/core/homepage_elements/featured/forms.py
```diff
@@ -0,0 +1,66 @@
...
+        choices=(
+            ('weekly', 'Weekly'),
+            ('monthly', 'Monthly'),
+            ('yearly', 'Yearly'),
+        ),
+        label='Number of Most Downloaded Articles to Display',
+        help_text='Determines how many popular articles we should display.',
+    )
+
+    def save(self, journal, commit=True):
```
mauromsl: I don't think we should override the signature of `save` to take in a journal. Especially since you already expect a journal to be passed to the constructor

src/core/homepage_elements/featured/hooks.py
```diff
@@ -1,10 +1,30 @@
...
+    if homepage_elements is not None and homepage_elements.filter(
+            name='Featured Articles',
+    ).exists():
 
-        return {'featured_articles': featured_articles}
+        most_downloaded, number, time = logic.get_popular_article_settings(
+            request.journal
+        )
+
+        if most_downloaded:
```
mauromsl: this means we can't have both most downloaded and featured articles at the same time.
Perhaps the "most downloaded" should be a homepage element of its own, so you can choose to have, neither, one, the other or both

PR #1351, opened 2019-11-14
Wraps email popup in an IF. 

### LINKED ISSUE #1350: Metadata Block
**Describe the bug**
NoReverseMatch at /manage/archive/article/1/
Reverse for 'send_user_email' with arguments '('',)' not found. 1 pattern(s) tried: ['email/user/(?P<user_id>\\d+)/$']

Error is caused by a f_author not having an email address which should be handled with an IF.

**Janeway version**
1.3.6+

**To Reproduce**
1. Add a new author with no email
2. Error occurs

ajrbyers:
Closes #1350


ajrbyers:
Thanks to @hinefuku for reporting.

PR #1352, opened 2019-11-18
Crossref forward links

ajrbyers:


src/metrics/management/commands/fetch_forward_links.py
```diff
@@ -0,0 +1,143 @@
...
+            'crossref_prefix',
+            journal,
+        ).value
+
+        url = 'https://doi.crossref.org/servlet/getForwardLinks?usr={usr}&pwd={pwd}&doi={doi}&startDate={date}'.format(
+            usr=usr,
+            pwd=pwd,
+            doi=doi,
+            date=date
+        )
```
mauromsl: should use `urlencode` to avoid encoding errors:
https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode

src/metrics/management/commands/fetch_forward_links.py
```diff
@@ -0,0 +1,143 @@
...
+
+        print('Making request.', end='... ')
+        r = requests.get(url)
+        print('[ok]')
+
+        print('Souping response', end='... ')
+        soup = BeautifulSoup(r.text, 'lxml')
+        print('[ok]')
+
+        print('Finding forward links', end='...')
```
mauromsl: Very minor: No separator after `...`

src/metrics/management/commands/fetch_forward_links.py
```diff
@@ -0,0 +1,143 @@
...
+        forward_links = soup.find_all('forward_link')
+        print('[ok]')
+
+        print('Looping through links', end='...')
```
mauromsl: as above

src/metrics/management/commands/fetch_forward_links.py
```diff
@@ -0,0 +1,143 @@
...
+            article = sm.Article.get_article(
+                journal,
+                'doi',
+                doi,
+            )
+
+            if article:
+                if type == 'journal_article':
+                    process_article(link, article)
+                else:
```
mauromsl: I think we should be defensive here and handle books explicitely

src/metrics/models.py
```diff
@@ -55,6 +55,45 @@ def remove_one_download(self):
...
+        ('book', 'Book'),
+        ('article', 'Article'),
+    )
+
+
+class ForwardLink(models.Model):
+    article = models.ForeignKey('submission.Article', blank=True, null=True)
+    doi = models.CharField(max_length=255)
+    object_type = models.CharField(max_length=10, choices=object_types())
+    year = models.CharField(max_length=5)
```
mauromsl: This model should be declared as an abstract base (Its common to name them accordingly as well: `AbstractForwardLink`)
https://docs.djangoproject.com/en/1.11/topics/db/models/#abstract-base-classes

src/submission/models.py
```diff
@@ -1045,6 +1045,13 @@ def production_assignment_or_none(self):
         except ObjectDoesNotExist:
             return None
 
+    @property
+    def citations(self):
+        for citation in self.forwardlink_set.all():
+            print(citation.pk)
```
mauromsl: Looks like a debug print statement

src/submission/models.py
```diff
@@ -1045,6 +1045,13 @@ def production_assignment_or_none(self):
...
+
+        return self.forwardlink_set.all()
```
mauromsl: This is a Question: Does this actually work? I'd have thought you'd need to build a queryset of `BookLink` and `ArticleLink`, since I don't think querysets can mixed types

src/themes/OLH/templates/journal/article.html
```diff
@@ -161,7 +161,7 @@ <h4 class="number">
                                                 </div>
                                                 <div class="large-4 columns">
                                                     <h4 class="number">
-                                                        {{ article.metrics.alm.total }}<span>{% trans "Alt Metrics" %}</span>
+                                                        {{ article.citations|length }}<span>{% trans "Citations" %}</span>
```
mauromsl: another question: Does this work? docs suggest that it can handle lists and strings,plus IIRC querysets don't implement `__len__`, you have to call queryset.count() instead.

src/themes/default/templates/journal/article.html
```diff
@@ -171,6 +171,8 @@ <h4>{% trans "Metrics" %}</h4>
                             <li>{% trans "Wikipedia" %}: {{ article.metrics.alm.wikipedia }}</li>{% endif %}
                         {% if article.metrics.alm.reddit %}
                             <li>{% trans "Reddit" %}: {{ article.metrics.alm.reddit }}</li>{% endif %}
+                        {% if article.citations %}
+                            <li>{% trans "Citations" %}: {{ article.citations|length }}</li>{% endif %}
```
mauromsl: as with the other theme

src/themes/material/templates/journal/article.html
```diff
@@ -147,6 +147,14 @@ <h5>
                                         <p><i class="fa fa-reddit"></i> Reddit</p>
                                     </div>
                                 {% endif %}
+                                {% if article.citations %}
+                                    <div class="alm">
+                                        <h5>
+                                            {{ article.citations|length }}
```
mauromsl: as with the other theme

src/metrics/management/commands/fetch_forward_links.py
```diff
@@ -0,0 +1,143 @@
...
+        'issue': issue,
+        'journal_issn': issn,
+    }
+
+    models.ArticleLink.objects.get_or_create(
+        article=article,
+        doi=doi,
+        object_type='article',
+        defaults=defaults,
+    )
```
mauromsl: Might be worth also updating the object if any of the metadata attributes have changed

src/metrics/models.py
mauromsl: year shouldn't be an instance of  `models.CharField` I don't think :P

src/themes/OLH/templates/journal/article.html
ajrbyers: Yes it does work. This filter wraps len() around the passed var. Could change to count tho.

src/metrics/models.py
ajrbyers: I had this as a positive integer field but some of the data from crossref sucks so covering arses by using a charfield.

src/themes/OLH/templates/journal/article.html
mauromsl: > Yes it does work. This filter wraps len() around the passed var. Could change to count tho.

That's what I meant, Queryset doesn't implement '__len__' so `len()` won't work...

src/metrics/models.py
mauromsl: What other kinds of 'year' have you seen? I think we should instead not store the year if the format is not valid and log it

mauromsl: What else can be returned other than year? I think we should not store the year if the format is not valid.

src/themes/OLH/templates/journal/article.html
ajrbyers: QuerySets do.implement len. Anyway, swapping to .count() is recommended.

mauromsl: Oh I see, `len()` queries all the rows and then counts them instead of issuing a count query to the DB. Yeah, worth swapping for a `count()` still

src/metrics/management/commands/fetch_forward_links.py
ajrbyers: Yes, added that to a new issue as want to get this thru: https://github.com/BirkbeckCTP/janeway/issues/1353

src/submission/models.py
```diff
@@ -1045,6 +1045,13 @@ def production_assignment_or_none(self):
         except ObjectDoesNotExist:
             return None
 
+    @property
+    def citation_count(self):
+        article_link_count = self.articlelink_set.all().count()
```
mauromsl: Should be `self.articlelink_set.count()` but I think the ORM is smart enough to not fetch the rows and just execute the count

PR #1354, opened 2019-11-20
Added a new homepage element for Journals and HTML block.

### LINKED ISSUE #650: Make randomizable list of journals in Huddersfield-style
Hudds press looks great. We need to build the grid of journals into a randomizable format.

ajrbyers:
Closes #650

### DIFF
--- a/src/core/homepage_elements/journals/hooks.py
+++ b/src/core/homepage_elements/journals/hooks.py
@@ -14,7 +14,7 @@ def get_random_journals():
         hide_from_press=False,
     )
 
-    sample_size = 6 if journals.count() > 6 else journals.count()
+    sample_size = min(6, journals.count())
 
     return random.sample(set(journals), sample_size)
 
--- a/src/core/homepage_elements/journals/hooks.py
+++ b/src/core/homepage_elements/journals/hooks.py
@@ -10,18 +10,13 @@ from utils.function_cache import cache
 
 
 def get_random_journals():
-    journals = journal_models.Journal.objects.filter(hide_from_press=False)
-    journal_pks = [journal.pk for journal in journals]
-    random_journal_pks = list()
+    journals = journal_models.Journal.objects.filter(
+        hide_from_press=False,
+    )
 
-    random.shuffle(journal_pks)
-    for i in range(0, 6):
+    sample_size = 6 if journals.count() > 6 else journals.count()
 
-        if journal_pks:
-            choice = journal_pks.pop()
-            random_journal_pks.append(choice)
-
-    return journals.filter(pk__in=random_journal_pks)
+    return random.sample(set(journals), sample_size)
 
 
 def yield_homepage_element_context(request, homepage_elements):
--- a/src/core/homepage_elements/journals_and_html/hooks.py
+++ b/src/core/homepage_elements/journals_and_html/hooks.py
@@ -18,7 +18,7 @@ def get_random_journals():
         hide_from_press=False,
     )
 
-    sample_size = 6 if journals.count() > 6 else journals.count()
+    sample_size = min(6, journals.count())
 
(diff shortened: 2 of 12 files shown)


src/core/homepage_elements/journals_and_html/hooks.py
```diff
@@ -0,0 +1,55 @@
...
+    journals = journal_models.Journal.objects.filter(hide_from_press=False)
+    journal_pks = [journal.pk for journal in journals]
+    random_journal_pks = list()
+
+    random.shuffle(journal_pks)
+    for i in range(0, 6):
+
+        if journal_pks:
+            choice = journal_pks.pop()
+            random_journal_pks.append(choice)
```
mauromsl: I think you are re-inventing `random.sample` here:
https://docs.python.org/3/library/random.html#random.sample

src/core/homepage_elements/journals_and_html/plugin_settings.py
```diff
@@ -0,0 +1,50 @@
...
+            'yield_homepage_element_context': {
+                'module': 'core.homepage_elements.journals_and_html.hooks',
+                'function': 'yield_homepage_element_context',
+                'name': PLUGIN_NAME,
+            }
+        }
+    except OperationalError:
+        # if we get here the database hasn't yet been created
+        return {}
+    except BaseException:
```
mauromsl: We should never capture `BaseException` as this is the base type of `SystemExit` and can leave us with a hanging process

src/core/homepage_elements/journals_and_html/hooks.py
```diff
@@ -0,0 +1,55 @@
...
+
+def yield_homepage_element_context(request, homepage_elements):
+    if homepage_elements is not None and homepage_elements.filter(
+            name=plugin_settings.PLUGIN_NAME,
+    ).exists():
+
+        if request.press.random_featured_journals:
+            featured_journals = get_random_journals()
+        else:
+            featured_journals = request.press.featured_journals.all()
```
mauromsl: I don't understand this bit, isn't this the responsibility of the featured journals homepage element?

src/core/homepage_elements/journals_and_html/hooks.py
ajrbyers: random.sample does not work on querysets. It also doesn't work when the population is less than the sample.

src/core/homepage_elements/journals_and_html/hooks.py
ajrbyers: This HPE also uses featured journals in a side by side with an HTML block. Its an implementation of the hack for Huddersfield's homepage.

ajrbyers:
See comments.

src/core/homepage_elements/journals/hooks.py
```diff
@@ -10,18 +10,13 @@
...
-    journals = journal_models.Journal.objects.filter(hide_from_press=False)
-    journal_pks = [journal.pk for journal in journals]
-    random_journal_pks = list()
+    journals = journal_models.Journal.objects.filter(
+        hide_from_press=False,
+    )
 
-    random.shuffle(journal_pks)
-    for i in range(0, 6):
+    sample_size = 6 if journals.count() > 6 else journals.count()
```
mauromsl: Very minor: `min(6, journals.count())` to avoid the double call to `journals.count()`

src/core/homepage_elements/journals_and_html/plugin_settings.py
mauromsl: I think you removed a different `BaseException` :smile: 
It'd be best to replace it for `Exception` and log the exception out for now. Otherwise we might break current sites that are silently swallowing these exceptions.

PR #1314, opened 2019-10-28
650 randomised journal list

ajrbyers:
Closes #650

PR #1355, opened 2019-11-20
Small change to ensure HPEs don't try to get context for disabled ele…

ajrbyers:
…ments.

These changes were represented in #1314 but have been moved here. This could be taken further by also moving the calling of HPEs into the model or logic but this is enough to stop the errors from being raised for now.

### DIFF
--- a/src/press/views.py
+++ b/src/press/views.py
@@ -12,7 +12,12 @@ from django.contrib import messages
 from django.core.management import call_command
 from django.http import HttpResponse, Http404
 
-from core import files, models as core_models, plugin_loader
+from core import (
+    files,
+    models as core_models,
+    plugin_loader,
+    logic as core_logic,
+)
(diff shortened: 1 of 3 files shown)


src/press/views.py
```diff
@@ -42,12 +47,13 @@ def index(request):
 
     # call all registered plugin block hooks to get relevant contexts
     for hook in settings.PLUGIN_HOOKS.get('yield_homepage_element_context', []):
-        hook_module = plugin_loader.import_module(hook.get('module'))
-        function = getattr(hook_module, hook.get('function'))
-        element_context = function(request, homepage_elements)
+        if hook.get('name') in homepage_element_names:
```
mauromsl: This hides the problem rather than solving it, we should at least log a warning when 'name' is not in homepage_element_names alerting of the bad configuration

PR #1366, opened 2019-11-27
Improvements for the new XSL files feature

mauromsl:


src/production/logic.py
```diff
@@ -86,7 +86,7 @@ def save_galley(article, request, uploaded_file, is_galley, label=None, save_to_
             f.write(cleaned_html)
             f.truncate()
 
-    elif new_file.mime_type == files.XML_MIMETYPES:
```
ajrbyers: I noticed this one yesterday when I was trying to upload an XML file during the demo! Good catch.

PR #1385, opened 2020-01-03
Pr fix migration reference

Beethro:
There are two  0035 migrations for submission.
- remove the older one ( I assume that's the one that should go)
- correct the reference in 0038


ajrbyers:
Hi @Beethro I think this has been fixed elsewhere.

PR #1392, opened 2020-01-14
Clean theme

ajrbyers:
This adds a new theme called clean, based on Bootstrap that adheres more closely to web content accessibility guidelines. It still requires more work but I wanted to get it into master for the next release.


src/themes/default/templates/journal/print.html
```diff
@@ -7,6 +7,21 @@
...
+    <script type="text/x-mathjax-config">
+  MathJax.Hub.Config({
+    tex2jax: {
+      inlineMath: [ ['$','$'], ["\\(","\\)"] ],
+      processEscapes: true
+    }
+  });
+
+    </script>
+
```
mauromsl: seems unrelated, perhaps from another feature branch?

src/static/OLH/css/app.css
```diff
@@ -4650,6 +4650,9 @@ main section {
   main section.no-padding {
     padding: 0; }
 
+#article {
+  word-wrap: break-word; }
+
```
mauromsl: this changes the OLH theme, is it intended?

src/static/OLH/css/app.css
```diff
@@ -5250,3 +5253,11 @@ button.list-group-item-danger.active:focus {
...
 .feature-article-image {
   object-fit: cover; }
+
+.alignright {
+  float: right;
+  padding: 5px; }
+
+.alignleft {
+  float: left;
+  padding: 5px; }
```
mauromsl: as above

src/core/plugin_loader.py
```diff
@@ -114,9 +115,15 @@ def get_plugin(module_name, permissive):
 
         return plugin
     try:
+        print(module_name)
```
mauromsl: print statement

ajrbyers:
> Nice!
> Are we replacing the default theme with this or will we be supporting the 4 themes from now on?
> There are also a couple of changes from another issue it seems

I'm open to options. See as its built off of the default theme it might be worth considering it a replacement, though if someone is already using the default theme it might cause them some bother.

ajrbyers:
As discussed in Discord we will close this PR and move Clean to its own Repository.

PR #1398, opened 2020-01-16
767 keyword view

### LINKED ISSUE #767: List of Keywords and List of Articles by Keyword
Have a view to see all the keywords.
Another view to see a list of articles that the keyword appears in the model (not a text search) with links to those articles.
Ability to toggle the view via the dashboard.

ajrbyers:
Closes #767

### DIFF
--- a/src/journal/models.py
+++ b/src/journal/models.py
@@ -394,6 +394,7 @@ class Journal(AbstractSiteModel):
         except core_models.WorkflowElement.DoesNotExist:
             return False
 
+    @property
     def published_articles(self):
         return submission_models.Article.objects.filter(
             journal=self,
@@ -403,7 +404,7 @@ class Journal(AbstractSiteModel):
 
     def article_keywords(self):
         return submission_models.Keyword.objects.filter(
-            article__in=self.published_articles()
+            article__in=self.published_articles
         ).order_by('word')
 
 
--- a/src/journal/models.py
+++ b/src/journal/models.py
@@ -394,6 +394,18 @@ class Journal(AbstractSiteModel):
         except core_models.WorkflowElement.DoesNotExist:
             return False
 
+    def published_articles(self):
+        return submission_models.Article.objects.filter(
+            journal=self,
+            stage=submission_models.STAGE_PUBLISHED,
+            date_published__lte=timezone.now(),
+        )
+
+    def article_keywords(self):
+        return submission_models.Keyword.objects.filter(
+            article__in=self.published_articles()
+        ).order_by('word')
+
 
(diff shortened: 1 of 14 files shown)


src/journal/views.py
```diff
@@ -359,6 +359,50 @@ def print_article(request, identifier_type, identifier):
...
+@keyword_page_enabled
+def keyword(request, keyword_id):
+    """
+    Displays a list of articles that use a given keyword.
+    :param request: HttpRequest object
+    :param keyword_id: Keyword object PK
+    :return: a rendered template
+    """
+    keyword = get_object_or_404(submission_models.Keyword, pk=keyword_id)
+    articles = submission_models.Article.objects.filter(
```
mauromsl: Now it can be `request.journal.published_articles().filter(keywords__pk=keyword.pk)` to avoid repeating the published articles logic :)

src/themes/OLH/templates/journal/keyword.html
```diff
@@ -0,0 +1,24 @@
...
+        <div class="row">
+            <div class="large-8 columns border-right">
+                <h1>{% trans "Keyword" %}: {{ keyword.word }}</h1>
+                <p>
+                    <small><a href="{% url 'keywords' %}">{% trans 'Back to Keywords List' %}</a></small>
+                </p>
+                <p>{% trans "Articles that use this keyword are listed below." %}</p>
+
+                {% for article in articles %}
+                    {% include "elements/journal/box_article.html" %}
```
mauromsl: I'd be a bit more explicit just in case
`{% include "elements/journal/box_article.html" with article=article %}`

src/journal/models.py
```diff
@@ -394,6 +394,18 @@ def element_in_workflow(self, element_name):
         except core_models.WorkflowElement.DoesNotExist:
             return False
 
+    def published_articles(self):
```
mauromsl: sounds like a `@property` more than a method

src/security/test_security.py
```diff
@@ -3317,6 +3319,36 @@ def test_section_editor_production_no_or_bad_article_id(self):
...
+            self.press,
+        )
+
+        self.assertIsInstance(decorated_func(request), HttpResponseRedirect)
+
+    def test_loading_keyword_page_success(self):
+        func = Mock()
+        decorated_func = decorators.keyword_page_enabled(func)
+
+        setting_handler.save_setting('general', 'keyword_list_page', self.journal_one, 'on')
```
mauromsl: This alters the state of the test, it should be either reverted or mocked (if this test runs before `test_loading_keyword_page_fail` I think the test would fail)

PR #1404, opened 2020-01-17
copyeditor_files can now be blank on a form. closes #1397

### LINKED ISSUE #1397: Copyedit Admin Bug
**Describe the bug**
When editing a copyediting assignment in admin you are required to select a file.

**Janeway version**
1.3.7

ajrbyers:
closes #1397


src/copyediting/models.py
```diff
@@ -39,7 +39,7 @@ class CopyeditAssignment(models.Model):
     date_decided = models.DateTimeField(null=True, blank=True)
 
     files_for_copyediting = models.ManyToManyField('core.File', related_name='files_for_copyediting')
-    copyeditor_files = models.ManyToManyField('core.File')
+    copyeditor_files = models.ManyToManyField('core.File', blank=True)
```
mauromsl: Just a Q. Does the admin it let you save the form if it the field is not nullable but you post a blank field?

ajrbyers: Yeah, its stored in a secondary table so doesn't need to be nulled but the without blank the form forces you to select one.

mauromsl: weird, I didn't know this quirk about M2M fields

PR #1407, opened 2020-01-17
checks that galley is in html/xml list before rendering image block

### LINKED ISSUE #1405: Edit Galley
**Describe the bug**
A second block wrapped in an `{% if galley.has_missing_image_files %}` errors out when editing a PDF.

**Janeway version**
1.3.7

ajrbyers:
closes #1405


ajrbyers:
Fixed elsewhere.

PR #1410, opened 2020-01-21
1115 carousel issues

ajrbyers:
Closes https://github.com/BirkbeckCTP/janeway/issues/1115


ajrbyers:
Documentation led me to believe it can't be loaded by non admins but I will
double check and report back. I may also want to lock these behind editor
permissions.

PR #1429, opened 2020-02-13
Fixes signature on review acceptance ack email template

mauromsl:


src/utils/install/journal_defaults.json
```diff
@@ -206,7 +206,7 @@
             "type": "rich-text"
         },
```
ajrbyers: Should probably use reviewassignment.editor see: https://github.com/BirkbeckCTP/janeway/blob/4ae006429900b643786b6cc19a07b22bc73d3faa/src/review/models.py#L74

mauromsl: Ah nice! hadn't noticed the editor was part of that model

Issue #172, opened 2017-09-22
Funding Metadata


ajrbyers:

ajrbyers:
This will be closed by PR #1451

PR #1451, opened 2020-03-09
Feature plan s

MartinPaulEve:
This adds Fundref functionality to new articles. It does not completely close the Plan S issue but would recommend a partial merge


ajrbyers:
Closes #172

src/submission/migrations/0042_funder_funding_id.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
ajrbyers, line 1: This will require a migration merge.

PR #1483, opened 2020-03-26
Handle CMYK images

### LINKED ISSUE #1482: CMYK images break on upload: 'cannot write mode CMYK as PNG'
**Describe the bug**
When a user uploads a `jpg` image with `CMYK` color mode, a 500 error is raised.and the stored image file is corrupted
The error is raised during transformation in `core.logic.resize_and_crop` when saving the image as `png`

**Janeway version**
v1.3.7

**To Reproduce**
1. Go to edit an issue's large image
2. Upload a CMYK image
3. See error

**Expected behavior**
The image should be transformed correctly into PNG

```
File "/usr/local/lib/python3.5/site-packages/PIL/PngImagePlugin.py", line 802, in _save
    raise IOError("cannot write mode %s as PNG" % mode)
OSError: cannot write mode CMYK as PNG
```

mauromsl:
closes #1482

### DIFF
--- a/src/core/logic.py
+++ b/src/core/logic.py
@@ -111,8 +111,9 @@ def resize_and_crop(img_path, size, crop_type='middle'):
         img = img.crop(box)
     else:
         img = img.resize((size[0], size[1]), Image.ANTIALIAS)
-        if img.mode == "CMYK":
-            img = img.convert("RGB")
+
+    if img.mode == "CMYK":
+        img = img.convert("RGB")
 
     img.save(img_path, "png")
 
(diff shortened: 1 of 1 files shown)


src/core/logic.py
```diff
@@ -111,6 +111,8 @@ def resize_and_crop(img_path, size, crop_type='middle'):
         img = img.crop(box)
     else:
         img = img.resize((size[0], size[1]), Image.ANTIALIAS)
+        if img.mode == "CMYK":
```
ajrbyers: Should this be indented under the else here?

mauromsl: indeed

PR #1492, opened 2020-04-03
Feature hooks

ajrbyers:
- PEP8-ifys a few views that were pretty bad
- Adds `LANGUAGE_CODE` to global settings as this was breaking the Clean theme
- Fixed `core.models.BASE_ELEMENTS` and `core.models.WorkflowElement` handshake_url to correct: this should be the "list" page for the element (migrations added to fix the DB)
- Fixes `core.models.has_missing_image_files` for files that aren't XML.
- Added two methods to `core.models.WorkflowElement` that let us see what stages are part of a element and which articles are in those stages
- the hook tag in `core.templatetags.hooks` no takes `*args` and `**kwargs`
- added a new method `submission.models.Article.current_stage`
- Moved the editor block of the dashboard to its own block, updated it to render based on what elements are in `request.journal.workflow`
- Updated various checks on the dashboard to ensure they only display when their Workflow element is present in `request.journal.workflow`
- Updated left hand nav to render workflow based on request.journal.workflow`
- Fixed an admin link to proofing
- `utils.install.update_settings` now allows for a path to load other settings with a default `utils/install/journal_defaults.json`
- `utils.management.commands.install_plugins` now accepts an option argument `plugin_name` so you can fire the install process for a single plugin eg. `./manage.py install_plugins typesetting`


src/core/models.py
```diff
@@ -846,6 +846,9 @@ def render(self):
         return files.render_xml(self.file, self.article, xsl_file=self.xsl_file)
 
     def has_missing_image_files(self, show_all=False):
+        if not self.file.mime_type in files.MIMETYPES_WITH_FIGURES:
```
ajrbyers, line 849: This probably should be in its own PR.

src/submission/models.py
```diff
@@ -949,6 +952,8 @@ def current_stage_url(self):
             return reverse('proofing_article', kwargs=kwargs)
         elif self.stage == STAGE_READY_FOR_PUBLICATION:
             return reverse('publish_article', kwargs=kwargs)
+        else:
+            return reverse(self.current_stage.jump_url, kwargs=kwargs)
```
mauromsl, line 956: all of the above branches can go now, except for `STAGE_UNASSIGNED` ( we could use a comment explaining that there is no `workflow_element` for the unassigned stage.

src/submission/models.py
```diff
@@ -1022,6 +1027,15 @@ def workflow_stages(self):
         from core import models as core_models
         return core_models.WorkflowLog.objects.filter(article=self)
 
+    @property
+    def current_stage(self):
```
mauromsl, line 1031: rename to `current_workflow_element` for clarity?

src/utils/plugins.py
```diff
@@ -0,0 +1,50 @@
...
+    version = None
+    janeway_version = None
+
+    is_workflow_plugin = False
+    jump_url = None
+    handshake_url = None
+    article_pk_in_handshake_url = False
+    press_wide = False
+
+    kanban_card = '{plugin_name}/kanban_card.html'.format(
```
mauromsl, line 23: where is this used?

ajrbyers: It will be used by the Typesetting Plugin.

src/submission/migrations/0041_auto_20200203_1547.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('submission', '0040_article_projected_issue'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='article',
+            name='stage',
+            field=models.CharField(choices=[('Unsubmitted', 'Unsubmitted'), ('Unassigned', 'Unassigned'), ('Assigned', 'Assigned to Editor'), ('Under Review', 'Peer Review'), ('Under Revision', 'Revision'), ('Rejected', 'Rejected'), ('Accepted', 'Accepted'), ('Editor Copyediting', 'Editor Copyediting'), ('Author Copyediting', 'Author Copyediting'), ('Final Copyediting', 'Final Copyediting'), ('Typesetting', 'Typesetting'), ('Proofing', 'Proofing'), ('pre_publication', 'Pre Publication'), ('Published', 'Published'), ('preprint_review', 'Preprint Review'), ('preprint_published', 'Preprint Published'), ('Back Content', 'Back Content Plugin'), ('typesetting_plugin', 'Typesetting Plugin')], default='Unsubmitted', max_length=200),
```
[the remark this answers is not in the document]
ajrbyers: It will need to be part of the choice field it will show up as empty for any article in the TypesettingPlugin stage.

src/submission/migrations/0043_merge_20200314_1336.py
```diff
@@ -0,0 +1,16 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('submission', '0041_auto_20200203_1547'),
+        ('submission', '0042_auto_20200227_1051'),
```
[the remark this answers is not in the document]
ajrbyers: Perhaps even better, as its not being used by anyone I could delete and squash them all?

src/submission/migrations/0041_auto_20200203_1547.py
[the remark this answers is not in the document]
mauromsl: We need the `choices to be a callable that can consider stages from plugins then (while somehow not generating database migrations).

ajrbyers:
> Some comments inline
> 
> I think we are missing the dashboard hooks for the typesetter, production_manager, proofreader no? as well as the swimlanes on the kanban view.

Yes dashboard cards for typesetter and proofreader yes. Swimlanes are already done (handled by `card.html`. Will push an update soon.

src/submission/migrations/0041_auto_20200203_1547.py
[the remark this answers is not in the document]
ajrbyers: > We need the choices to be a callable that can consider stages from plugins then (while somehow not generating database migrations).

I'm not sure it's possible - it looks like django always evaluates the possible choices for a choice field. Perhaps we could make a validator and custom form element so its not viewed as a choice field by migrations?

ajrbyers:
Should be ready for re-review now.

src/core/workflow.py
```diff
@@ -208,3 +212,33 @@ def remove_element(request, journal_workflow, element):
...
+            'kanban_card': getattr(settings_module, 'KANBAN_CARD', ''),
+            'dashboard_template': getattr(
+                settings_module, 'DASHBOARD_TEMPLATE', ''
+            )
+        }
+
+    except (ImportError, KeyError) as e:
+        if settings.DEBUG:
+            print(e)
+        pass
```
mauromsl, line 240: `logger.error(e)`?

src/submission/models.py
```diff
@@ -298,6 +299,30 @@ def get_queryset(self):
...
+    def validate(self, value, model_instance):
+        """
+        Validates value and throws ValidationError.
+        """
+        try:
+            super().validate(value, model_instance)
+        except exceptions.ValidationError:
+            # Check if the value is in dynamic choices and remove the
+            # error message if it is
+            if 'invalid_choice' in self.error_messages and value in self.dynamic_choices:
```
mauromsl, line 322: Does this work? I think `value` is a string whilst `self.dynamic_choices` is a tuple of tuples

src/core/workflow.py
```diff
@@ -169,11 +169,15 @@ def articles_in_workflow_stages(request):
             workflow_list[element.element_name] = element_dict
         except (KeyError, AttributeError) as e:
             if settings.DEBUG:
```
mauromsl, line 171: we should log this in production as well, no?

src/submission/models.py
```diff
@@ -298,6 +299,34 @@ def get_queryset(self):
...
+
+    def validate(self, value, model_instance):
+        """
+        Validates value and throws ValidationError.
+        """
+        try:
+            super().validate(value, model_instance)
+        except exceptions.ValidationError as e:
+            # Check if the value is in dynamic choices and remove the
+            # error message if it is
```
mauromsl, line 321: This is slightly inaccurate now

PR #1508, opened 2020-04-16
Plugin link available to editors

ajrbyers:
Closes #1429

### DIFF
--- a/src/journal/views.py
+++ b/src/journal/views.py
@@ -132,9 +132,11 @@ def articles(request):
     pinned_articles = [pin.article for pin in models.PinnedArticle.objects.filter(
         journal=request.journal)]
     pinned_article_pks = [article.pk for article in pinned_articles]
-    article_objects = submission_models.Article.objects.filter(journal=request.journal,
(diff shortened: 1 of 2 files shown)


src/journal/views.py
```diff
@@ -132,9 +132,11 @@ def articles(request):
...
         journal=request.journal)]
     pinned_article_pks = [article.pk for article in pinned_articles]
-    article_objects = submission_models.Article.objects.filter(journal=request.journal,
-                                                               date_published__lte=timezone.now(),
-                                                               section__pk__in=filters).prefetch_related(
+    article_objects = submission_models.Article.objects.filter(
+        journal=request.journal,
+        stage=submission_models.STAGE_PUBLISHED,
+        date_published__lte=timezone.now(),
+        section__pk__in=filters).prefetch_related(
```
mauromsl, line 139: This seems to come from another local branch

ajrbyers: Yeah looks like https://github.com/BirkbeckCTP/janeway/pull/1507

PR #1517, opened 2020-04-23
1515 added a new view that returns XML if an article has an XML galley.

### LINKED ISSUE #1515: Add XML download URL
Add a url to allow download of associated XML or 404

ajrbyers:
Closes #1515

### DIFF
--- a/src/journal/views.py
+++ b/src/journal/views.py
@@ -1959,3 +1959,25 @@ def doi_redirect(request, identifier_type, identifier):
         raise Http404()
 
     return redirect(article_object.local_url)
+
+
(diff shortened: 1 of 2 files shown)


src/journal/views.py
```diff
@@ -1959,3 +1959,25 @@ def doi_redirect(request, identifier_type, identifier):
...
+        request.journal,
+        identifier_type,
+        identifier,
+    )
+
+    if not article_object:
+        raise Http404
+
+    galleys = article_object.galley_set.all()
+    xml_galley = get_object_or_404(
```
mauromsl, line 1975: We could potentially get a `MultipleObjectsReturned` exception here.

PR #1536, opened 2020-05-01
Updates Galley Upload

### LINKED ISSUE #1452: Add support for setting galley labels for "Other" galleys
When a galley that is not PDF or XML is created, the label is set to "Other" which results on the text "Download Other" being rendered on the article page. It should be possible to set a custom label instead. Additionally, labels can't currently be updated on the admin side either since there is a required "Type" field that might not list one's galley type

A field box on the upload galley modal to set the custom label for an "Other" galley.

N/A

ajrbyers:
- Galley upload now has only one upload field
- Galley upload also has a label field
- We attempt to work out the best label if none is supplied
closes #1452

### DIFF
--- a/src/production/logic.py
+++ b/src/production/logic.py
@@ -117,6 +117,8 @@ def save_galley(article, request, uploaded_file, is_galley, label=None, save_to_
         sequence=article.get_next_galley_sequence(),
     )
 
+    print(new_galley)
+
     return new_galley
 
 
--- a/src/production/logic.py
+++ b/src/production/logic.py
@@ -117,8 +117,6 @@ def save_galley(article, request, uploaded_file, is_galley, label=None, save_to_
         sequence=article.get_next_galley_sequence(),
     )
 
-    print(new_galley)
-
     return new_galley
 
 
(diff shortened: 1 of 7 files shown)


src/production/logic.py
```diff
@@ -87,17 +96,29 @@ def save_galley(article, request, uploaded_file, is_galley, label=None, save_to_
...
     new_galley = core_models.Galley.objects.create(
-        article=article, file=new_file, label=label, type=type, sequence=article.get_next_galley_sequence()
+        article=article,
+        file=new_file,
+        label=label,
+        type=type_,
+        sequence=article.get_next_galley_sequence(),
     )
 
+    print(new_galley)
```
mauromsl, line 120: print statment

src/production/views.py
```diff
@@ -291,42 +291,24 @@ def production_article(request, article_id):
     galleys = logic.get_all_galleys(production_assignment.article)
 
     if request.POST:
-
+        print(request.FILES)
```
mauromsl, line 294: print statement

ajrbyers:
Now also closes #1452

PR #1539, opened 2020-05-04
Added File.parent FK back.

ajrbyers:
Closes https://github.com/BirkbeckCTP/typesetting/issues/58


src/production/logic.py
```diff
@@ -145,13 +145,12 @@ def remove_css_from_html(source_html):
...
-        new_file.parent = galley.file
-        new_file.save()
-        galley.file = new_file
-        galley.save()
+
+        files.overwrite_file(
+            uploaded_file,
+            galley.file,
+            ('articles', article.pk),
+        )
```
mauromsl, line 153: from what I can see, `galley.file` would still be pointing at the wrong file (the old one)
Shouldnt't this be something like:
```
galley.file = files.overwrite file(...)
galley.save()
```

PR #1541, opened 2020-05-04
1540 ArticleAuthorOrder is now set on creation. On snapshot we use th…

### LINKED ISSUE #1540: Author Order Retention
**Describe the bug**
Author order is not being retained when Frozen Author records are being created.

**Janeway version**
1.3.7

ajrbyers:
…e default db insertion order rather than the M2M which was returning alpha sorting.

Closes #1540

### DIFF
--- a/src/submission/logic.py
+++ b/src/submission/logic.py
@@ -212,3 +212,19 @@ def order_fields(request, fields):
         order = ids.index(field.pk)
         field.order = order
         field.save()
+
+
+def save_author_order(request, article):
+    author_pks = [int(pk) for pk in request.POST.getlist('authors[]')]
+    print(author_pks)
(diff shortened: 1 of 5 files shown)


src/submission/logic.py
```diff
@@ -19,8 +19,15 @@
 def add_self_as_author(user, article):
     new_author = user
     article.authors.add(new_author)
-    models.ArticleAuthorOrder.objects.create(article=article, author=new_author)
-
+    models.ArticleAuthorOrder.objects.create(
+        article=article,
+        author=new_author,
+    )
```
mauromsl, line 25: I believe this should go away, unless I'm missing something here

ajrbyers:
test this please

PR #1560, opened 2020-05-17
Preprint remodel

ajrbyers:
This PR is a WIP for comment.

- I've removed article object managers as these are no longer required.
- Preprints has been remodelled to make it more or less independent
    - It still used submission.Keyword and submission.Licence as there is no reason not to continue using these.
- Added a new Repository model that will represent a Preprint silo eg. Humanities Preprints or English Literature Preprints

Two things to note:

1. I have not run `makemigrations` on this code yet, I'm not sure how it will work.
2. This breaks pretty much the whole workflow


src/preprint/models.py
```diff
@@ -3,24 +3,221 @@
...
+def width_choices():
+    return (
+        (3, '3'),
+        (6, '6'),
+        (9, '9'),
+        (12, '12'),
+    )
+
+
+fs_path = os.path.join(settings.BASE_DIR, 'files')
```
mauromsl, line 34: Should probably be a relative path to avoid generating migrations between different installs:
`fs_path = 'files/preprints'`

hardyoyo:
I've attempted to look at a few of these conflicts via command line Git by doing an interactive rebase of the preprint-remodel branch on top of the current master branch... after about 30 minutes, I've yet to see anything that looks serious. A much faster way to resolve these conflicts would be via the GitHub interface, but, I lack permissions to use the tool on this repository. I don't think there's any serious problem with this PR, you just need to rebase it on master, and resolve any minor conflicts. I can help with the .lando.yml file. Maybe a screenshare working session would help?

PR #1570, opened 2020-05-19
#1522 featured journals are now sorted into alpha order.

### LINKED ISSUE #1522: Press Home - Journals in Alpha
**Describe the bug**
Even when selecting a list of random journals they should probably be in alphabetical order.

**Janeway version**
1.3.7

ajrbyers:
Closes #1522

### DIFF
--- a/src/core/homepage_elements/journals/hooks.py
+++ b/src/core/homepage_elements/journals/hooks.py
@@ -27,6 +27,6 @@ def yield_homepage_element_context(request, homepage_elements):
         else:
             featured_journals = request.press.featured_journals.all()
 
-        return {'featured_journals': featured_journals}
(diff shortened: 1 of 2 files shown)


src/core/homepage_elements/journals/hooks.py
```diff
@@ -27,6 +27,6 @@ def yield_homepage_element_context(request, homepage_elements):
         else:
             featured_journals = request.press.featured_journals.all()
 
-        return {'featured_journals': featured_journals}
+        return {'featured_journals': sorted(featured_journals, key=lambda x: x.name)}
```
mauromsl, line 30: This is what I meant when we spoke about this:
https://docs.python.org/3/library/operator.html#operator.attrgetter

The lambda is fine too though

PR #1574, opened 2020-05-19
Adds custom how to cite

### LINKED ISSUE #1565: Custom "How to cite"
Add support for customizing the How to cite section

A free-form textbox where users can add their own custom "how to cite" block when the one generated from Janeway is not suitable. If such field is present in an article, we render it, otherwise we render the current auto-generated one.

Allow editors to use a regex/template to create their own patterns for the how to cite block. This might not be a very user friendly approach, perhaps something we can add later as a separate feature

mauromsl:
closes #1565


src/submission/models.py
```diff
@@ -436,6 +436,12 @@ class Article(models.Model):
     # Agreement, this field records the submission checklist that was present when this article was submitted.
     article_agreement = models.TextField(default='')
 
+    custom_how_to_cite = models.TextField(
+        blank=True, null=True,
+        help_text="Custom 'how to cite' text. To be used only if the block"
+            " generated by Janeway is not suitable."
```
ajrbyers, line 442: Missing comma here

src/submission/models.py
```diff
@@ -471,6 +477,32 @@ class Article(models.Model):
     class Meta:
         ordering = ('-date_published', 'title')
 
+    @property
+    def how_to_cite(self):
```
ajrbyers, line 481: Would it be easier/neater to use a template and render_to_string for this as we're generating HTML?

PR #1580, opened 2020-05-20
#1579 Tweaks how we fetch issue articles for some instances.

### LINKED ISSUE #1579: Issue Numbers
**Describe the bug**
In some cases we fetch issue.articles for public display rather than enforce publication checks.

**Janeway version**
1.3.7

ajrbyers:
Closes #1579 

Also includes a couple of tweaks or issue zip download.

### DIFF
--- a/src/core/files.py
+++ b/src/core/files.py
@@ -751,12 +751,15 @@ def zip_files(files, article_specific=False):
     os.makedirs(_dir, 0o775)
 
     for file in files:
-        if article_specific and file.article_id:
-            folder_name = '{id} - {title}'.format(id=file.article_id, title=strip_tags(file.article.title))
-            article_dir = os.path.join(_dir, folder_name)
-            if not os.path.exists(article_dir):
-                os.makedirs(article_dir, 0o775)
-            shutil.copy(file.self_article_path(), article_dir)
+        if file.article_id:
+            if article_specific:
+                folder_name = '{id} - {title}'.format(id=file.article_id, title=strip_tags(file.article.title))
+                article_dir = os.path.join(_dir, folder_name)
+                if not os.path.exists(article_dir):
+                    os.makedirs(article_dir, 0o775)
+                shutil.copy(file.self_article_path(), article_dir)
+            else:
+                shutil.copy(file.self_article_path(), _dir)
 
     zip_path = '{dir}.zip'.format(dir=_dir)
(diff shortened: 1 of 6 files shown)


src/core/files.py
```diff
@@ -751,15 +751,12 @@ def zip_files(files, article_specific=False):
...
     for file in files:
-
         if article_specific and file.article_id:
             folder_name = '{id} - {title}'.format(id=file.article_id, title=strip_tags(file.article.title))
             article_dir = os.path.join(_dir, folder_name)
             if not os.path.exists(article_dir):
                 os.makedirs(article_dir, 0o775)
             shutil.copy(file.self_article_path(), article_dir)
-        else:
-            shutil.copy(file.self_article_path(), _dir)
```
mauromsl, line 762: Doesn't this change render the `article_specific` kwarg useless? and maybe rename this to `zip_article_files` ?

src/core/files.py
```diff
@@ -751,12 +751,15 @@ def zip_files(files, article_specific=False):
     os.makedirs(_dir, 0o775)
 
     for file in files:
-        if article_specific and file.article_id:
-            folder_name = '{id} - {title}'.format(id=file.article_id, title=strip_tags(file.article.title))
-            article_dir = os.path.join(_dir, folder_name)
-            if not os.path.exists(article_dir):
-                os.makedirs(article_dir, 0o775)
-            shutil.copy(file.self_article_path(), article_dir)
+        if file.article_id:
```
mauromsl, line 754: I  think we should still rename this to `zip_article_files` for clarity. I call `zip_files` and pass it a list of non article files, it will return an empty zip, no? I would make it error if a file without an article_id is passed as well, but this might uncover hidden bugs at runtime

ajrbyers: > I think we should still rename this function to `zip_article_files` for clarity. I call `zip_files` and pass it a list of non article files, it will return an empty zip, no? I would make it error if a file without an article_id is passed as well, but this might uncover hidden bugs at runtime

This is also used for zip downloading review files which have article_ids though aren't loaded into article specific folders for the zip.

mauromsl: So those are also article files, the problem is that it will now ignore files without an article_id, but that is only clear after you follow all the if statements in the function, not at all clear from the name

Issue #1512, opened 2020-04-20
Display last review date


ajrbyers:
**Is your feature request related to a problem? Please describe.**
Editors don't want to contact users over and over.

**Describe the solution you'd like**
We should display the last time a user did a peer review on the assignment page. We should also update the roles page to display the user history link.

PR #1600, opened 2020-06-02
Completes work on #1512

mauromsl:


src/review/logic.py
```diff
@@ -74,28 +93,15 @@ def get_assignment_content(request, article, editor, assignment):
     return render_template.get_message_content(request, email_context, 'editor_assignment')
 
 
-def get_review_url(request, review_assignment):
```
ajrbyers, line 77: Not sure you meant to delete this?

PR #1643, opened 2020-06-19
#1642 Adds an option backend kwarg to ArticleInfo form.

### LINKED ISSUE #1642: Submission Configurator - Edit Metadata
**Describe the bug**
If a field is disabled in the submission config it still shows in Edit Metadata but cannot be changed.

**Janeway version**
1.3.8

**Expected behavior**
We should stop the ArticleInfo form from popping fields when editing in the backend but display a note to warn the user it is disabled on submission.

ajrbyers:
- Adds an optional boolean kwarg `backend` to `submission.forms.ArticleInfo`.
- If set to True this will stop the form from over-writing metadata with the defaults set in the submission configuration tool.

Closes #1642

### DIFF
--- a/src/submission/forms.py
+++ b/src/submission/forms.py
@@ -86,6 +86,7 @@ class ArticleInfo(KeywordModelForm):
         elements = kwargs.pop('additional_fields', None)
         submission_summary = kwargs.pop('submission_summary', None)
         journal = kwargs.pop('journal', None)
+        self.backend = kwargs.pop('backend', False)
 
         super(ArticleInfo, self).__init__(*args, **kwargs)
         if 'instance' in kwargs:
@@ -114,7 +115,7 @@ class ArticleInfo(KeywordModelForm):
                 self.fields['non_specialist_summary'].required = True
 
             # Pop fields based on journal.submissionconfiguration
-            if journal:
+            if journal and not self.backend:
                 if not journal.submissionconfiguration.subtitle:
                     self.fields.pop('subtitle')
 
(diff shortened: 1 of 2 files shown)


src/submission/forms.py
```diff
@@ -86,6 +86,7 @@ def __init__(self, *args, **kwargs):
         elements = kwargs.pop('additional_fields', None)
         submission_summary = kwargs.pop('submission_summary', None)
         journal = kwargs.pop('journal', None)
+        self.backend = kwargs.pop('backend', False)
```
mauromsl, line 89: I think this can just be a local variable instead of an instance attribute.
Can we add a docstring explaining what it means? Or maybe use a name more specific to the form instead of specific to the caller e.g (`pop_disabled_fields`)

src/submission/forms.py
```diff
@@ -316,8 +318,7 @@ class ConfiguratorForm(forms.ModelForm):
 
     def __init__(self, *args, **kwargs):
         super(ConfiguratorForm, self).__init__(*args, **kwargs)
-        self.fields[
-            'default_section'].queryset = models.Section.objects.filter(
+        self.fields['default_section'].queryset = models.Section.objects.language().fallbacks('en').filter(
```
mauromsl, line 321: should be settings.LANGUAGE_CODE instead of  `'en'`

src/submission/forms.py
ajrbyers: Setting it as an instance attribute means I can also use it in the save method (which we do). I can change the variable name and add a docstring though!

src/submission/forms.py
ajrbyers: Roger that.

Issue #175, opened 2017-09-22
Complete COUNTER Implementation


ajrbyers:

MartinPaulEve:
NB @ajrbyers to do this as a special favour for @martinpauleve who's in hospital and hates writing COUNTER implementations

ajrbyers:
Adding this to Trello, please ignore.

mauromsl:
BARF Notes: Recover the notes we had from Trello and bring this for a future BARF session.

PR #1645, opened 2020-06-23
#1644: Retry concurrent read/writes on Article Access

### LINKED ISSUE #1644: OperationalError: could not serialize access due to read/write dependencies among transactions
**Describe the bug**
1.3.8 resolved a concurrent update issue when logging article access records.
However it uncovered another problem: When an article is visited for the first time in some time (every 30 seconds) and then receives two or more concurrent requests, the transaction to create the new record will conflict with another transaction trying to update the previous record. This leads to a race condition in which, regardless of what transaction is comitted first, an OperationalError will be raised by the ORM.

Additionally, MariaDB doesn't support SKIP LOCKED while MySQL doesn't support NOWAIT. So ideally we want a solution that doesn't involve this syntax.

Worth mentioning that these errors only happen when using the serializable isolation level, since the underlying records state would have mutated by the time one of the concurrent transactions commits. 

**Janeway version**
v1.3.8

mauromsl:
This PR closes #1644 and resolves an incompatibility with MariaDB

Behaviour of this module will be fixed in #175

### DIFF
--- a/src/utils/decorators.py
+++ b/src/utils/decorators.py
@@ -1,0 +1,43 @@
+"""
+Utility decorators
+"""
+__copyright__ = "Copyright 2020 Birkbeck, University of London"
+__author__ = "Birkbeck Centre for Technology and Publishing"
+__license__ = "AGPL v3"
+__maintainer__ = "Birkbeck Centre for Technology and Publishing"
+
+import time
+from functools import wraps
+
+from utils.logger import get_logger
+
+
+logger = get_logger(__name__)
+
+
+def retry(attempts=3, throttle=0, exc=Exception):
+    """ Retries calling a function when certain exceptions are raised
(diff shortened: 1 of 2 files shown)


src/utils/decorators.py
```diff
@@ -0,0 +1,43 @@
...
+
+def retry(attempts=3, throttle=0, exc=Exception):
+    """ Retries calling a function when certain exceptions are raised
+    :param attempts: maximum number of times to retry
+    :param throttle_ms: optional throttling between calls in seconds
+    :param exc: An exception (or tuple of exceptions) on which to retry
+    """
+    def wrapper(f):
+        f_name = f.__name__
+        msg = "%s failed with %s: %s, Retrying in %d s..."
```
ajrbyers, line 26: Should probably use .format() here.

PR #1652, opened 2020-06-25
Fixed the carousel M2M widgets.

ajrbyers:
- Exposes translation JS as a URL
- Adds it to template

### DIFF
--- a/src/core/include_urls.py
+++ b/src/core/include_urls.py
@@ -17,11 +17,6 @@ from cms import views as cms_views
 from submission import views as submission_views
 from journal import views as journal_views
 
-js_info_dict = {
-    'domain': 'djangojs',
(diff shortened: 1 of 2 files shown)


src/core/include_urls.py
```diff
@@ -189,6 +190,8 @@
     url(r'^download/file/(?P<file_id>\d+)/$', journal_views.download_journal_file, name='journal_file'),
 
     url(r'^set-timezone/$', core_views.set_session_timezone, name='set_timezone'),
+
+    url(r'^jsi18n/$', JavaScriptCatalog.as_view(), name='javascript-catalog'),
```
mauromsl, line 194: Sounds like we should cache this view, since it will parse the translation files on every request.

Issue #1654, opened 2020-06-26
Dashboard server error (AttributeError: 'NoneType' object has no attribute 'element')


mauromsl:
**Describe the bug**
Some editors are getting a 500 server error when getting to the journal dashboard. 
Seems to be related to d9ebe492a8a56c34f72068389d8ce09e7341687e , in particular here: https://github.com/BirkbeckCTP/janeway/blob/2008f8152a6b2c9f4491c334a40b6cf6b61e66c8/src/submission/models.py#L1131

If there are no `WorkflowLog` entries for the article, this line throws an `AttributeError` which leads to the 500 on the dashboard page. I have only seen this happen to one article from Orbit that was imported in 2018, but It could potentially happen to articles imported into the workflow without following the regular submission path.

**Janeway version**
v1.3.8

PR #1657, opened 2020-06-26
#1654 changes how we handle lookups of elements and handles edge cases.

ajrbyers:
- Adds a map of stages to workflow elements for core and plugin elements.
- Hardcoded link added for published articles
- When an article has no stage and thus we cannot ascertain what stage the article would be in an error is displayed.

### DIFF
--- a/src/core/workflow.py
+++ b/src/core/workflow.py
@@ -23,6 +23,20 @@ ELEMENT_STAGES = {
     'prepublication': [submission_models.STAGE_READY_FOR_PUBLICATION]
 }
 
+STAGES_ELEMENTS = {
+    submission_models.STAGE_ASSIGNED: 'review',
+    submission_models.STAGE_UNDER_REVIEW: 'review',
+    submission_models.STAGE_UNDER_REVISION: 'review',
+
+    submission_models.STAGE_EDITOR_COPYEDITING: 'copyediting',
+    submission_models.STAGE_AUTHOR_COPYEDITING: 'copyediting',
+    submission_models.STAGE_FINAL_COPYEDITING: 'copyediting',
+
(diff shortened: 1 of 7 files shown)


src/core/workflow.py
```diff
@@ -23,6 +23,20 @@
...
+    submission_models.STAGE_UNDER_REVISION: 'review',
+
+    submission_models.STAGE_EDITOR_COPYEDITING: 'copyediting',
+    submission_models.STAGE_AUTHOR_COPYEDITING: 'copyediting',
+    submission_models.STAGE_FINAL_COPYEDITING: 'copyediting',
+
+    submission_models.STAGE_TYPESETTING: 'typesetting',
+    submission_models.STAGE_PROOFING: 'proofing',
+    submission_models.STAGE_READY_FOR_PUBLICATION: 'prepublication',
+}
```
mauromsl, line 38: We could generate this from ELEMENTS_STAGES to be a bit more DRY but I'll leave that as optional.

src/submission/models.py
```diff
@@ -1125,11 +1118,37 @@ def workflow_stages(self):
...
+            return reverse('review_unassigned_article', kwargs=kwargs)
+        elif self.stage == STAGE_PUBLISHED:
+            return reverse('manage_archive_article', kwargs=kwargs)
+        else:
+            element = self.current_workflow_element
+            if element:
+                return reverse(element.jump_url, kwargs=kwargs)
+            else:
+                logger.error(
+                    'Article #{} on Journal {} has no Stage.'.format(
```
mauromsl, line 1146: There are two different problems being obfuscated here:

- Article has no stage (can be ammended by a staff member)

- current_workflow_element returns None: Probably a bug that needs patching.
We should probably log that accordingly and tweak the nostage page as well

PR #1662, opened 2020-07-01
1612 submission only

### LINKED ISSUE #1612: Submission Only
- [x] Disable standard front end pages
- [x] Redirect to a version of the Submission Page that uses a backend template
- [x] Make it clear that this is a submission system
- [x] Link to the Journal URL

ajrbyers:
Closes #1612


src/journal/decorators.py
```diff
@@ -0,0 +1,23 @@
...
+    If journal.disable_front_end is enabled this decorator redirects to the
+    submission page.
+
+    :param func: the function to callback from the decorator
+    :return: either the function call or raises an Http404
+    """
+
+    def frontend_enabled_wrapper(request, *args, **kwargs):
+        if request.journal and not request.journal.disable_front_end:
+            return func(request, *args, **kwargs)
```
mauromsl, line 15: We should update the wrapper here. `functools.wraps` should do.

PR #1668, opened 2020-07-04
#1667 Updated core logic to handle error.

### LINKED ISSUE #1667: ImportError in core.logic.get_available_elements
**Describe the bug**
If a folder in `src/plugins` isn't an actual plugin this stop Janeway from running. Although this is unlikely to happen it was encountered today.

**Janeway version**
1.3.8

**To Reproduce**
1. Add a python module that isn't a Janeway plugin to `src/plugins`
2. Run server

**Expected behavior**
Janeway should catch the import error and log it.

ajrbyers:
- Adds a try/except block to `get_available_elements`
- Logs output of try/except
- Closes #1667

### DIFF
--- a/src/core/logic.py
+++ b/src/core/logic.py
@@ -29,6 +29,9 @@ from utils import render_template, notify_helpers, setting_handler
 from submission import models as submission_models
 from comms import models as comms_models
 from utils import shared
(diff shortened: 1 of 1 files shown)


mauromsl:
test this please (jenkins)

PR #1679, opened 2020-07-09
#1678 Fixes bug by always setting querysets.

### LINKED ISSUE #1678: Creating a new carousel
**Describe the bug**
Can't create a brand new Carousel. The form does not render.

**Janeway version**
1.3.8

**To Reproduce**
1. Delete a journal's carousel
2. Try to make a new one

**Expected behavior**
A carousel form should render.

ajrbyers:
Closes #1678

### DIFF
--- a/src/core/homepage_elements/carousel/forms.py
+++ b/src/core/homepage_elements/carousel/forms.py
@@ -48,7 +48,6 @@ class CarouselForm(forms.Form):
         request = kwargs.pop('request', None)
 
         article_list, news_list = self.load(request)
-        print(article_list, news_list)
 
         super(CarouselForm, self).__init__(*args, **kwargs)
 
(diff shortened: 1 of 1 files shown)


src/core/homepage_elements/carousel/forms.py
```diff
@@ -1,124 +1,145 @@
...
+            False,
+            attrs={'rows': '2'},
+        )
+    )
+
+    def __init__(self, *args, **kwargs):
+        request = kwargs.pop('request', None)
+
+        article_list, news_list = self.load(request)
+        print(article_list, news_list)
```
mauromsl, line 51: Print statement here

ajrbyers:
>  I didn't know one had to set an empty queryset for the form elements

You can pass an empty list normally but when you're using the widget we are it will cause an error so better to pass an empty QS.

Issue #1680, opened 2020-07-10
Author order breaks when publishing articles


hinefuku:
**Describe the bug**
When publishing articles, the author order reverts to alpha order by first name when publishing/scheduling publication. (This affects articles whose author order were changed to alpha by first name during submission).

Andy identified the issue as the call to article.snapshot_authors(article) in publish.

**Janeway version**
1.3.8

**To Reproduce**
1. In Pre-Publication, Edit Metadata and rearrange author order. Click Update Metadata.
2. Complete pre-publication and publish article.
3. Author order is incorrect on next screen.

**Expected behavior**
Author order retains changes made during metadata edits.

**Screenshots**
Initial author order:
<img width="767" alt="10369-authororder1" src="https://user-images.githubusercontent.com/45498398/87163711-69d58680-c28d-11ea-85f8-7e1259431f18.png">
Reordered author order:
<img width="745" alt="10369-authororder2" src="https://user-images.githubusercontent.com/45498398/87163739-735eee80-c28d-11ea-8eeb-9befd095e8e4.png">
Author order after publication:
<img width="729" alt="10369-authororder3" src="https://user-images.githubusercontent.com/45498398/87163783-807bdd80-c28d-11ea-84d8-aaf12bc574c9.png">

## Front-end Issues
If the issue is front-end specific please add the following details:

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #1681, opened 2020-07-10
#1680 fixes bug

ajrbyers:
- Reformats some code for PEP8
- When a FrozenAuthor is found we now set the frozen_record['order'] variable to be the FAs current order to maintain order.

### DIFF
--- a/src/core/models.py
+++ b/src/core/models.py
@@ -400,7 +400,10 @@ class Account(AbstractBaseUser, PermissionsMixin):
 
     def snapshot_self(self, article):
         try:
-            order = submission_models.ArticleAuthorOrder.objects.get(article=article, author=self).order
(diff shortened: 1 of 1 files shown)


src/core/models.py
```diff
@@ -419,14 +422,18 @@ def snapshot_self(self, article):
 
         if frozen_author:
             for k, v in frozen_dict.items():
+                frozen_dict['order'] = frozen_author.order
```
mauromsl, line 425: This should be done outside of the loop, right after `if frozen_author:` above(unless I'm missing something)
It could also use a comment (if my understanding is correct, we want to prioritize `frozen_author.order` over `ArticleAuthorOrder` because the reordering tool will, after submission is completed, save the ordering to `frozen_author.order` only and not `ArticleAuthorOrder`)

PR #1699, opened 2020-07-17
The submit button no longer displays for journals that have disabled …

ajrbyers:
…submission.


src/themes/OLH/templates/press/press_journals.html
```diff
@@ -14,6 +14,8 @@
         <div class="row">
             <div class="large-8 columns">
                 {% for current_journal in journals %}
+                    {% setting_var current_journal 'disable_journal_submission' as submission_disabled %}
+                    {{ submission_disabled }}
```
mauromsl, line 18: Looks like a debug statement.

PR #1703, opened 2020-07-20
Allow longer article titles

mauromsl:
Some articles imported from OJS recently have titles longer than 300+ characters. They use the TEXT type so they could potentially be much longer than this but I feel like 999 is more than enough for now


mauromsl:
> @mauromsl can you test this is ok on MySQL if you haven't already?

yeah its fine, The limit on MySQL would be around 21k with the settings we recommend

PR #1714, opened 2020-07-21
#1707 Swapped LogEntry.to for a model ToAddress

### LINKED ISSUE #1707: Error creating log entry with large number of TO addresses
**Describe the bug**
When sending an email with too many addresses in the TO field an error is generated as the LogEntry to field is not long enough.

**Janeway version**
1.3.8-RC8

```
'to': ['tech@openlibhums.org', 'maj52@aber.ac.uk', 'nea4@aber.ac.uk', 'olliebevington@gmail.com', 'martin@martineve.com', 'martin.eve@openlibhums.org', 'caroline.edwards@openlibhums.org', 'chris.richards@press.wales.ac.uk', 's.lewis@press.wales.ac.uk', 's.chapman@press.wales.ac.uk', 'andy.byers@openlibhums.org', 'tech+editor@openlibhums.org', 's.chapman@wales.ac.uk', 'diana.wallace@southwales.ac.uk']```

### PR DESCRIPTION
**Pull request by ajrbyers**
Closes #1707

### CODE REVIEW

File: src/utils/migrations/0017_auto_20200721_1421.py [hunk 1]
```diff
@@ -0,0 +1,50 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.29 on 2020-07-21 12:07
+from __future__ import unicode_literals
+import re
+
+from django.db import migrations, transaction
+
+def parse_email_addresses(entry):
+    match = re.findall(r'[\w\.-]+@[\w\.-]+\.\w+', entry.to)
```
mauromsl, line 9: The good old problem of matching all email addresses by regex...
I'd just add a `\+` to the allowed characters in the local part

src/utils/migrations/0017_auto_20200721_1421.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def move_to_m2m(apps, schema_editor):
+    LogEntry = apps.get_model('utils', 'LogEntry')
+    ToAddress = apps.get_model('utils', 'ToAddress')
+
+    email_log_entries = LogEntry.objects.filter(to__isnull=False)
+
+    for entry in email_log_entries:
+        with transaction.atomic():
```
mauromsl, line 20: Docs suggest that the `Migration` class has to be flagged as not atomic in order to execute parts of the migration within a transaction of their own.
https://docs.djangoproject.com/en/1.11/howto/writing-migrations/#non-atomic-migrations

That said, I don't understand why we run a transaction for each entry, is the length of the table problematic? If so, should we do the same in `move_to_strings`?

PR #1739, opened 2020-07-30
Revise object-related text for repository pages.

### LINKED ISSUE #1738: Potentially wrong indefinite article in use for submission form on the repository home page
**Describe the bug**
If one defines a repository with an object that begins with a vowel (like, say "article"), the repository home page template in the default material theme contains a hard coded indefinite article "a" which results in a submission button that reads: "Submit a article"

**Janeway version**
The preprint-remodel branch, current head at dd20ec71f

**To Reproduce**
1. configure a repository with an object of "article/articles"
2. visit the home page of the repository
3. see the submission button that reads "Submit a article"

**Expected behavior**
A grammatically correct home page.

**Possible solutions**
Change the button text to always read "Start New Submission" or use the plural form of the repository object "Submit articles" or add some logic to select the correct indefinite article (which would be complicated for translators to handle).

hardyoyo:
- Revise the submit button text on the repository home page to avoid
using a potentially wrong indefinite article
- fixes #1738

### DIFF
--- a/src/static/common/css/repo.css
+++ b/src/static/common/css/repo.css
@@ -19,7 +19,6 @@
 
 nav ul a {
     color: #616161 !important;
-    text-transform: capitalize;
 }
 
 nav ul li {
--- a/src/static/common/css/repo.css
+++ b/src/static/common/css/repo.css
@@ -19,6 +19,7 @@
 
 nav ul a {
     color: #616161 !important;
+    text-transform: capitalize;
 }
 
 nav ul li {
--- a/src/themes/material/templates/repository/home.html
+++ b/src/themes/material/templates/repository/home.html
@@ -22,7 +22,7 @@
 
             <div class="center spacer">
                 <a class="btn waves-effect waves-light"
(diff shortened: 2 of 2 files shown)


src/static/common/css/repo.css
```diff
@@ -19,6 +19,7 @@
 
 nav ul a {
     color: #616161 !important;
+    text-transform: capitalize;
```
mauromsl, line 22: For custom CMS pages, this will force nav items like "Become a Reviewer" to be rendered as "Become A Reviewer".

I think its best to be more explicit here and use a filter on the specific nav item we want to fix which I believe is this one:
https://github.com/BirkbeckCTP/janeway/blob/19036ec3717d3440330f5fff20638bc712e780dd/src/themes/material/templates/repository/nav.html#L20

We can use `capfirst`:
https://docs.djangoproject.com/en/1.11/ref/templates/builtins/#capfirst

hardyoyo: Fair enough, this fix isn't even part of #1738, so I'll remove it from this PR, it can be its own ticket and wait for a future PR.

Issue #1590, opened 2020-05-27
Preprints - Links to external supplementary files


mauromsl:
**Is your feature request related to a problem? Please describe.**
The preprint system should allow authors to add externaly hosted supplementary files (e.g.: Driad)

**Describe the solution you'd like**
An interface for adding a link to an external supplementary file that is then related to the submission

**Additional context**
Not sure if we would need versioning support here.

tingletech:
I'll take a look at this one

ajrbyers:
What remains on #1789 to finish this?

tingletech:
`/repository/dashboard/1010/` pages 
`/repository/manager/1010/edit/` pages

PR #1749, opened 2020-08-03
WIP start of external supplementary files

tingletech:
Preprints - Links to external supplementary files #1590

I didn't get to far on this, but I wanted to get some feedback to see if this looks like it is in the right direction.

### DIFF
--- a/src/repository/models.py
+++ b/src/repository/models.py
@@ -281,6 +281,12 @@ class Preprint(models.Model):
         null=True,
         on_delete=models.SET_NULL,
     )
+    submission_supplementary_files = models.ManyToManyField(
+        'PreprintSupplementaryFile',
+        related_name='submission_supplementary_files',
+        blank=True,
+        null=True,
+    )
     meta_image = models.ImageField(
         blank=True,
         null=True,
@@ -531,6 +537,11 @@ class PreprintFile(models.Model):
             'repository_download_file',
             kwargs=self.reverse_kwargs(),
         )
+
+
+class PreprintSupplementaryFile(models.Model):
+    preprint = models.ForeignKey(Preprint)
+    url = models.URLField()
 
 
 class PreprintAccess(models.Model):


Bbkctp:

src/repository/models.py
```diff
@@ -281,6 +281,12 @@ class Preprint(models.Model):
         null=True,
         on_delete=models.SET_NULL,
     )
+    submission_supplementary_files = models.ManyToManyField(
```
mauromsl, line 284: I think a one-to-many relationship works best here, even if it leads to some duplication across multiple articles. Some examples:
- removing a supplementary file from a preprint: We need to check all the pointers from the supplementary file or risk having orphan supplementary files.
- Ordering supplementary files: If we want the authors to re-order the supplementary files, we can do so with an attribute on the `SupplementaryPreprintFile` model. With a M2M we would need an intermediate model that holds this attribute.

src/repository/models.py
```diff
@@ -281,6 +281,12 @@ class Preprint(models.Model):
...
+        'PreprintSupplementaryFile',
+        related_name='submission_supplementary_files',
+        blank=True,
+        null=True,
```
mauromsl, line 288: If we end up going for a M2M, we don't need to set `null=True`

src/repository/models.py
```diff
@@ -533,6 +539,11 @@ def download_url(self):
         )
 
 
+class PreprintSupplementaryFile(models.Model):
+    preprint = models.ForeignKey(Preprint)
```
mauromsl, line 543: This would do the job if we go for a one-to-many relatioship. It's not needed on a M2M, since django will create the intermediate table and keys automatically

tingletech:
yea, you are right -- a preprint can have multiple supplementary files potentially -- but two different preprints probably would very rarely share the same supplementary files.

PR #1751, opened 2020-08-04
Marks publisher notes as safe.

ajrbyers:
- Publisher notes can now handle HTML.


src/themes/material/templates/journal/article.html
```diff
@@ -1,4 +1,4 @@
-{% extends "core/base.html" %}
+\\wsl$\Ubuntu-20.04\home\ajrbyers\code\janeway\src\themes\OLH\templates\journal\article.html{% extends "core/base.html" %}
```
mauromsl, line 1: this looks like some WSL wizardry

PR #1759, opened 2020-08-07
UP Importer fixes for PIA

mauromsl:
Importer fixes:
 - Thumbnail url might not match article id, which leads to either a missing thumbnail or in the worst case a wrong one
 - Adds support for alphanumeric issue numbers (e.g. Issue "S1"). This involves changing the `Issue.issue` into a CharField which sounds scary but works quite well actually since there is only one case of integer operations applied to this attribute
 - Handles fake/place holder email addresses by freezing author metadata from the source data rather than from the account (similar to what we did in the new OJS importer).
 
New feature:
 - new `scrape_up_all` command: The OAI feed from UP is unreliable, there are lots of contents missing and sometimes pointers to non-standard article urls which don't have all the metadata we need. We already have a command to scrape an individual article and another to scrape the issue archive for sorting the articles. I've put 1 + 1 together and created a new command that will import the articles from the urls found in the issue archive, which is better at article discovery than the OAI feed...


src/utils/management/commands/scrape_up_all.py
```diff
@@ -0,0 +1,40 @@
...
+        :return: None
+        """
+        parser.add_argument('url')
+        parser.add_argument('journal_code')
+        parser.add_argument('user_id')
+        parser.add_argument('--delete',
+                            action='store_true',
+                            dest='delete',
+                            default=1,
+                            help='Delete all articles and non-superusers in the database before import')
```
ajrbyers, line 24: comma and new line here?

src/utils/importers/up.py
```diff
@@ -319,7 +320,15 @@ def map_review_recommendation(recommentdation):
...
 
 
-def import_issue_images(journal, user, url):
+def import_issue_images(journal, user, url, import_missing=False):
+    """ Imports all issue images and other issue related content
+    Currently also reorders all issues, articles and sections within issues,
+    article thumbnails and issue titles.
+    :param journal: a journal.models.Journal
+    :param user: the owner of the imported content as a core.models.Account
+    :url: the base url of the journal to import from
```
ajrbyers, line 329: :param url: ?

ajrbyers (commented):
Two comments, nothing major.

PR #1766, opened 2020-08-12
1633 search feature

ajrbyers:
Kudos to @hardyoyo for getting this started.

### DIFF
--- a/src/repository/views.py
+++ b/src/repository/views.py
@@ -207,7 +207,7 @@ def repository_list(request, subject_slug=None):
     return render(request, template, context)
 
 
-def preprint_search(request, search_term=None):
+def repository_search(request, search_term=None):
     """
     Searches for and displays a list of Preprints.
     """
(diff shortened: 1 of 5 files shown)


src/repository/views.py
```diff
@@ -207,51 +207,61 @@ def repository_list(request, subject_slug=None):
...
-
-        from_author = core_models.Account.objects.filter(
-            (Q(first_name__in=split_search_term) |
-             Q(last_name__in=split_search_term) |
-             institution_query)
+        # Create a filter for author affiliation using the split term to check
+        # if each individual word is present.
+        affiliation_query = reduce(
+            operator.and_,
+            (Q(author__affiliation__icontains=x) for x in split_search_term)
```
mauromsl, line 244: If I search for "University of Never-land" this is likely to return results for preprints from authors affiliated with:
    "Trump University",
    "League of Justice"

Makes me wonder if the split term search would be at all useful (on top of the search overhead for search terms with many words)

It does make more sense to keep it for author names and keywords though

PR #1793, opened 2020-08-25
Login hotfix duex

ajrbyers:


src/core/forms.py
```diff
@@ -133,7 +133,6 @@ def clean_password_2(self):
 
     def save(self, commit=True):
         user = super(RegistrationForm, self).save(commit=False)
-        user.username = self.cleaned_data['email']
```
mauromsl, line 136: This should do it. Can you add a test for the registration view with mixed case email addresses?
I added one for the `core_add_user` view but it would be best to have one for registration as well for peace of mind:

https://github.com/BirkbeckCTP/janeway/blob/1352418743ec8c51022397551390fc3de32d371a/src/core/tests/test_app.py#L47

mauromsl (approved):
Looks good!
Worth adding a test for registration/login of mixed case email addresses as inidicated inline

ajrbyers:
test this please

PR #1815, opened 2020-09-22
[Preprints]: sitemap refactored like press.index

tingletech:
seems to work on my test instance


hardyoyo (approved):
👍 looks good, works on our main dev instance.

PR #1841, opened 2020-10-02
Repository manager fix

### LINKED ISSUE #1840: [Preprints] Assigning a Repo Manager has no effect
**Describe the bug**
As above

**Janeway version**
preprint-remodel

**To Reproduce**
Assign a repo manager
Login as them
Try to do a manager task

ajrbyers:
Closes #1840

### DIFF
--- a/src/core/models.py
+++ b/src/core/models.py
@@ -392,8 +392,8 @@ class Account(AbstractBaseUser, PermissionsMixin):
     def is_proofing_manager(self, request):
         return self.check_role(request.journal, 'proofing_manager')
 
-    def is_repository_manager(self, request):
-        if self in request.repository.managers.all():
+    def is_repository_manager(self, repository):
+        if self in repository.managers.all():
             return True
 
         return False
--- a/src/core/models.py
+++ b/src/core/models.py
@@ -392,6 +392,12 @@ class Account(AbstractBaseUser, PermissionsMixin):
     def is_proofing_manager(self, request):
         return self.check_role(request.journal, 'proofing_manager')
 
+    def is_repository_manager(self, request):
+        if self in request.repository.managers.all():
+            return True
+
+        return False
+
     def is_preprint_editor(self, request):
         if self in request.press.preprint_editors():
             return True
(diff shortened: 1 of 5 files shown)


src/core/models.py
```diff
@@ -392,6 +392,12 @@ def is_typesetter(self, request):
     def is_proofing_manager(self, request):
         return self.check_role(request.journal, 'proofing_manager')
 
+    def is_repository_manager(self, request):
```
mauromsl, line 395: I'd change `request` for `repository` since that is what we actually want

src/repository/logic.py
```diff
@@ -354,7 +354,7 @@ def get_unpublished_preprints(request):
         'preprintauthor_set'
     )
 
-    if request.user.is_staff:
+    if request.user.is_staff or request.user in request.repository.managers.all():
```
mauromsl, line 357: I think would be nice to use the function you just added above:
`if request.user.is_staff or request.user.is_repository_manager(request.repository)`

src/repository/logic.py
```diff
@@ -367,7 +367,7 @@ def get_published_preprints(request):
...
```
mauromsl, line 370: I think would be nice to use the function you just added above:
`if request.user.is_staff or request.user.is_repository_manager(request.repository)`

src/security/decorators.py
```diff
@@ -1007,6 +1007,9 @@ def wrapper(request, *args, **kwargs):
         if request.user in preprint.subject_editors():
             return func(request, *args, **kwargs)
 
+        if request.user in request.repository.managers.all():
```
mauromsl, line 1010: I think would be nice to use the function you just added above:
`if request.user.is_staff or request.user.is_repository_manager(request.repository)`

src/security/templatetags/securitytags.py
```diff
@@ -69,6 +69,12 @@ def is_author(context):
     return request.user.is_author(request)
 
 
+@register.simple_tag(takes_context=True)
+def is_repository_manager(context):
+    request = context['request']
+    return request.user.is_repository_manager(request)
```
mauromsl, line 75: If you make the change I proposed earlier, then we need:
`return request.user.is_repository_manager(request.repository)`

mauromsl (commented):
Looks good, just a couple of suggestions

PR #1846, opened 2020-10-05
1842 admi dash load

ajrbyers:
This PR:

- Removes the First Author from New and Published Preprints
- Removes the Views and Downloads from Published Preprints
- Reduces the number of queries from 2480 to 15.

The solution is just a suggestion, instead we could limit the number of Published preprints to say 100 and add a search option.

Comments welcome @justingonder @hardyoyo @tingletech

### DIFF
--- a/src/repository/logic.py
+++ b/src/repository/logic.py
@@ -334,17 +334,17 @@ def delete_file(request, preprint):
         )
 
 
-def subject_article_pks(request):
+def subject_article_pks(user):
     prepint_pks = []
     # TODO: Update this implementation
-    for subject in request.user.preprint_subjects():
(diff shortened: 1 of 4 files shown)


ajrbyers:
Suggestion from @justingonder:

- Limit the published articles to a reasonable amount - most likely the latest 50?
- Add a search function where Mods can search by Title and Author name like the standard search
- Reinstate the author first name

hardyoyo:
Those suggestions sound great. I'd consider this existing PR a good incremental step, and the suggestions from @justingonder a great place to end up. Maybe merge this now, and then do a new PR with the additional changes?

ajrbyers:
@hardyoyo agreed.

PR #1847, opened 2020-10-05
Adds colour coded status to review home page.

ajrbyers:


src/templates/admin/review/home.html
```diff
@@ -25,24 +25,37 @@ <h2>Articles in Review</h2>
                         <th>Editors</th>
                         <th>Section</th>
                         <th>Projected Issue</th>
-                        <th>No. Active Reviews</th>
+                        <th>Review Status</th>
```
mauromsl, line 28: When we talked about this we said we would add a setting IIRC. I'm happy for it to be the default behaviour, but then we should probably remove the next column (No. Complete Reviews) since its a bit redundant.

Otherwise, add a setting and show either column depending on the setting value?

src/utils/shared.py
```diff
@@ -69,3 +69,10 @@ def yes_or_no(question):
             return True
         if reply[0] == 'n':
             return False
+
+
+def day_month(date):
+    return '{}-{}'.format(
+        date.day,
+        date.month,
```
mauromsl, line 77: Should probably use `.strftime` instead:
`return date.strftime("%d-%m")`

Also, worth rendering month names for the benefit of our fellow American friends?
`return date.strftime("%d-%B")`

src/static/admin/css/admin.css
```diff
@@ -462,4 +466,12 @@ th {
 
 .file p {
     word-break: break-all;
+}
+
+.no-pad td {
+    padding: 0.1rem;
```
mauromsl, line 472: Maybe we can use a better name than `no-pad` when we are adding some padding

mauromsl (approved):

src/core/logic.py
```diff
@@ -251,6 +251,10 @@ def get_settings_to_edit(group, journal):
                 'name': 'enable_one_click_access',
                 'object': setting_handler.get_setting('general', 'enable_one_click_access', journal),
             },
+            {
+                'name': 'enable_one_click_access',
```
ajrbyers, line 255: Need to update the setting name here.

ajrbyers (commented):
Tiny comment.

Issue #1851, opened 2020-10-07
Easy Order Review Form Elements


ajrbyers:
**Is your feature request related to a problem? Please describe.**
Reordering review form elements is hard.

**Describe the solution you'd like**
Editors should be able to drag and drop order elements.

PR #1853, opened 2020-10-07
#1851 adds drag and drop ordering for review form elements

ajrbyers:
This PR:

- Adds a generic function for reordering objects (if approved a new Issue to be created to convert others to use it)
- Adds a generic sortable html template for inclusion in `block_js`
- Adds sortable stuff to edit review form view
- Adds default ordering to the `ReviewFormElement` model


src/templates/admin/core/kanban.html
```diff
@@ -1,5 +1,5 @@
 {% extends "admin/core/base.html" %}
-{% load static from staticfiles %}
```
mauromsl, line 2: What do this backticks do?

src/utils/shared.py
```diff
@@ -69,3 +69,20 @@ def yes_or_no(question):
...
+def set_order(objects, order_attr_name, pk_list):
+    """
+    A generic implementation of model object ordering.
+    :param: objects: a queryset or list of model objects
+    :param: order_attr_name: string the model object's order attr name
+    :param: pk_list: list of object PKs in order
+    """
+    ids = [int(_id) for _id in pk_list]
+
+    for _object in objects:
```
mauromsl, line 83: Nitpick: PEP8 way would be `object_` instead of `_object`

mauromsl (changes requested):
This now conflicts with master, needs rebasing/merging
Couple of nitpicks inline

src/review/migrations/0008_auto_20201007_1002.py
```diff
@@ -0,0 +1,19 @@
...
+    dependencies = [
+        ('review', '0007_null_help_text'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='reviewformelement',
+            options={'ordering': ('order', 'name')},
+        ),
+    ]
```
mauromsl, line 19: This migration is now a diverging node on the graph, can you recreate it?

mauromsl (changes requested):

mauromsl (approved):

PR #1860, opened 2020-10-13
[preprints] links on repository manager dashboard should work

tingletech:
Description: Hello,

I seem to be having trouble accessing the moderation part of the EA
dashboard. I can access the repository manager no problem; however, if I
try to access X Versions Awaiting Moderation or View Orphaned Preprints, I
encounter the following:

```
You are authenticated as ____, but are not authorized to
access this page. Would you like to login to a different account?
```

Any advice would be appreciated.

### DIFF
--- a/src/repository/views.py
+++ b/src/repository/views.py
@@ -1140,7 +1140,7 @@ def repository_delete_subject(request):
     )
 
 
-@staff_member_required
+@is_repository_manager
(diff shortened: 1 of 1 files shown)


hardyoyo:
+1 obviously correct

mauromsl (approved):

PR #1875, opened 2020-10-21
Removes cache-control header from file serving

### LINKED ISSUE #1874: Replacing manuscript in "Documents" does not work
**Describe the bug**
If you go to the "Documents" page for an article and replace a manuscript file for an article in the review stage, the file appears to upload. However, when you click "Download", you are presented with the original file, not the new one.

**Janeway version**
The version running on OLH.

**To Reproduce**
1. Go to Documents on an article that is in review
2. Click on the replace icon next to the MS file
3. Upload a new document
4. Go back to the Documents view and press the download icon next to the MS. You will get the original file, not the new one.

**Expected behavior**
Expect the new file to show up.

mauromsl:
Fixes #1874 
Mitigates a bug where replaced files such as manuscripts were being cached by the browser.
Chrome and Firefox will ignore the cache-control directive when making the same request from the same tab, so the only way to reproduce the bug originally reported is to:
 - Download a file.
 - Replace the file.
 - Download the file again from a different tab.

I've taken an abrupt approach in that I've removed the cache-control header from all `serve_ file_to_browser` requests (except images) but perhaps a more selective approach might be best, knowing that cache invalidation is the hardest problem in computer science ;)

Any thoughts @ajrbyers  ?

### DIFF
--- a/src/core/files.py
+++ b/src/core/files.py
@@ -21,7 +21,7 @@ from django.http import StreamingHttpResponse, HttpResponseRedirect, HttpRespons
 from django.shortcuts import get_object_or_404
 from django.utils.html import strip_tags
 from django.utils.text import slugify
-from django.views.decorators.cache import cache_control
+from django.views.decorators.cache import patch_cache_control
 
 from utils import models as util_models
 from utils.logger import get_logger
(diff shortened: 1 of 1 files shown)


ajrbyers (approved):
@mauromsl I think this is fine.

PR #1884, opened 2020-10-23
Prompt production manager that article has no projected issue

### LINKED ISSUE #1877: Projected issue prompt needed for before articles are typeset
At present editors can miss the need to set a projected issue for an article at the point of (ie prior to) sending the article to be typeset, meaning that the typesetters don't have the necessary information to hand when typesetting the volume and issue number. Would it be possible to add in a feature to prompt the editor to check this before sending please? At present it's possible to send an article to be typeset without doing this, which can mean that the typesetters don't have the correct issue / volume number needed. A popup automatically flagging which volume and issue the article will be typeset for (or warning the editor that none has been set and with a link to where to do this) would be very useful indeed here - thanks for your help.

mauromsl:
Closes #1877

### DIFF
--- a/src/templates/admin/production/assign_typesetter.html
+++ b/src/templates/admin/production/assign_typesetter.html
@@ -18,6 +18,20 @@
                 {% include "elements/forms/errors.html" with form=form %}
                 {% csrf_token %}
             </div>
+            {% if not article.projected_issue %}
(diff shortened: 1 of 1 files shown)


ajrbyers (approved):
@mauromsl should we also add this to https://github.com/BirkbeckCTP/typesetting ?

PR #1885, opened 2020-10-23
Add Self as Author button: ensure the user's orcid is copied, too

### LINKED ISSUE #1878: Add self as author button does not seem to copy the user's ORCID to the submission's author
**Describe the bug**
Clicking the add self as author button does not seem to copy the user's ORCID to the submission's author.

**Janeway version**
Discovered while running on the current head of the preprint-remodel branch.

**To Reproduce**
1. Ensure an ORCID is added to the user account you will use for testing
2. Start a submission
3. When it comes time to add authors (step 2), click the 'Add Self as Author' button
4. Accept preprint
5. View preprint in repository
6. Observe that no ORCID badge appears next to the name of the author

**Expected behavior**
An ORCID badge should appear next to any author entry which includes an ORCID

hardyoyo:
- adds the user's orcid to the author dictionary for new authors
- updates the author object with the user's orcid to catch an edge case

fixes #1878

### DIFF
--- a/src/repository/models.py
+++ b/src/repository/models.py
@@ -465,7 +465,7 @@ class Preprint(models.Model):
             defaults=author_dict,
         )
 
-        #if the author object already exists, but does not have an orcid, the default given above will not be used... solution: go back and update the orcid
+        # if the author object already exists, but does not have an orcid, the default given above will not be used... solution: go back and update the orcid
         if not author.orcid and user.orcid:
             author.orcid = user.orcid
             author.save()
--- a/src/repository/models.py
+++ b/src/repository/models.py
@@ -217,10 +217,10 @@ class Repository(model_utils.AbstractSiteModel):
             return self._site_path_url(path)
 
         return logic.build_url(
-                netloc=self.domain,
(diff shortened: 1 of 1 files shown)


src/repository/models.py
```diff
@@ -451,16 +451,25 @@ def display_authors(self):
...
             'affiliation': user.affiliation() if user.affiliation else '',
+            'orcid': user.orcid if user.orcid else '',
         }
+
         author, a_created = Author.objects.get_or_create(
             email_address=user.email,
             defaults=author_dict,
         )
+
+        #if the author object already exists, but does not have an orcid, the default given above will not be used... solution: go back and update the orcid
```
ajrbyers, line 468: Strictly speaking this should be `# if the author`

ajrbyers (changes requested):
One nasty PEP8 comment.

src/repository/models.py
hardyoyo: I double-checked my VSCode config, it's set up to use autopep8, and autopep8 is installed. Oddly enough, this space after a hash issue is not flagged. But, I trust you, and have made the requested change.

Issue #1873, opened 2020-10-20
[Preprints] how can I modify the title and abstract of a pre-existing pre-print on the new system?


tingletech:
**Is your feature request related to a problem? Please describe.**

https://twitter.com/DrCraigMagee/status/1315551768818987008 

**Describe the solution you'd like**
Some way for authors to edit the metadata (and for the moderators to approve it)

**Describe alternatives you've considered**
The question is do we want to store the history? If yes we may need to fundamentally change the models.
If not, we can store it in VersionQueue and then simply over-write the title and abstract.

**Additional context**

I started to look at the VersionQueue solution 
https://gist.github.com/tingletech/c8206ff653f648509dc89c5917c54980 

the text in the help box made it sound like it would not trigger a new version " If this is a completely new version of this manuscript, upload it and a new version will be generated and the older version retained for posterity."  If every metadata edit triggers a new version, that seems confusing.

Andy suggested: "Could change it to allow a quick replace with correction." but I could not see an obvious way of how to do that.

justingonder:
@tingletech @ajrbyers 
In testing, the following issues were discovered with this patch:
- From the [Repository Manager dashboard](https://dev.eartharxiv.org/repository/manager/), clicking on the title of an Unpublished Preprint that has pending revisions results in a 404 (e.g. MS 1735 "Revision Test")
- From the [Versions Moderation queue](https://dev.eartharxiv.org/repository/manager/versions/), clicking on View Detail for an item with a pending **Metadata Correction** has no effect. (Clicking View Detail for an item with a pending **Text Correction** results in a modal appearing that details the changes, which is what is expected when clicking the same for a Metadata Correction.) Moderators can therefore not see what changes have been made before approving those changes.
- In the Version Moderation queue, the ID column in the table contains unexpected values that do not relate to the manuscript's actual ID. This would make it difficult to understand which entries in the table relate to the same MS, especially in cases when multiple revisions for a single entry are present in the table.
- As an author, there is no clear indication that a Metadata Update has been successfully entered. Clicking the Submit Update button simply returns the user to the Preprint #---- Summary page. A notification that the update has been submitted would help here. From that page, an author can click through to immediately submit another Metadata Update, and the previously submitted update is not reflected (i.e. the _original_ title and abstract are still displayed). This may cause a user to submit unnecessary duplicate changes since they are not getting UI feedback that their changes are now saved and pending.
[...]

PR #1891, opened 2020-10-27
1873 added a base solution for this problem.

ajrbyers:
This is a potential solution for #1873 as agreed with @mauromsl.

- We now store a Title and Abstract against the VersionQ object.
- If a VersionQ is approved the Preprint.current_version is updated to hold the current Title and Abstract and the Preprint Title and Abstract are overwritten by the version in VersionQ.
- Also fixes the Moderation Queue duplicate indicator


ajrbyers:
Welcome comments from @hardyoyo, @justingonder and @tingletech

I think the only thing missing here is a 'View History' button.

src/repository/models.py
```diff
@@ -842,9 +854,46 @@ class VersionQueue(models.Model):
...
+        current_version.title = self.preprint.title
+        current_version.abstract = self.preprint.abstract
+
+        # Create a new PreprintVersion, this will now be the current_version.
+        PreprintVersion.objects.create(
+            preprint=self.preprint,
+            file=self.file,
+            version=self.preprint.next_version_number(),
+            moderated_version=self,
+        )
```
mauromsl, line 889: Shouldn't we add the title and abstract when the `PreprintVersion` is created, instead of back-populating the previous one?

src/repository/migrations/0018_auto_20201027_1110.py
```diff
@@ -0,0 +1,37 @@
...
+            model_name='preprintversion',
+            name='abstract',
+            field=models.TextField(blank=True, help_text='Please avoid pasting content from word processors as they can add unwanted styling to the abstract. You can retype the abstract here or copy and paste it into notepad/a plain text editor before pasting here.', null=True),
+        ),
+        migrations.AddField(
+            model_name='preprintversion',
+            name='title',
+            field=models.CharField(default='', help_text='Your article title', max_length=300),
+            preserve_default=False,
+        ),
```
mauromsl, line 25: I think we need a migration to back-populate all existing `PreprintVersion` and `VersionQueue`.

Alternatively assume an empty field as "unchanged" when we render the history (in which case we should handle that in `VersionQueue.approve` too). This approach is a bit more implicit application-wise but does save us from having dobule copies of abstract and title for each version.

src/repository/models.py
```diff
@@ -842,9 +854,46 @@ class VersionQueue(models.Model):
...
+
+        # Overwrite the preprint's metadata now we have a historical record.
+        self.preprint.title = self.title
+        self.preprint.abstract = self.abstract
```
mauromsl, line 893: Without the migration I mentioned below, I think this will wipe the title and abstract of all preprints that currently have a VersionQueue in progress.

mauromsl (commented):
Added a couple of comments in-line. Also a question:

Should we allow authors to create new versions in which they just request an abstract and/or title change? Currently, the file field is required in the model, the form and in the handling of the form POST

src/repository/models.py
ajrbyers: I had thought of that but there are other ways a Title or Abstract can be changed to wanted it set at the point the version was no longer the `current_version`.

src/repository/models.py
```diff
@@ -842,9 +854,49 @@ class VersionQueue(models.Model):
...
+
     def approve(self):
         self.date_decision = timezone.now()
         self.approved = True
+
+        # Update the current version to have the Preprint's current title
+        # and abstract.
+        current_version = self.preprint.current_version
+        current_version.title = self.preprint.title
+        current_version.abstract = self.preprint.abstract
```
mauromsl, line 881: Now that we established 'blank' means unchanged, should we handle it here and in the VersionForm as well? Otherwise, all future `VersionQueue` and `PreprintVersion` objects will still have those redundant titles/abstracts. If this is desirable, then I'd suggest still adding the migration to backpopulate the past records for avoidance of doubt.

src/templates/admin/elements/repository/version_detail.html
```diff
@@ -29,10 +29,10 @@ <h2>Current Metadata</h2>
...
                 </div>
                 <div class="content">
-                    <p><strong>Title{% if version.title != version.preprint.title %} (changed){% endif %}</strong></p>
-                    <p>{{ version.title|safe }}</p>
-                    <p><strong>Abstract{% if version.abstract != version.preprint.abstract %} (changed){% endif %}</strong></p>
-                    <p>{{ version.abstract|safe|linebreaksbr }}</p>
+                    <p><strong>Title{% if version.title and version.title != version.preprint.title %} (changed){% endif %}</strong></p>
+                    <p>{{ version.title|safe }}{% if not version.title %}Title is unchanged{% endif %}</p>
+                    <p><strong>Abstract{% if version.abstract and version.abstract != version.preprint.abstract %} (changed){% endif %}</strong></p>
+                    <p>{{ version.abstract|safe|linebreaksbr }}{% if not version.abstract %}Abstract is unchanged{% endif %}</p>
```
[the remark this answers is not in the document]
ajrbyers: Based on our discussion earlier I’ll do a check if the they match their counterpart and mark them as unchanged.

src/repository/models.py
ajrbyers: Based on discussion earlier:

- add a migration to update existing version queue objects
- make fields not null

src/templates/admin/repository/author_article.html
```diff
@@ -259,7 +259,7 @@ <h2>Metrics Summary</h2>
          data-animation-out="slide-out-down">
         <div class="card">
             <div class="card-divider">
-                <h4><i class="fa fa-upload">&nbsp;</i>Galley {{ galley.pk }}</h4>
+                <h4><i class="fa fa-upload">&nbsp;</i>Update Preprint</h4>
```
tingletech, line 262: FWIW, I just noticed that `Update {{ request.repository.object_name }}` is used above on line 140

PR #1899, opened 2020-10-29
1898 added subject page and made subject filtering bette

### LINKED ISSUE #1898: Filtering preprints by sub-discipline
On the old EarthArXiv website, we used to be able to browse preprints and filter them by discipline/tag, e.g. "Geophysics and Seismology". This was useful for quickly browsing new preprints most relevant to the user.

Currently, a workaround is to search for the "geophysics and seismology" tag, which sort of works, but the search results cannot be sorted by date, so it is not easy to quickly browse the newest preprints which are most appropriate to you. The sub-discipline tickbox options on the old EarthArXiv website worked well, I think.

ajrbyers:
- Adds a subject list page
- Updates subject filter page
Closes #1898


src/repository/views.py
```diff
@@ -274,7 +294,7 @@ def repository_search(request, search_term=None):
         )]
 
         preprints = list(set(list(preprint_search) + preprints_from_author))
-
+        preprints.sort(key=lambda x: x.date_published, reverse=True)
```
mauromsl, line 297: Worth using `operator.attrgetter` here to squeeze some nanoseconds out of the sorting.

src/themes/material/templates/repository/elements/subject_tree.html
```diff
@@ -0,0 +1,8 @@
+{% for subject in subjects %}
+    <li><a href="{% url 'preprints_list_subject' subject.slug %}">{{ subject }}</a></li>
```
mauromsl, line 2: So it looks like the view routed from this URL predates this PR. I had a look at it and I don't think it does what it is supposed to do:
https://github.com/BirkbeckCTP/janeway/blob/d89a6114b0ed280b63f210316937ec0dcd989ea8/src/repository/views.py#L199

It fetches the subject but does not apply any filters on the queryset.

While we are at it, is it worth renaming this to repository_list_subject to be consistent with the other views?

shicks-seismo:
Nice - speedy work - thanks!

src/themes/material/templates/repository/elements/subject_tree.html
```diff
@@ -0,0 +1,8 @@
+{% for subject in subjects %}
```
mauromsl, line 1: Just a nice to have: we could add a count of articles on each subject

mauromsl (changes requested):

src/themes/material/templates/repository/elements/subject_tree.html
ajrbyers: It grabs the subject and then fetches its preprint_set see:

https://github.com/BirkbeckCTP/janeway/blob/d89a6114b0ed280b63f210316937ec0dcd989ea8/src/repository/views.py#L201

mauromsl: :+1:

PR #1958, opened 2020-12-17
Update Lando configs to use MySQL instead of Postgres

hardyoyo:
- update Python version to a more modern 3.6.8
- install all dependencies for using MySQL with Janeway
- configure a MySQL database for the Lando dev environment
- adjust all default configuration environment variables


mauromsl (commented):
Hello @hardyoyo 

Even though we are aiming for full compatibility for MySQL, we want to promote postgresql as the preferred database backend is there a reason to change the default option here?

hardyoyo:
> Hello @hardyoyo
> 
> Even though we are aiming for full compatibility for MySQL, we want to promote postgresql as the preferred database backend is there a reason to change the default option here?

At CDL we use MySQL. I can further customize this so that Lando "does the right thing" no matter which DB is chosen, that would be a fun project.

hardyoyo:
closing, I'll just use a .lando.local.yml file to set up mysql as the DB, after I figure it out, I'll write it up.

PR #1982, opened 2021-01-19
bump requests to 2.25.1 to make current pip happy

hardyoyo:
As documented on [HackerNews](https://news.ycombinator.com/item?id=25145396) the Pip package manager for Python recently had its dependency algorithm tweaked. This lead to a situation where the current requirements.txt file could not be used with the latest version of pip without the load of dependencies failing. This change bumps up the version for the requests package to a version that keeps the current pip happy. Tests pass with this version of the library.


mauromsl:
Thanks @hardyoyo !

PR #1990, opened 2021-01-22
Removes the markdown cheat sheet.

ajrbyers:


mauromsl (commented):
I'm now thinking that we could use the old Markdown block to give some instructions here, I'll add a commit with that

ajrbyers:
I just made it full screen so there was more real-estate for the CMS page. I think this is good - maybe add a modal with help on it ?

PR #1994, opened 2021-01-25
#1993 adds defaults to issue-article ordering get_or_create

### LINKED ISSUE #1993: Issue Order
**Describe the bug**
When articles are imported they may not have an ArticleOrdering when added to an issue. This causes a get_or_create to try a create operation without a section.

**Janeway version**
1.3.9

**To Reproduce**

**Expected behavior**
Article ordering should work.

ajrbyers:
Closes #1993

### DIFF
--- a/src/journal/views.py
+++ b/src/journal/views.py
@@ -1490,11 +1490,13 @@ def issue_article_order(request, issue_id=None):
             order_obj, c = models.ArticleOrdering.objects.get_or_create(
                 issue=issue,
                 article=article,
+                defaults={
+                    'order': order,
+                    'section': article.section,
+                }
             )
-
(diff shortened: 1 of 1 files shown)


src/journal/views.py
```diff
@@ -1490,11 +1490,13 @@ def issue_article_order(request, issue_id=None):
...
+                defaults={
+                    'order': order,
+                    'section': article.section,
+                }
             )
-
-            order_obj.order = order
-            order_obj.section = article.section
             order_obj.save()
+            print(order_obj.pk)
```
mauromsl, line 1499: Print statement

mauromsl (changes requested):

PR #2011, opened 2021-02-02
Adds a new suppress_citations_metric setting.

ajrbyers:
Requested in support ticket 145.


mauromsl (changes requested):
This seems to only handle the OLH theme.

I can see that material uses a different logic, it only renders the citation block if there are any citations.

Whichever approach we prefer, we should make it consistent across all themes.
Personally, I prefer what we already have in the material theme (only display the block when there are citations). There is already another setting
`disable_metrics_display` at the journal level for the entire metrics block.

mauromsl (approved):

PR #2027, opened 2021-02-10
[preprints] Preprint remodel metadata edit bug

tingletech:
----

fixes some errors on `preprints-remodel` that were keeping us from pushing the latest changes.

Justin reported on https://github.com/BirkbeckCTP/janeway/issues/1873#issuecomment-741984981

I think we have not pushed code to prod since before  #1891 ?

yea, our prod is running c1e628c5c92bec288580278d4e544fa8b941a66e right now

---

I kept getting different errors between dev and localhost, after some false leads, turned out it was because debug mode made the error manifest differently.  Tried `path` and `host` modes and different versions of python before I figured out it was debug mode that was making me not be able to reproduce the same exact errors.

### DIFF
--- a/src/repository/models.py
+++ b/src/repository/models.py
@@ -889,6 +889,7 @@ class VersionQueue(models.Model):
     def approve(self):
         self.date_decision = timezone.now()
         self.approved = True
+        current_version = None
 
         # Update the current version to have the Preprint's current title
         # and abstract.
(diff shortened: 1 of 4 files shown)


ajrbyers:
@mauromsl requested a review but to confirm @tingletech is this still a WIP? If yes can you mark it as a draft?

tingletech:
I think the third commit fixed it, and I'm not getting 500 errors on dev or localhost while doing metadata only updates on version-less preprints

hardyoyo (commented):
this all looks correct to me

hardyoyo (approved):

ajrbyers (approved):

ajrbyers:
Thanks @tingletech

PR #2038, opened 2021-02-15
Adds a KBART metadata endpoint.

### LINKED ISSUE #2035: Add KBART metadata export
Some providers, such as EBSCO, require a KBART export
[KBART_Global_Template_2016-04-26 (1).xlsx](https://github.com/BirkbeckCTP/janeway/files/5981982/KBART_Global_Template_2016-04-26.1.xlsx)

MartinPaulEve:
Closes #2035


ajrbyers:
Just looking at the lines - they look quite long. Can you bring them under 90 chars per PEP8.

src/api/views.py
```diff
@@ -116,3 +118,45 @@ def oai(request):
...
+        if issues.count() > 0:
+            journal_line['date_first_issue_online'] = '{:%Y-%m-%d}'.format(issues.order_by('date')[0].date)
+            journal_line['num_first_vol_online'] = issues.order_by('date')[0].volume
+            journal_line['num_first_issue_online'] = issues.order_by('date')[0].issue
+            journal_line['date_last_issue_online'] = '{:%Y-%m-%d}'.format(issues.order_by('-date')[0].date)
+            journal_line['num_last_vol_online'] = issues.order_by('-date')[0].volume
+            journal_line['num_last_issue_online'] = issues.order_by('-date')[0].issue
+
+        journal_line['title_url'] = journal.full_url()
+        journal_line['coverage_depth'] = 'fulltext'
```
ajrbyers, line 152: Will this be true of all journals?

src/api/views.py
```diff
@@ -116,3 +118,45 @@ def oai(request):
...
+        journal_line['publisher_name'] = request.press.name
+        journal_line['publication_type'] = 'serial'
```
ajrbyers, line 154: Also not necesarily true for all journals, some of them on Iowa are conferences.

src/api/views.py
```diff
@@ -116,3 +118,45 @@ def oai(request):
...
+        journal_line['title_id'] = journal.issn
+        journal_line['access_type'] = 'F'
```
ajrbyers, line 156: What is F? Maybe a comment here inline.

ajrbyers (changes requested):

src/api/views.py
MartinPaulEve: The KBART spec only allows "serial" or "monograph". As conference papers are published with an ISSN, I think this is all we can do.

src/api/views.py
MartinPaulEve: Added

src/api/views.py
MartinPaulEve: It may _technically_ be possible for a user to add a table of contents that points to non-free content using overlay functionality. Not sure how best we can cover that eventuality...

MartinPaulEve:
Fixed line lengths -- pycharm misled me!

ajrbyers:
> Fixed line lengths -- pycharm misled me!

Pycharm is a bit fast and loose with line lengths!

src/api/views.py
mauromsl: Ignoring remote (`is_remote`) journals might be a good start.

src/api/views.py
mauromsl: We can do a bit more but checking the `issue.issue_type` on line https://github.com/BirkbeckCTP/janeway/blob/cd9ebd1716e3a8592b5d021dd1fe994df449264e/src/api/views.py#L156

A good start would be to filter on `issue_type__code='issue'` but ideally we would add a new property `journal.models.IssueType.is_serial` that is always true when `self.code == issue`

src/api/views.py
```diff
@@ -116,3 +118,72 @@ def oai(request):
...
+    writer = csv.DictWriter(response, fieldnames=fields, delimiter=delim)
+    response['Content-Disposition'] = 'attachment; filename="kbart.csv"' \
+        if delim == ',' else 'attachment; filename="kbart.tsv"'
+
+    writer.writeheader()
+
+    for journal in request.press.journals():
+        journal_line = {}
+        issues = journal.issues()
+        if issues.count() > 0:
```
mauromsl, line 157: Minor nitpick `issues.exists()` rather than `issues.count() > 0`

src/api/views.py
```diff
@@ -116,3 +118,72 @@ def oai(request):
...
+            journal_line['date_last_issue_online'] = \
+                '{:%Y-%m-%d}'.format(issues.order_by('-date')[0].date)
+            journal_line['num_last_vol_online'] = \
+                issues.order_by('-date')[0].volume
+            journal_line['num_last_issue_online'] = \
+                issues.order_by('-date')[0].issue
+
+        journal_line['title_url'] = journal.full_url()
+        journal_line['coverage_depth'] = 'fulltext'
+        journal_line['publisher_name'] = request.press.name
```
mauromsl, line 173: journals have a setting `journal_settings.general.publisher_name`

src/api/views.py
```diff
@@ -116,3 +118,72 @@ def oai(request):
...
+              "date_monograph_published_online", "monograph_volume",
+              "monograph_edition", "first_editor",
+              "parent_publication_title_id", "preceding_publication_title_id",
+              "access_type"]
+
+    response = HttpResponse(content_type='text/csv')
+
+    writer = csv.DictWriter(response, fieldnames=fields, delimiter=delim)
+    response['Content-Disposition'] = 'attachment; filename="kbart.csv"' \
+        if delim == ',' else 'attachment; filename="kbart.tsv"'
```
mauromsl, line 150: I would flip this logic around `.tsv` for delim == `\t` else `.csv`
(since your function allows arbitrary values for `delim`

src/api/views.py
```diff
@@ -116,3 +118,72 @@ def oai(request):
...
+              "num_last_vol_online", "num_last_issue_online",
+              "title_url", "first_author", "title_id", "embargo_info",
+              "coverage_depth", "notes", "publisher_name", "publication_type",
+              "date_monograph_published_print",
+              "date_monograph_published_online", "monograph_volume",
+              "monograph_edition", "first_editor",
+              "parent_publication_title_id", "preceding_publication_title_id",
+              "access_type"]
+
+    response = HttpResponse(content_type='text/csv')
```
mauromsl, line 146: needs handling based on delimeter

src/api/views.py
```diff
@@ -116,3 +118,72 @@ def oai(request):
...
+
+    writer = csv.DictWriter(response, fieldnames=fields, delimiter=delim)
```
mauromsl, line 148: Using a regular writer + an ordered dict saves us from maintaining two sets of keys (`fields` + each key added to the dictionary. It also makes the absence of a value for a key explicit rather than implicit (e.g. did Martin forget to add `"first_editor"` or has he left it out on purpose)

`journal_dict["first_editor"] = None # We don't support this because ABCXYZ`

src/api/views.py
```diff
@@ -116,3 +118,72 @@ def oai(request):
...
+            journal_line['num_first_issue_online'] = \
+                issues.order_by('date')[0].issue
+            journal_line['date_last_issue_online'] = \
+                '{:%Y-%m-%d}'.format(issues.order_by('-date')[0].date)
+            journal_line['num_last_vol_online'] = \
+                issues.order_by('-date')[0].volume
+            journal_line['num_last_issue_online'] = \
+                issues.order_by('-date')[0].issue
+
+        journal_line['title_url'] = journal.full_url()
```
mauromsl, line 171: `journal.full_url()` is deprecated in favour of `journal.site_url()`

https://github.com/BirkbeckCTP/janeway/blob/cd9ebd1716e3a8592b5d021dd1fe994df449264e/src/journal/models.py#L237

mauromsl (changes requested):
Thank you @MartinPaulEve !
I see the spec has other data fields that have been omitted but are necessary according to the spec:

> The KBART recommended practice includes twenty-five fields that are necessary to accurately describe resources for inclusion in a knowledgebase.

e.g.:  `publication_title` 
> For both serials and monographs, the publication_title field is used to give the full name of a publication as it appears on the print edition or homepage

Some comments from my end in line as well

src/api/views.py
MartinPaulEve: Done

src/api/views.py
MartinPaulEve: I've used this now but note that journal_settings.general.publisher_name does not seem to be auto-populated at install-time in any way.

src/api/views.py
MartinPaulEve: Not sure about the merit of this -- realistically, although it allows delim to be specified, it's only going to be called with either tab or comma.

src/api/views.py
MartinPaulEve: Done (except maintained OrderedDict with late initialisation)

src/api/views.py
MartinPaulEve: Changed

src/api/views.py
MartinPaulEve: Done

ajrbyers:
test this please

mauromsl (approved):

PR #2069, opened 2021-02-26
Uploading zipped galleys returns an error

mauromsl:


ajrbyers (commented):
Should we continue to avoid the word “Galleys” in the errors?

mauromsl:
That is something we changed for typesetting plugin but not for Janeway

PR #2082, opened 2021-03-08
#2081 update order and section when a create doesnt occur.

### LINKED ISSUE #2081: Issues: Article ordering
**Describe the bug**
When you attempt to order an article on the issue manager it will not save a new order.

**Janeway version**
1.3.8+

**To Reproduce**
1. Go to the issue manager
2. Try to drag and drop an article

**Expected behavior**
Article are re-ordered within their section

ajrbyers:
Closes #2081

### DIFF
--- a/src/journal/views.py
+++ b/src/journal/views.py
@@ -1487,7 +1487,7 @@ def issue_article_order(request, issue_id=None):
                     " %s" % ids
                 )
                 continue
-            order_obj, c = models.ArticleOrdering.objects.get_or_create(
+            models.ArticleOrdering.objects.update_or_create(
                 issue=issue,
                 article=article,
                 defaults={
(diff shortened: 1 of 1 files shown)


src/journal/views.py
```diff
@@ -1495,9 +1495,13 @@ def issue_article_order(request, issue_id=None):
                     'section': article.section,
                 }
             )
+            if not c:
+                # When an object is not created, update it.
+                order_obj.order = order
+                order_obj.section = article.section
```
mauromsl, line 1501: sounds like we should just use `update_or_create`

mauromsl (approved):
Just a minor comment, looks good otherwise

mauromsl (approved):

PR #2109, opened 2021-03-24
1967 contact button

### LINKED ISSUE #1967: Add a button to 'Email reviewer' from the review detail page
When an editor is reading the review from a peer reviewer, they don't have any links to contact them from this page (they have to instead go back to the review page and hit the "envelope" icon next to the corect author.

Add a button to contact the reviewer from the View Review page. Additionally, allow the sender to include other email addresses in the CC list. It would be useful to be able to edit the CC list on all of the email screens.

ajrbyers:
- Adds a contact button to the view_review page
- Adds a CC field to the form
- Validates email addresses
- Adds CC to `send_email_with_body_from_user` defaults to None
- Closes #1967


src/journal/forms.py
```diff
@@ -91,6 +91,7 @@ def __init__(self, *args, **kwargs):
 
 
 class EmailForm(forms.Form):
+    cc = forms.CharField(max_length=1000, help_text='Seperate email addresses with ;')
```
mauromsl, line 94: Typo: "Separate"

Worth using tagit instead of relying on the semi-colon?

src/journal/logic.py
```diff
@@ -505,6 +505,8 @@ def resend_email(article, log_entry, request, form):
 
 
 def send_email(user, form, request, article):
+    cc_list = [x.strip() for x in form.cleaned_data['cc'].split(';') if x]
+    valid_cc_email_addresses = validate_to_list(cc_list)
```
mauromsl, line 509: This function can raise a ValidationError, which would hit the top level and throw a 500
We should do the validation in the form instead.

Also: if you follow my recommendation above about using tagit, we could have front-end validation too.

mauromsl (changes requested):
Couple of suggestions inline on the use of the CC field

mauromsl (approved):

PR #2110, opened 2021-03-24
#2103 added new option to display reviewer information to authors. Of…

### LINKED ISSUE #2103: Add option to disable peer review data on Author page
When an author logs onto the dashboard and looks at their article they can see lots of information about peer review.

- Make this optional and off by default as Editors may not want this information available.
- Add text to the block informing the author that due dates do not mean that they will receive feedback  instantly.

ajrbyers:
- Adds new setting to enable in depth data
- Setting is off by default
- When off only info on complete reviews are shown

Closes #2103

### DIFF
--- a/src/utils/install/journal_defaults.json
+++ b/src/utils/install/journal_defaults.json
@@ -2944,7 +2944,7 @@
             "name": "general"
         },
         "setting": {
-            "description": "If enabled, authors will see detailed (anonymous) information about peer reviews inc. due and completed dates on the article information page.",
+            "description": "If enabled, authors will see the requested, due, and complete date for their papers' review assignment. Otherwise, they will only have access to the review comments.",
             "is_translatable": false,
             "name": "enable_peer_review_data_block",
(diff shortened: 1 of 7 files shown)


src/utils/install/journal_defaults.json
```diff
@@ -2938,5 +2938,20 @@
         "value": {
             "default": "1"
         }
+    },
+    {
+        "group": {
+            "name": "general"
+        },
+        "setting": {
+            "description": "If enabled, authors will see detailed (anonymous) information about peer reviews inc. due and completed dates on the article information page.",
```
mauromsl, line 2947: Hmm, it might not be clear what happens when this is disabled: 
"If enabled, authors will see the requested, due, and complete date for their papers' review assignment. Otherwise, they will only have access to the review comments"


I would avoid using the word "anonymous" since it it won't be true for open reviews.

mauromsl (changes requested):
One minor comment about the help text

Can you update the docs accordingly as well? we should make a habit of this when we add new settings.

PR #2111, opened 2021-03-25
#1975 when making a decison, editors are informed of reviews that are…

### LINKED ISSUE #1975: Warn editors when article is rejected without review comments being made available
Often times, the editors might make a decision on an article, having forgotten to make the peer review comments visible to the author. When this happens, the author receives a link to the review page where they receive a permission denied (Although they are able to see a usefull message saying that the peer review comments haven't been made available).

When the editor makes a decision before making some of the review comments available to the author, we should bring up an alert prompt. The editor should still be able to make their decision regardless.

**Additional context**
Resolution of the situation described above is a bit complex for the editors when they are unfamiliar with Janeway. They have to find the rejected article in the archive, jump back to the peer review stage open each review comment and then hit "Allow author to see"

ajrbyers:
… not available to the author

- Closes #1975

### DIFF
--- a/src/submission/models.py
+++ b/src/submission/models.py
@@ -1287,6 +1287,15 @@ class Article(models.Model):
 
         return article_link_count + book_link_count
 
+    def hidden_completed_reviews(self):
+        return self.reviewassignment_set.filter(
+            is_complete=True,
+            date_complete__isnull=False,
+            for_author_consumption=False,
+        ).exclude(
+            decision='withdrawn',
+        )
+
 
 class FrozenAuthor(models.Model):
     article = models.ForeignKey('submission.Article', blank=True, null=True)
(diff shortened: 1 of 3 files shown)


src/submission/models.py
```diff
@@ -1287,6 +1287,15 @@ def citation_count(self):
 
         return article_link_count + book_link_count
 
+    def hidden_completed_reviews(self):
+        return self.reviewassignment_set.filter(
+            is_complete=True,
+            date_complete__isnull=False,
```
mauromsl, line 1293: we might as well warn them of incomplete reviews, no?

mauromsl (commented):
Just a question

src/submission/models.py
ajrbyers: When combined with the new decision helper (which displays incomplete reviews) I think this will be fine.

mauromsl:
@ajrbyers  you've merged this into master, not  into 1.3.10  branch

PR #2114, opened 2021-03-26
#2101 the review page now includes metadata for the paper.

### LINKED ISSUE #2101: Peer Review Page
When undertaking a review the peer reviewer is not given enough information.

Ensure the following are included on the review form page:
- Title
- Abstract
- Section
- Keywords

ajrbyers:
Closes #2101

### DIFF
--- a/src/templates/admin/elements/review/review_meta_block.html
+++ b/src/templates/admin/elements/review/review_meta_block.html
@@ -1,0 +1,37 @@
+<p><strong>The editor handling this piece is:</strong></p>
+
+<ul>
+    {% if review_request.article.section_editors %}
+        {% for editor in review_request.article.section_editors %}
+            <li>{{ editor }} <a href="mailto:{{ editor.email }}?subject={{ review_request.article.journal.name }} Review"><i
+                    class="fa fa-envelope"></i></a></li>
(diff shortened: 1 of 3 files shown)


src/templates/admin/elements/review/review_meta_block.html
```diff
@@ -0,0 +1,37 @@
+<p><strong>The editor handling this piece is:</strong></p>
+
+<ul>
+    {% if review_request.article.section_editors %}
+        {% for editor in review_request.article.section_editors %}
+            <li>{{ editor }} <a href="mailto:{{ editor.email }}?subject={{ review_request.article.journal.name }} Review"><i
+                    class="fa fa-envelope"></i></a></li>
+        {% endfor %}
+    {% elif review_request.article.editors %}
```
mauromsl, line 9: There could be both editors and section editors in an article no? (`if` vs `elif`)

We could also UNION (with the `.union()` method https://docs.djangoproject.com/en/1.11/ref/models/querysets/#union ) all these querysets to avoid the awkward logic.

mauromsl (changes requested):
Just one minor comment, then feel free to merge

PR #2124, opened 2021-04-09
Correct Pending Updates table heading

justingonder:
The heading 'delete' in the Pending Updates table is incorrect as the actions in that column are 'Manage' not 'Delete.' Changing heading to 'Manage.'


mauromsl (approved):
Thank you @justingonder !

PR #2127, opened 2021-04-12
1809 draft decisions

### LINKED ISSUE #1809: Draft Decisions
Some updates are required to Draft Decisions:

- [x]  Add help text for draft decisions.
- [x]  Update email template to use the email to the Author not the email to the Editor.
- [x] Check how its handled if no senior editor is assigned [if none: select from dropdown?].
- [x] Add reviews to the Draft page so that Senior Editor can check and edit the reviews.

ajrbyers:
Closes #1809


src/review/logic.py
```diff
@@ -192,21 +192,24 @@ def get_decision_content(request, article, decision, author_review_url):
...
+        do_revisions_url = request.journal.site_url(path=reverse(
+            'do_revisions',
+            kwargs={
+                'article_id': article.pk,
+                'revision_id': revision.pk,
+            }
+        ))
+        email_context['do_revisions_url'] = do_revisions_url
+    else:
+        email_context['do_revisions_url'] = "{{ do_revisions_url }}"
```
mauromsl, line 214: How does this work? presumably the template already has the variable `{{ do_revisions_url }}` . If we don't render the context variable here, how does it get rendered? Is it expecting to be processed by the event later?

src/review/forms.py
```diff
@@ -13,18 +13,33 @@
 from review.logic import render_choices
 from core import models as core_models
 from utils import setting_handler
-from utils.forms import FakeModelForm
+from utils.forms import FakeModelForm, HTMLDateInput
 
 
 class DraftDecisionForm(forms.ModelForm):
+    widgets = {'revision_request_due_date': HTMLDateInput()}
```
mauromsl, line 20: Should pass ISO format to `format` argument

src/review/forms.py
```diff
@@ -13,18 +13,33 @@
...
 
     def __init__(self, *args, **kwargs):
-        email_message = kwargs.pop('email_message', None)
+        self._newly_created = kwargs.get('instance') is None
+        message_to_editor = kwargs.pop('message_to_editor', None)
+        editors = kwargs.pop('editors', [])
         super(DraftDecisionForm, self).__init__(*args, **kwargs)
-        self.fields['email_message'].initial = linebreaksbr(email_message)
+        self.fields['message_to_editor'].initial = linebreaksbr(message_to_editor)
+        self.fields['revision_request_due_date'].widget = HTMLDateInput()
```
mauromsl, line 34: Should pass ISO format to `format` argument

mauromsl (changes requested):
So far so good, a few minor issues pointed inline.

I think we will need to update the import plugin since we are removing a field from DecisionDraft.

src/review/logic.py
ajrbyers: We retain the placeholder `{{ do_revisions_url }}` in the drafted template so it can be rendered later as the request does not exist at this point.

src/review/logic.py
```diff
@@ -238,10 +240,10 @@ def log_revision_event(text, user, revision_request):
     revision_request.actions.add(action)
 
 
-def get_draft_email_message(request, article):
+def get_draft_email_message(request, article, ):
```
mauromsl, line 243: Looks like an accidental comma

mauromsl (changes requested):
One minor thing left in line

mauromsl (approved):

PR #2130, opened 2021-04-12
#2122 updates the review home page.

### LINKED ISSUE #2122: Update to review list page
The details on the review page are great but a couple of suggestions: 

1) Is it possible to adjust the invitation status from "accept/decline" to "yes/no." Accept/decline are somewhat confusing since they are so similar with manuscript decisions (rather than invitation to review decisions). 

2) Past conversations with Janeway suggested an intention to make peer review decisions visible on the review page -- any updated? 

3) is it possible to distinguish between review rounds on the review page to avoid confusion.

ajrbyers:
Closes #2122

### DIFF
--- a/src/templates/admin/review/home.html
+++ b/src/templates/admin/review/home.html
@@ -13,6 +13,7 @@
         <div class="title-area">
             <h2>Articles in Review</h2>
             {% if filter == 'me' %}<a class="button" href="{% url 'review_home' %}?filter=all">All Submissions</a>{% else %}<a class="button" href="{% url 'review_home' %}?filter=me">My Assignments</a>{% endif %}
+            <a href="#" data-open="statuskey" class="button">Review Status Key</a>
         </div>
         <div class="content">
             <table class="small scroll" id="unassigned">
@@ -45,12 +46,11 @@
                         {% if journal_settings.general.enable_expanded_review_details %}
                         <td>
                             {% if article.reviewassignment_set.exists %}
-
                                 <table style="width:100%; min-width:230px;" class="table table-stack small no-pad no-bottom-margin">
-                                    {% for review in article.reviewassignment_set.all %}
-
+                                    {% for review in article.reviews_not_withdrawn %}
                                         <tr>
-                                            <td>({{ review.review_round.round_number }}) <a href="{% url 'review_view_review' article.pk review.pk %}">{{ review.reviewer.initials }}</a></td>
+                                            <td>{{ review.review_round.round_number }}</td>
+                                            <td><a href="{% url 'review_view_review' article.pk review.pk %}">{{ review.reviewer.initials }}</a></td>
                                             <td>
                                                 <span class="{{ review.status.span_class }}">
                                                     {% if review.status.code == 'complete' %}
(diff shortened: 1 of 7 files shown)


src/templates/admin/review/home.html
```diff
@@ -50,8 +50,18 @@ <h2>Articles in Review</h2>
...
+                                            <td>({{ review.review_round.round_number }}) <a href="{% url 'review_view_review' article.pk review.pk %}">{{ review.reviewer.initials }}</a></td>
+                                            <td>
+                                                <span class="{{ review.status.span_class }}">
+                                                    {% if review.status.code == 'complete' %}
+                                                    <span data-tooltip class="top" tabindex="2" title="Recommendation: {{ review.get_decision_display }}">
+                                                    {% endif %}
+                                                    {{ review.status.display|safe }}
+                                                    {% if review.status.code == 'complete' %}
+                                                    </span>
+                                                    {% endif %}
```
mauromsl, line 62: As discussed:
- Ignore reviews with status 'withdrawn'
- Instead of rendering the round, only show review assignments relevant to this round
Can we also add a different message when the article is under revisions (There exists an open RevisionRequest)

mauromsl (changes requested):

src/templates/admin/review/home.html
ajrbyers: @mauromsl I've updated this in consultation with @rhb123. Decision was made to continue to show all rounds though.

src/utils/shared.py
```diff
@@ -89,4 +89,4 @@ def set_order(objects, order_attr_name, pk_list):
 
   
 def day_month(date):
-    return date.strftime("%d-%b")
+    return date.strftime("%d-%m")
```
mauromsl, line 92: The original formatting was set in ff3d7e8bb since it was confusing for thoe in countries where the default date  is represented as month-date.

I'd advise reverting this change

src/templates/admin/review/home.html
```diff
@@ -45,17 +46,42 @@ <h2>Articles in Review</h2>
...
+                                                </span>
+                                            </td>
                                             <td>{{ review.status.date }}</td>
                                             <td>{% if review.status.reminder %}<a href="{% url 'review_send_reminder' article.pk review.pk review.status.reminder %}">{% endif %}Due {{ review.date_due|date:"d-m" }}{% if review.status.reminder %}</a>{% endif %}</td>
                                         </tr>
                                     {% endfor %}
+                                    {% if article.active_revisions %}
```
mauromsl, line 69: I think we should aslo considere completed revisions within this round too here.

Once revisions are completed, it won't be apparent to the editor that the article is ready to act on

mauromsl (changes requested):
Looks good, on comment and one change request in line

src/utils/shared.py
ajrbyers: I think if we revert this we should have the due date match so there aren't two weird date formats right next to one another?

src/utils/shared.py
```diff
@@ -89,4 +89,4 @@ def set_order(objects, order_attr_name, pk_list):
 
   
 def day_month(date):
-    return date.strftime("%d-%m")
+    return date.strftime("%d-%M")
```
mauromsl, line 92: Should be `%b`

mauromsl (changes requested):

src/utils/shared.py
ajrbyers: For some reason %b is returning with a capitalised month eg `Jan` and when using `%b` in the template it correctly returns a lowercase month eg `jan` so to hamonise I've set them both to `%M` which according to Django docs returns:

`M | Month, textual, 3 letters. | 'Jan'`

mauromsl: `stfrtime` is a python `datetime` function and `%M` renders minutes, not the month in textual form.

ajrbyers: 🤣

mauromsl (approved):

PR #2150, opened 2021-04-19
Revision request for missing editors

mauromsl:
When a revision request has no editor assigned no email notification was sent
Now we fallback to article editors -> journal editors


ajrbyers (approved):
🎖️

PR #2152, opened 2021-04-20
2149 enhancement

### LINKED ISSUE #2149: Users Page
The purpose of this page is not clear.

- Update the page name to be "Enrolled users"
- Add a description of what the page is used for
- Change "Roles" to "Enroll Users"

ajrbyers:
Closes #2149

### DIFF
--- a/src/templates/admin/core/manager/users/enrol_users.html
+++ b/src/templates/admin/core/manager/users/enrol_users.html
@@ -1,0 +1,77 @@
+{% extends "admin/core/base.html" %}
+{% load static from staticfiles %}
+
+{% block breadcrumbs %}
+    {{ block.super }}
+    <li><a href="{% url 'core_manager_index' %}">Manager</a></li>
+    <li><a href="{% url 'core_manager_users' %}">Enrol Users</a></li>
+{% endblock %}
+
+{% block title %}Enrol Users{% endblock %}
+{% block title-section %}Enrol Users{% endblock %}
+{% block title-sub %}Enrol users from a list of all users across the press.{% endblock %}
+
+{% block body %}
+    <div class="large-12 columns">
+        <div class="box">
+            <div class="title-area">
+                <h2>Enrol Users</h2>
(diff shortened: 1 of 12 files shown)


ajrbyers:
Ignore for now.

ajrbyers:
This is not perfect - but better than it was previously - I think.

src/templates/admin/core/manager/users/enrol_users.html
```diff
@@ -0,0 +1,94 @@
...
+{% block title %}Enrol Users{% endblock %}
+{% block title-section %}Enrol Users{% endblock %}
+{% block title-sub %}Search for an enrol user accounts in your journal.{% endblock %}
+
+{% block body %}
+    <div class="large-12 columns">
+        <div class="box">
+            <div class="title-area">
+                <h2>Enrol Users</h2>
+                <a href="{% url 'core_manager_index' %}" class="button">< Go back</a>
```
mauromsl, line 19: I think we link to this page from various places now, we might need a GET parameter so we can return them to the right place.

src/templates/admin/elements/core/user_search_list.html
```diff
@@ -0,0 +1,46 @@
...
+        <td>{{ user.username }}</td>
+        <td>{{ user.first_name }}</td>
+        <td>{{ user.last_name }}</td>
+        {% if settings.DEBUG %}
+            <td>
+                <button type="submit" class="tiny button"
+                        formaction="/hijack/{{ user.id }}/?next={% url 'core_dashboard' %}">
+                    Hijack User
+                </button>
+            </td>{% endif %}
```
mauromsl, line 24: We should add the email address to the table, since it is used in the form for searches.

One other header that would be usefull here is "is_active" which is one that trips editors a lot too.

mauromsl (commented):
Agree that this is definately a big improvement over what we have ,even with the overhead of the backend search.
I added a couple of suggestions inline that I think are worth considering before  merging

PR #2162, opened 2021-04-23
Se download links

ajrbyers:
Adds a SE clause to the badly named `article_stage_accepted_or_later_or_staff_required`


src/submission/models.py
```diff
@@ -220,6 +220,13 @@ def article_media_upload(instance, filename):
     STAGE_REJECTED,
 }
 
+SECTION_EDITOR_STAGES = {
+    STAGE_UNASSIGNED,
+    STAGE_ASSIGNED,
+    STAGE_UNDER_REVIEW,
+    STAGE_UNDER_REVISION
+}
+
```
mauromsl, line 229: This is not really the case, section editors can handle other stages like copyediting. I think it is best to delete this, since it not needed anyways (as per my comment above)

src/security/decorators.py
```diff
@@ -542,6 +542,9 @@ def wrapper(request, *args, **kwargs):
             deny_access(request)
         elif article_object is not None and (request.user.is_editor(request) or request.user.is_staff):
             return func(request, *args, **kwargs)
+        elif article_object.stage in models.SECTION_EDITOR_STAGES and request.user in article_object.section_editors():
```
mauromsl, line 545: I think it is safe to remove the stage check and just check `request.user in article_object.section_editors()`

mauromsl (changes requested):

ajrbyers:
Ready for re-review.

mauromsl (approved):

PR #2177, opened 2021-05-04
Two minor fixes for 1.3.10.

ajrbyers:


src/review/models.py
```diff
@@ -70,12 +70,12 @@ def __repr__(self):
 
     def active_reviews(self):
         return self.reviewassignment_set.exclude(
-            Q(date_declined__isnull=False) | Q(date_accepted__isnull=False) | Q(decision='withdrawn')
+            Q(date_declined__isnull=False) | Q(date_accepted__isnull=True) | Q(decision='withdrawn')
```
mauromsl, line 73: I think we need to exclude based on `date_completed` rather than on `date_accepted`. 
I think in english we'd say that a review is active if:
 - it has not been declined
 - It has not been completed
 - It has not been withdrawn
 
 Whether it has been accepted or not shouldn't be considered no?

mauromsl (changes requested):

mauromsl (approved):

PR #2218, opened 2021-05-19
Make date_time from prepub timezone aware

ajrbyers:


src/journal/logic.py
```diff
@@ -222,8 +223,7 @@ def handle_set_pubdate(request, article):
     date_time_str = "{0} {1}".format(date, time)
 
     try:
-        date_time = dateparser.parse(date_time_str)
-
+        date_time = make_aware(dateparser.parse(date_time_str))
```
mauromsl, line 226: We need to set the correct timezone here, otherwise `make_aware` uses the default timezone:
https://docs.djangoproject.com/en/1.11/_modules/django/utils/timezone/#make_aware

mauromsl (changes requested):

PR #2245, opened 2021-06-07
Email normalisation fix

ajrbyers:
Fixes an issue where the QuickUser form was setting the username to equal the email address rather than let the model handle this. Stops the handler function from setting defaults to institution (now included in the form) and country (not a required field anyway).


mauromsl (changes requested):
Looks good, we need a copy of the core 0042 migration for those accounts created in the interim

mauromsl (approved):

PR #2272, opened 2021-06-23
1 4 working branch

### LINKED ISSUE #1708: Refactor core.views.edit_settings_group
This view has a number of problems
 It serves 4 different purposes:

- Create a new journal from the press
- Edit a journal from the press manager
- Edit a journal from the journal manager
- Edit a setting group

- Break the view into a single view for each use case
- Get rid of the mutation of the global state of `request.journal`

### LINKED ISSUE #1895: Remove Default Theme
- [x] Copy any default-only templates into Clean
- [x] Move Defaullt to its own repository

### LINKED ISSUE #1896: Move Clean to Core
- [x] Move clean to core themes folder

### LINKED ISSUE #1897: Translations
- [x] Remove default from core themes
- [x] Update clean translations
- [x] Update material translations
- [x] Update OLH translations
- [x] Update Submission translations
- [x] Add welsh language
- [x] Regenerate PO files
- [x] Add i18n to builtins

### LINKED ISSUE #2014: Apply suppress_citations_metric logic to clean theme
In 1.3.9 we updated OLH and Material to allow editors to suppress the Citation metric block accross the journal independently from disabling the whole block or displaying if there is a metric.

Apply this to clean which is joining Core in 1.4

None.

### LINKED ISSUE #2016: HVAD to ModelTranslations
### General
- [x] Add Model Translations
- [x] Configure translatable fields
- [x] Add data migration
- [x] Remove HVAD
- [x] Leave fields encase something goes wrong - drop in 1.5?
- [x] Create JanewayTranslatableForm
- [x] When a field is translatable show [translatable] in the label

### Settings
- [x] Update Setting Handler
- [x] Rehome nav settings

### Plugin Settings
- [x] Rediect Plugin Settings to Settings
- [x] Each plugin should use a unique group name

### HVAD
- [x] Remove .language & .fallbacks from queries

### Sections

- [x] Update Section manager
- [x] Update Section filters, get, get_or_create etc
- [x] Update objects manager and get_or_create and update_or_create, if these are called outside the base language we should throw an error
- [x] Add page to view all articles in a given section

### Setting Interfaces
- [x] Update group settings interface
- [x] Update individual settings interface
- [x] Update email template settings interface
- [x] Update Delete Override (replace wording `Reset to Default (as above)` )
- [x] Flag settings as multi lingual as translatable
- [x] Regroup Group Settings and layout pages properly

### LINKED ISSUE #2033: 1.4 clean theme needs to merge glossing
The CSS in common.css to enable glossing should be merged into the clean theme for 1.4

### LINKED ISSUE #2078: Copyeditor/author can't access all files presented by the editor
**Describe the bug**
When a copyedit task is created, the editor can select which files should be accessible/editable by the copyeditor. This does not work, the copyeditor is only presented with the manuscript

**Janeway version**
v1.3.8

**To Reproduce**
1. Move an article into copyediting with at least a manuscript and a figure file
2. Create a copyediting selecting all of the files as available for the copyeditor
3. Login as the copyeditor and get to the task

**Expected behavior**
All selected files should be made available to the copyeditor, as well as the author when asking them to review the results of the copyediting process.

In fact, it might be best to allow the copyeditor/author access to all figures and supporting files rather than having the editor select them ahead of time (similar to what we did for typesetting plugin)

**Screenshots**
Original task:
image
Task recorded in the database
image
End result:
image

### LINKED ISSUE #2088: Identifiers.doi_pattern setting is flagged as translatable
**Describe the bug**
The setting `doi_pattern` is flagged as `ìs_translatable=true` which seems incorrect. This leads to a number of problems on journals  that had I18N enabled at some point, since they now might not have a setting value set for their default language (e.g if the language context of the journal manager was not set to the default language at the time when the setting was first entered)

As a result, those journals now raise an error whenever the setting is fetched in the default language context (or any other language even)

**Janeway version**
v1.3.9

**Expected behavior**
The setting should be flagged as not translatable at installation time.
A migration is required to fix current installs where:
 - A value is set for the default language
 - The installed setting `is_translatable` attribute has to change to `false`

### LINKED ISSUE #2155: Error when loading the News homepage plugin
**Describe the bug**
```
16:15:26 Traceback (most recent call last):
16:15:26   File "/vol/janeway/src/core/tests/test_app.py", line 269, in setUp
16:15:26     call_command('install_plugins')
16:15:26   File "/usr/local/lib/python3.5/site-packages/django/core/management/__init__.py", line 131, in call_command
16:15:26     return command.execute(*args, **defaults)
16:15:26   File "/usr/local/lib/python3.5/site-packages/django/core/management/base.py", line 330, in execute
16:15:26     output = self.handle(*args, **options)
16:15:26   File "/vol/janeway/src/utils/management/commands/install_plugins.py", line 52, in handle
16:15:26     plugin_settings.install()
16:15:26   File "/vol/janeway/src/core/homepage_elements/news/plugin_settings.py", line 48, in install
16:15:26     pretty='Number of Articles',
16:15:26 AttributeError: 'Setting' object has no attribute 'value'
```

**Janeway version**
2016-translation

**To Reproduce**
1. Run the test suite

### LINKED ISSUE #2159: Add language switcher to all themes
The OLH is the only theme that renders a form for switching the language of the front end.

Add language switcher to:

- [x] Material
- [x] Clean

### LINKED ISSUE #2160: switch_language setting wrongfully flagged as translatable
**Describe the bug**
The switch_language setting is flagged as translatable, despite being a boolean

**Janeway version**
v1.3.9

**Expected behavior**
The setting should not be flagged as translatable
A migration will be required to fix the setting as installed in all current journals

### LINKED ISSUE #2217: Add more customization options to the "Submissions" page template
At the moment (in version 1.3 I suppose), there are a number of predefined sections in the "Submission Settings" page (...manager/settings/submission/), but the range of options is quite limited and does not exactly suit our needs. For example, we would like to include a short section on "manuscript preparation" and on "Data and code" here. 

We believe the possibility to add custom sections (with freely defineable title plus a text field) would make this section a lot more flexible. In addition, it would be very useful to be able to re-order the sections on this page. For example, we would like to include a short section on "manuscript preparation" and on "Data and code" here.

One possible alternative, i.e. adding the additional information we would like to provide to authors in the uppermost field, does not work for us. One reason is that some of the information should not really be placed right at the top of the page, as this overly prioritizes it. The second reason is that subsections we add in the uppermost field do not appear in the table of contents column, which means it is harder to find and impossible to provide a direct link to it. 

**Additional context**
This issue may be related to, although it seems to me that it is distinct from, issues #969 and #122. 
If this or a similar feature is already planned for version 1.4, then please ignore the issue.

### LINKED ISSUE #2281: User Profile: Hide Publications header when user has none
On the user profile page the Publications header will show even if the user has none.

Hide this header when there are no publications.

ajrbyers:
Closes #2217
Closes #2088
Closes #2016 
Closes #1897 
Closes #1895 
Closes #1896 
Closes #1708 
Closes #2155
Closes #2160
Closes #2014
Closes #2033
Closes #2078
Closes #2159
Closes #2281


src/core/homepage_elements/html/hooks.py
```diff
@@ -13,7 +13,7 @@ def yield_homepage_element_context(request, homepage_elements):
     plugin = models.Plugin.objects.get(name='HTML')
     try:
         html_block_content = setting_handler.get_plugin_setting(plugin, 'html_block_content', request.journal).value
-    except (IndexError, models.PluginSettingValue.DoesNotExist) as e:
+    except (IndexError) as e:
```
mauromsl, line 16: I think this is now `AttributeError` based on what we did for the about element

src/core/migrations/0047_remove_dots_from_urls.py
```diff
@@ -5,16 +5,26 @@
...
-        setting.value = setting.value.replace("url }}.", "url }} ")
-        setting.save()
+    try:
+        SettingValueTranslation = apps.get_model('core', 'SettingValueTranslation')
+        settings = SettingValueTranslation.objects.filter(master__setting__group__name="email")
+        for setting in settings:
+            setting.value = setting.value.replace("url }}.", "url }} ")
+            setting.save()
+    except LookupError:
+        translation.activate(django_settings.LANGUAGE_CODE)
```
mauromsl, line 22: Should be enabled as a context: `with translation.activate` just in case

src/core/migrations/0048_add_article_details_to_review_assignment_email.py
```diff
@@ -3,26 +3,48 @@
...
+        for setting in setting_values:
+            # IF it hasn't been modified, just update it with the new value
+            if setting.value == OLD_VALUE:
+                setting.value = NEW_VALUE
+            elif "article_detail" not in setting.value:
+                # otherwise, append the metadata at the end
+                setting.value += ("<br/>{{ article_details }}")
+            setting.save()
+    except LookupError:
+        translation.activate(django_settings.LANGUAGE_CODE)
```
mauromsl, line 31: as above

src/journal/models.py
```diff
@@ -886,7 +896,7 @@ def __str__(self):
 @receiver(post_save, sender=Journal)
 def setup_default_section(sender, instance, created, **kwargs):
     if created:
-        submission_models.Section.objects.language('en').get_or_create(
+        submission_models.Section.objects.get_or_create(
```
mauromsl, line 899: we should save the value of name for settings.LANGUAGE_CODE, otherwise creating a journal from a different language than the default will leave us without a default language entry

src/journal/models.py
```diff
@@ -90,10 +91,21 @@ class Journal(AbstractSiteModel):
     enable_correspondence_authors = models.BooleanField(default=True)
     disable_html_downloads = models.BooleanField(default=False)
     full_width_navbar = models.BooleanField(default=False)
-    is_remote = models.BooleanField(default=False)
+    is_remote = models.BooleanField(
+        default=False,
+        help_text='When enabled the journal is marked as not hosted in Janeway.',
```
mauromsl, line 96: Shall we make all these new help_texts translatable?

Also, could use a comma: "When enabled, the journal..."

src/journal/models.py
```diff
@@ -157,10 +169,9 @@ def get_setting(self, group_name, setting_name):
         return setting_handler.get_setting(group_name, setting_name, self, create=False).processed_value
 
     @property
-    @cache(300)
     def name(self):
         try:
-            return setting_handler.get_setting('general', 'journal_name', self, create=False, fallback='en').value
+            return setting_handler.get_setting('general', 'journal_name', self, create=False).value
```
mauromsl, line 174: Doesn't this raise an attribute error now? Also create is deprecated and I think we want default=True

src/journal/models.py
```diff
@@ -171,7 +182,7 @@ def name(self, value):
 
     @property
     def publisher(self):
-        return setting_handler.get_setting('general', 'publisher_name', self, create=False, fallback='en').value
+        return setting_handler.get_setting('general', 'publisher_name', self, create=False).value
```
mauromsl, line 185: create is deprecated and we want the default to be loaded

src/journal/models.py
```diff
@@ -180,7 +191,7 @@ def publisher(self, value):
     @property
     @cache(120)
     def issn(self):
-        return setting_handler.get_setting('general', 'journal_issn', self, create=False, fallback='en').value
+        return setting_handler.get_setting('general', 'journal_issn', self, create=False).value
```
mauromsl, line 194: as above

src/journal/models.py
```diff
@@ -189,8 +200,7 @@ def use_crossref(self):
             return setting_handler.get_setting('Identifiers',
                                                'crossref_prefix',
                                                self,
-                                               create=False,
-                                               fallback='en').processed_value
+                                               create=False).processed_value
```
mauromsl, line 203: as above

src/press/views.py
```diff
@@ -117,8 +117,11 @@ def manager_index(request):
             install.update_license(new_journal)
             install.update_issue_types(new_journal)
             new_journal.setup_directory()
-            return redirect("{0}?journal={1}".format(reverse('core_edit_settings_group', kwargs={'group': 'journal'}),
-                                                     new_journal.pk))
+            return redirect(
+                new_journal.site_url(
+                    path=reverse('core_manager_index')
```
mauromsl, line 122: Should this redirect to the journal settings instead?

src/templates/admin/cms/nav.html
```diff
@@ -15,8 +18,18 @@
 {% block body %}
     <div class="large-6 columns">
         <div class="box">
+            {% include "admin/elements/translations/form_tabs.html" with object=nav_item_to_edit %}
             <form method="POST">
                 {% csrf_token %}
+                <small>
+                    <p>To create a navigation link to a CMS page:</p>
+                    <ul>
+                        <li>{% if settings.URL_CONFIG == 'path' and request.journal %}Prefix the link with {{ request.journal.code }}/{% endif %}site/</li>
```
mauromsl, line 27: I'm thinking we should do this when rendering the navigation rather than having users type in the journal code

src/utils/template_override_middleware.py
```diff
@@ -49,24 +49,24 @@ def get_theme_dirs(self):
                 try:
                     theme_setting = self.query_theme_dirs(_local.request.journal)
                 except BaseException:
-                    theme_setting = 'default'
+                    theme_setting = 'clean'
```
mauromsl, line 52: Should be OLH, also replace BaseException above with Exception

mauromsl (changes requested):
313 files reviewd! 😅 

Let's not do this again

src/journal/models.py
ajrbyers: Fixed the AttributeError already but yes swapped create for default.

src/utils/template_override_middleware.py
ajrbyers: This is the default for journals only, I've changed the default for presses and added a migration for anyone who still has default set.

PR #2288, opened 2021-07-05
Fixes save_galley for file paths.

ajrbyers:


src/production/logic.py
```diff
@@ -69,9 +69,15 @@ def save_source_file(article, request, uploaded_file):
 
 
 def save_galley(article, request, uploaded_file, is_galley, label=None, save_to_disk=True):
-    mime = files.guess_mime(uploaded_file.name)
+    #  When save_to_disk is false the uploaded_file is a string/path.
+    if save_to_disk:
+        mime = files.guess_mime(uploaded_file.name)
+    else:
+        mime = files.file_path_mime(uploaded_file)
```
mauromsl, line 76: Isn't it better to check if `uploaded_file` is a string? Might save us from another surprise

`save_file_to_article` which is called later with these same articles is checking the type of `uploaded_file` as well

mauromsl (changes requested):
One comment/question

src/production/logic.py
ajrbyers: DOOOONE

mauromsl (approved):

Issue #1924, opened 2020-11-13
Section Editor Unable to Download Files From Review after Assingment


hachacha:
**Describe the bug**
When someone with the Section Editor role is assigned to review a submission they are unable to download the file from the review page but instead get permission denied.

**Janeway version**
v1.3.9-RC-2
509b344b26075b2480872489903a4d3cd3282d8a

**To Reproduce**
submit an article
assign a user with the section editor role to review
sign in as the reviewer
try to examine the files associated with the article
unable to download
able to add files to review
from that page the reviewer (section editor) is still unable to download.

**Expected behavior**
section editor should be able to download the files before applying them to review and after.

**Additional context**
:)

PR #2293, opened 2021-07-07
Hotfix PUBD-209 section editors should be able to download assigned files

hardyoyo:
Backports the three commits from https://github.com/BirkbeckCTP/janeway/pull/1925 to the preprint-remodel branch, which is due to be merged soon, but we happen to have preprint-remodel deployed to production, and would feel more comfortable using an official branch than a bespoke one.


hardyoyo:
Tested locally, passes tests, *and* also squashes the bug known internally at CDL as PUBD-209, for Janeway, the bug is known as #1924

ajrbyers:
Test this please.

PR #2299, opened 2021-07-15
Add 'Mx' as a salutation

alexdutton:
'Mx' is the most predominantly used gender-neutral title, and is widely accepted by organisations, including the UK Government.

This PR makes Janeway systems slightly more inclusive.

See [the Wikipedia article](https://en.wikipedia.org/wiki/Mx_%28title%29) for more information.

### DIFF
--- a/src/core/models.py
+++ b/src/core/models.py
@@ -55,6 +55,7 @@ SALUTATION_CHOICES = (
     ('Ms', 'Ms'),
     ('Mrs', 'Mrs'),
     ('Mr', 'Mr'),
+    ('Mx', 'Mx'),
     ('Dr', 'Dr'),
     ('Prof.', 'Prof.'),
 )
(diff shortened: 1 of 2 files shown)


Bbkctp:

mauromsl:
jenkins - add to whitelist

src/core/models.py
```diff
@@ -55,6 +55,7 @@ def profile_images_upload_path(instance, filename):
     ('Ms', 'Ms'),
     ('Mrs', 'Mrs'),
     ('Mr', 'Mr'),
+    ('Mx', 'Mx'),
```
mauromsl, line 58: We are going to need a migration  on `core` for this change I think.

mauromsl (changes requested):

mauromsl (changes requested):
Thanks @alexdutton !

Happy to include this change on the next release

src/core/models.py
alexdutton: Ooh, yes. I think I can just add that retrospectively to the initial migration, as it doesn't have any actual effect on the database schema (i.e. no `ALTER` needs to be emitted).

Is that okay, or a bit unintuitive compared to a fresh migration?

mauromsl: yep, that is fine by me

alexdutton:
@mauromsl I've just force-pushed an updated commit with the migration change and an updated commit message explaining why the change is on the initial migration instead of a new migration. Should be good for a re-review now :-)

alexdutton:
Thank you @mauromsl and @ajrbyers! :rainbow:

ajrbyers:
> Thank you @mauromsl and @ajrbyers! rainbow

Happy to merge this and slightly embarrassed that we hadn't previously added it!

PR #2318, opened 2021-07-28
Move return under if.

ajrbyers:


src/metrics/logic.py
```diff
@@ -237,20 +237,7 @@ def store_article_access(request, article, access_type, galley_type='view'):
...
-                    galley_type=galley_type,
-                ).order_by('-accessed')[0]
-
-                access.accessed = timezone.now()
-                access.save()
-
-            else:
-
+            if not exists:
                 access = models.ArticleAccess.objects.create(
```
mauromsl, line 241: A bit of an edge-case, but we should also pass in the original accessed time here, capturing it before the logic starts.

src/metrics/logic.py
```diff
@@ -237,20 +237,7 @@ def store_article_access(request, article, access_type, galley_type='view'):
                 galley_type=galley_type,
             ).exists()
 
-            if exists:
-                # get the most recent access attempt and reset its accessed to now.
-                access = models.ArticleAccess.objects.filter(
-                    identifier=identifier,
```
mauromsl, line 243: I can't ping the above code but counter recommends IP + user agent as the identifier when a tracking session/ cookie is not available

mauromsl (changes requested):
looks good, just a couple minor adjustments

mauromsl (approved):

PR #2322, opened 2021-07-29
Adds support for JATS outputting <span class=underline for <underline>

ajrbyers:


mauromsl (approved):

mauromsl (changes requested):
Since these are identical, I think it is best to move the CSS classes to common/css/common.css

ajrbyers:
Done.

PR #2324, opened 2021-07-29
Updates typography for admin and OLH theme header sizes.

ajrbyers:


mauromsl (changes requested):
One minor comment

src/static/admin/css/app.css
```diff
@@ -579,7 +579,7 @@ body {
   margin: 0;
   padding: 0;
   background: #f5f6f9;
-  font-family: "Merriweather", serif;
+  font-family: "Open Sans", serif;
```
mauromsl, line 582: Isn't it weird to fallback a sans font with a serif font?

PR #2331, opened 2021-08-05
#2328 update

### LINKED ISSUE #2328: Make article and abstract fields translatble on submission.Article
In order to fully support migration of multilingual content from other platforms, we need to at least be able to translete these two fields.

At first, making the fields translatable and allowing entering translations via metadata editor should suffice. We would have to enforce the default language in various views such as the submission system, outgoing emails and APIs /metadata deposits
.
Alternatively, we could fully implement support for these multilingual attributes

ajrbyers:
Closes #2328


src/utils/shared.py
```diff
@@ -107,18 +107,37 @@ def language_override_redirect(request, url_name, kwargs):
...
+        kwargs,
+        additional_query_strings=None,
+):
     reverse_string = "{}?language={}".format(
         reverse(url_name, kwargs=kwargs),
         request.override_language,
     )
+
+    for qs in additional_query_strings:
+        reverse_string = reverse_string + '&{}'.format(qs)
```
mauromsl, line 128: Should use `urllib.parse.urlencode here` to make sure query strings don't get mangled

mauromsl (changes requested):
One minor problem but otherwise looks great!

PR #2336, opened 2021-08-06
2315 bugfix

### LINKED ISSUE #2315: Articles in the "Unassigned" stage show a link to the review stage
When an article is in the unassigned stage we show a link to the review stage at the top. If the editor hits the link without moving the article to review, they will get a confusing message when attempting to assign a reviewer: "This article has already been accepted or declined."

One or more of the following are needed:
 - The link to the review stage should be disabled until a first decision is made. eg:
image
- If the editor manages to land onto the review page, they should be told that the article is not in the review stage yet and perhaps give them the option to move it to review at that point
- Alternatively, let the editor create review assignments as normal and "Move the article to review" when they do so.

ajrbyers:
- Disabled the review header link when an article is unassigned
- If somehow an editor makes it review there is now a callout warning (image below), this only works for articles in the Unassigned stage.
- Warning messages are now improved

Closes #2315 

image

### DIFF
--- a/src/review/views.py
+++ b/src/review/views.py
@@ -380,6 +380,10 @@ def in_review(request, article_id):
 
     if request.POST:
 
+        if 'move_to_review' in request.POST and article.stage == submission_models.STAGE_UNASSIGNED:
+            article.stage = submission_models.STAGE_UNDER_REVIEW
+            article.save()
+
         if 'new_review_round' in request.POST:
 
             # Complete all existing review assignments.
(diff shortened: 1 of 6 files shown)


src/review/views.py
```diff
@@ -380,6 +380,10 @@ def in_review(request, article_id):
 
     if request.POST:
 
+        if 'move_to_review' in request.POST and article.stage == submission_models.STAGE_UNASSIGNED:
+            article.stage = submission_models.STAGE_UNDER_REVIEW
+            article.save()
```
mauromsl, line 385: Should we check if there is an editor assigned as well?  This is what we do on the "Unassigned" screen.

mauromsl (approved):
One minor comment, otherwise good to merge!

ajrbyers:
Could assign the current user if they are an editor?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

mauromsl (approved):

Issue #2337, opened 2021-08-06
Make issue field optional rather than required


lstachew:
**Is your feature request related to a problem? Please describe.**
Many of our journals only use volume numbers, but the current system requires both a volume and issue number for a new set of content. 

**Describe the solution you'd like**
It would be great if the issue number field could be optional rather than required, even if this is assessed on a journal-by-journal basis

**Describe alternatives you've considered**
As a workaround, we've been entering "0" into the issue field for journals that only use volume numbers, but this means that on the journal's page, the volume is listed as "Volume 15, Issue 0" for instance.

PR #2338, opened 2021-08-09
2337 issue volume zero

mauromsl:
When the issue volume/number is set to zero, avoid displaying it.

It doesn't make the fields optional yet as per #2337, but I'm not sure defaulting to zero when a value is not provided makes a lot of sense. what do you think @ajrbyers


ajrbyers (approved):
I think this is a neat solution for now.

PR #2345, opened 2021-08-11
Preprint merge

ajrbyers:
Merges the Repository branch into master.


src/repository/views.py
```diff
@@ -0,0 +1,1755 @@
...
+    top_level_subjects = models.Subject.objects.filter(
+        parent__isnull=True,
+    ).prefetch_related('editors')
+
+    template = 'admin/repository/subjects.html'
+    context = {
+        'top_level_subjects': top_level_subjects,
+        'form': form,
+        'subject': subject,
+        'active_users': core_models.Account.objects.all()
```
ajrbyers, line 1216: missing comma

src/repository/views.py
```diff
@@ -0,0 +1,1755 @@
...
+        models.Subject,
+        pk=subject_id,
+        repository=request.repository,
+    )
+    subject.delete()
+
+    messages.add_message(
+        request,
+        messages.SUCCESS,
+        'Subject deleted. Any associated articles will have been orphaned.'
```
ajrbyers, line 1236: missing comma

src/repository/views.py
```diff
@@ -0,0 +1,1755 @@
...
+    """
+    Displays a list of preprints that have bee orphaned from subjects.
+    :param request: HttpRequest object
+    :return: HttpResponse
+    """
+    orphaned_preprints = repository_logic.list_articles_without_subjects()
+
+    template = 'admin/repository/orphaned_preprints.html'
+    context = {
+        'orphaned_preprints': orphaned_preprints
```
ajrbyers, line 1276: missing comma

src/repository/views.py
```diff
@@ -0,0 +1,1755 @@
...
+    preprint_authors = models.PreprintAuthor.objects.filter(
+        preprint=preprint,
+    )
+
+    for preprint_author in preprint_authors:
+        order = posted_author_pks.index(preprint_author.pk)
+        author_order, c = models.PreprintAuthor.objects.get_or_create(
+            preprint=preprint,
+            account=preprint_author.account,
+            defaults={'order': order}
```
ajrbyers, line 1350: missing comma

src/themes/OLH/templates/core/login.html
```diff
@@ -32,6 +32,11 @@
                         <h4 class="text-center">{% trans "Log in with your account" %}</h4>
                         <a href="{{ settings.ORCID_URL }}?client_id={{ settings.ORCID_CLIENT_ID }}&response_type=code&scope=/authenticate&redirect_uri={% orcid_redirect_uri %}" class="success button expanded">{% trans "Log in with ORCiD" %}</a>
                         {% endif %}
+                        <p>
+                            <small>
+                                Note: Login credentials are case sensitive. Please ensure you enter them in the same case you registered.
```
ajrbyers, line 37: This should be removed.

ajrbyers:
test this please

src/repository/models.py
```diff
@@ -0,0 +1,1033 @@
...
+
+
+fs_path = os.path.join('files/')
+preprint_file_store = JanewayFileSystemStorage(location=fs_path)
+preprint_media_store = JanewayFileSystemStorage()
+
+
+def preprint_file_upload(instance, filename):
+    try:
+        uuid_filename = str(uuid.uuid4()) + '.' + str(filename.split('.')[1])
```
mauromsl, line 65: should use os.path.splitext

src/repository/models.py
```diff
@@ -0,0 +1,1033 @@
...
+        uuid_filename = str(uuid.uuid4())
+
+    path = os.path.join('repos', str(instance.preprint.pk), uuid_filename)
+    instance.original_filename = filename
+    return path
+
+
+def repo_media_upload(instance, filename):
+    try:
+        filename = str(uuid.uuid4()) + '.' + str(filename.split('.')[1])
```
mauromsl, line 76: should use os.path.splitext

src/repository/models.py
```diff
@@ -0,0 +1,1033 @@
...
+        null=True,
+        storage=preprint_media_store,
+        upload_to=repo_media_upload,
+    )
+    hero_background = models.ImageField(
+        blank=True,
+        null=True,
+        storage=preprint_media_store,
+        upload_to=repo_media_upload,
+    )
```
mauromsl, line 117: Could use help text here

src/repository/models.py
```diff
@@ -0,0 +1,1033 @@
...
+    custom_js_code = models.TextField(
+        blank=True,
+        null=True,
+        help_text=_('The contents of this field are output into the JS area'
+                    'at the foot of every Repository page.')
+    )
+    live = models.BooleanField(
+        default=False,
+        verbose_name='Repository is Live?'
```
mauromsl, line 130: Might need help text here saying that the repository won't be accessible by anyone otherwise (not even managers)

src/repository/models.py
```diff
@@ -0,0 +1,1033 @@
...
+
+class PreprintAuthorManager(models.Manager):
+    def get_queryset(self):
+        return super().get_queryset().select_related('author')
+
+
+class PreprintAuthor(models.Model):
+    preprint = models.ForeignKey('Preprint')
+    account = models.ForeignKey('core.Account', null=True)
+    author = models.ForeignKey('Author', blank=True, null=True)
```
mauromsl, line 700: can we document somehow the difference between these two? I'm assuming they behave similar to frozen authors in journals, but the implementation is different in this case (e.g There are 3 affiliation records for an "author": one in account, one in Author and one in PreprintAuthor)

src/repository/models.py
```diff
@@ -0,0 +1,1033 @@
...
+    order = models.PositiveIntegerField(default=0)
+    objects = PreprintAuthorManager()
+    affiliation = models.TextField(blank=True, null=True)
+
+    class Meta:
+        ordering = ('order',)
+        unique_together = ('account', 'preprint')
```
mauromsl, line 707: Do we need a unique_together for 'author' and 'preprint'?

src/repository/models.py
```diff
@@ -0,0 +1,1033 @@
...
+    class Meta:
+        ordering = ('-date_time', '-pk')
+
+    def __str__(self):
+        return 'Comment by {author} on {article}'.format(
+            author=self.author.full_name(),
+            article=self.preprint.title,
+        )
+    
+    def mark_public(self):
```
mauromsl, line 847: `toggle_public` sounds like a better name, since it can also mark as not public this metod

src/repository/models.py
```diff
@@ -0,0 +1,1033 @@
...
+
+        # Create a new PreprintVersion, this will now be the current_version.
+        # If the current VersionQueue has no file (in the case of Metadata
+        # updates) use the preprint's current version's file.
+        PreprintVersion.objects.create(
+            preprint=self.preprint,
+            file=self.file if self.file else this_file,
+            version=self.preprint.next_version_number(),
+            moderated_version=self,
+            published_doi=self.published_doi,
```
mauromsl, line 957: Is it really worth tracking changes to this?

src/repository/models.py
```diff
@@ -0,0 +1,1033 @@
...
+
+@receiver(models.signals.post_delete, sender=PreprintFile)
+def auto_delete_file_on_delete(sender, instance, **kwargs):
+    """
+    Deletes file from filesystem
+    when corresponding `PreprintFile` object is deleted.
+    """
+    if instance.file:
+        if os.path.isfile(instance.file.path):
+            os.remove(instance.file.path)
```
mauromsl, line 1002: Is this desirable? This is not what we do when deleting on the journal side, we always keep the files.

In any case, it should probably catch the exception in case the file does not exist.

mauromsl (changes requested):

ajrbyers:
In discussion with @mauromsl we agreed that a test for the repo site should be added to make sure the pres/journal/repo domain/path middleware is operating as expected.

src/repository/models.py
ajrbyers: No, author is vestigial and is being removed.

PR #2363, opened 2021-08-23
#2362 add new event for article access.

### LINKED ISSUE #2362: Event for viewing and downloading galleys

In implementing the [NGLP plugin](https://github.com/NGLPteam/nglp-janeway-plugin) we'd like to be able to capture artcle download and view events.

In discussion with @ajrbyers I think we agreed that adding a new event type and raising it from `store_article_access` would provide a solution. This would cover `download_galley` and `view_galley` views, and take care of the/our COUNTER requirement.

We'd started implementing this as middleware (introspecting the resolver match to determine the view and article ID, but it's not particularly elegant.

ajrbyers:
Add a new event `ON_ARTICLE_ACCESS`

Closes #2362

### DIFF
--- a/src/events/logic.py
+++ b/src/events/logic.py
@@ -227,8 +227,8 @@ class Events:
     # raised when a workflow element completes to hand over to the next one
     ON_WORKFLOW_ELEMENT_COMPLETE = 'on_workflow_element_complete'
 
-    # kwargs:
-    # raised when an article is viewed or downlaoded
+    # kwargs: request, article, article_access
+    # raised when an article is viewed or downloaded
     ON_ARTICLE_ACCESS = 'on_article_access'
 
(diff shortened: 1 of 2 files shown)


src/events/logic.py
```diff
@@ -227,6 +227,10 @@ class Events:
     # raised when a workflow element completes to hand over to the next one
     ON_WORKFLOW_ELEMENT_COMPLETE = 'on_workflow_element_complete'
 
+    # kwargs: request, article, article_access
+    # raised when an article is viewed or downloaded
```
mauromsl, line 231: I'd add a comment mentioning that this is raised only for accesses that pass counter compliance checks.

mauromsl (approved):
Just one minor comment.

PR #2384, opened 2021-09-01
#2383 check if the abstract is not None

### LINKED ISSUE #2383: Abstract Header
**Describe the bug**
If an article's abstract is `None` the header appears with `None` as the abstract.

**Janeway version**
1.4-RC1

ajrbyers:
Closes #2383

### DIFF
--- a/src/themes/OLH/templates/journal/article.html
+++ b/src/themes/OLH/templates/journal/article.html
@@ -119,9 +119,9 @@
                     <h3>{{ article.title|safe }}</h3>
                 <hr/>
                 {% endif %}
-                            {% if article.abstract != '' %}
+                            {% if article.abstract and article.abstract != ''%}
                                 <h2>{% trans "Abstract" %}</h2>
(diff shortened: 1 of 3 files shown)


src/themes/OLH/templates/journal/article.html
```diff
@@ -119,9 +119,9 @@ <h3>{{ article.title|safe }}</h3>
                     <h3>{{ article.title|safe }}</h3>
                 <hr/>
                 {% endif %}
-                            {% if article.abstract != '' %}
+                            {% if article.abstract and article.abstract != ''%}
```
mauromsl, line 122: Doesn't `if article.abstract` already account for falsey values?

mauromsl (commented):
Just a comment, otherwise good to go

src/themes/OLH/templates/journal/article.html
ajrbyers: I think we've seen wierdness with this in the past so I've left if in. Can remove though if you think it'll be ok.

PR #2431, opened 2021-09-15
#2430 updated the nav item page to make it clearer when you are editi…

### LINKED ISSUE #2430: Nav Redirect
**Describe the bug**
After creating a new nav item users should be redirected to the base page rather than to the edit page for the current item.

ajrbyers:
…ng vs creating

Closes #2430

### DIFF
--- a/src/templates/admin/cms/nav.html
+++ b/src/templates/admin/cms/nav.html
@@ -22,8 +22,8 @@
             <form method="POST">
                 {% csrf_token %}
                 <div class="title-area">
-                    <h2>{% if nav_item_to_edit %}Currently Editing {{ nav_item_to_edit.link_name }}{% else %}Create New Nav Item{% endif %}</h2>
-                    {% if nav_item_to_edit %}<a href="{% url 'cms_nav' %}" class="button"><span class="fa fa-plus"></span> Add a New Nav Item</a>{% endif %}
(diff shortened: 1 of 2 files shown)


src/templates/admin/cms/nav.html
```diff
@@ -21,21 +21,27 @@
...
-                        <li>For top level nav items that you don't want to appear in the dropdown leave the Link field empty.</li>
-                    </ul>
-                </small>
-                {% include "admin/elements/forms/errors.html" %}
-                {{ form|foundation }}
-                <button type="submit" name="submit" class="success button">
-                    {% if nav_item_to_edit  %}Update{% else %}Create{% endif %}
-                </button>
+                <div class="title-area">
+                    <h2>{% if nav_item_to_edit %}Currently Editing {{ nav_item_to_edit.link_name }}{% else %}Create New Nav Item{% endif %}</h2>
```
mauromsl, line 25: rather than nav_item_to_edit, shall we check `form.instance`?
I think it is a bit more explicit and, in fact, I think this would be wrong in the case when a nav item is deleted? (there will be a `nav_item_to_edit` but `form.instance` will be `None`

mauromsl (changes requested):
Just one comment that might be a bug

mauromsl (approved):

Issue #2424, opened 2021-09-13
Preprint manager item page should have link to live article


alainna:
**Is your feature request related to a problem? Please describe.**
As a moderator, I would like to have a convenient link to the live version of a preprint from the moderator view. 

**Describe the solution you'd like**
Repository manager's item page should have a "View live article" link, similar to author manager's item page. 

**Additional context**
mockup: 
![image](https://user-images.githubusercontent.com/15632391/133135482-c42f6f82-810a-45d2-a9df-e59b1a9387e0.png)

ajrbyers:
Should be a nice simple PR if @tingletech or @hardyoyo have 10 mins!

tingletech:
@alainna can you tell me the URL path you see in the browser when you are on this page?

tingletech:
<img width="1150" alt="Screen Shot 2021-09-22 at 9 23 08 AM" src="https://user-images.githubusercontent.com/227374/134383282-100195a3-74f5-4e1f-af02-40b8092885c5.png">

alainna:
I'm seeing this feature as live in master so let's close it 
https://github.com/BirkbeckCTP/janeway/blob/master/src/templates/admin/repository/article.html#L22

PR #2455, opened 2021-09-22
add "View Live Article" link as per #2424

tingletech:


ajrbyers (changes requested):
Hey @tingletech  if you add class="button" to the <a> tag you don't need the div or the float!

A

PR #2463, opened 2021-09-23
Remove the sitemap link as its for comps not people

ajrbyers:


mauromsl (changes requested):
I think we still need this to be present so that crawlers can find it, just not displayed, though the recommendation from google scholar was to add the sitemap url onto robots.txt

Also, seems material is missing it?

ajrbyers:
@mauromsl I've added a link to the head portion which should be better than having an invisible link.

mauromsl (approved):

PR #2470, opened 2021-09-30
Embeds a Vimeo player via XSLT

### LINKED ISSUE #2330: [XSLT]Vimeo integration
Some journals at the OLH publish video articles (JER). They use Vimeo to host the videos and embed a video player in the rendered article. Janeway's XSLT does not support rendering an embedded player for video platforms and Vimeo could be a good start.

An XSLT directive that detects a media object with a vimeo URL and renderes and embeded player.

The alternative is to create an HTML galley just to include the embeded player but this is a cumbersome process that requires involving the tech department for what should be a typesetting/editorial task

**Additional context**
Videos are usually uploaded as supplementary files. Then it is up to a member of staff to upload the video to Vimeo and then passing the URL and other details to the typesetter. Instead, we could upload the video to Vimeo via their API and generate a DOI automatically for it. It could be done with a plugin that provides a "transform" option during typesetting which would create a remote galley, uploads the video and mints a DOI for it.

mauromsl:
Closes #2330

### DIFF
--- a/src/journal/views.py
+++ b/src/journal/views.py
@@ -383,7 +383,7 @@ def article(request, identifier_type, identifier):
         article_object.abstract = (
             "<p><strong>This is an accepted article with a DOI pre-assigned"
             " that is not yet published.</strong></p>"
-        ) + article_object.abstract or ""
+        ) + (article_object.abstract or "")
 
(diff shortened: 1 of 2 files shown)


ajrbyers:
This looks okay - how hard would it be to add support for Youtube at the same time as we have this coming up with Michigan?

A

PR #2479, opened 2021-10-04
Allows Galleys to marked as not for public

ajrbyers:
Tied to:

- https://github.com/BirkbeckCTP/typesetting/pull/141
- https://github.com/BirkbeckCTP/back_content/pull/27


src/journal/logic.py
```diff
@@ -111,14 +111,15 @@ def create_galley_from_file(file_object, article_object, owner=None):
...
     Attempts to get the best galley possible for an article
     :param article: Article object
     :param galleys: list of Galley objects
     :return: Galley object
     """
-    if article.render_galley:
+    if article.render_galley and article.render_galley.public:
         return article.render_galley
 
     try:
```
mauromsl, line 125: queries below also need checking for "publicness"

mauromsl (changes requested):
Looks good! there is one other case to filter which is when the article.render_galley is not set explicitely

mauromsl (approved):

PR #2508, opened 2021-10-13
Adds new bepress metadata to article page

### LINKED ISSUE #736: Crossref Journal Name
We should add a new field for "Crossref Journal Name" so we can override the default name if this field is supplied.

mauromsl:
closes #736 
- Adds a new field for publication title
- Renders the new field in the article page across themes (page numbers, publisher, publication title and article number)
- exposes new metadata fields via crossref/OAI


mauromsl:
closes #736

PR #2566, opened 2021-10-29
OAI_record now displays full names.

### LINKED ISSUE #2556: OAI creator should include full name
In my mind this is a bug, but it probably is a feature request. The OAI creator field is not displaying the complete author name. this will give poor results when sharing metadata as the initial plus surname form can be easily conflated with other authors.

For example, in https://pubs.lib.uiowa.edu/wwqr/api/oai/?verb=ListRecords&metadataPrefix=oai_dc
we have <dc:creator>Whitman, W.</dc:creator> for https://doi.org/10.13008/0737-0679.2415 -- the author name in the article is Whitman, Walt
and
<dc:creator>Mong, D.</dc:creator> for https://doi.org/10.13008/0737-0679.2420 -- the author name in the article is Mong, Derek

ajrbyers:
Closes #2556

### DIFF
--- a/src/submission/models.py
+++ b/src/submission/models.py
@@ -1462,10 +1462,15 @@ class FrozenAuthor(models.Model):
     @property
     def dc_name_string(self):
         name_string = ""
+
+        if self.is_corporate:
+            return self.corporate_name
+
         if self.last_name:
-            name_string += "{}{}, ".format(
+            name_string += "{}{}{} ".format(
                 self.last_name,
                 " {}".format(self.name_suffix) if self.name_suffix else '',
+                "," if self.first_name else '',
             )
         if self.first_name:
             name_string += "{}".format(self.first_name)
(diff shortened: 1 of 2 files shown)


ajrbyers:
@mauromsl we should add this one to RC8!

mauromsl:
> 
> 
> @mauromsl we should add this one to RC8!

I think so, perhaps move it to a property as you had suggested? (`FrozenAuthor.dc_name_string`)

src/submission/models.py
```diff
@@ -1459,6 +1459,21 @@ def full_name(self):
             full_name = "%s %s" % (full_name, self.name_suffix)
         return full_name
 
+    @property
+    def dc_name_string(self):
+        name_string = ""
+        if self.last_name:
+            name_string += "{}{}, ".format(
```
mauromsl, line 1466: The comma here should be added when there is a first name only, no?

We also need to consider the case for `is_corporate` (just call `self.corporate_name`)

mauromsl (changes requested):
couple minor comments

mauromsl (approved):

PR #2571, opened 2021-11-02
Addtional rows also now watch for size to make each row equal in height

ajrbyers:


mauromsl (approved):

mauromsl (changes requested):
The popular homepage element needs the same change
An optional comment - popular and featured are almost identical except for the title, worth moving them to the same shared template?

Issue #141, opened 2017-09-08
Peer Review: Display on Article Page


ajrbyers:
If a review was Open, we should add a link to display it on the article page.
closes #141

ajrbyers:
https://github.com/BirkbeckCTP/janeway/pull/2602

PR #2602, opened 2021-11-11
Add open peer review. #141

ajrbyers:


ajrbyers:
- [ ] Add Public/Open status to the new review assignment element template
- [ ] Add front end code to the clean theme as its missing from here currently
- [ ] Test this looks okay/works

ajrbyers:
Closes #141

src/themes/OLH/templates/elements/public_reviews.html
```diff
@@ -0,0 +1,32 @@
...
+                                <br/><br/>
+                            </p>
+                        {% endif %}
+                    {% endfor %}
+                    <p><b>Note:</b><br/>This review refers to round {{ review.review_round.round_number }} of peer
+                        review and may pertain to an earlier version of the document.</p>
+                    <button class="close-button" data-close aria-label="Close reveal" type="button">
+                        <span aria-hidden="true">&times;</span>
+                    </button>
+                </div>
```
joemull, line 26: How do we feel about adding review files here and in the other themes?

ajrbyers: Personally, I think we should push the structured review but I'm not against adding the file for download if its the general consensus.

ajrbyers:
- [ ] Add document on how this should work, editors should disable file upload etc.

src/submission/models.py
```diff
@@ -1273,6 +1273,16 @@ def active_reviews(self):
     def completed_reviews(self):
         return self.reviewassignment_set.filter(is_complete=True, date_declined__isnull=True)
 
+    @property
+    def completed_reviews_with_permission(self):
+        return self.reviewassignment_set.filter(is_complete=True, date_declined__isnull=True,
+                                                permission_to_make_public=True)
+
+    @property
+    def public_reviews(self):
```
mauromsl, line 1282: can be more DRY with `return self.completed.reviews_wih_permission.filter(display_public=True)`

src/submission/models.py
```diff
@@ -1273,6 +1273,16 @@ def active_reviews(self):
     def completed_reviews(self):
         return self.reviewassignment_set.filter(is_complete=True, date_declined__isnull=True)
 
+    @property
+    def completed_reviews_with_permission(self):
+        return self.reviewassignment_set.filter(is_complete=True, date_declined__isnull=True,
+                                                permission_to_make_public=True)
```
mauromsl, line 1279: can be more DRY with `return self.completed.reviews_wih_permission.filter(permission_to_make_public=True)`

src/utils/install/journal_defaults.json
```diff
@@ -2939,6 +2939,21 @@
             "default": "<p>You may upload a file in lieu of completing the review form. Please ensure you've covered all of the required fields below.</p>"
         }
     },
+    {
+        "group": {
+            "name": "general"
+        },
+        "setting": {
+            "description": "Enable open peer review mode on this journal.",
```
mauromsl, line 2947: We could use a better description here, explaining what this setting actually does

mauromsl (changes requested):
A couple minor changes in line.

It would be good to also update the journal settings section of the docs with perhaps a screenshot or two

joemull:
FYI I'm going to take this one @ajrbyers .

joemull (approved):
Implemented changes requested by @mauromsl

PR #2607, opened 2021-11-16
2601 enhancement

### LINKED ISSUE #2601: Journals can switch languages even if the language switcher is off
**Describe the bug**
In path mode, when you change a journal's language, the language, the cookie will be shared amongst every journal under that domain. This leads to journals getting translated to other languages even when they shouldn't be.

This also happened on non path journals, perhaps by matching the browser locale

**Janeway version**
v1.4

1. In a path mode install, switch one journal to german
2. visit another journal with language switching off
3. Journal is in German without option to revert back to english

**Expected behavior**
When switching to another journal, consider if the language selection from the previous journal is compatible with the current one.

ajrbyers:
Closes #2601

### DIFF
--- a/src/journal/middleware.py
+++ b/src/journal/middleware.py
@@ -17,7 +17,7 @@ class LanguageMiddleware(object):
         """
         Checks that the currently set language is okay for the current journal.
         """
-        if request.journal:
+        if request.journal and settings.USE_I18N:
             current_language = translation.get_language()
             available_languages = request.journal.get_setting(
                 group_name='general',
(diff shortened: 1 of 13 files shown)


src/journal/middleware.py
```diff
@@ -0,0 +1,45 @@
...
+logger = get_logger(__name__)
+
+
+class LanguageMiddleware(object):
+    @staticmethod
+    def process_request(request):
+        """
+        Checks that the currently set language is okay for the current journal.
+        """
+        if request.journal:
```
mauromsl, line 20: `if request.journal and settings.USE_I18N`?

mauromsl (changes requested):
Noice!

just 1 comment

mauromsl (approved):

PR #2618, opened 2021-11-18
#2585 removed excess <p> tags.

### LINKED ISSUE #2585: Extra empty p tags in references
Reference lists are rendered with extra `<p>` tags not in the XML. Creates the unwanted effect of prompting CSS to add padding between references and is also bad for accessibility. Reported by @pgoussy.

Version 1.4.0

1. Go to https://journals.publishing.umich.edu/tia/plugins/typesetting/preview_galley/article/964/galley/264/#heading3
2. Inspect HTML
3. See extra tags

Screenshot from 2021-11-04 12-23-19
Screenshot from 2021-11-04 12-25-10

ajrbyers:
- Removes P tags that wrap references
- Closes #2585

### DIFF
--- a/src/transform/xsl/default.xsl
+++ b/src/transform/xsl/default.xsl
@@ -1582,13 +1582,13 @@
   <xsl:template match="ref">
     <xsl:choose>
       <xsl:when test="count(element-citation)=1">
-          <p id="{@id}">
+          <p id="{parent::*/@id}">
             <xsl:apply-templates select="element-citation | nlm-citation"/>
           </p>
       </xsl:when>
       <xsl:otherwise>
         <xsl:for-each select="element-citation | nlm-citation | mixed-citation">
-            <p id="{@id}">
+            <p id="{parent::*/@id}">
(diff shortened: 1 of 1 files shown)


src/transform/xsl/default.xsl
```diff
@@ -1588,12 +1588,10 @@
       </xsl:when>
       <xsl:otherwise>
         <xsl:for-each select="element-citation | nlm-citation | mixed-citation">
-            <p id="{@id}">
```
mauromsl, line 1591: I can see the `template match="mixed-citation">` duplicates this logic, but `element-citation` and `nlm-citation` do not. Those citations wouldn't render any HTML elements with an `id` so I suspect internal links would stop working those objects. Should we remove the redundant <p> tag from mixed-citation instead?

mauromsl (changes requested):

ajrbyers:
image
Now looking much cleaner.

mauromsl (approved):

PR #2631, opened 2021-11-23
Add support email settings for manager page

### LINKED ISSUE #2586: Add tech support email address field on manager page
Currently there isn't a great spot for tech support email addresses for production service providers who use Janeway. We could add a field for this on the manager page that would be controlled by the settings file.

joemull:
Fixes #2586


joemull:
Need to add this to the press manager page

joemull:
@ajrbyers and @mauromsl as originally envisioned, can you remind me whether we wanted this on the press manager page, the journal manager page, or both? The issue was created for the press manager, but the work that has been done already and put into this PR is for the journal manager. For my part I think the press level is most important, then journal level

mauromsl (approved):
This is a very neat feature!

PR #2634, opened 2021-11-23
Make file submission help text a setting

### LINKED ISSUE #2496: Allow edits to file upload info text
Journal editors would like to be able to edit the text users are shown when uploading files here:
https://github.com/BirkbeckCTP/janeway/blob/master/src/templates/admin/elements/submit/file.html

I.e. the text that says "Please upload your manuscript file, it should be in PDF..."

Allow editors and other staff roles to edit the text for this upload form using a settings option.
Bildschirmfoto 2021-09-29 um 18 25 16

joemull:
Fixes #2496


src/templates/admin/elements/submit/file.html
```diff
@@ -8,17 +8,11 @@ <h5><i class="fa fa-upload">&nbsp;</i>{% trans "Add New File" %}</h5>
...
         <div class="card-section">
             <small>{% include "elements/forms/errors.html" with form=form %}</small>
-            {% if modaltype == 'data' %}
-                <p>{% trans "Please supply any original figure files or supporting information. The journal urges you to submit this document to a repository such as Figshare so that a DOI can be assigned to it, but this is not required." %}</p>
-            {% else %}
-                {% if not journal_settings.general.limit_manuscript_types %}
-                    <p>{% trans "Please upload your manuscript file, it should be in PDF, RTF, ODT or DOC/X format. LaTeX may be accepted later in the process but a simple, readable file is required at this stage." %}</p>
-                {% else %}
-                    <p>{% trans "Please upload your manuscript file, it should be in RTF, ODT or DOC/X format. LaTeX may be accepted later in the process but a simple, readable file is required at this stage." %}</p>
+            {% if modaltype == 'manuscript' %}
```
mauromsl, line 11: We loose the handling of `modaltype == 'data'` here, is that intended?

src/templates/admin/elements/submit/file.html
```diff
@@ -8,17 +8,11 @@ <h5><i class="fa fa-upload">&nbsp;</i>{% trans "Add New File" %}</h5>
...
+                {% if journal_settings.general.limit_manuscript_types %}
+                    <p>{% trans "File types limited to Doc, Docx, RTF, and ODT." %}</p>
```
mauromsl, line 13: Changing this text would require requesting new translations to our contributors, which makes me wonder if it is at all necessary since the text is not very different from what we had previously

src/templates/admin/submission/submit_files.html
```diff
@@ -20,19 +20,17 @@
         <div class="row expanded">
             <div class="large-12 columns">
 
-                <p><strong>{% trans "Submission guidelines" %}</strong>: {% blocktrans %}To upload a file, select it
-                    using one of the 'Choose file' buttons, then upload it with the 'Upload file' button next to it. You
-                    will then be asked for some additional information (label, description etc.){% endblocktrans %}</p>
+                <p>{{ file_submission_guidelines }}</p>
```
mauromsl, line 23: We loose the "**Submission Guidelines**" bold text here, which is not necessarily a problem

src/templates/admin/elements/submit/file.html
joemull: Yes, it is intentional. I removed it since the one file submission guidelines field will take the place of all this help text. The only reason we still need `if modaltype== 'manuscript'` is to show the restriction on filetypes when the limit_manuscript_types setting is checked.

src/templates/admin/elements/submit/file.html
joemull: The change is to fix a grammatical "error" that many editors might find distracting. We don't copyedit things before sending them for translation, do we?

src/templates/admin/submission/submit_files.html
joemull: Yes, that is intentional. I wanted to give editors control over all the text that appears here.

mauromsl: makes sense

src/templates/admin/elements/submit/file.html
mauromsl: We do not.

In my opinion we should only use variable names here rather than the actual text and store the english text in the .po file instead. (the .po files is where we store the translation strings)

Alternatively, you can change the name of this string (`msgid`) on the PO file for the other languages so that we can preserve the translations that we currently have (e.g.: https://github.com/BirkbeckCTP/janeway/blob/84a0d087926ef4f779bc37c6d78ad15df5f3c7f2/src/core/locales/nl/LC_MESSAGES/django.po#L812)

src/templates/admin/elements/submit/file.html
mauromsl: Sounds good to make it configurable.

PR #2660, opened 2021-12-10
Add frozen_biography and biography() to FrozenAuthor

### LINKED ISSUE #2659: Add frozen_biography to FrozenAuthor
FrozenAuthor objects do not have biographies associated with them. We need to add these in order to import back content containing biographies.

joemull:
Fixes #2659 --this is related to simultaneous work on imports


src/submission/models.py
```diff
@@ -1386,6 +1386,7 @@ class FrozenAuthor(models.Model):
 
     institution = models.CharField(max_length=1000)
     department = models.CharField(max_length=300, null=True, blank=True)
+    frozen_biography = models.TextField(null=True, blank=True, verbose_name=_('Biography'))
```
mauromsl, line 1389: @joemull this change should require a migration, you can generate one with `python src/manage.py makemigrations submission`

mauromsl (changes requested):
Looks good! just needs a migration

src/submission/models.py
```diff
@@ -1386,6 +1386,7 @@ class FrozenAuthor(models.Model):
 
     institution = models.CharField(max_length=1000)
     department = models.CharField(max_length=300, null=True, blank=True)
+    frozen_biography = models.TextField(null=True, blank=True, verbose_name=_('Frozen Biography'))
```
mauromsl, line 1389: One last change needed @joemull:
We need to update the `submissions.forms.EditFrozenAuthor` class and add the `frozen_biography` field so that it can be edited by regular means if needed.

I would also add a help_text to this field similar to what we have for frozen_orcid, explaining how account/frozen author biography fields play with each other.

mauromsl (changes requested):
Great, one last change and we are good to go!

src/submission/models.py
joemull: OK I've added these things

PR #2684, opened 2022-01-07
Allow combining domain and path modes

### LINKED ISSUE #2256: Add support for combining path and domain mode journals
Janeway includes a handy middleware for running sites in either path mode or domain mode. Some partners would like to enhance this feature by allowing some journals to run in domain mode and others in path mode within the same press. This is currently not possible because when the press runs on path mode, our custom middleware injects the journal code as the PATH_INFO, leading to all URLs generated from Janeway to have a leading /journal_code/ path. 

As a staff member, I can flag a journal as "domain" URL mode, despite path mode having been enabled in the settings.py file.
When a journal has domain mode enable, the custom middleware will ignore path mode rules and generate URLs using domain mode rules  instead.

**Additional context**
Once this feature is added, it might be possible to remove the URL_MODE setting and rely only on the configuration set for the journal.

mauromsl:
- The domain field is now optional
 - Logic for retrieving a site object from path or domain moved from middleware to concrete models.
 -  Journals and Repos are always available under their path mode URL
 - Journals and repos with a domain will resolve correctly when their domain is used
 - the URL_CONFIG variable now only indicates if the preferred URL is the path or the domain one
 - Included migrations to clear out domain placeholders on path based installations (they had press/journal_code as their domain)
 - Updates various pieces of help text that referenced the old behaviour of URL_CONFIG
 - Updates forms and views that expected the site domain as a mandatory field
closes #2256


ajrbyers:
test this please

ajrbyers (approved):

PR #2686, opened 2022-01-07
Add default journal support message to press manager view and template

### LINKED ISSUE #2673: Add support contact message to press manager
Do this but for the press mananger: https://github.com/BirkbeckCTP/janeway/issues/2586

joemull:
Fixes #2673

### DIFF
--- a/src/core/logic.py
+++ b/src/core/logic.py
@@ -846,7 +846,7 @@ def render_nested_setting(
     setting = setting_handler.get_setting(
         setting_group,
         setting_name,
-        request.journal if request.journal else None,
+        request.journal,
     ).value
 
     setting_context = {}
--- a/src/core/logic.py
+++ b/src/core/logic.py
@@ -846,7 +846,7 @@ def render_nested_setting(
     setting = setting_handler.get_setting(
         setting_group,
         setting_name,
-        request.journal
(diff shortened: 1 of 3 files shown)


src/core/logic.py
```diff
@@ -846,7 +846,7 @@ def render_nested_setting(
     setting = setting_handler.get_setting(
         setting_group,
         setting_name,
-        request.journal
+        request.journal if request.journal else None,
```
mauromsl, line 849: Is this change necessary?

If request.journal is None, then there is no need for the if statement.
This would only be relevant if request.journal was set to `False` or a Falsey value like empty string or empty list

mauromsl (commented):
Just a question

src/core/logic.py
joemull: I wasn't sure about how the request object was created and whether it would have been assigned anything for `request.journal` at the press level, and figuring that out would have taken some time. If you say it's not needed, then I'll take out the line.

For future work with request objects, where in the codebase are they created?

ajrbyers: request.journal and request.press always exist in the request/response cycle and are set in `core.middleware`.

mauromsl (approved):

PR #2689, opened 2022-01-10
Adds a code field to Issue allowing for verbose urls

mauromsl:
- Adds Issue.code field
 - Tidies up the edit issue form showing a preview of the issue images and removing the width limit on the modal


src/journal/models.py
```diff
@@ -459,6 +464,15 @@ class Issue(models.Model):
         through='IssueEditor',
     )
 
+    code = models.CharField(
```
ajrbyers, line 467: Could this not just be a slugfield for simplicity?

ajrbyers (approved):
One question but looks ok.

src/journal/models.py
mauromsl: > Could this not just be a slugfield for simplicity?

yep, that works. I always think of SlugField as an auto-generated field, but it doesn't need to be

PR #2709, opened 2022-01-17
Tweak the FN layout to make scrolling better

ajrbyers:


src/transform/xsl/default.xsl
```diff
@@ -430,18 +430,18 @@
     </xsl:template>
 
     <xsl:template match="fn-group/fn">
-        <li>
+        <xsl:variable name="fn-number">
+            <xsl:number level="any" count="fn[not(ancestor::front)]" from="article | sub-article | response"/>
+        </xsl:variable>
+        <li id="fn{$fn-number}">
+            <span class="footnotemarker" id="n{$fn-number}"></span>
```
mauromsl, line 437: Doesn't this lead to two DOM elements having identical IDs?

mauromsl (changes requested):

src/transform/xsl/default.xsl
ajrbyers: On is `fn{id}` and ther other is `n{id}` the link in the text uses `nm{id}`. I think we added the n{id} one as other stuff was using the `fn{id}` one and we didn't want to mess around with it.

src/transform/xsl/default.xsl
```diff
@@ -430,18 +430,18 @@
...
             <xsl:apply-templates/>
+            [<a href="#nm{$fn-number}"><sup>^</sup></a>]
```
mauromsl, line 439: The span now closes before this is rendered, so the footnotemarker is no longer inside an element with the class `footnotemarker` which might screw up CSS overrides

src/transform/xsl/default.xsl
mauromsl: sounds good

src/transform/xsl/default.xsl
ajrbyers: From what I can see there are no references to footnotemarker, I could add it to the <a>?

mauromsl: > From what I can see there are no references to footnotemarker in the css, I could add it to the though?

I believe some of the new journals used it to tweak the colour and size so it is worth adding it to the `<a>` here.

Also, not much point having it on the `<span>>` anymore since it is empty

mauromsl (approved):

Issue #2713, opened 2022-01-18
Templates create divs for empty text settings


joemull:
**Describe the bug**
HTML templates are sometimes populated with `<p><br></p>` when there shouldn't be anything. This happens when someone edits a text setting to be blank, since summernote saves this value for empty fields.

**Janeway version**
1.4.0

PR #2714, opened 2022-01-18
Ignore empty p and br tags from empty summernote fields

joemull:
What do you think of this as a fix for #2713 -- is it the right way to go about it?

### DIFF
--- a/src/core/models.py
+++ b/src/core/models.py
@@ -32,6 +32,7 @@ from copyediting import models as copyediting_models
 from submission import models as submission_models
 from utils.logger import get_logger
 from utils import logic as utils_logic
+from core.janeway_global_settings import SUMMERNOTE_SENTINEL
 
 fs = JanewayFileSystemStorage()
 logger = get_logger(__name__)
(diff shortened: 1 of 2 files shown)


src/core/models.py
```diff
@@ -32,6 +32,7 @@
 from submission import models as submission_models
 from utils.logger import get_logger
 from utils import logic as utils_logic
+from core.janeway_global_settings import SUMMERNOTE_SENTINEL
```
mauromsl, line 35: @joemull this is a django antipattern: The settings object is not a module, but rather a lazy loaded singleton. In order to ensure it is only loaded once and only when expected, it needs to be imported as `from django.conf import settings` (https://docs.djangoproject.com/en/1.11/topics/settings/#using-settings-in-python-code)

Janeway always loads the `janeway_global_settings.py` module when the `settings` object is first loaded so you can then call `settings.SUMMERNOTE_SENTINEL` even for those using a custom settings file.

All that said, I think it might be simpler to store the SUMMERNOTE_SENTINEL as a constant either on `core.forms` or `core.models`

mauromsl (changes requested):

src/core/models.py
joemull: Got it--thanks for catching this. I moved it to `core.models`.

mauromsl (approved):

PR #2740, opened 2022-02-01
Make Account.institution and FrozenAuthor.institution optional

### LINKED ISSUE #2304: Removing author institution requirement in metadata
The author institution field is currently required for article metadata, but many of our authors do not have any institutional affiliation.

It would be great if (for Michigan Publishing's instance on Janeway) that the institution field is optional rather than required.

In the meantime, I've been entering "N/A" in the institution field for authors without affiliation, but this means that "N/A" shows up next to the author's name on their article

joemull:
Fixes #2304


ajrbyers (commented):
Just to check: have you looked at templates to ensure there is nowhere we assume an institution is available?

joemull:
Yes, I checked the templates and author.institution is either called in parallel with other fields that can be blank (like author.department) and so will work just like they do, or it is called inside if statements that check it exists. So I believe this should be OK, but I was going to ask you if there is anything else we should look at.

ajrbyers:
> Yes, I checked the templates and author.institution is either called in parallel with other fields that can be blank (like author.department) and so will work just like they do, or it is called inside if statements that check it exists. So I believe this should be OK, but I was going to ask you if there is anything else we should look at.

Ace. We will probably need to check all the plugins too.

joemull:
Mmm that makes sense. I can do that.

joemull:
OK, I went through the plugins and all the calls of institution look OK.

mauromsl (approved):

PR #2758, opened 2022-02-14
Allow author enrolement to be vetted by a staff member

### LINKED ISSUE #2451: Manage Registrations
Some journals would like to approve new registrations.

- Add a setting to enable or disable this feature
- If enabled, use an alternative template that doesn't have a link for the Registrant to activate their account
- Email the Admins/Staff letting them know there is a new registration
- Allow editors to create additional registration questions
- Update that inactive users page to display details and activate buttons for new registrants
- Allow editors to remove accounts (but not to delete so they cannot recreate an account)

ajrbyers:
When the new setting is enabled, all users wanting to acquire the author role must request it.

Close #2451


ajrbyers:
Can I request that someone has a good look over the changes to journal_defaults.json? The conflict here was a total nightmare and I want fresh eyes on it.

mauromsl (changes requested):
@ajrbyers Can we have some unit tests, testing the submission start page and role acquisition with the setting on and off?

src/security/decorators.py
```diff
@@ -1164,3 +1164,39 @@ def keyword_page_enabled_wrapper(request, *args, **kwargs):
...
+        if request.journal and request.journal.get_setting('general', 'limit_access_to_submission'):
+            if not request.user.is_author(
+                request,
+            ):
+                return redirect(
+                    reverse(
+                        'request_submission_access'
+                    )
+                )
+
```
mauromsl, line 1199: Shoudl editors and stafff/superusers have the ability to submit even if the setting is enabled and they lack the author role?

mauromsl (changes requested):
Thanks for the tests, a couple last questions:

 1. I think the new decorator `submission_authorised_wrapper` would block anyone without the author role to submit, including editors/staff members. I think it would make sense to let them pass the checks in this decorator as per the other reviews, no?
 2. When a user adds further authors to a paper, they'll grant those users the author role regardless of the new feature implemented on this branch (https://github.com/BirkbeckCTP/janeway/blob/f3a0b414923e5a9b538439cf8b5b20af4653fbbf/src/submission/logic.py#L32) We might need to check the setting value for the site in the request context before doing so (e.g `if give_role and not request.site_type.limit_access_to_submission:`)

src/security/decorators.py
ajrbyers: Good catch. `is_author` works for staff but not for editors and obviously works for neither on repos here.

Suggest: `if staff or (repo and is moderator) or (journal and is editor): return func` right at the start before the logic ?

mauromsl: ỳeah, there is an `editor_or_manager` decorator in this file we might be able to re-purpose

mauromsl (approved):

PR #2767, opened 2022-02-21
Robots & Sitemaps

ajrbyers:
Adds robots and sitemap support.

- Files must be generated or the built in views return a 404
- For faster serving nginx should be configured to serve the generated static files


src/templates/common/journal_sitemap.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+
+    # Issues
```
mauromsl, line 4: I don't think this is valid XML

src/templates/common/repo_sitemap.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+
+    # Subjects
```
mauromsl, line 4: I don't think this is valid XML

src/templates/common/site_map_index.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+
+    # Journals
```
mauromsl, line 4: I don't think it is valid XML

src/templates/common/site_map_index.xml
```diff
@@ -0,0 +1,16 @@
...
+    {% for journal in journals %}
+    <sitemap>
+        <loc>{{ journal.site_url }}/sitemap.xml</loc>
+    </sitemap>
+    {% endfor %}
+    # Repositories
```
mauromsl, line 10: I don't think it is valid XML

src/utils/management/commands/generate_sitemaps.py
```diff
@@ -0,0 +1,117 @@
...
+                    'common/subject_sitemap.xml',
+                    {'subject': subject},
+                )
+                subject_file_path = os.path.join(
+                    repo_storage_path,
+                    '{}_sitemap.xml'.format(subject.pk)
+                )
+                with open(subject_file_path, 'w') as file:
+                    file.write(subject_sitemap)
+                    file.close()
```
mauromsl, line 117: I think it would be best to move the generation logic to its own function, then a wrapper or parallel function to write to disk. It would make the code easier to test and re-use.

This command would not have any business logic and would just act as CLI interface

docs/source/robotsandsitemaps.rst
```diff
@@ -0,0 +1,63 @@
...
+
+Janeway has a built in view that can handle the serving of the sitemaps files but you can also configure your webserver to serve these files for you, this can be quite complex when in domain mode and may best be left to Janeway to handle however.
+
+Custom Robots/Sitemaps
+----------------------
+If you don't want Janeway to serve robots or sitemap files you can configure your webserver to handle the URL routes that Janeway uses.
+
+Cron
+----
+Generation of sitemap files needs to be regular to ensure they are up to date. Janeway's ``install_cron`` command has been updated to install this command for you if you're using crontab. However you will need to schedule this manually otherwise, currently we recommend you regenerate files every 30 minutes.
```
mauromsl, line 63: Is this true?

mauromsl (changes requested):
I've added a few comments inline.

On top of that, we'd need at least a test for the XML generation of an issue and the press site

docs/source/robotsandsitemaps.rst
ajrbyers: It is now. I forgot to push the change.

PR #2770, opened 2022-02-23
Fix domain journal url rendering while in browsing from path

### LINKED ISSUE #2769: URLs in emails sometimes include the journal code for domain mode sites
**Describe the bug**
When browsing a domain journal in path mode, some rendered URLs will still include the journal code in the URL, leading to a 404.

For example:
 - Journal has a domain of `journal.com`
 - Press has a domain of `press.com`

**Janeway version**
v1.4.1-RC-1

**To Reproduce**
1. Create a journal with code `test` and domain `journal.com`
2. Configuration URL_CONFIG is set to `domain` 
3. Visit the journal on its path mode URL: `http://press.com/test`
4. Emails sent from the system will have the following structure: `http://journal.com/test`

**Expected behavior**
URLs sent with domain mode while browsing in path mode should be well formed and not include the unnecessary journal code in the path.

mauromsl:
closes #2769

### DIFF
--- a/src/journal/models.py
+++ b/src/journal/models.py
@@ -299,6 +299,11 @@ class Journal(AbstractSiteModel):
 
     def site_url(self, path=""):
         if self.domain and not settings.URL_CONFIG == 'path':
+
+            # Handle domain journal being browsed in path mode
+            site_path = f'/{self.code}'
+            if path and path.startswith(site_path):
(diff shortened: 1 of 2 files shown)


ajrbyers (approved):
Hello f-strings. 🎖️

PR #2778, opened 2022-03-03
Fixes Keywords not saving due to cleaned data not being mutable

mauromsl:
Resolves a bug introduced in 0c38a3fd59725429cba34458636cae87322ea4df due to a pointless overprotective feature


joemull (approved):
All good--submitting new article works fine now.

PR #2780, opened 2022-03-03
Add option to display page numbers and article numbers in on issue pages

joemull:


src/journal/models.py
```diff
@@ -574,29 +585,32 @@ def is_serial(self):
...
+        if journal.display_article_number and article and article.article_number:
+            article_number = ugettext("Article") + " {}".format(article.article_number)
+        if journal.display_article_page_numbers and article:
+            if article.page_range:
+                page_numbers = article.page_range
+            elif article.total_pages:
+                if article.total_pages > 1:
+                    page_numbers = str(self.total_pages) + ugettext('pages')
+                else:
+                    page_numbers = str(self.total_pages) + ugettext('page')
```
mauromsl, line 612: I believe we need to add a space in these string concatenations, perhaps a format string would make it more clear, though I see it wasn't done that way above.

src/journal/migrations/0049_auto_20220303_1636.py
```diff
@@ -0,0 +1,25 @@
...
+
+    dependencies = [
+        ('journal', '0048_merge_20220110_1933'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='journal',
+            name='display_article_number',
+            field=models.BooleanField(default=False, help_text='Some journals give each article in a volume its own number rather than subdividing the volume into issues. If you display article numbers, it is recommended to turn off display of issue numbers. Note that article number is an optional field separate from article ID and can be set in Edit Metadata.'),
```
mauromsl, line 18: Just curious about the tip `If you display article numbers, it is recommended to turn off display of issue numbers`. Is that the case? We have a number of journals that have used Volume, Issue, and article numbers in the past

mauromsl (changes requested):
One last minor thing to change

src/journal/migrations/0049_auto_20220303_1636.py
joemull: Ah interesting--in that case I will change the tip.

src/journal/models.py
joemull: I reworked this and added a few assertion statements to the test to cover the different possible combos.

mauromsl (approved):

PR #2783, opened 2022-03-04
Author display name handles empty first or last name fields

joemull:


ajrbyers:
I think it should be fine but we will need to check if Crossref will break.

joemull:
@ajrbyers sounds good--would you like me to check into that or make it part of review?

ajrbyers (approved):
Neat approach. 🎖️

Issue #2799, opened 2022-03-17
Sitemaps: lastmod should be better


ajrbyers:
**Is your feature request related to a problem? Please describe.**
Sitemaps currently display the pub date, this isn't ideal. We should display a better date.

**Describe the solution you'd like**
Consider:

- Adding a last modified stamp to article
- Checking upload date of galleys?

PR #2804, opened 2022-03-21
#2799 Applies new last modified model to get a better lastmod date for articles.

ajrbyers:


src/core/model_utils.py
```diff
@@ -358,3 +360,59 @@ def is_svg(f):
...
+                remote_field = manager.source_field_name
+                related_filter = {remote_field: self}
+                objects = manager.through.objects.filter(**related_filter)
+            elif field.one_to_many:
+                accessor = getattr(self, field.get_accessor_name())
+                objects = accessor.all()
+
+            for object in objects:
+                if hasattr(object, 'last_modified'):
+                    dates.append(getattr(object, 'last_modified'))
```
mauromsl, line 414: I suspect this needs to be a recursive call, right? Or have an argument to make it optional at least.

mauromsl (changes requested):
Looks good, I quite like the approach :)

One important comment regarding the calculation of the most recentd date_modified: I think this process needs to be recursive.
Also, we have no tests for the new behaviour

src/core/model_utils.py
```diff
@@ -358,3 +360,59 @@ def is_svg(f):
...
+class AbstractLastModifiedModelManager(models.Manager):
+    def get_queryset(self):
+        # this is to use your custom queryset methods
+        return AbstractLastModifiedModelQuerySet(self.model, using=self._db)
+
+
+class AbstractLastModifiedModel(models.Model):
+    last_modified = models.DateTimeField(
+        auto_now=True,
+        auto_now_add=False,
```
mauromsl, line 380: We should drop this as well, as want the date_modified to be populated on creation too.

ajrbyers:
> Looks good, I quite like the approach :)

I'm glad you like the approach you recommended ;)

ajrbyers:
@mauromsl can we carve some time tomorrow to take a look at the issues we discussed yesterday namely:

1. Recursive call
2. FKs not beeing looked up

PR #2805, opened 2022-03-22
Support multi-graphic figures

mauromsl:


ajrbyers:
@mauromsl is this for 1.4.1?

PR #2806, opened 2022-03-23
db backend names typos

gamboz:
In the comments, the backend name for postgres was missing the "ql" at
the end, and the one for mysql was pointing to sqlite.


Bbkctp:

ajrbyers:
Thanks @gamboz

PR #2584, opened 2021-11-04
#2583 removed whitespace, ensured that each group of 4 editors is on …

ajrbyers:
…their own line.

Closes #2583

PR #2824, opened 2022-04-01
#2584 adds support for OIDC login.

### LINKED ISSUE #2548: Keycloak Integration
NGLP are using Keycloak as their SSO solution.

To make Janeway work nicely with other NGLP software (AD, WDP etc) we should add the ability to use Keycloak for authentication. Authorisation will remain part of Janeway.

N/a

**Additional context**
NGLP requires this.

ajrbyers:
- Adds support for OIDC thru Mozilla's Django OIDC package.
- Closes #2548


docs/source/oidc.rst
```diff
@@ -0,0 +1,53 @@
...
+        'utils.oidc.JanewayOIDCAB',
+    )
+
+.. note:: All of the urls above are examples derived from a Keycloak instance. You can get the real URLs by using your service providers .well-known url which will look something like:     https://exmaple/org/.well-known/openid-configuration. You can get your Client ID and Secret from your service provider admin console or can request one from your technical staff. If you use RS256 for OIDC_RP_SIGN_ALGO you must also complete OIDC_OP_JWKS_ENDPOINT otherwise login will fail.
+
+New Accounts
+------------
+When a user without an account on Janeway logs in a new active account is generated for them with their Email address as the django username, first name and last name are also copied.
+
+If you want to stop OIDC from creating accounts you can add `OIDC_CREATE_USER = False` to your settings file.
```
mauromsl, line 43: Just curious as to what would happen in Janeway if this is set to False and I log in with an account that does not exist in Janeway

src/utils/oidc.py
```diff
@@ -0,0 +1,43 @@
...
+
+        return user
+
+    def update_user(self, user, claims):
+        user.first_name = claims.get('given_name', '')
+        user.last_name = claims.get('family_name', '')
+        user.is_active = True
+        user.save()
+
+        return user
```
mauromsl, line 39: Can we have a test case for these?

mauromsl (changes requested):

src/utils/oidc.py
ajrbyers: No you can’t. Go away.

docs/source/oidc.rst
ajrbyers: Dropped this line.

mauromsl (approved):
exciting!

PR #2842, opened 2022-04-07
Light copyedits

joemull:
Made these light edits while comparing the documentation content to the behavior of the 1.4.1 interface.


hardyoyo (approved):
👍 obviously correct copy edits

hardyoyo:
I think a bunch of commits got pulled in to this branch by mistake?

joemull:
Hi @hardyoyo , sorry about the messy commit history--I rebased this branch onto the latest "master", which has the effect of committing everything that has happened on "master" from when the branch was originally created up to now. I might scrap this branch and start a new one, though, as we are discussing whether to convert the end-user documentation to pure markdown at the same time as I do light copyedits. I will check with Andy and Mauro.

joemull:
Now that I know more about our desired behavior for `git rebase`, I can see this is not what we want. Closing!

PR #2855, opened 2022-04-21
Add base class for filterable class-based view

joemull:
Here is the work in progress for the CBV base class. I've put it on core and chosen a url and template name ('article_list`) that will work well with the default naming conventions of CBVs in later versions of Django.

I also implemented pagination via a template element, forming the template so it works with existing CSS.

What else should be included at this stage, before we create more specialized views?

### DIFF
--- a/src/templates/common/elements/pagination.html
+++ b/src/templates/common/elements/pagination.html
@@ -1,19 +1,39 @@
 {% load pages %}
 
-<div class="pagination-block">
-    <ul class="pagination">
-        {% if page_obj.has_previous %}
-            <li class="arrow"><a href="{{ request.path }}?page={{ page_obj.previous_page_number }}">&laquo;</a></li>
-        {% endif %}
+<div class="pagination-block input-group">
+    <form method="GET">
+        <div class="input-group-field small-8 columns">
+            {% if is_paginated %}
+                <ul class="pagination">
+                    {% if page_obj.has_previous %}
+                        <li class="arrow"><button name="page" value="{{ page_obj.previous_page_number }}">&laquo;</button></li>
+                    {% endif %}
 
-        {% for page in page_obj|slice_pages:3 %}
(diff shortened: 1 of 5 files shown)


src/templates/common/elements/pagination.html
```diff
@@ -0,0 +1,19 @@
...
+        {% for page in page_obj|slice_pages:3 %}
+            <li class="{% if page_obj.number == page.number %}current{% endif %}">
+                <a href="?page={{ page.number }}">{{ page.number }}</a>
+            </li>
+        {% endfor %}
+
+        {% if page_obj.has_next %}
+            <li class="arrow"><a href="{{ request.path }}?page={{ page_obj.next_page_number }}">&raquo;</a></li>
+        {% endif %}
+    </ul>
```
mauromsl, line 18: This is a usefull common element. One other thing I would add is a button for first and last page, which you can workout with `article_list.num_pages`.

In order to make it more re-usable, I would use a name like `paginated_results` and then include this element above with:
`{% include "common/elements/pagination.html"  with paginated_results=article_list%}`

joemull: I've now added the button for first and last via a new template tag, `slice_pages_with_first_last_ellipsis`.

As for reusability, I'm not clear on how `paginated_results` that would make it more reusable. I don't refer to article_list in this template, just `page_obj` (and with the other commit this morning, `is_paginated` and `paginate_by`, all three of which are now passed by the CBV base class.

Maybe related--should this be reusable beyond CBVs?

mauromsl (approved):

src/templates/common/elements/pagination.html
mauromsl: With your new template tag, my previous comment is irrelevant. I also like using the paginator rather than the queryset for consistency with what `slice_pages` used to do. Well done!

PR #2867, opened 2022-05-11
Fixes captcha display on the disabled front end contact form.

ajrbyers:


hardyoyo (approved):
Can confirm this PR puts a captcha on the journal contact page, on a site that has the "front end" disabled. 👍

mauromsl (approved):

PR #2870, opened 2022-05-11
2869 adds additional filters to limit the scope of views to the current repository where required

### LINKED ISSUE #2869: Preprints: Add support for multitenancy dashboards
**Describe the bug**
In a Janeway instance with multiple preprints repositories, a single preprint repository displays all preprints from the instance -- there is no filtering for only that one repository. 

**Janeway version**
Janeway 1.4.1 Preprints 

**To Reproduce**
1. In a Janeway instance which has one preprints repository (Repository A), add a second preprints repository (Repository B)
2. Log into Repository B using a moderator account which also has submitted preprints to Repository A. 
3. View author dashboard. See preprints from Repository A in Repository B
4. View moderator dashboard. See preprints from Repository A in Repository B

**Expected behavior**
Repository B would only display Repository B preprints. 

**Screenshots**
Demo Preprints repository displays EarthArXiv preprints
image

ajrbyers:
- Closes #2869

### DIFF
--- a/src/repository/logic.py
+++ b/src/repository/logic.py
@@ -344,6 +344,7 @@ def get_unpublished_preprints(request, user_subject_pks):
         date_submitted__isnull=False,
         date_declined__isnull=True,
         date_accepted__isnull=True,
+        repository=request.repository,
     ).prefetch_related(
         'preprintauthor_set'
     )
@@ -357,7 +358,9 @@ def get_unpublished_preprints(request, user_subject_pks):
 def get_published_preprints(request, user_subject_pks):
     published_preprints = models.Preprint.objects.filter(
         date_published__isnull=False,
-        date_submitted__isnull=False).prefetch_related(
+        date_submitted__isnull=False,
+        repository=request.repository,
+    ).prefetch_related(
         'preprintauthor_set'
     )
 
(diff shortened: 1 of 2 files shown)


src/repository/logic.py
```diff
@@ -382,10 +386,11 @@ def get_preprint_if_id(preprint_id):
 
 # TODO: Update this
 @cache(300)
-def list_articles_without_subjects():
+def list_articles_without_subjects(repository):
```
mauromsl, line 389: Should this be an optional argument as a kwarg that defaults to `None`?

src/repository/logic.py
```diff
@@ -367,12 +370,13 @@ def get_published_preprints(request, user_subject_pks):
         return published_preprints.filter(pk__in=user_subject_pks)
 
 
-def get_preprint_if_id(preprint_id):
+def get_preprint_if_id(preprint_id, repository):
```
mauromsl, line 373: Is this function even eeded? sounds like the one view using it could do the `if preprint_id` check itself

mauromsl (approved):

src/repository/logic.py
ajrbyers: 👍🏻

src/repository/logic.py
ajrbyers: Dropped. Thought it would be used more, was wrong XD

mauromsl (approved):

PR #2879, opened 2022-05-19
2835 Repository managers can copy a preprint into a journal stage

### LINKED ISSUE #2835: Create Submission from Preprint
- [x] Repo manager can approve a preprint
- [ ] Repo manager can select a journal to send the preprint to
- [ ] Repo manager can select the workflow element to generate a submission in

### LINKED ISSUE #2836: Link Article and Preprint
- Add a one to one field between a preprint and an article
- Display link between preprint and article

ajrbyers:
- Preprint managers can send a preprint to a journal stage
- Stage choice is generated based on the journal's workflow
- Closes #2835 and Closes #2836


src/repository/models.py
```diff
@@ -623,6 +632,52 @@ def local_url(self):
...
+
+            # copy preprints latest file and add it as a MS file to the article
+            file = files.copy_preprint_file_to_article(
+                self,
+                article,
+            )
+            article.manuscript_files.add(file)
+
+            # save and return the article
+            self.article = article
```
mauromsl, line 674: I'm a bit confused about what the `force` argument does here, presumably it leaves any previously linked article orphaned from the preprint, is this a behaviour we want? Wouldn't it be best to first ask the editor to delete the article or unlink it from the preprint explicitely?

src/repository/models.py
```diff
@@ -623,6 +632,52 @@ def local_url(self):
...
+
+            # snapshot authors
+            article.snapshot_authors()
+
+            # copy preprints latest file and add it as a MS file to the article
+            file = files.copy_preprint_file_to_article(
+                self,
+                article,
+            )
+            article.manuscript_files.add(file)
```
mauromsl, line 671: I think it makes sense for the function above (`copy_preprint_file_to_article`) to take an argument like `as_manuscript=True` so that the dev doesn't need to remember to add the file to article.manuscripts

src/repository/models.py
```diff
@@ -623,6 +632,52 @@ def local_url(self):
...
+
+            # save and return the article
+            self.article = article
+            self.save()
+            return article
```
mauromsl, line 676: Do we need to copy over supplementary materials from the `preprint` into the `article`?

src/repository/forms.py
```diff
@@ -519,3 +519,28 @@ def save(self, commit=True):
...
+
+class PreprinttoArticleForm(forms.Form):
+    license = forms.ModelChoiceField(queryset=submission_models.Licence.objects.none())
+    section = forms.ModelChoiceField(queryset=submission_models.Section.objects.none())
+    stage = forms.ChoiceField(
+        choices=()
+    )
+    force = forms.BooleanField(
+        required=False,
+        help_text='If you want to force the creation of a new article object even if one exists, check this box.',
```
mauromsl, line 532: Perhaps we can explain what happens to the old article here

src/repository/models.py
```diff
@@ -623,6 +632,52 @@ def local_url(self):
...
+                abstract=self.abstract,
+                license=journal_license,
+                section=journal_section,
+                date_submitted=timezone.now(),
+                comments_editor='Submitted from {}'.format(self.repository.name),
+                stage=workflow_stage,
+            )
+
+            # copy authors to submission
+            for preprint_author in self.preprintauthor_set.all():
```
mauromsl, line 654: Do we not also need to `article.authors.add(preprint_author)` here?

src/repository/models.py
```diff
@@ -623,6 +632,52 @@ def local_url(self):
 
         return url
 
+    def create_article(self, journal, workflow_stage, journal_license, journal_section, force=False):
```
mauromsl, line 635: Can we have a Unit Test for this function please (with and without `force` behaviour)

src/repository/models.py
```diff
@@ -401,6 +402,14 @@ class Preprint(models.Model):
...
     current_step = models.IntegerField(default=1)
 
+    article = models.OneToOneField(
+        'submission.Article',
+        blank=True,
+        null=True,
+        on_delete=models.SET_NULL,
+        help_text='Linked article of this preprint.',
+    )
+
```
mauromsl, line 412: We should probably also display this relationship in article archive and the preprint page.

src/repository/models.py
```diff
@@ -401,6 +402,14 @@ class Preprint(models.Model):
     date_updated = models.DateTimeField(blank=True, null=True)
     current_step = models.IntegerField(default=1)
 
+    article = models.OneToOneField(
```
mauromsl, line 405: This means I can only submit the preprint to a single journal, which presumably it is okay?

mauromsl (changes requested):
Overall looks good, a few comments/questions inline

src/repository/models.py
ajrbyers: The relationship is already on the preprint page. From what I remeber there was no request for it to be on the article page though.

src/repository/models.py
ajrbyers: Yep.

src/repository/models.py
ajrbyers: Yes it will unlink the other article so there and replace it with the new one. If they want to delete the article thats another story (no other area of Janeway gives anyone the ability to delete an article except via admin), I think, but isn't this unlink pretty explicit given the help text that runs along with the field?

ajrbyers: I think, as you comment below, making it explicitly clear what happens to the old article in the help text will clear this up.

src/repository/models.py
ajrbyers: 👍🏻

src/repository/forms.py
ajrbyers: 👍🏻

src/repository/models.py
ajrbyers: 👍🏻

mauromsl (approved):

Issue #2734, opened 2022-01-27
Register backlist DOIs in batches


joemull:
**Is your feature request related to a problem? Please describe.**
Publishers would like to be able to register DOIs in batches for back content. They may need to do this when they load a backlist into Janeway for a journal that never had DOIs, for example. 

**Describe the solution you'd like**
We could create a tool/button that allows you to tell Janeway: please register all the DOIs for this journal for which no validated/registered DOI yet exists, as long as use_crossref is enabled.

Currently, you can register or re-register all the DOIs in a journal via the command line with register_all_crossref, but this command creates one XML submission per article, rather than concatenating the articles into a single Crossref submission, which would be better for an upgraded batch submission tool.

@mauromsl sees two parts to this problem: one is batch-exporting articles to Crossref. The second one is a query tool that let's you more precisely choose which articles to export (perhaps an entire issue, an entire journal, all the articles that don't have a successful receipt etc.). The second feature could be used for a myriad of things like depositing via DOAJ API or updating metadata.

joemull:
We are going to use class-based views for the querying and filtering functionality. @mauromsl explains: "One base class like `FilteredArticlesListView(ListView)` that we can reuse for `CrossrefArticlesView`, `DOAJArticlesView`, `SearchSubmissionsView`."

PR #2888, opened 2022-05-20
DOI Manager

### LINKED ISSUE #1470: Identifier Management
There is not overview of identifiers in Janeway and the Crossref view is not great either.

A single page that lists all of a Press' or Journal's DOIs and their status. It should also be able to show articles that do not have DOIs or DOIs that do not resolve properly.

joemull:
This branch addresses #1470 and #2734 with a new class-based view for filtering articles into sets, registering DOIs for them in batches, and polling Crossref for status in batches.

It includes changes to the CrossrefDeposit model, including storing the deposit document on the model, and separating Identifier from CrossrefDeposit with the intermediary CrossrefStatus.

CrossrefStatus knows about all the deposits that included a given identifier, but it keeps track of the current status by reading Crossref's response to the latest deposit when the user polls for a new status.

Users can filter by pub date, registration status, primary issue, and journal (if at press level).

Errors are displayed using the messages framework, including if Crossref settings are missing.

For the first time, we have unit tests for validating Crossref deposits. This code also updates our Crossref schema version to 4.8.1.

Includes other selected unit tests and end-user documentation.

Where possible, work on the filters has been incorporated into core to be re-usable by other class-based views in the future.

Fixes #1470 and #2734.


src/core/templatetags/classname.py
```diff
@@ -0,0 +1,11 @@
+from django import template
+
+register = template.Library()
+
+
+@register.filter
+def classname(obj):
+    try:
+        return obj.__class__.__name__
+    except:
```
ajrbyers, line 10: This exception is very wide, could it be narrowed?

src/identifiers/logic.py
```diff
@@ -21,42 +23,134 @@
...
+                identifier = generate_crossref_doi_with_pattern(article)
+            if identifier:
+                identifiers.append(identifier)
+        except AttributeError as e:
+            logger.debug(f'Error with article {article.pk}: {e}')
+    return identifiers
+
+
+def poll_dois_for_articles(articles):
+    from utils.shared import clear_cache
```
ajrbyers, line 123: Can this import be moved to the top of the module?

ajrbyers (changes requested):

src/identifiers/admin.py
```diff
@@ -17,9 +17,17 @@ class Identifier(admin.ModelAdmin):
     raw_id_fields = ('article',)
 
 
+class CrossrefStatus(admin.ModelAdmin):
```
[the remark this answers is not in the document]
joemull: Added some list tools.

PR #2906, opened 2022-06-06
2781 Janeway now stores ORCIDs in a standard format of 0000-0000-0000-000X

### LINKED ISSUE #2781: Update DOI field to support inputs in both short and long (URI) form
Currently some inputs to the DOI field cause problems with DOI registration, or don't behave as expected in other ways.

Use regular expressions to detect various DOI forms and store a uniform version of the DOI, even if it is still served or displayed in various ways through the GUI.

ajrbyers:
Closes #2781

### DIFF
--- a/src/submission/forms.py
+++ b/src/submission/forms.py
@@ -454,7 +454,7 @@ def utility_clean_orcid(form, field_name='orcid'):
             orcid = result.group(0)
         else:
             form.add_error(
-                'orcid',
+                field_name,
                 'An ORCID must be in the pattern https://orcid.org/0000-0000-0000-0000 or'
                 ' 0000-0000-0000-0000',
             )
--- a/src/submission/forms.py
+++ b/src/submission/forms.py
@@ -3,6 +3,8 @@ __author__ = "Martin Paul Eve & Andy Byers"
 __license__ = "AGPL v3"
 __maintainer__ = "Birkbeck Centre for Technology and Publishing"
 
+import re
+
 from django import forms
 from django.utils.translation import ugettext, ugettext_lazy as _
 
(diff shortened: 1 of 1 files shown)


src/submission/forms.py
```diff
@@ -432,3 +438,25 @@ def __init__(self, *args, **kwargs):
...
+        if result:
+            orcid = result.group(0)
+        else:
+            form.add_error(
+                field_name,
+                'An ORCID must be in the pattern https://orcid.org/0000-0000-0000-0000 or'
+                ' 0000-0000-0000-0000',
+            )
+
+    return orcid
```
mauromsl, line 462: can this function take a string instead of a form and a field_name? It'd be easier to test and more handy to use elswhere.
Speaking of which, a unit test would be good too.

The retrieval and error handling could be done in the form's `clean_*` method

src/submission/forms.py
```diff
@@ -432,3 +438,25 @@ def __init__(self, *args, **kwargs):
     class Meta:
         model = models.Article
         fields = ('projected_issue',)
+
+
+def utility_clean_orcid(form, field_name='orcid'):
+    """
+    Utility function that cleans an ORCID ID.
+    """
+    orcid = form.cleaned_data[field_name]
```
mauromsl, line 447: If I'm reading the below correctly, this will still return the uncleaned value if a malformed orcid is passed. Is that desired?

mauromsl (changes requested):
A couple of comments inline

src/submission/forms.py
ajrbyers: Yes I believe so as the malformed value is shown to the user with the error.

src/submission/forms.py
ajrbyers: Sure - will add.

mauromsl (approved):

PR #2907, opened 2022-06-07
Adds a new homepage element that renders a search bar

mauromsl:
The homepage element renders a search box that connects with the regular search view


src/core/homepage_elements/search_bar/urls.py
```diff
@@ -0,0 +1,5 @@
+from django.conf.urls import url
```
ajrbyers, line 1: Could probably drop this as url_patterns is empty?

ajrbyers (changes requested):
Could drop: urls.py and views.py as they're empty. Otherwise looks good.

PR #2913, opened 2022-06-09
Git-ignore emacs' backup files

gamboz:
By default, the emacs editor leaves copies of modified files that end with "~". These can safely be ignored by git.


Bbkctp:

PR #2929, opened 2022-06-16
#2737 Added new archive stage.

### LINKED ISSUE #2737: Dark archive button
Occasionally the need arises to move articles out of the workflow without publishing them, and also without deleting them entirely.

It might make sense to have a dark archive button. It's an open question where this button would be best located.

ajrbyers:
- Adds a new stage called Archived.
- Archived papers show with Rejected in the Back Content List.
- Archived papers who a warning along the top of workflow pages.
- Workflow tasks by non-editors cannot be completes for archived papers.
- Closes #2737


src/submission/models.py
```diff
@@ -1015,11 +1018,20 @@ def get_article(journal, identifier_type, identifier):
             # resolve an article from an identifier type and an identifier
             if identifier_type.lower() == 'id':
                 # this is the hardcoded fallback type: using built-in id
-                article = Article.objects.filter(id=identifier, journal=journal)[0]
+                article = Article.objects.filter(
+                    id=identifier,
+                    journal=journal,
+                ).exclude(
+                    stage=STAGE_ARCHIVED,
```
mauromsl, line 1025: doing this via the default manager might be good here too
In this case we would need a custom manager to retrieve the rejrected/archived articles explicitely

src/workflow/logic.py
```diff
@@ -79,12 +79,9 @@ def stages_in_between(from_stage, to_stage, article):
     return stages[:stage_index+1]
 
 
-def move_to_stage(from_stage, to_stage, article):
+def move_to_stage(from_stage, to_stage, article, all=False):
```
mauromsl, line 82: This doesn't seem to do anything

src/templates/admin/workflow/manage_article_workflow.html
```diff
@@ -45,6 +48,20 @@
...
+                <div class="title-area">
+                    <h2>Archive</h2>
+                </div>
+                {% if article.stage == 'Archived' %}
+                <p>This article has already been archived.</p>
+                {% else %}
+                <p>Archiving an article will remove it from the workflow, essentially tombstoning it. You can still access it through the Article Archive.</p>
+                <form method="POST">
+                    {% csrf_token %}
+                    <button name="archive" class="danger button">Archive Article</button>
```
mauromsl, line 61: It would be good to use Joe's new `ConfirmableForm` here

mauromsl (changes requested):
I added an optional comment about abstracting the filter logic for the archived article into managers. I think it is easier to address the exception than having to always remember the special stage when making a query for articles or article-related models.

A couple other comments inline

ajrbyers:
Test this please

src/templates/admin/workflow/manage_article_workflow.html
joemull: Will do so in 1.5.1: https://github.com/BirkbeckCTP/janeway/issues/3150

PR #2931, opened 2022-06-17
Load urls from homepage elements only if possible.

gamboz:
Adding a check on the existence of `homepage/elements/BLOCK/urls.py` otherwise I get 
```
Error loading a block: search_bar, No module named 'core.homepage_elements.search_bar.urls'
```
during django setup.

**NB** I'm not sure if it is ok that `urls.py` is missing 🫤

### DIFF
--- a/src/core/include_urls.py
+++ b/src/core/include_urls.py
@@ -215,10 +215,12 @@ blocks = plugin_loader.load(os.path.join('core', 'homepage_elements'), prefix='c
 if blocks:
     for block in blocks:
         try:
-            urlpatterns += [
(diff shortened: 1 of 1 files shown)


Bbkctp:

mauromsl:
Thank you @gamboz !
We did notice this error being logged on CI. We have implemented a similar fix to yours here: ca8a45a551893cd6b44ef88a057911d320ada8eb

As you suggested, it shouldn't be a hard requirement for plugins/Homepage elements to include a `urls.py` as per with any other django application. Closing this in lieu of ca8a45a551893cd6b44ef88a057911d320ada8eb

PR #2943, opened 2022-06-21
518 adds new reader role to which users can add themselves, they will then receive notifications when new articles are published

### LINKED ISSUE #518: Reader Group and Article Notifications
- [x] When a user registers or edits their account give them the option to enroll as a reader
- [x] Add new setting that will allow an editor to enable emails to be sent when either a new article is published
    - [x] Add a new email setting
    - [x] Add a new email subject setting
    - [x] Add a new boolean to flag this feature on or off
    - [x] Add a new management command to send these emails
    - [x] Update CRON installation to add this new command 
- [x] Alow users to cancel their subscription
- [x] Add page where Editors can see readers and see notifications sent

ajrbyers:
Closes #518


src/utils/install/journal_defaults.json
```diff
@@ -3403,6 +3403,51 @@
...
+        "setting": {
+            "description": "If enabled Janeway will notify readers of new published articles.",
+            "is_translatable": false,
+            "name": "send_reader_notifications",
+            "pretty_name": "Send Reader Notifications",
+            "type": "rich-text"
+        },
+        "value": {
+            "default": ""
+        }
```
joemull, line 3450: This should have type `boolean` and default `false`, right?

joemull (changes requested):
I managed to catch a couple copy-paste issues but I didn't go deep into the design on this one as I'm still not super familiar with the email and role frameworks. Let's have Mauro take a look at it before merging.

PR #2945, opened 2022-06-21
#2028 adds a feature flag to disable the Reviews block on the author's article page before acceptance/rejection

### LINKED ISSUE #2028: Feature Flag Reviews on Author Page
When an Author looks at the Article page they can see if there are assigned reviewers. Some journals like authors to be able to see this information, others do not.

Add a feature flag to hide author reviews until they are released by Editors. This should be on by default so reviews will not show up until the Editor marks them as for Author Consumption.

In the interim 1.3.9 will have a note telling authors that they will be informed once the Editors release the reviews.

ajrbyers:
Closes #2028


src/templates/admin/core/article.html
```diff
@@ -22,6 +22,7 @@ <h2>{{ article.title }}</h2>
         </div>
     </div>
 
+    {% if not journal_settings.general.hide_review_data_pre_release or article.is_accepted or article.date_declined %}
```
joemull, line 25: I think this logic means that checking the boolean will *always* hide reviews in the review stage, even if the editor marks a review as for author consumption. Am I reading it right? Do we want to make it so reviews released for author consumption show up during review, independent of the new setting's value?

src/utils/install/journal_defaults.json
```diff
@@ -3074,6 +3074,21 @@
...
+            "description": "Enable if you want to hide all review data from authors until released.",
+            "is_translatable": false,
+            "name": "hide_review_data_pre_release",
+            "pretty_name": "Hide Review Data from Authors Before Release",
+            "type": "boolean"
+        },
+        "value": {
+            "default": "on"
+          }
+    },
```
joemull, line 3091: Just thinking about how folks might misunderstand this (because they will!), I wonder if we could make the description read something like this: 

Enable if you want to hide all review data from authors through the end of the review stage, except reviews that are individually shared with authors during review.

joemull (changes requested):
I have a question about the template logic and I made a suggestion for the help text.

src/templates/admin/core/article.html
```diff
@@ -22,12 +22,13 @@ <h2>{{ article.title }}</h2>
         </div>
     </div>
 
+    {% if not journal_settings.general.hide_review_data_pre_release or article.peer_reviews_for_author_consumption %}
```
joemull, line 25: With the latest change, did you decide to ignore whether an article has passed the review stage? It will no longer show the block once the article has been accepted or declined, as it did before. Is that what you want?

src/templates/admin/core/article.html
```diff
@@ -22,12 +22,13 @@ <h2>{{ article.title }}</h2>
...
     <div class="box">
         <div class="title-area">
             <h2>Reviews</h2>
         </div>
         <div class="box">
-            {% if journal_settings.general.enable_peer_review_data_block %}
+            {% if journal_settings.general.enable_expanded_review_details %}
```
joemull, line 31: What's the difference between these settings? Have we documented them? I think we need an explainer on how to use these settings in combination, since we now have so many:

- `hide_review_data_pre_release`
- `enable_peer_review_data_block`
- `enable_expanded_review_details`

joemull (commented):
Asked a few questions about the differences between the various review visibility settings you're using and whether we could document them.

src/utils/install/journal_defaults.json
```diff
@@ -3064,16 +3064,31 @@
...
-            "name": "hide_review_metadata_from_authors",
-            "pretty_name": "Hide Review Metadata from Authors",
+            "name": "enable_peer_review_data_on_review_page",
+            "pretty_name": "Enable Peer Review Data on Review Pages",
             "type": "boolean"
         },
         "value": {
             "default": ""
           }
     },
```
joemull, line 3076: This needs a data migration to delete the old setting in people's databases, right?

docs/source/manager/review/index.rst
```diff
@@ -27,10 +27,23 @@ Under review settings we can control how peer review operates for the given jour
...
+Janeway has various controls for how review data is displayed to Authors. These settings are:
+
+- Hide Review Data Pre Release
+    - When this setting is enabled authors will so no information about active peer reviews until the editor makes them available, otherwise they will see a list of review assignments.
+- Enable Peer Review Data Block
+    - When this setting is enabled authors will have access to view more information about the reviews for their article (requested, due, and complete dates) on the Article page. If it is left off they will only have access to the review comments.
+- Enable Peer Review Data on Review Page
+    - When this setting is enabled authors will have access to view more information about their reviews on the review page.
+
+The settings "Enable Peer Review Data Block" and "Enable Peer Review Data on Review Page" do similar things but on different pages. "Enable Peer Review Data Block" is used on the author's article page, which they have access to from the point of submission. "Enable Peer Review Data on Review Page" is used on the review page which authors only get access to after an editor grants the author access. Both are off by default.
```
joemull, line 46: This is good but I can't quite square it with the options available on the review settings page:

Screenshot from 2022-09-06 10-33-33

To avoid confusion these will probably need to correspond more closely.

joemull (changes requested):
I believe we need to make few more changes to clean up the various settings and their documentation.

joemull (approved):

ajrbyers:
test this please

joemull:
jenkins: test this please

ajrbyers:
Test this please

joemull:
Jenkins: test this please

ajrbyers:
jenkins: test this please

PR #2947, opened 2022-06-21
Add new hooks "extra_corefields" in all themes.

gamboz:
So that plugins and apps can customize the user profile form.

The extra code will appear in the "Core data" section of the user profile form.\
Example with hook function defined [here](https://github.com/gamboz/wjs-profile-project/blob/61ce9752d7450a73fac9b59384d6d0eb1670c4dc/wjs/jcom_profile/hooks.py)

example

### DIFF
--- a/src/themes/OLH/templates/elements/accounts/user_form.html
+++ b/src/themes/OLH/templates/elements/accounts/user_form.html
@@ -1,6 +1,7 @@
 {% load foundation %}
 {% load static %}
 {% load i18n %}
+{% load hooks %}
 
 <h5>{% trans 'Core Data' %}</h5>
 <div class="row">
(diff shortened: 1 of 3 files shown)


gamboz:
Hmmm... I noticed now that if an app modifies a form introducing a mandatory field, any previous tests that use that form will fail because the request payload did not include the new mandatory field.

E.g. `core.tests.test_app.CoreTests.test_mixed_case_email_address_correct_username` that does a POST toward "core_register" fails if [`wjs.jcom_profile`](https://github.com/gamboz/wjs-profile-project/tree/hooks) is installed.

gamboz:
Closing to better organize. Will probably re-open in future :)

PR #2950, opened 2022-06-21
Add confirmation pane to author-facing task submission buttons

### LINKED ISSUE #2753: Create confirmation box to prevent new user slips before submission
Newcomers to Janeway sometimes click submit on tasks before they have fully understood what is being requested of them. For example, an author may receive a revision request and not realize they need to upload a file or replace one of the current manuscript files. Or they may simply forget to upload it. Clicking submit on this task results in a closed task that they cannot re-enter on their own to fix the error. In cases like these we should try to help them prevent the error from happening in the first place.

We can create a box that shows users what is about to happen and asks them to confirm it. This can be a reusable template element with a recognizable, consistent design that we can add if appropriate to a variety of different interaction points with new or intermittent users, such as copyeditors, reviewers, authors, typesetters, and proofreaders.

joemull:
Closes #2753.

Screenshots:
Screenshot from 2022-06-21 16-39-49
Screenshot from 2022-06-21 14-12-56
Screenshot from 2022-06-21 10-31-41


joemull:
Will also need to merge https://github.com/BirkbeckCTP/typesetting/pull/160

mauromsl (approved):
I'm very happy with how this new form class turned out. Good job @joemull 

Something we can do to encapsulate the widget assets further is to declare a `Media` class inside the form to bundle the javascript. (rather than explicitiely having to introduce logic under `{{ block js }}`. This however would involve refactoring the template we currently use for all modals, so it is probably not worth doing as part of this PR

https://docs.djangoproject.com/en/1.11/topics/forms/media/#media-on-forms

joemull:
That's an interesting idea @mauromsl! For sure will have to keep that in mind for the future.

PR #2953, opened 2022-06-22
2936 ported popup email logic and form to core and added view to repositories so managers can email preprint owners from inside Janeway.

### LINKED ISSUE #2936: Preprints: Port internal "email submitting author" messaging to preprints
As a preprints moderator, I would like to be able to send a message to a submitting author directly from within Janeway. 

When preprints moderator clicks "Contact the Owner/Submitting Author", they should be taken to an internal messaging system which allows them to send a message to the submitting author. Once sent, that message should appear in the preprint's logs. 

The "Discussions" feature could also be built out to serve this purpose. 

**Additional context**
image

ajrbyers:
Closes #2936

### DIFF
--- a/src/core/forms.py
+++ b/src/core/forms.py
@@ -15,6 +15,7 @@ from django.utils.translation import ugettext_lazy as _, get_language
 from django.conf import settings
 from django.contrib.auth.forms import UserCreationForm
 from django.db.models import Q
+from django.core.validators import validate_email, ValidationError
 
 from django_summernote.widgets import SummernoteWidget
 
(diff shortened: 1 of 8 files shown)


hardyoyo (commented):
This all looks just perfect, I'll run it through a test, but based on a visual inspection alone, I'm :+1:

hardyoyo:
Local testing worked fine, and I've confirmed that this passes tests (after rebasing this branch on master, which has a fix for a failing test on MySQL). We're running more tests on our dev instance, to give it a fair shake, but signs are looking good.

hardyoyo (approved):
Looks good to me :+1: Ideally CC info would also be included in the log file entry, but that shouldn't block the merging of this desired functionality, we can add CC into to the log entry later.

src/core/forms.py
```diff
@@ -619,3 +620,27 @@ def order_by(self, queryset, facet, fks):
...
+        required=False,
+        max_length=1000,
+        help_text='Separate email addresses with ;',
+    )
+    subject = forms.CharField(max_length=1000)
+    body = forms.CharField(widget=SummernoteWidget)
+
+    def clean_cc(self):
+        cc = self.cleaned_data['cc']
+        if not cc or cc == '':
```
mauromsl, line 636: The second operand here is redundant, the expression should shortcircuit on the first conditional when `cc` is set to `''` (thought I see this code was copied verbatim from `journal/views.py`)

mauromsl (approved):

hardyoyo:
This looks approved, I'd feel better if it were *merged* :-) hint hint

PR #2974, opened 2022-06-28
#2972 move password rules markup from register to reset_password templates

### LINKED ISSUE #2972: password length is hardcoded throughout most of Janeway, despite being configurable
**Describe the bug**
There is a press configuration for password length, but a large portion of Janeway still has a hard-coded notion of 12 characters being the only acceptable password length. This old code needs to be revised to use the press setting for password length.

**Janeway version**
Current master, with HEAD at `43a7955b7`

**To Reproduce**
[Search the code ](https://github.com/BirkbeckCTP/janeway/search?q=%27characters+long%27) for the phrase "characters long," see many instances of the value `12` being hard coded.

**Expected behavior**
the Press configuration for required password length should be used.

**Screenshots**
n/a

hardyoyo:
Fixes: #2972 by copying password rules from the registration page to the reset password page

### DIFF
--- a/src/themes/OLH/templates/core/accounts/reset_password.html
+++ b/src/themes/OLH/templates/core/accounts/reset_password.html
@@ -24,10 +24,15 @@
                 <div class="callout">
                     <div class="row column log-in-form">
                         <h5>{% trans "Enter your new password twice to complete the reset process" %}:</h5>
-                        <p>{% blocktrans %}Your password should be at minimum 12 characters long. It does not need to
-                            contain specific
-                            characters but you should make it as long as possible. For more information read our
-                            <a href="#" data-open="password-modal">password guide</a>{% endblocktrans %}.</p>
+                        <p>Password Rules:</p>
+                    <ul class="no-top-margin">
+                        <li>- {% trans "Your password should be" %} {{ request.press.password_length }} {% trans "characters long" %}.</li>
+                        {% if request.press.password_upper %}<li>- {% trans "Your password should contain at least one upper case letter." %}</li>{% endif %}
+                        {% if request.press.password_number %}<li>- {% trans "Your password should contain at least one number." %}</li>{% endif %}
+                    </ul>
+                    <p>{% blocktrans %}For more information read our
+                        <a href="#passwordmodal" class="modal-trigger">password guide</a>
+                        .{% endblocktrans %}</p>
                         {{ form|foundation }}
                         <p>
                             <button type="submit" class="button expanded">{% trans "Reset Password" %}</button>
--- a/src/themes/clean/templates/core/accounts/reset_password.html
+++ b/src/themes/clean/templates/core/accounts/reset_password.html
@@ -14,11 +14,15 @@
                             {% csrf_token %}
                             <div class="form-group">
                                 <h5>{% trans "Enter your new password twice to complete the reset process" %}:</h5>
-                                <p>{% blocktrans %}Your password should be at minimum 12 characters long. It does not
-                                    need to contain specific
-                                    characters but you should make it as long as possible. For more information read our
-                                    <a href="#" data-toggle="modal" data-target="#passwordmodal">password
-                                        guide</a>.{% endblocktrans %}</p>
+                                <p>Password Rules:</p>
+                                <ul class="no-top-margin">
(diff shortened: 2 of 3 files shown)


mauromsl:
Thanks @hardyoyo! We release this in 1.4.3 to give our translators time to adjust the translations for the front end.

src/themes/OLH/templates/core/accounts/reset_password.html
```diff
@@ -24,10 +24,15 @@
                 <div class="callout">
                     <div class="row column log-in-form">
                         <h5>{% trans "Enter your new password twice to complete the reset process" %}:</h5>
-                        <p>{% blocktrans %}Your password should be at minimum 12 characters long. It does not need to
-                            contain specific
-                            characters but you should make it as long as possible. For more information read our
-                            <a href="#" data-open="password-modal">password guide</a>{% endblocktrans %}.</p>
+                        <p>Password Rules:</p>
```
mauromsl, line 27: This text needs to be wrapped in a `{% trans %}` template tag (here and in the other themes

mauromsl (changes requested):
Thanks, @hardyoyo, one minor change pointed inline

src/themes/OLH/templates/core/accounts/reset_password.html
hardyoyo: This code is copied directly from the register template, you'll need to change the register template as well. Still want this change for this PR?

src/themes/OLH/templates/core/accounts/reset_password.html
```diff
@@ -24,10 +24,15 @@
...
+                    <ul class="no-top-margin">
+                        <li>- {% trans "Your password should be" %} {{ request.press.password_length }} {% trans "characters long" %}.</li>
```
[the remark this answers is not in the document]
hardyoyo: This code is copied directly from the register template, you'll need to change the register template as well. Still want this change for this PR?

src/themes/OLH/templates/core/accounts/reset_password.html
mauromsl: Yes, that would help. With identical text the same translation will work across both pages

src/themes/OLH/templates/core/accounts/reset_password.html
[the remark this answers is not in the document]
mauromsl: Yes, that would help. With identical text the same translation will work across both pages

src/themes/OLH/templates/core/accounts/reset_password.html
hardyoyo: OK, you'd still need to change the register template for that to work. Would you like me to add that to this PR?

hardyoyo:
> Thanks @hardyoyo! We release this in 1.4.3 to give our translators time to adjust the translations for the front end.

OK... I believe the translations would work as-is, since this code is copied from the register template, so it should just use the translation from the register template. I think you may want to make a new issue for the translation refactor, as it will likely require refactoring more than just this one copy/paste?

hardyoyo:
I'm going to share a bit about why I'm pushing for this change: if 2974 isn't merged for 1.4.2, I'll need to either refactor the Glossa Psycholinguistics theme files before we upgrade to 1.4.2, or we'll have to make a fork of Janeway and pull PR 2974 into it. I'd prefer not to do either of those things, but, I'll live with it.

mauromsl (approved):

mauromsl:
> I'm going to share a bit about why I'm pushing for this change: if 2974 isn't merged for 1.4.2, I'll need to either refactor the Glossa Psycholinguistics theme files before we upgrade to 1.4.2... and, actually, all I'll need to do is merge this PR into a new copy of the OLH theme, and I'll be done. That's not so bad. I talked myself out of it. :-) Proceed as you see fit.
> 
> I will make the requested change for the reset_password and register templates as part of this PR, probably won't be this week.

Fair enough, since the translation keys are the same as in the registration page, it shouldn't be a problem then. We can look into updating both sets of templates later.

hardyoyo:
FYI, the suggested translation-related changes made above were assembled in this pull request: https://github.com/BirkbeckCTP/janeway/pull/2984

Issue #2971, opened 2022-06-27
Add a new setting that allows inserting a custom message in the registration page


mauromsl:
**Is your feature request related to a problem? Please describe.**
Journals and presses may want to add a custom message/notice on the registration page similar to the setting that is available in the login page: `journal_settings.general.login_page_notice`

**Describe the solution you'd like**
A new setting that allows a press or journal to add custom HTML onto the registration page (e.g: journal_settings.general.registration_page_notice)

This block should be added at the bottom of the registration form

**Additional context**
Add any other context or screenshots about the feature request here.ves you've considered
![image](https://user-images.githubusercontent.com/10760678/175990562-ecffa4a7-5487-472f-886b-9c0c3b2b9bf2.png)

PR #2975, opened 2022-06-28
2971 Add custom message setting register_page_notice

hardyoyo:
fixes: #2971


mauromsl:
Thanks @hardyoyo ! We are including this into v1.4.2 since we have another RC coming today

mauromsl (approved):

PR #2981, opened 2022-07-01
Doi Manager performance optimization

### LINKED ISSUE #2967: Optimize DOI Manager SQL queries
- [x] Optimize SQL queries
- [x] Chunk the DOI Manager queryset into batches in logic.py (make own issue if can't finish in time for 1.4.2)

joemull:
Fixes #2967. 

The new defaults are:
Cap on number of articles per action: 200
Cap on batch size: 100
Time-out on polling: 30


src/identifiers/logic.py
```diff
@@ -120,28 +133,41 @@ def get_dois_for_articles(articles, create=False):
...
+
     status = ''
     error = False
     identifiers = get_dois_for_articles(articles)
     polled = set()
-    for identifier in identifiers:
+    for i, identifier in enumerate(identifiers):
+
+        # Time out gracefully
+        if datetime.datetime.now() > start + datetime.timedelta(seconds=30):
```
mauromsl, line 150: It seems to me that the 30 seconds used here are matching the cache invalidation term used on the decorators above.
If so, it would be more explicit to have a constant at the top of this module like:

`CROSSREF_TIMEOUT_SECONDS = 30`

Also, we want the timeout to be an optional argument to this function, since it is only relevant when executing under an HTTP request, but not when executing from command line or a Django command. (The default can be `timeout=CROSSREF_TIMEOUT_SECONDS`) which would allow command line users to set it as `timeout=None`

Finally, you don't need to work with datetime objects here, it is best to work with `time.time()` (integers) since you are replicating the behaviour of a timer here. e.g.:

```
start_time = time.time()
for item in tems:
    #do some work
    if time.time() - start_time > 30:
         break
```
Or you could even write a `Timer` context manager for bonus pythonic points ;)

src/core/views.py
```diff
@@ -2262,38 +2266,46 @@ def get_facet_queryset(self):
...
+        if actions:
+            start = datetime.datetime.now()
+
+            action_status = ''
+            action_error = False
+
             querysets = []
-            for journal in {article.journal for article in queryset}:
-                querysets.append(queryset.filter(journal=journal))
+            queryset = self.get_queryset(params_querydict=params_querydict)
```
mauromsl, line 2294: We are not sanitising `params_string`, meaning that users would be able to inject their own database filters here with `params_querydict` outside of those you have intended.

We should make sure the user has not sent additional 'facets' than those we expect them to. This can be done in `.get_queryset` but also validating `forms.CBVFacetForm` with `request.POST`/ `request.GET` data as appropriate

src/journal/models.py
```diff
@@ -494,6 +495,7 @@ class Issue(AbstractLastModifiedModel):
     volume = models.IntegerField(default=1)
     issue = models.CharField(max_length=255, default="1")
     issue_title = models.CharField(blank=True, max_length=300)
+    display_title = models.CharField(blank=True, max_length=300)
```
mauromsl, line 498: Should probably set this as editable=False, since user input will always be overriden by .save (with a comment)

mauromsl (changes requested):
Nice work, a few comments inline

src/journal/models.py
```diff
@@ -1077,6 +1082,12 @@ def setup_default_form(sender, instance, created, **kwargs):
             default_review_form.elements.add(main_element)
 
 
+@receiver(post_save, sender=Journal)
+def update_issue_display_title(sender, instance, created, **kwargs):
+    for issue in Issue.objects.filter(journal=instance):
+        issue.update_display_title()
+        issue.save()
```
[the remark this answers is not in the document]
joemull: It's not redundant since `update_display_title()` doesn't call `save()`. This was to avoid a circular call, since `save()` calls `update_display_title()`.

Or is django smart enough to avoid a loop between the two functions?

src/identifiers/logic.py
```diff
@@ -254,6 +280,7 @@ def create_crossref_issues_context(journal, identifiers):
     return crossref_issues
 
 
+@cache(30)
```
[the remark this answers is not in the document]
joemull: On closer inspection, this cache doesn't provide much performance gain, because calls to the function are already sorted and split up by issue. In other words, the cache isn't used because the function is only ever called with unique parameters. So I will remove it.

src/core/views.py
joemull: Yes, this occurred to me, and I made sure to filter out unexpected parameters in `get_queryset` with this line: https://github.com/BirkbeckCTP/janeway/blob/d6718e24980d3b4d5d448fb337dc1ddb57041231/src/core/views.py#L2233

I'll add a comment above it specifying that's why it's there. Let me know if you spot a problem with it.

src/journal/models.py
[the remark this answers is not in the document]
joemull: Ah, sorry, misread your earlier comment. Will remove the call to update.

PR #2982, opened 2022-07-04
Give section editors access to list views

### LINKED ISSUE #2229: Search Submissions for Section Editors
Allow section editors to see and use the Search Submissions screen.

joemull:
Closes #2229.

I talked to three editors who had worked with section or guest editors, and the consensus was that they should be able to search submissions and view the workflow list pages for review and copyediting, but only the articles to which they are assigned should be visible.


src/core/logic.py
```diff
@@ -919,3 +919,11 @@ def render_nested_setting(
     )
 
     return rendered_string
+
```
ajrbyers, line 922: There should be an extra newline here.

src/core/logic.py
```diff
@@ -919,3 +919,11 @@ def render_nested_setting(
...
+def restrict_articles_to_editor_assigned(request, articles):
```
ajrbyers, line 923: Might be worth renaming this, it doesn't actually restrict anything and could be used for other things.

src/core/templatetags/nose_tools.py
```diff
@@ -0,0 +1,14 @@
+from nose import tools
+from django import template
+
+register = template.Library()
+
+class NoseNode(template.Node):
+
+    def render(self, context):
+        tools.set_trace()         # Debugger will stop here
```
ajrbyers, line 9: This is nice!

ajrbyers (approved):

ajrbyers:
This looks good, a few comments and a conflict to resolve.

joemull:
After it was approved, this branch was corrupted with a rebase that did not do what we wanted. It is superseded by https://github.com/BirkbeckCTP/janeway/pull/3115

PR #3000, opened 2022-07-12
Removed status logic from manager_review_status_change setting.

ajrbyers:


src/utils/transactional_emails.py
```diff
@@ -1617,6 +1617,7 @@ def preprint_review_status_change(**kwargs):
     request = kwargs.get('request')
     review = kwargs.get('review')
     status_change = kwargs.get('status_change')
+    status_text = None
```
joemull, line 1620: If this value is sent to the template, will it display "None" as text?

ajrbyers: It will be, but the withdrawl template doesn't use {{ status_text }} so it wont be displayed.

ajrbyers: It will be sent to the template as None, however, the withdrawl template doesn't use this variable and all other routes set it.

PR #3007, opened 2022-07-14
#3004 swap .difference for .exclude in when searching for repo reviewers

### LINKED ISSUE #3004: Fix .difference query function not working for non-postgres backends
**Describe the bug**
MySQL does not support .difference and we use it in a few places across the codebase.

**Janeway version**
1.4.2

ajrbyers:
Closes #3004


mauromsl (approved):
One minor comment, otherwise good to merge.

I think there was another use of `.difference()` in the codebase that might be worth fixing?

PR #3018, opened 2022-07-21
#2711 review due dates are now in the default invitation and on the review page.

### LINKED ISSUE #2711: Peer reviewers with one-click access cannot see due date
An editor reported a reviewer could not find the due date for the review anywhere. This can happen if one-click-access is turned on for reviewers, as the review request view does not include the date. If the review request email also does not include the due date, as is default, the reviewer won't see it anywhere.

We can:
- [x] Add the due date to the review request view, including the one-click endpoint
- [x] Add the due date to the default review request email template: `{{ review_request.date_due }}`
- [x] Add the template variable for due date to the documentation or help text for the review reqeuest template setting.
- [x] Where a review_assignment template has not been editted, overwrite it with the new value

ajrbyers:
Closes #2711


src/core/migrations/0074_auto_20220721_1448.py
```diff
@@ -0,0 +1,49 @@
...
+    Updates the review_assignment setting for a journal where it has not been edited.
+    """
+    with translation.override(django_settings.LANGUAGE_CODE):
+        SettingValue = apps.get_model('core', 'SettingValue')
+        setting_value = SettingValue.objects.get(
+            setting__name='review_assignment',
+            journal=None,
+        )
+        language_var = "value_{}".format(django_settings.LANGUAGE_CODE)
+        setattr(setting_value, language_var, NEW_VALUE)
```
joemull, line 28: I may not be reading this right, but it would seem to save NEW_VALUE and tag it as French for an installation that was set to French. Wouldn't we want to avoid that and only change English values?

src/core/migrations/0074_auto_20220721_1448.py
```diff
@@ -0,0 +1,49 @@
...
+def update_default_setting(apps, schema_editor):
+    """
+    Updates the review_assignment setting for a journal where it has not been edited.
+    """
+    with translation.override(django_settings.LANGUAGE_CODE):
+        SettingValue = apps.get_model('core', 'SettingValue')
+        setting_value = SettingValue.objects.get(
+            setting__name='review_assignment',
+            journal=None,
+        )
```
joemull, line 26: Doesn't this get all the setting's values, whether or not they're customized? As a result, won't they all be overwritten?

docs/source/manager/other/template_variables.rst
```diff
@@ -0,0 +1,151 @@
...
+
+{{ review_assignment.date_due }}
+
+If I wanted to display the title of the issue this article is projected to be in I can use:
+
+{{ article.projected_issue.display_title }}
+
+If I wanted to display an article's journal's name I would use:
+
+{{ article.journal.name }}
```
joemull, line 151: This is a great idea!

joemull (commented):
Asked some questions about migration logic.

src/core/migrations/0074_auto_20220721_1448.py
ajrbyers: As this is an AND not an OR it will only get settings where review_assignment is the setting_name and the journal=None.

src/core/migrations/0074_auto_20220721_1448.py
ajrbyers: Specifically for the default value. At the moment default values are only loaded in English. Still, may be for the best to set this to EN.

joemull: OK, let's set this to EN then.

joemull (changes requested):
How about we just change the language to English on that line?

joemull (approved):

PR #3020, opened 2022-07-21
#2819 adds description to Review Files block to avoid confusion and adds the latest manuscript and figure files inline below a revision request.

### LINKED ISSUE #2819: Review/Revision Page enhancements
Based on feedback from @MartinPaulEve we can make this page easier for editors as some of the context is not always clear.

1. Add a descriptio to the Review Round files block e.g.: "These files are the versions sent out to reviewers in Round X of review. They are not the most recent revised versions, which can be found on the Revisions page"
2. When a revision task is complete display the current version of all files in a small inline block of links below the request details.

Not doing this.

ajrbyers:
Closes #2819

### DIFF
--- a/src/submission/models.py
+++ b/src/submission/models.py
@@ -8,6 +8,7 @@ from urllib.parse import urlparse
 import uuid
 import os
 from dateutil import parser as dateparser
+from itertools import chain
 
 from django.urls import reverse
 from django.db import connection, models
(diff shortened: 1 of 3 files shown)


joemull (changes requested):
I think this is ready to go once we make the strings translatable.

PR #3035, opened 2022-07-25
#2840 added setting to form, updated to work on form.

### LINKED ISSUE #2840: Render HTML galley but suppress HTML file download
One of our journal editors requested that the HTML galley be displayed on the article page, but that the HTML file was not available to users for download. The reasoning was that since the HTML file did not include the images within the article, the PDF file was a better option for readers to download.

ajrbyers:
- Adds the existing setting to the Article settings page
- Moves exclude logic to the view, after render content is retrieved
- Closes #2840

### DIFF
--- a/src/journal/views.py
+++ b/src/journal/views.py
@@ -437,8 +437,6 @@ def article(request, identifier_type, identifier):
         # exclude any HTML galleys.
         galleys = galleys.exclude(
             file__mime_type='text/html',
-            type='html',
-            label='HTML'
         )
 
     template = 'journal/article.html'
--- a/src/journal/views.py
+++ b/src/journal/views.py
@@ -433,6 +433,14 @@ def article(request, identifier_type, identifier):
     if article_object.is_published:
         store_article_access(request, article_object, 'view')
 
+    if request.journal.disable_html_downloads:
+        # exclude any HTML galleys.
+        galleys = galleys.exclude(
+            file__mime_type='text/html',
+            type='html',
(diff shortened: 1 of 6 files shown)


src/journal/views.py
```diff
@@ -433,6 +433,14 @@ def article(request, identifier_type, identifier):
...
         store_article_access(request, article_object, 'view')
 
+    if request.journal.disable_html_downloads:
+        # exclude any HTML galleys.
+        galleys = galleys.exclude(
+            file__mime_type='text/html',
+            type='html',
+            label='HTML'
+        )
+
```
joemull, line 443: The `label` field is free text, so if somebody types something else in for an HTML galley, it might not match `HTML` exactly. Can we avoid filtering on label here?

src/templates/admin/elements/forms/group_article.html
```diff
@@ -6,10 +6,20 @@ <h2>General Article Display</h2>
...
+
     {% include "admin/elements/forms/field.html" with field=edit_form.hide_author_email_links %}
 </div>
 
+
+<div class="title-area">
+    <h2>Download/View Links</h2>
+</div>
+<div class="content">
+    <p>{% trans "Settings to control the display of view and download links on article pages." %}.</p>
```
joemull, line 18: There's an extra period

joemull (approved):
Looks great! I just have a logic question and I spotted a typo.

PR #3042, opened 2022-07-26
Fix logic on issue assignment during prepublication checklist

joemull:
In addition to the logic changes specified in the bug description, I've made the primary issue visible on the modal:

Peek 2022-07-26 16-33

I've also added "Primary" before the primary issue in the OLH and material theme sidebars for consistency, since we prepend the name of the issue type (e.g. "Collection" or "Issue") for subsequent issues in the issue list).

### DIFF
--- a/src/journal/logic.py
+++ b/src/journal/logic.py
@@ -200,6 +200,9 @@ def handle_assign_issue(request, article, issues):
         if issue_to_assign in issues:
             issue_to_assign.articles.add(article)
             issue_to_assign.save()
+            if not article.primary_issue:
+                article.primary_issue = issue_to_assign
+                article.save()
             messages.add_message(request, messages.SUCCESS, 'Article assigned to issue.')
         else:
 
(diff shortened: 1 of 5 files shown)


src/journal/logic.py
```diff
@@ -215,6 +218,12 @@ def handle_unassign_issue(request, article, issues):
         if issue_to_unassign in issues:
             issue_to_unassign.articles.remove(article)
             issue_to_unassign.save()
+            if issue_to_unassign == article.primary_issue:
+                if article.issues_list().exists():
+                    article.primary_issue = article.issues_list().first()
+                else:
+                    article.primary_issue = None
+                article.save()
```
ajrbyers, line 226: This is good, however, it might be better to handle this in the Signal we already have listening to articles being added and removed from `issue.articles`, that way even if its done in the admin area this will be executed. The signal is: `journal.models.issue_articles_change`

ajrbyers (changes requested):

ajrbyers (approved):

PR #3054, opened 2022-08-02
Editors will be warned when they attempt to assign a task to a user whose account is not active.

### LINKED ISSUE #2822: Tell editors before sending task requests to unactivated email addresses
We often get support requests for situations in which an editor has sent a task to someone who has had trouble accessing that task in Janeway. Often this happens when the person has duplicate accounts: the task was sent to one account, but the user is logging in with the other account. Often, the account the user is _not_ logging into is inactive.

To facilitate better communication about what email address to use, we could check whether an account is active before Janeway creates a task and ask the task creator to confirm the action. This may head off many of these situations because it will let them know that the task will be created for an account the user does not (yet) have access to.

We could use individualised task links that allow Janeway to figure out whether the person has logged into the correct account for a given task. This would be a larger feature for down the road.

joemull:
Closes #2822

### DIFF
--- a/src/api/serializers.py
+++ b/src/api/serializers.py
@@ -23,7 +23,7 @@ class FrozenAuthorSerializer(serializers.HyperlinkedModelSerializer):
 
     class Meta:
         model = submission_models.FrozenAuthor
-        fields = ('first_name', 'middle_name', 'last_name', 'institution', 'department', 'country')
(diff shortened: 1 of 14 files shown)


ajrbyers:
@joemull if you're happy with the changes I made this can be merged once the conflicts are dealt with.

PR #3056, opened 2022-08-03
cached Issue display titles are now stored as translatable strings

### LINKED ISSUE #3047: Issue titles are no longer translatable
**Describe the bug**
Issue titles are no longer translatable, they output always in the same language

**Janeway version**
v1.4.2.1

**To Reproduce**
1. Create a journal issue with the journal set to any language
2. Visit the issues page in a different language
3. See issue title is always displayed as the language used in step 1

**Expected behavior**
Issue titles should remain translatable

**Additional context**
This was introduced in the fix for #2967, in particular this optimization: 7d943167e4fccd7713d7f1b07cee2cb2a9086a06
Since the generated title is stored in the database, it has become static and cannot be translated by the rendering engine.

One solution we've discussed is to pre-generate all translated titles when saving an issue. (relying on django-modeltransaltions to store each different translation). The trade-off being that these titles would not get updated whenever new translations are introduced through `.po` files.

An alternative approach would be to store the translation string in the database (e.g: `{% trans 'Volume' %} 1 &bull {% trans 'Issue' %} 1`), deferring the translation to Django's template rendering engine. It would require us to store the title in a separate field and turn `display_title` into a property that can return a rendered string when called.

mauromsl:
closes #3047


mauromsl:
jenkins: test this please

mauromsl:
Jenkins: test this please

src/journal/models.py
```diff
@@ -592,46 +600,62 @@ def is_serial(self):
...
             if article.page_range:
                 page_numbers = article.page_range
             elif article.total_pages:
-                if article.total_pages != 1:
-                    label = ugettext('pages')
-                else:
-                    label = ugettext('page')
                 num_pages = str(article.total_pages)
-                page_numbers = f'{num_pages} {label}'
+                page_numbers = f"{num_pages} {{% trans 'label' %}}"
```
joemull, line 643: I think we'd want 'label' to be changed to 'pages' here, since that is the translatable string, right?

src/journal/models.py
```diff
@@ -592,46 +600,62 @@ def is_serial(self):
...
 
         return [volume, issue, year, issue_title, article_number, page_numbers]
```
joemull, line 645: With the changes in this function there is a failing test in submission: `submission.tests.SubmissionTests.test_article_issue_title`. 

And the function it tests, `article.issue_title`, needs some changes:
Screenshot from 2022-08-04 11-13-31

joemull (changes requested):
Definitely an improvement--thank you. It looks like we need to make some changes to `article.issue_title` as well as handle translation of `pages` in `issue_title_parts`.

joemull (approved):

PR #3060, opened 2022-08-04
Adds controls for handling articles in Accepted stage

### LINKED ISSUE #3051: Articles getting stuck in stage Accepted
**Describe the bug**
When an article is accepted, several events are triggered behind the scenes such as depositing the article with Crossref and moving the article to the next workflow element. If an error happens during these events, there are a few side effects that we'd like to avoid and help users recover from.

A server error is raised, rather than a custom message, so the user has to click Back to get to a page they can take any further actions. The article is also moved into the Accepted stage, which typically is only a pass-through stage on the way to Copyediting, so it isn't very well handled for articles that get stuck there. If the user clicks Back to the article page, there are no actions they can take:

Screenshot from 2022-07-29 13-46-48

Screenshot from 2022-07-29 13-47-33

If the user tries to find the article in Search Submissions, the link to View Article does not take them anywhere:

Screenshot from 2022-07-29 12-55-56
 
Screenshot from 2022-07-29 12-56-23

Articles with stage Accepted also don't appear in the Review list page:

Screenshot from 2022-07-29 13-41-51

**Janeway version**
1.4.2

**Planned changes**

We discussed this and want to make the following changes:
- [x] When acceptance events fail, redirect back to Review page
- [x] Include articles with Accepted stage in Review list page
- [x] Fix "View Article" button for Accepted stage on Search Submissions page
- [x] Replace "Make a Decision" button with "Move to Next Stage" on Review page
- [x] Create a modal for Review page that opens with info if article in Accepted stage

joemull:
Closes #3051.

Screenshot from 2022-08-04 12-11-06

Screenshot from 2022-08-04 13-42-31

Screenshot from 2022-08-04 13-42-43

Screenshot from 2022-08-04 13-43-30


ajrbyers:
This is great! 🥇

ajrbyers (approved):

src/templates/admin/review/in_review.html
```diff
@@ -224,6 +224,14 @@ <h2>Actions</h2>
                         <li><a data-open="summarymodal"><i class="fa fa-server action-icon">&nbsp;</i>View Metadata</a>
                         </li>
                         <li><a href="{% url 'document_management' article.pk %}?return={{ request.path }}"><i class="fa fa-file-archive-o">&nbsp;</i>Document Management</a></li>
+                    {% if article.is_accepted %}
+                        <li>
+                            <a href="{% url 'move_to_next_workflow_element' article.pk %}">
+                                <i class="fa fa-arrow-circle-right action-icon">&nbsp;</i>
```
ajrbyers, line 230: As discussed, swap this for a <form method="POST" action="{% url '' %}">

src/workflow/views.py
```diff
@@ -52,3 +53,26 @@ def manage_article_workflow(request, article_id):
     }
 
     return render(request, template, context)
+
+
+@has_journal
+@editor_user_required
+def move_to_next_workflow_element(request, article_id):
```
ajrbyers, line 60: This function should have the  @require_POST decorator!

ajrbyers (changes requested):
Looks good, a couple of notes in the review.

src/templates/admin/review/in_review.html
joemull: Thanks for catching this. I've implemented POST as well as a current stage check.

ajrbyers (approved):

PR #3067, opened 2022-08-08
Enable preprint moderators to un-reject preprint

alainna:
Does this work to enable both moderators and super users/staff to un-reject preprints? Only preprint moderators should be able to access this page, no? 

https://github.com/BirkbeckCTP/janeway/issues/2539


ajrbyers (approved):
This is great, thanks! 🎖️

PR #3069, opened 2022-08-10
3059 when an editor completes a workflow element they will automatically be moved onto the next one rather than being directed to the dashboard

### LINKED ISSUE #3059: Stay with articles when moving between workflow elements
Janeway's workflow elements seem to have been designed to keep you with an article when moving between workflow elements. The workflow logic attempts to get the appropriate URL for the article page here: 

https://github.com/BirkbeckCTP/janeway/blob/ab6696fa8b3a75347ea54257aed107e3f75fa179/src/core/workflow.py#L94-L100

But the reverse matching is failing at present when moving to stages including Copyediting, Typesetting, and Prepublication. As a result you are sent to the dashboard: 

https://github.com/BirkbeckCTP/janeway/blob/ab6696fa8b3a75347ea54257aed107e3f75fa179/src/core/workflow.py#L102-L104

Find out what's not working with handshake URLs that prevents reverse matching from succeeding and fix it.

ajrbyers:
Closes #3059 

**Note:** we are aware that this does not work for Production Managers in the old Production Proofing workflow, this is a limitation that cannot be handled at the moment.

### DIFF
--- a/src/core/workflow.py
+++ b/src/core/workflow.py
@@ -64,9 +64,10 @@ def workflow_next(handshake_url, request, article, switch_stage=False):
     :param handshake_url: Current workflow element's handshake url
     :param request: HttpRequest object
     :param article: Article object
+    :param switch_stage: Boolean
     :return: HttpRedirect
     """
-
+    response = None
     workflow = models.Workflow.objects.get(journal=request.journal)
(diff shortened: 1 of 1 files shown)


joemull (approved):
This is great!

Issue #2875, opened 2022-05-18
Preprints: OAI-PMH feed for preprints


alainna:
**Is your feature request related to a problem? Please describe.**
As a Preprints server Moderator/Admin, I want external systems to have persistent access to metadata about our preprints via a data feed. 

**Describe the solution you'd like**
Develop OAI-PMH support for Janeway Preprints

Required fields: 

* Source (Repository name) 
* Publisher (?? Press name??) 
* Preprint publish date
* Date last updated 
* Preprint title  
* Preprint Identifier (DOI)
* Repository ID (Preprint ID)  
* Document type 
* Preprint authors 
* Preprint subject/categories 
* Preprint language (??set at preprint server level??) 

Optional fields: 

* Preprint licence  
* Preprint abstract 
* Preprint version 
* Preprint version publish date 
* Keywords 
* Preprint author identifiers 
* Preprint author affiliations 
* Published version identifier (Publisher DOI)

alainna:
https://github.com/BirkbeckCTP/janeway/pull/3098

alainna:
The OAI-PMH feed is now live. We should be able to consider this closed -- no?

PR #3098, opened 2022-08-23
#2875 oai pmh endpoint for preprints

everreau:
When users go to OAI-PMH endpoints for preprint repositories respond with appropriate preprint articles and metadata.

### DIFF
--- a/src/api/oai/views.py
+++ b/src/api/oai/views.py
@@ -272,7 +272,7 @@ class OAIListPreprintRecords(OAIPagedModelView):
         queryset = super().get_queryset()
 
         if self.request.repository:
-            queryset = queryset.filter(repository=self.request.repository)
+            queryset = queryset.filter(repository=self.request.repository).order_by('date_published')
         return queryset
 
     def filter_is_published(self, qs):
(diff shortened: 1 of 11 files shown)


Bbkctp:

ajrbyers:
Test this please.

everreau:
I have added tests for the preprint oai views.

src/api/oai/views.py
```diff
@@ -31,7 +32,10 @@ def oai_view_factory(request, *args, **kwargs):
             verb = DEFAULT_ENDPOINT
         elif verb not in ROUTES:
             raise exceptions.OAIBadVerb()
-        view = ROUTES.get(verb)
+        if request.repository:
+            view = PREPRINT_ROUTES.get(verb)
+        else:
+            view = ROUTES.get(verb)
```
mauromsl, line 38: One issue with splitting the preprint routes from the previous views is that there won't be a unified feed from the press site anymore (preprints will only be returned from the OAI feed of the preprint site). This would be problematic for some installations where a publisher is hosting a combination of journals and preprints in the same Janeway instance.

mauromsl (approved):
Thank you so much for your work here @everreau , it is pretty good!

There is a high level issue with the way in which the views have been split between Journal and other OAI views, which will lead the press feed to ignore preprints for now. We intend on eventually merging the `repository.models.Preprint` and `submission.models.Article` models, so I'm thinking it is best to ignore this problem for now.

src/api/oai/views.py
everreau: Thanks @mauromsl ! Yes, it will be great to have preprint and article merged and will make it a lot easier to create a unified feed.

PR #3108, opened 2022-08-31
Fixup test_oai.py

hardyoyo:
- Add a blank init so Nose realizes it needs to run tests in the API folder
- Fix some broken tests
- Address lint errors as reported by VScode
- Add docstrings for each test
- Rename one duplicate method name (test_get_record_jats)
- Drop unused variable i in for loop
- Revise some fixtures to match current output of OAI endpoint

NOTE: after merging this change, you will see one remaining test failure, for the test_list_sets test, and one error. This is not a complete fix for broken tests, it is a partial one, and will help y'all see the remaining failure and error.

Merging this *ought* to cause Jenkins to fail... which is correct, it should have been failing all along. The build *should* be broken because there are test failures.


hardyoyo:
I think the most important fix included in this PR (enabling the API tests to be detectable by nose) has been made in another PR elsewhere. And the other fixes are just window dressing, for the most part... AND the API tests have changed in other ways, or have other pending pull requests... so... this open/draft PR is muddying the waters, and is no longer necessary. I'm closing this.

PR #3120, opened 2022-09-07
Added new settings to disable article thumbnails and article large image independantly 

mauromsl:
A migration will toggle both settings on for installations that had toggled the `disable_article_images` boolean on the journal


ajrbyers (approved):
This looks great. I've approved but there are conflicts with `core/forms.py`

PR #3123, opened 2022-09-07
RSS feed titles and descriptions are now not terrible.

ajrbyers:


src/rss/views.py
```diff
@@ -51,13 +54,17 @@ def item_link(self, item):
...
 
     def get_object(self, request, *args, **kwargs):
         return request.journal
 
+    def title(self, obj):
+        return "{} Article Feed".format(obj.name)
+
+    def description(self, obj):
+        return "A feed of the 10 latest articles from {}".format(obj.name)
+
```
mauromsl, line 67: can `obj` be `None` ? It seems that way from `self.get_object`, in which case it would raise an `AttributeError` here

mauromsl:
jenkins: test this please

ajrbyers:
Looks like this one is good to go!

mauromsl:
> Looks like this one is good to go!

Waiting on Jenkins at the moment, there were a few tests still failing

ajrbyers:
> > Looks like this one is good to go!
> 
> Waiting on Jenkins at the moment, there were a few tests still failing

Ah, it showed passed to me!

mauromsl:
> > > Looks like this one is good to go!
> > 
> > 
> > Waiting on Jenkins at the moment, there were a few tests still failing
> 
> Ah, it showed passed to me!

Shows as pending to me, but you are right, build has passed! 💚

PR #3124, opened 2022-09-07
2841 fixes article links on profile pages

### LINKED ISSUE #2841: Public profile page links to broken article URLs from other journals
**Describe the bug**
On the user public profile page, publications from other journals are rendered using the `article_listing` component.
This component renders article urls using the `journal_url` template tag which uses the current journal (`request.journal`) to construct the URL. In situations where the journal is configured with a custom domain, this leads to a broken URL.

The same problem occurs with the thumbnail URL which won't load properly.

We should either use absolute URLs for the articles (`submission,models.Article.url`) in `article_listing.html` or drop articles from other journals being listed from the user's public profile.

**Janeway version**
v1.4.1-RC-6

**To Reproduce**
1. Create a user with publications on two or more journals
2. visit the user profile page from one of the journals
3. See error on article URLs from other journals

**Expected behavior**
Either the URLs resolve correctly to the other journals or this articles to be omitted

**Screenshots**
image

## Front-end Issues

ajrbyers:
Closes #2841

### DIFF
--- a/src/press/migrations/0014_auto_20171127_1502.py
+++ b/src/press/migrations/0014_auto_20171127_1502.py
@@ -16,7 +16,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='press',
             name='homepage_preprints',
-            field=models.ManyToManyField(to='submission.Article'),
+            field=models.ManyToManyField(to='submission.Article', blank=True),
         ),
         migrations.AddField(
             model_name='press',
--- a/src/press/models.py
+++ b/src/press/models.py
@@ -112,7 +112,7 @@ class Press(AbstractSiteModel):
     preprint_decline = models.TextField(blank=True, null=True, default=press_text('decline'))
 
     random_homepage_preprints = models.BooleanField(default=False)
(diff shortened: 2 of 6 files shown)


src/press/migrations/0014_auto_20171127_1502.py
```diff
@@ -16,7 +16,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='press',
             name='homepage_preprints',
-            field=models.ManyToManyField(to='submission.Article'),
+            field=models.ManyToManyField(to='submission.Article', blank=True),
```
joemull, line 19: I might be missing something but these changes don't seem to match the description for this commit: https://github.com/BirkbeckCTP/janeway/pull/3124/commits/07de9335c1e351529068181d33e448ade8928932

joemull (commented):
I have a question about one of the commits.

src/press/migrations/0014_auto_20171127_1502.py
joemull: We discussed this on a call, and it is a safe change from another issue.

joemull (approved):

PR #3128, opened 2022-09-09
Adds support for Issue and Journal DOIs to Crossref Integration

### LINKED ISSUE #2817: Feature to register issue DOI's
We would like to register issue DOI's for some of our journal

implement a solution for DOI registration of issue doi's, much in the same as is the case now for articles

**Additional context**

### LINKED ISSUE #2826: Title Level DOI
Some journals may not have an ISSN so we should support title level DOIs.

Allow editors to insert a title level DOI and if present push it with the articles.

**Additional context**
We don't need to mint them, just use them.

mauromsl:
For  journal issues, it includes settings and an event that allows an issue DOI to be automatically assigned, similar to how article DOIs are autogenerated from a pattern. The event is raised when an article is first added to an issue.
Closes #2817 
closes #2826


src/identifiers/logic.py
```diff
@@ -625,3 +625,28 @@ def register_preprint_doi(request, crossref_enabled, identifier):
...
+        'Identifiers', 'use_crossref', issue.journal,
+        default=True,
+    ).processed_value
+    if auto_assign_enabled and not issue.doi:
+        issue.doi = generate_issue_doi_from_logic(issue)
+        issue.save()
+
+
+def on_article_assign_to_issue(article, issue, user):
+    auto_assign_issue_doi(issue)
```
joemull, line 652: This event isn't fired when adding articles via the issue manager, and I think it should be.

src/templates/common/identifiers/crossref_journal_issue.xml
```diff
@@ -7,6 +7,12 @@
                 {% if crossref_issue.journal.journal_issn and crossref_issue.journal.journal_issn != '0000-0000' %}
                 <issn media_type="electronic">{{ crossref_issue.journal.journal_issn }}</issn>
                 {% endif %}
+                {% if crossref_issue.journal.doi %}
+                <doi_data>
+                    <doi>{{ crossref_issue.journal.doi }}</doi>
+                    <resource>{{ crosref_issue.journal.url }}</resource>
+                </doi_data>
+                {% endif %}
```
joemull, line 15: The `journal` here is different from the standard journal object. It is created by `create_crossref_journal_context` above (in order to handle custom titles in ISSNs). So the `doi` attribute will need to be added there before it shows up in this template.

joemull (changes requested):
This is really great--I think it will be very useful.

There are a few potential bugs that I managed to catch. Comments inline.

joemull (approved):

PR #2941, opened 2022-06-20
Isolate code that loads hooks.

gamboz:
So that django apps (i.e. not Janway plugins) can register their own hooks.

E.g.
```python
# myapp/apps.py

    def ready(self):
        """Call during initialization."""
        ...
        self.register_hooks()

    def register_hooks(self):
        """Register my functions to Janeway's hooks."""
        hooks = [
            dict(nav_block=dict(module='myapp.hooks',
                                function='example_hook')),
        ]
        # NB: do not `import core...` before `ready()`,
        # otherwise django setup process breaks
        from core import plugin_loader
        plugin_loader.register_hooks(hooks)
```

PR #3149, opened 2022-09-21
Isolate code that loads hooks.

gamboz:
So that django apps (i.e. not Janway plugins) can register their own hooks.

This is the same as PR #2941 (unfortunately I wasn't able to just re-open the old one).

### DIFF
--- a/src/core/plugin_loader.py
+++ b/src/core/plugin_loader.py
@@ -25,6 +25,25 @@ def get_dirs(directory):
     return dirs
 
 
+def register_hooks(hooks: list):
+    """Register plugin hooks."""
+    # Hooks are dictionaries. For a description,
+    # see "Technical Configuration" (docs/source/configuration.rst)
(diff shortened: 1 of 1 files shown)


gamboz:
Sorry, I see that I have some failing checks. I've set the PR as draft, I'll fix them and be back :(

gamboz:
Can't replicate the errors locally.
I think the PR is ok :)

ajrbyers:
I think they are both fixed in master so should be fine.

ajrbyers:
Test this please

joemull:
@ajrbyers I think this needs to be rebased if we want the test to pass. Shall I rebase it?

ajrbyers:
> @ajrbyers I think this needs to be rebased if we want the test to pass. Shall I rebase it?

Yeah go ahead, I don't think Mr Jenkins is listening to me so maybe the rebase process will push the test.

PR #3158, opened 2022-09-26
Test fixes for version 1.4.3-release-candidate-0

joemull:


joemull:
@ajrbyers This should be good to go. As a warning, [bc31ffc](https://github.com/BirkbeckCTP/janeway/pull/3158/commits/bc31ffce31a5038196c03c97da09b42196bb9b50) represents a kind of fanatical but ultimately necessary standardization of how we create and query role names and slugs, at least in testing.

ajrbyers:
@joemull this looks good, the only thing to check is whether the typesetting plugin uses this role. I don't think it does though.

joemull:
> @joemull this looks good, the only thing to check is whether the typesetting plugin uses this role. I don't think it does though.

Good catch--I just checked and there aren't any uses of role in the typesetting plugin that would run afoul of these changes.

Issue #3010, opened 2022-07-18
Preprint endpoint in API


alainna:
**Is your feature request related to a problem? Please describe.**
It would be beneficial to external connections to have a preprint endpoint in the Janeway API

**Describe the solution you'd like**
Preprint endpoint in Janeway API; get preprint metadata in tasty XML/Json format. yay!

alainna:
@ajrbyers FYI, added the tables with data I'd expect that we would wish to be accessible via API.

hardyoyo:
@alainna I wonder if you could have a look at the API endpoint added by #3166 and let me know what's still missing? The Django REST Framework mostly adheres to the Django object models, so... getting anything more than the preprint model onto the preprint endpoint will take some creative coding.

PR #3166, opened 2022-09-27
Add preprint API endpoint

hardyoyo:
Add a read-only preprints endpoint to the existing Janeway API. This is mostly the work of @tingletech as the commits show, with a few changes from me to get this endpoint to work with the current version of Janeway (with considerable help and cheerleading from @ajrbyers).


hardyoyo:
Addresses most of #3010 but I think misses a few points

alainna:
Thanks @hardyoyo ! 

There are two major issues that I would consider blockers on the preprints API endpoint based on my current tests, as well as some suggestions for how it might be improved: 

1. Lack of ability to filter by repository: Currently the preprints endpoint displays data about all preprints in Janeway, and not just the specific repository one may be accessing. It may therefore be preferable to either a) add a repository filter, or b) add a repository endpoint which should be the main method of obtaining data about preprints in that repository. One should be able to query the repository endpoint to learn of the active public repositories, then query eg /repository/[id]/ or /repository/[short-name]/ to obtain the preprints under that repository. 

2. Release of data that probably shouldn't be public: The author section seems to pull all data from the account table, including some data that users may not be aware is being shared (and which isn't shared via the UX in some cases): 
   * Email address (for both submitting and non-submitting authors); 
   * Whether account is active (for all authors). 

The second is the most critical to resolve. 

There is additional metadata that I would expect to see in the preprint API endpoint. However, I don't consider those immediately blockers, so I'll include that in the feature request, then reference that comment here.

hardyoyo:
This PR has been updated to address the two major issues identified above: e-mail and is_active are now hidden from author data, and the api URL for each respository limits the output to just the contents of that repository. This code is currently deployed on our dev instance, and can be viewed here: 

- https://dev.eartharxiv.org/api/preprints/
- https://dev.ecoevorxiv.org/api/preprints/

hardyoyo:
This is just a start, I know that the output might not be quite sufficient, but changes can be made fairly easily.

alainna:
@hardyoyo This is epic to see, great work!! There are a few things missing that we'd like to see in a "GET" version, as the primary use case is to allow a manuscript submission system to make a single GET call to obtain metadata for a specific preprint (``https://dev.eartharxiv.org/api/preprints/1234`` ) or series of preprints (``https://dev.eartharxiv.org/api/preprints/``). 

Regarding metadata to add: I think the essence of what we want is already in the JATS version of the OAI feed for preprints: https://github.com/BirkbeckCTP/janeway/pull/3098

It would be good to double check whether the OAI feed also includes the custom fields a Repository may ask during submission, e.g. the "Data availability" statement that we see in EA/EER. I believe those are in ``repository_repositoryfield`` and ``repository_repositoryfieldanswer/``. Those should also be included. :)

hardyoyo:
From a quick perusal of the OAI feed, I think mostly this API is missing a link to download files, and a publisher field (though that could be inferred from the source of the API). The OAI feed does have a "description" field, but I believe this is "abstract" rebranded.

ajrbyers:
We're tagging RC-7 tomorrow, so you want to sneak this one in it will need to be finished up today. Otherwise we can fold it in 1.5.1 in Spring 2023.

hardyoyo:
Ugh, sorry, only just now seeing this. I can try to focus on this tomorrow, but I have a site upgrade scheduled first thing in the morning.

On Nov 14, 2022 6:19 AM, Andy Byers ***@***.***> wrote:

CAUTION: EXTERNAL EMAIL

We're tagging RC-7 tomorrow, so you want to sneak this one in it will need to be finished up today. Otherwise we can fold it in 1.5.1 in Spring 2023.

hardyoyo:
So, my main question is, does it really have to be perfect before it gets in to a release?

On Nov 15, 2022 9:42 PM, Hardy Pottinger ***@***.***> wrote:
Ugh, sorry, only just now seeing this. I can try to focus on this tomorrow, but I have a site upgrade scheduled first thing in the morning.

On Nov 14, 2022 6:19 AM, Andy Byers ***@***.***> wrote:

CAUTION: EXTERNAL EMAIL

We're tagging RC-7 tomorrow, so you want to sneak this one in it will need to be finished up today. Otherwise we can fold it in 1.5.1 in Spring 2023.

hardyoyo:
Adding the additional model properties to this REST endpoint will be a journey of discovery, and may take some time (it will take _this_ developer some time to figure out how to even add such a property/field to a dev repository). Here's a breadcrumb courtesy of @ajrbyers : https://stackoverflow.com/questions/17066074/modelserializer-using-model-property

alainna:
Hey there, the CDL team just discussed this and will make a retraction of the requirement to include custom repository fields in the API for now given the difficulty. Let's add this instead as a new Feature Request to improve the API functionality in the next phase.

ajrbyers:
If someone can tidy up the conflicts I could probably add the custom fields for you for the 1.5 release.

hardyoyo:
I'll get the conflicts done, I think probably a new ticket for the improvements is the best route, and then you can do that ticket at your convenience?

hardyoyo:
Revising the title of this PR, since pagination was added in another PR

src/api/views.py
```diff
@@ -141,6 +142,20 @@ def get_queryset(self):
         return queryset
 
 
+class PreprintViewSet(viewsets.ModelViewSet):
+    """
+    API Endpoint for preprints.
+    """
+    pass
```
ajrbyers, line 149: Probably should remove this.

src/api/serializers.py
```diff
@@ -85,7 +102,7 @@ class JournalSerializer(serializers.HyperlinkedModelSerializer):
 
     class Meta:
         model = journal_models.Journal
-        fields = ('pk', 'code', 'name', 'publisher', 'issn', 'description', 'current_issue', 'default_cover_image',
```
ajrbyers, line 88: Undo this line.

ajrbyers (commented):
Two comment inline, other than that good to go,

ajrbyers:
@hardyoyo I've added two comments, and there is a test failing on this branch.

ajrbyers:
test this please

hardyoyo:
The custom fields part of the original request has been moved to a new issue #3371

src/api/serializers.py
```diff
@@ -67,13 +68,27 @@ class Meta:
...
+class PreprintSupplementaryFileSerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = repository_models.PreprintSupplementaryFile
+        fields = ('url', 'label',)
 class IssueSerializer(serializers.HyperlinkedModelSerializer):
 
     class Meta:
         model = journal_models.Issue
-        fields = ('pk', 'volume', 'issue', 'issue_title', 'date', 'issue_type', 'issue_description',
```
ajrbyers, line 75: Can you tell us why you removed this whole line?

src/api/serializers.py
```diff
@@ -85,8 +100,6 @@ class JournalSerializer(serializers.HyperlinkedModelSerializer):
 
     class Meta:
         model = journal_models.Journal
-        fields = ('pk', 'code', 'name', 'publisher', 'issn', 'description', 'current_issue', 'default_cover_image',
```
ajrbyers, line 88: Can you tell us why you removed this whole line?

ajrbyers (changes requested):
See comments.

src/api/serializers.py
hardyoyo: I'm pretty sure you asked me to 'remove this, please', in an earlier line comment

ajrbyers: I think I asked you to undo the change you made on that line...

hardyoyo:
I believe it is, I don't know why GitHub persists in saying changes are requested

src/api/serializers.py
hardyoyo: I've since put this line of code back. I plead ignorance + distraction, I thought you told me to delete it, so I did. Whoops! I put it right back and marked this conversation as resolved. On the off chance that GitHub really wanted me to answer your question, I'm answering it now.

hardyoyo:
erm... is this going to be merged into 1.5.1?

PR #3172, opened 2022-09-28
#3171 adds an email sent message whenever an email has a request and request.user object

### LINKED ISSUE #3171: Email sent confirmation message
**User Story**
This is based on a production editor's feedback.

As an email writer, I am sometimes not 100% sure whether an email was sent, right after I click a button that should have sent it. I see other messages in the bottom right-hand corner like "Review assignment created" or "Typesetter assigned", but these don't tell me about the email. I had the option to skip the email, so I know it's a separate thing. Why would Janeway tell me only about one thing happening, if two things were supposed to happen? It raises just a tiny bit of doubt in the back of my mind.

Peek 2022-09-28 15-29

What if Janeway showed me an email-related message in the bottom-right hand corner, like the other green progress bar things? That would help give me confidence that the emails are going out.

Of course, the email log exists, and if I know it's there, I can go check it. But really that's beside the point. It's about the consistency and thoroughness of the little boxes that appear in the bottom right, allowing me to [recognize something I know should be there](https://www.nngroup.com/articles/recognition-and-recall/).

Add a message like "Email sent" to the messages framework whenever someone prompts an email to go out. This would be in addition to the message about task creation. The "Email sent" message would need to be consistently implemented across all email interfaces since if folks see it some places but don't see it others, they will think some emails are not being sent.

**Additional context**
It is tempting to consider things like this as unneeded, given our issue backlog. But they make people feel better, calmer, and more in control when using Janeway, and the value of that for our project cannot be overstated.

- https://www.nngroup.com/articles/visibility-system-status/
- https://www.nngroup.com/articles/recognition-and-recall/

ajrbyers:
Closes #3171

### DIFF
--- a/src/utils/notify_plugins/notify_email.py
+++ b/src/utils/notify_plugins/notify_email.py
@@ -5,6 +5,7 @@ from collections import Iterable
 from django.conf import settings
 from django.core.mail import EmailMultiAlternatives
 from django.utils.html import strip_tags
+from django.contrib import messages
 
 from utils import setting_handler
 from utils import notify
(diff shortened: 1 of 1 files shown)


joemull (approved):
A thing of beauty

Issue #3185, opened 2022-10-05
v1.4.2.1 python manage.py install_janeway / No module named 'nose'.


cederom:
**Describe the bug**
During `python manage.py install_janeway` missing `nose` module error occurs.

**Janeway version**
`v1.4.2.1`

**To Reproduce**
1. `python manage.py install_janeway`
2. See `nose` module missing.

**Expected behavior**
No error.

**Screenshots**
```
$ python manage.py install_janeway
Operations to perform:
  Apply all migrations: admin, auth, carousel, cms, comms, contenttypes, copyediting, core, cron, discussion, django_summernote, featured, identifiers, journal, metrics, press, production, proofing, repository, review, sessions, submission, utils
Running migrations:
  Applying core.0001_initial... OK
(..)
  Applying journal.0053_display_article_images_settings...Parsed setting publication_fees
Parsed setting publisher_url
Parsed setting privacy_policy_url
Parsed setting publisher_name
Parsed setting copyright_notice
Parsed setting acceptance_criteria
Parsed setting collection_name
Parsed setting collection_name_plural
Parsed setting submission_checklist
Parsed setting from_address
Parsed setting replyto_address
Traceback (most recent call last):
  File "/home/XXX/.local/venv3.6janeway/lib/python3.6/site-packages/django/db/models/query.py", line 464, in get_or_create
    return self.get(**lookup), False
  File "/home/XXX/.local/venv3.6janeway/lib/python3.6/site-packages/django/db/models/query.py", line 380, in get
    self.model._meta.object_name
core.models.DoesNotExist: SettingValue matching query does not exist.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/XXX/.local/venv3.6janeway/lib/python3.6/site-packages/django/template/utils.py", line 65, in __getitem__
    return self._engines[alias]
KeyError: 'django'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/XXX/.local/venv3.6janeway/lib/python3.6/site-packages/django/template/backends/django.py", line 126, in get_package_libraries
    module = import_module(entry[1])
  File "/usr/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
[...]

PR #3189, opened 2022-10-05
#3185 Add django-nose to requirements.txt

joemull:


joemull:
We are going to change tack on this to try to avoid requiring `django-nose`.

src/core/templatetags/debug_tools.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+
+    def render(self, context):
+        try:
+            from nose import tools
+            tools.set_trace()         # Debugger will stop here
+        except ImportError:
+            import pdb
+            pdb.set_trace()           # Debugger will stop here
+        except ImportError:
```
mauromsl, line 25: This would need to live within a nested try/except block, otherwise it will not be handled.

```
try:
    foo()
except Exception:
    try:
        bar()
    except Exception:
        pass
```

That said, `pdb` is part of the standard library, so there is no need to handle an import error here, so you can remove the second exception handler

mauromsl (changes requested):
Thanks Joe, one comment in line

src/core/templatetags/debug_tools.py
joemull: Thanks for catching that. Will remove it.

mauromsl (approved):

PR #3229, opened 2022-10-19
Fix bug that let submission editors see other articles in Search Submissions

### LINKED ISSUE #3228: Section editors can see all articles in Search Submissions if they reorder them
Section editors can see all articles in Search Submissions if they use the Order dropdown:

Image

They cannot click to the article page, as they get permission denied.

Still, on the search page, they should only be able to see articles to which they've been assigned.

joemull:
Fixes #3228.

### DIFF
--- a/src/templates/admin/core/active_submissions.html
+++ b/src/templates/admin/core/active_submissions.html
@@ -15,7 +15,7 @@
 
         <div class="row expanded">
             {% include "admin/elements/no_stage.html" %}
-            <div class="large-8 columns end">
+            <div class="{% if editor %}large-8{% else %}large-12{% endif %} columns end">
                 <div class="box">
                     <div class="title-area">
                         <h2>Submission List</h2>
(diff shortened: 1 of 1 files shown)


src/templates/admin/core/active_submissions.html
```diff
@@ -31,14 +31,12 @@ <h2>Submission List</h2>
...
-            <div class="large-4 columns">
-                <div class="box">
-
-                    <div class="content">
-
-                        <div class="row">
-                            <form id="filter_form" class="autoSubmit" method="POST">
-                                {% if editor %}
+            {% if editor %}
+                <div class="large-4 columns">
```
mauromsl, line 35: With this 4 column div dissappearing, the row only has 8 columns declared, which can mess up the grid when introducing other rows or on other port sizes. It is worth setting the sibling to `large-12` when this div is not rendered

mauromsl (changes requested):
Thanks Joe, one small recommendation and we are good to merge

PR #3246, opened 2022-10-25
Fix bug that caused broken link on OLH press nav

### LINKED ISSUE #3245: Empty link field still causes problems in nav
**Describe the bug**
If you leave the Link field blank when creating a top-level nav item (as described in #3234), the idea was that it would just function as a dropdown, not a link in itself. While the fix for #3234 took out the sub-nav item, it did not remove the link from the top-level item. As a result, you can still click on it, leading to a 404. We will need to remove the anchor element in the templates if the link field is empty.

joemull:
Fixes #3245.

### DIFF
--- a/src/cms/models.py
+++ b/src/cms/models.py
@@ -64,10 +64,7 @@ class NavigationItem(models.Model):
         return self.link_name
 
     def sub_nav_items(self):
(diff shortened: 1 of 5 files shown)


ajrbyers (approved):
🎖️

PR #3252, opened 2022-10-26
Fixes a bug where a carousel limits were not being applied when latest_articles is set.

ajrbyers:


ajrbyers:
test this please

src/core/homepage_elements/carousel/models.py
```diff
@@ -91,6 +91,8 @@ def get_items(self):
                 articles |= core_logic.latest_articles(self, 'press')
             elif hasattr(self, 'journal'):
                 articles |= core_logic.latest_articles(self, 'journal')
+            if self.article_limit > 0:
+                articles = articles[:self.article_limit]
```
mauromsl, line 95: I think this makes the slice in line 101 redundant now

We also need to apply the same change to the news query below, I think

mauromsl (changes requested):

mauromsl (approved):

PR #3258, opened 2022-10-31
Password length not showing in pwd reset form

gamboz:
I think it's because we are accessing an object attribute from the `blocktrans` [block](https://docs.djangoproject.com/en/1.11/topics/i18n/translation/#blocktrans-template-tag).

It probably happens in many other forms also.

For completeness, django 4.1 uses the tag [`blocktranslate`](https://docs.djangoproject.com/en/4.1/topics/i18n/translation/#blocktranslate-template-tag) (instead of `blocktrans`).


ajrbyers:
Thanks @gamboz!

PR #3273, opened 2022-11-09
Fixes a bug where a reminder might be sent when an article is no longer under review. Also ensures the review_url and do_revisions_url variables are available to reminder templates.

### LINKED ISSUE #3272: Review Reminders Bug
**Describe the bug**
A reviewer recieved a review reminder and then could not find the review once logging into Janeway. The email also did not have a link in it.

**Janeway version**
1.4.3

**Expected behavior**
The article in question was no longer in review (it had been rejected) so:

1. The email should not have been sent &
2. The email _should_ have had a link in it

**Screenshots**

Screenshot 2022-11-09 at 12 09 27

ajrbyers:
Closes #3272

### DIFF
--- a/src/cron/models.py
+++ b/src/cron/models.py
@@ -8,10 +8,13 @@ from django.db import models
 from django.utils import timezone
 from django.db.models import Q
 from datetime import timedelta
+from django.shortcuts import reverse
 
 from cron import logic
 from journal import models as journal_models
 from utils import render_template, notify_helpers
+from submission import models as submission_models
+from review import logic as review_logic
 
 
 class CronTask(models.Model):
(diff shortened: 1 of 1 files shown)


src/cron/models.py
```diff
@@ -172,7 +193,7 @@ def send_reminder(self, test=False):
                 )
                 # Create a SentReminder object to ensure we don't do this more than once by accident.
                 SentReminder.objects.create(type=self.type, object_id=item.pk)
-                print('Reminder sent for {0}'.format(object))
+                print('Reminder sent for {0}'.format(item))
```
mauromsl, line 196: Should invoke the logger rather than `print`

mauromsl (commented):
One small comment

ajrbyers:
print() swapped with logger.info()

PR #3301, opened 2022-11-30
Add pagination for Django REST Framework to Janeway global settings

hardyoyo:
Add pagination for Django REST Framewark, so that all API endpoints paginate their results.


ajrbyers (approved):
Thanks @hardyoyo

PR #3324, opened 2022-12-16
Use environment variables for defaults for the install_janeway management command

hardyoyo:
- Add arguments for the install_janeway command for press_name, press_domain, press_contact, journal_name, journal_description, journal_code, journal_domain
- Use environment variables for default values
- Hard code the existing default values into the getenv method invocation for each env, so the script behaves the same as previous versions of Janeway


mauromsl (changes requested):
Thanks @hardyoyo! just one comment inline regarding the use in command line of these arguments

mauromsl (approved):

PR #3332, opened 2022-12-21
Add journal_manager role, and related is_journal_manager checks

### LINKED ISSUE #495: Trigger Ajax reload on slider click
On active submissions, clicking on any of the sliders should auto trigger the Ajax refresh

### LINKED ISSUE #945: Journal Manager Role
Current permission levels are not as clear as they should be, and utilising is_staff isn't the best solution as its a Django permission that gives site-wide access.

The introduction of a new role "Journal Manager" or similar so that we can limit access to certain areas/settings only to those with higher security clearance without granting site wide Django permission is_staff. JM role could be set only by staff/admin/superusers.

None come to mind in particular.

**Stuff to do**
A few things to do for 1.5:
- [x] Update the group settings getter to filter settings a user cannot edit
- [x] Tests
- [x] Documentation on how this whole things work
- [x] All Settings edit_key form should validate JSON input
- [x] Test all JSON fields on a clean DB

### LINKED ISSUE #3232: Optionally disallow editors from making changes with significant legal or operational impacts
Janeway distributes an impressive amount of responsibilities to editors. However, in some situations, this creates (for some organizations) unacceptable legal risks. Specifically, it is our opinion that the copyright statement and license options are only safely edited by administrators.

For Copyright:
- Located in Submission Settings, but also accessible from All Settings.
- There are two settings: copyright_notice and copyright_submission_label
- In addition to restricting user input, would also need to prevent this item being removed from Submission Fields Configurator

For Licenses:
- Has its own settings page: Licence Manager. It is not listed in All Settings.
- Is present in Submission Fields Configurator, so would need to prevent users from removing it there.

For Email:
- Settings > General > Email > From Address (This should not be changeable / needs to always be set to our no-reply address, i.e. a change here can impact the system's ability to send e-mail *at all*).

Trust and documentation are options, however, this presents us with an unacceptable legal risk. The more journals we launch, the greater the overall risk of a mistake being made.

**Additional context**
Obviously, not every Janeway installation will be this risk-averse. This feature should therefore be toggleable in some way. We leave the implementation details for the developer.

One possible approach would be to build out the role configuration, to facilitate fine-grained permissions for the fields listed above. However, this is not a requirement.  Being able to limit write access to these fields to admin users *is* the only requirement.

hardyoyo:
- Add journal_manager role migration
- Add is_journal_manager check to Account model
- Add journal_manager_user_required decorator for later convenience
- Add checks for journal_manager role to portions of group_journal,
group_submission, configurator and licenses admin templates

fixes #945
fixes #3232

This PR is replaced by https://github.com/BirkbeckCTP/janeway/pull/3374 so closing this


hardyoyo:
Oh, I can see two things wrong with this PR: it accidentally includes an unrelated commit, and it has a typo mention of an unrelated issue. I'll fix those things and push again

ajrbyers:
@hardyoyo you should be able to tweak the create function for this Serializer to stop Editors for assigning the role.

https://github.com/BirkbeckCTP/janeway/blob/d1cd3c9b98baf39197cecda4a75da0d1c2064798/src/api/serializers.py#L116

hardyoyo:
word is this is still not quite right, @justingonder will post with more details soon, do not merge, I'll mark this as a draft

hardyoyo:
Marked this as ready for review last week (removed the "draft" flag), as you can see from the flurry of commits, there were a fair number of tiny bugs to fix. I can squash all of the commits if that would make reviewing the PR easier. Let me know if you'd like this squashed.

src/api/serializers.py
```diff
@@ -121,9 +121,10 @@ class Meta:
 
     def create(self, validated_data):
         role = validated_data.get("role")
+        disallowed_roles = ['reader', 'journal-manager']
 
-        if role.slug == 'reader':
-            raise serializers.ValidationError({"role": "You cannot add a user as a reader via the API."})
+        if role.slug in disallowed_roles:
+            raise serializers.ValidationError({"role": "You cannot add a user as a " + role.slug + " via the API."})
```
mauromsl, line 127: I'd feel more comfortable if this logic took place inside a validator except of happening ad-hoc at `create` time:
https://www.django-rest-framework.org/api-guide/validators/

src/templates/admin/submission/manager/licenses.html
```diff
@@ -43,6 +45,7 @@ <h2>Current Licences</h2>
                     </form>
                 </div>
             </div>
+            {% if user_is_journal_manager %}
```
mauromsl, line 48: as above, we'd need backend validation too

src/templates/admin/submission/manager/configurator.html
```diff
@@ -16,6 +18,7 @@
     <div class="box">
         <div class="row expanded">
             <div class="large-6 columns">
+                {% if user_is_journal_manager %}
```
mauromsl, line 21: While this invocation will stop the inner HTML from rendering, it won't actually validate on the backend, we'd need validation to take place there.

mauromsl (changes requested):
I've added a couple of comments about the implementation inline.

As for the introduction of this new role, I don't think it is a simple change to introduce and needs a bit more thought:
Do we want to disallow editors from accessing the "All settings" interface? I think this approach is a bit extreme and I would rather opt for a more configurable approach. Instead, I would add a new attribute/row to the `'core.Setting'` model that determines the roles which can alter a given setting. The `journal_defaults.json` file would provide the default values while allowing individual installs to alter our defaults. Janeway. 
My reasoning is based around 3  lines of thought:
 1. I believe this approach would be more flexible towards the introduction of new built-in roles (e.g. managing editor vs editor) 
  2. Configurable permission settings would also take us one step closer towards the creation of custom roles (e.g. web content editor, review manager...). 
  3. A flexible system like the proposed would allow us to not turn this feature into a breaking change for folks currently using Janeway, while also allowing for the conditions of the use case presented here to be met.
  
Another discussion point is to decide what areas of the system should be accessible to a user who is granted the `journal-manager` role but not the editor role, since I can't see that case covered in this PR
  
Finally, we'd need to update the documentation and editor guides accordingly, clearly stating what is on/offlimits moving forward for the editor role. Some of the training videos will need re-recording too.

hardyoyo:
replaced by https://github.com/BirkbeckCTP/janeway/pull/3374 ... closing

PR #3345, opened 2023-01-11
Overhaul admin views to support searching and browsing

### LINKED ISSUE #3256: Brush up admin views
Many of the current admin views have incomplete filtering and searching in the list view, and unhelpful widgets in the item view. We can do a pass through all the models and clean some of this up, both on the main repository and the plugins.

joemull:
This overhaul of the admin views addresses navigation issues related to bad search fields, filters that populate thousands of items as choices, and a lack of connections between related models. Closes #3256.

Design goals:
- Display the most relevant fields in list views
- Search support for visible fields
- Always filter by journal or repository, if possible, using custom filters
- Add other filters that narrow down the content, based on likely lookups when providing tech support
- Use inlines on article and preprint models to represent metadata elements
- Use inlines to connect related models where relevant, but not when the inline could ever load hundreds of lines
- Add an admin_utils file to hold custom filter and inline classes
- Add unit tests via a meta-class to run a search query against every admin view, to make sure all prescribed fields are actually searchable


joemull:
This is ready for review, or, if we want to fix the underlying failing tests first, I can rebase it.

mauromsl (approved):
I know this was a lot of work, and I really appreciate the classes under utils/admin_utils.py which make the entire processes so much more DRY.

Well done @joemull

PR #3349, opened 2023-01-12
The manage workflow stage link is now available to editors

ajrbyers:
This view has been protected by `@editor_user_required` for a while so has been available for editors to use if the know the path. Editors want to make use of the archive function that appears on this view so we should make the page available.


ajrbyers:
@mauromsl can you include this in v1.4.4-RC-2, thanks!

mauromsl (approved):

PR #3350, opened 2023-01-12
Add templates missing from material theme

yakky:
The following files are missing from material theme:

- themes/material/templates/elements/journal/article_issue_list.html
- themes/material/templates/elements/journal/social_meta.html
- themes/material/templates/elements/article_meta_tags.html

If one is using `INSTALLATION_BASE_THEME` other than material the theme system is capable of fallback to the files in the base theme, but if `INSTALLATION_BASE_THEME = "material"` the fallback mechanism fails and `TemplateDoesNotExist` exception is raised

### DIFF
--- a/src/themes/material/templates/elements/article_meta_tags.html
+++ b/src/themes/material/templates/elements/article_meta_tags.html
@@ -1,0 +1,59 @@
+
+    <!-- Commented out meta fields are TODO -->
+
+    <!-- <meta name="DC.Coverage.spatial" xml:lang="en" content="London"/> -->
+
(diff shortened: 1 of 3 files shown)


Bbkctp:

ajrbyers:
@mauromsl is going to see if we can move these common files into a commons structure.

mauromsl:
Hi @yakky, @gamboz ,

I've moved these templates to the `templates/common` directory as part of #3370

This is to avoid having identical copies of the same template, as often they get out of sync with each other when they are present in multiple themes.

The templates can still be overriden on your own sub-themes if required with the usual notation

yakky:
@mauromsl thanks, it works for us

PR #3356, opened 2023-01-19
Fixes a bug that stopped user details from showing on the merge users page

ajrbyers:


mauromsl (changes requested):
@ajrbyers do we not need to also traverse any additional pages that may be available through the API?

ajrbyers:
> @ajrbyers do we not need to also traverse any additional pages that may be available through the API?

could do… but 100 records should be sufficient.

mauromsl:
> > @ajrbyers do we not need to also traverse any additional pages that may be available through the API?
> 
> could do… but 100 records should be sufficient.

Ah I see what you mean, because we use it for search, the user will always expand the query to narrow down the results beyond 100 anyways

mauromsl (approved):

PR #3370, opened 2023-01-26
Fixed a bug where some themes had no access to some templates when used as the base theme

mauromsl:
Some templates are identical across themes, such as those listing meta tags for semantic web data mining.
We used to hold a single copy of these in the base theme (OLH) but since now installations can pick any theme as their base, they need to be moved to the `templates/common` directory now.

Closes #3350


ajrbyers (changes requested):
The OLH theme also has some of these files, probably worth dropping them as part of this change?

mauromsl:
> The OLH theme also has some of these files, probably worth dropping them as part of this change?

That's what the PR does

ajrbyers:
> > The OLH theme also has some of these files, probably worth dropping them as part of this change?
> 
> That's what the PR does

🤦‍♂️ I missed that Git had picked it up as a file move so expected to see a deletion.

PR #3374, opened 2023-01-31
Journal manager role

ajrbyers:
Although this is done I'm not marking it as ready for review yet. I can tell @mauromsl will want the migrations condensed. 😄


src/utils/install/journal_defaults_permissions.json
```diff
@@ -0,0 +1,4733 @@
+[
```
mauromsl, line 1: What do we need this file for? seems to be a copy of journal_defaults.json

ajrbyers:
@mauromsl this is now ready for review.

mauromsl (approved):

PR #3377, opened 2023-02-01
Update README.md

ajrbyers:
Updates the readme to have more current information.


ajrbyers:
@mauromsl can we include this in the 1.4.4 pin?

PR #3379, opened 2023-02-01
Fix test_article_encoding_bibtex and test_article_encoding_ris with corrections to the expected results

hardyoyo:


mauromsl (approved):

src/submission/tests.py
```diff
@@ -609,7 +609,7 @@ def test_article_encoding_ris(self):
             PY  - 1990
             TI  - Test article: A RIS export test case
             T2  - {journal_name}
-            UR  - http://localhost/TST/article/id/{article_id}/
+            UR  - http://testserver/article/id/{article_id}/
```
mauromsl, line 612: For this change to work consistently, the test needs to run under domain mode explicitely, otherwise the test outcome can change based on user settings. See this for an example:

https://github.com/BirkbeckCTP/janeway/blob/5511797544a27e30751b156474b08da42f401fb6/src/repository/tests/test_views.py#L58

src/submission/tests.py
```diff
@@ -566,7 +566,7 @@ def test_article_encoding_bibtex(self):
                 title = {Test article: a test article},
                 volume = {2},
                 year = {1990},
-                url = {http://localhost/TST/article/id/%s/},
+                url = {http://testserver/article/id/%s/},
```
mauromsl, line 569: For this change to work consistently, the test needs to run under domain mode explicitely, otherwise the test outcome can change based on user settings. See this for an example:

https://github.com/BirkbeckCTP/janeway/blob/5511797544a27e30751b156474b08da42f401fb6/src/repository/tests/test_views.py#L58

mauromsl (changes requested):
Just one more change needed to ensure idempotency across test runs

mauromsl (approved):

PR #3381, opened 2023-02-02
Refactor tests to eliminate naive datetime warnings

hardyoyo:
The tests currently issue a number of warnings about "naive datetimes" ... warnings are distracting. This pull request eliminates them by using timezone.make_aware and timezone.datetime, to work with built-in methods for Django to generate datetimes with expected timezone information.

### DIFF
--- a/src/api/tests/test_oai.py
+++ b/src/api/tests/test_oai.py
@@ -7,6 +7,7 @@ from urllib.parse import unquote_plus
 from django.test import RequestFactory, TestCase, override_settings
 from django.urls import reverse
 from django.utils.http import urlencode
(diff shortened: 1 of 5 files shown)


ajrbyers:
I've assigned @mauromsl to get his thoughts on this.

mauromsl (approved):

mauromsl:
Thanks @hardyoyo !

At first I was going to suggest not using `make_aware` as it is sensitive to django's settings.py module and could lead to varying results when running the tests across different environments, however I don't think any of these tests we have now would be sensitive to timezone changes. So it is good to merge

PR #3389, opened 2023-02-08
Adds the base structure and migrations to support en-us locale

mauromsl:
- Adds database migration for en-us support on translatable model fields
- Adds .po files for en-us locale
- Adds en-us to the list of supported locales
-


src/core/migrations/0079_auto_20230208_1233.py
```diff
@@ -0,0 +1,45 @@
...
+            model_name='editorialgroup',
+            name='name_en_us',
+            field=models.CharField(max_length=500, null=True),
+        ),
+        migrations.AddField(
+            model_name='settingvalue',
+            name='value_en_us',
+            field=models.TextField(blank=True, null=True),
+        ),
+        migrations.AlterField(
```
ajrbyers, line 40: If nothing is actually being changed here, can we drop this?

ajrbyers (approved):
One comment in line, nothing major but worth considering. If you're happy as is, merge away.

PR #3391, opened 2023-02-08
2498 - Add RSS feed for preprints

### LINKED ISSUE #2498: RSS support should be extended to preprints/repository
As a preprints server administrator, I would like to have an RSS feed to conveniently share the latest published preprints. 

Extend RSS support to preprints/repository.

hardyoyo:
fixes #2498

### DIFF
--- a/src/rss/urls.py
+++ b/src/rss/urls.py
@@ -6,7 +6,8 @@ from django.conf.urls import url
 from rss import views
 
 urlpatterns = [
-    url(r'^', views.LatestArticlesFeed(), name='rss_articles'),
+    # url(r'^', views.LatestArticlesFeed(), name='rss_articles'),
     url(r'^news/$', views.LatestNewsFeed(), name='rss_news'),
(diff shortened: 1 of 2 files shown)


src/rss/urls.py
```diff
@@ -6,7 +6,8 @@
 from rss import views
 
 urlpatterns = [
-    url(r'^', views.LatestArticlesFeed(), name='rss_articles'),
```
ajrbyers, line 9: A slight hiccough with this change: anyone who is subscribed to this URL will stop seeing new content.

ajrbyers (changes requested):
One change suggested.

src/rss/urls.py
ajrbyers: Ah, just reorder them so the LatestArticlesFeed one comes last in the list.

hardyoyo: Ah, it *does* work if I move it to the end of the URLs, I'll change that

mauromsl (approved):

Issue #3402, opened 2023-02-21
publication_title shoudl be used in how to cite when present


ajrbyers:
**Describe the bug**
An article with a set publication_title attrbute will display an incorrect how to cite statement.

**Janeway version**
1.4.4

**Expected behavior**
When publication_title is set, how to cite should use this attribute and fallback to journal name.

PR #3403, opened 2023-02-21
#3402 use publication_title in how to cite if a given article has this field set.

ajrbyers:
Reasoning:

The purpose of this field is to override the default journal name, so this should include the how to cite display.

Note:

Requested in ticket #6382


mauromsl:
Cherry-picked this for 1.4.4 compatibility here: #3406

ajrbyers:
> Cherry-picked this for 1.4.4 compatibility here: #3406

Thief!

PR #3454, opened 2023-03-24
Funder form fixes

### LINKED ISSUE #3195: Hardcoded language in funder metadata
The word 'grant' is hardcoded on the article page, making it hard to control the overall funder statement:

https://github.com/BirkbeckCTP/janeway/blob/eacd5c858863a7400a03194faae125bf05709f27/src/themes/OLH/templates/journal/article.html#L217

https://github.com/BirkbeckCTP/janeway/blob/eacd5c858863a7400a03194faae125bf05709f27/src/themes/material/templates/journal/article.html#L142

Screenshot from 2022-10-11 10-59-22

joemull:
Closes #3195

### DIFF
--- a/src/submission/views.py
+++ b/src/submission/views.py
@@ -136,13 +136,12 @@ def submit_funding(request, article_id):
                 'name': request.POST.get('funder_name', None),
                 'fundref_id': request.POST.get('funder_doi', None),
                 'funding_id': request.POST.get('grant_number', None)
-            }
+            },
+            article=article,
         )
 
         if funder_form.is_valid():
-            funder = funder_form.save()
-            article.funders.add(funder)
-        article.save()
+            funder_form.save()
 
     template = 'admin/submission/submit_funding.html'
     context = {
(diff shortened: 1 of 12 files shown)


src/submission/views.py
```diff
@@ -131,14 +131,17 @@ def submit_funding(request, article_id):
...
+                'fundref_id': request.POST.get('funder_doi', None),
+                'funding_id': request.POST.get('grant_number', None)
+            }
         )
 
-        funder.save()
-        article.funders.add(funder)
+        if funder_form.is_valid():
+            funder = funder_form.save()
+            article.funders.add(funder)
```
ajrbyers, line 144: Could the funder form take an article argument and add this as part of the save process? It would make it easier to reuse the form.

ajrbyers (approved):
I've added a comment but its not essential. If you're happy then go ahead and merge.

src/submission/views.py
joemull: Good catch. Added some commits that move the saving routine inside the form

PR #3459, opened 2023-03-27
#3289 remove ID from section __str__

### LINKED ISSUE #3289: Section IDs should not be visible to submitting authors
Submitting authors are bewildered by the section ID numbers that appear when they are asked to select a section to submit to.

Seems like these ID numbers have no use to an author, so suggest removing them from the UI in the submission workflow to avoid confusion.

**Additional context**
Creating this issue to get Birkbeck feedback, then CDL are happy to work on it.
Screen_Shot_2022-09-01_at_3 34 41_PM

ajrbyers:
- Updates the Section model's __str__ method to remove the PK
- Closes #3289

### DIFF
--- a/src/submission/models.py
+++ b/src/submission/models.py
@@ -1991,7 +1991,9 @@ class Section(AbstractLastModifiedModel):
         ordering = ('sequence',)
 
     def __str__(self):
-        return self.name
+        if self.name:
+            return self.name
+        return f"Unnamed Section {self.pk}"
 
     def published_articles(self):
         return Article.objects.filter(section=self, stage=STAGE_PUBLISHED)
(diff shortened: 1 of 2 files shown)


mauromsl (approved):

src/submission/models.py
```diff
@@ -1991,10 +1991,7 @@ class Meta:
         ordering = ('sequence',)
 
     def __str__(self):
-        return "{} - {}".format(
-            self.pk,
-            self.name,
-        )
+        return self.name
```
mauromsl, line 1994: `Section.name` is nullable, which could lead to this method returning a `NoneType` object. We should return either an empty string, a sentinel value or a string with the section ID in that case

There is also a test case failing with this change.

mauromsl (changes requested):

ajrbyers:
Updates pushed, thanks @mauromsl

mauromsl (approved):

PR #3461, opened 2023-03-27
Add a test for whether /usr/bin/crontab exists, before attempting to run install_cron. Warn if crontab not found.

hardyoyo:
Crontab might be missing from certain environments (especially local developer environments). This PR makes running install_cron optional, and dependent on the existence of /usr/bin/crontab. Warns if the file isn't found.


ajrbyers:
As discussed on Discord shifting the check to the command will cover for people calling it manually.

ajrbyers (approved):

PR #3465, opened 2023-03-29
#1471 pass decisions through to the warning view so it can redirect the user properly

### LINKED ISSUE #1471: Workflow is incorrect when an article is fast accepted
**Describe the bug**
When you press "accept article" without any review process from the first review page (as a user might wish to do to fast-track a piece), you end up at the wrong page. Really we want to take the user straight to copyediting.

**Janeway version**
Master as of March 13

ajrbyers:
- Passes the decision a user makes into the warning page so it can then redirect the user on an override
- Closes #1471


src/review/urls.py
```diff
@@ -111,7 +111,7 @@
     re_path(r'^reviewer/(?P<assignment_id>\d+)/file_download/all/$', views.review_download_all_files,
         name='review_download_all_files'),
 
-    re_path(r'^article/(?P<article_id>\d+)/access_denied/$', views.review_warning, name='review_warning'),
+    re_path(r'^article/(?P<article_id>\d+)/decision/(?P<decision>accept|decline|undecline|review)/access_denied/$', views.review_warning, name='review_warning'),
```
mauromsl, line 114: Can we generate these out of the new Enum constants? You should be able to with something like `"|".join(item.value for item in review.const.ReviewDecisions)`

```

**Review by mauromsl — CHANGES_REQUESTED** [review]

File: src/review/views.py [hunk 2]
```diff
@@ -2052,10 +2052,30 @@ def review_warning(request, article_id):
                 task_object=article,
                 **kwargs
         )
-        return redirect(reverse('review_in_review', kwargs={'article_id': article.pk}))
+
+        if decision == 'review':
```
mauromsl, line 2056: we should use constants here too

src/security/decorators.py
```diff
@@ -77,14 +77,23 @@ def editor_is_not_author(func):
 
     def wrapper(request, *args, **kwargs):
         article_id = kwargs.get('article_id', None)
+        decision = kwargs.get('decision', 'review')
```
mauromsl, line 80: we should use constants here

mauromsl (changes requested):
Just a couple minor suggestions inline

ajrbyers:
Closed in favour of https://github.com/BirkbeckCTP/janeway/pull/3570

PR #3471, opened 2023-03-30
Modify article pages for all themes to support proofing accuracy

### LINKED ISSUE #3207: Articles should look identical in proofing and after publication
I'm coming back to [Typesetting issue 148](https://github.com/BirkbeckCTP/typesetting/issues/148) now that I have a better understanding of what causes an article to look different in proofing and after publication.

**User Story**
As a journal editor, I'm proofing an article, and I want to check how a given piece of metadata will display. For example, the “How to cite” block sometimes has issues, and I want to know whether I'll need to put in a custom how-to-cite. I want to do this in the proofing stage, before I publish the article, because I want the published article to look as good as it can *before* it is public.

However, some elements don't show up on the article page in proofing, and I don't know why. This is a cause of concern about whether I have set the metadata correctly. I go back and double-check, I compare it with other published articles, I spend more time than I should have, and I get a little frustrated. Finally I click publish because what else can I do, and then I immediately see a few small problems, and I have to go and edit things, right as the publication announcement has gone out and people are viewing the article for the first time. This is not ideal--I wish I could view the galley in typesetting and see *exactly* what I'll see on publication. This would help me trust the system and move more confidently and quickly through the task of proofreading.

**Proposed Solution**
We use `Article.is_published` to show or hide certain metadata in `journal/article.html` in all three themes. Example:
https://github.com/BirkbeckCTP/janeway/blob/9e979a496432b202cc53773b2280ca1d49f39617/src/themes/OLH/templates/journal/article.html#L176-L179

We can 
- [ ] Modify the template logic to handle the scenario of proofreading exactly like it handles published articles. This might involve adding to the context in the view
- [ ] Add a unit test to compare the HTML string rendered for proofing with the one rendered for a published article, with the same full set of metadata

**Context**
Visibility of System Status, https://www.nngroup.com/articles/visibility-system-status/
Memory Recognition and Recall in User Interfaces, https://www.nngroup.com/articles/recognition-and-recall/

joemull:
Fixes #3207 
See also https://github.com/BirkbeckCTP/typesetting/issues/148

### DIFF
--- a/src/themes/OLH/templates/journal/article.html
+++ b/src/themes/OLH/templates/journal/article.html
@@ -87,14 +87,16 @@
                 </ul>
             </div>
         </section>
+        <div id="author_biographies">
         {% for author in article.frozen_authors.all %}
             {% if author.biography %}
-                <div class="dropdown-pane author-info proofing-test-include" data-dropdown id="author{{ forloop.counter }}" data-hover="true"
+                <div class="dropdown-pane author-info" data-dropdown id="author{{ forloop.counter }}" data-hover="true"
                      data-hover-pane="true">
                     {{ author.biography }}
                 </div>
             {% endif %}
         {% endfor %}
+        </div>
     {% endif %}
 
         <div data-sticky-container class="row">
@@ -142,7 +144,7 @@
             </div>
         </div>
 
-        <section id="content" class="no-padding proofing-test-include">
+        <section id="content" class="no-padding">
             <div class="row">
                 <div class="large-8 columns border-right">
                     <div class="row">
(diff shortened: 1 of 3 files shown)


src/themes/OLH/templates/journal/article.html
```diff
@@ -89,7 +89,7 @@ <h3>{{ article.title|safe }}</h3>
         </section>
         {% for author in article.frozen_authors.all %}
             {% if author.biography %}
-                <div class="dropdown-pane author-info" data-dropdown id="author{{ forloop.counter }}" data-hover="true"
+                <div class="dropdown-pane author-info proofing-test-include" data-dropdown id="author{{ forloop.counter }}" data-hover="true"
```
mauromsl, line 92: I don't see this CSS class defined anywhere, where is it coming from?

mauromsl (changes requested):
Looks good, but I think there might be some CSS missing as some of the CSS classes used are not defined anywhere

src/themes/OLH/templates/journal/article.html
mauromsl: I just saw https://github.com/BirkbeckCTP/typesetting/pull/185/files and understood what you are trying to do here.

It'd be best to give these nodes an `id` attribute and then have the test define what elements it should see and which it should, rather than bundling this ghost articles into the live article template

(e.g.: give the how to cite block an `id="how-to-cite"`)

joemull: Sure--I've made the change

mauromsl (approved):
Awesome, thanks @joemull 

These IDs will be useful for other feature requests too that required identifiying these blocks so that they can be referenced from XML galleys.

ajrbyers (approved):

PR #3479, opened 2023-04-04
#3404 adds some enhancements to the repository submission file upload

### LINKED ISSUE #3404: Preprints: Step 3 (upload/attach supp files) should support custom text
As a preprints repository moderator or admin, I expect to provide explicit instruction to submitting authors about the requirements for new submissions on **Step 3: Upload Files** of the submission process. Currently preprint repositories may only provide such instruction on Step 1, which is hidden on another page from the submitting author. 

Add to Step 3 of the template the ability to add custom text in places where the author may upload files, such as the submission file as well as supplemental data (if that feature is built in). It would also be nice to be able to use custom wording in the "Repository requires that all author uploads be PDF files" statement.  

Finally, it would be nice to have a link to "Review submission requirements" which opens a popup or modal widow which displays the same requirements text as displayed in Step 1 (submission start text). 

**Mockup**
image

**Additional context**
This has some overlap with https://github.com/BirkbeckCTP/janeway/issues/3227; let me know if it would be preferable to split the two.

ajrbyers:
Closes #3404


mauromsl (approved):

src/repository/migrations/0033_file_upload_texts.py
```diff
@@ -0,0 +1,28 @@
...
+
+from django.conf import settings
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0078_auto_20230222_1317'),
```
joemull, line 11: Missing a dependency migration:

```shell
django.db.migrations.exceptions.NodeNotFoundError: Migration repository.0033_file_upload_texts dependencies reference nonexistent parent node ('core', '0078_auto_20230222_1317')
```

joemull (changes requested):
Wanted to run this to see what it looked like and got a migration dependency error.

src/repository/migrations/0033_file_upload_texts.py
ajrbyers: > Missing a dependency migration:
> 
> ```shell
> django.db.migrations.exceptions.NodeNotFoundError: Migration repository.0033_file_upload_texts dependencies reference nonexistent parent node ('core', '0078_auto_20230222_1317')
> ```

The dependencies for this migration are super weird. It shouldn't require this core migration or the press one. I've removed this. I've also added an example of what it looks like:

image

joemull: Thanks. It is now saying this needs merging with others on 1.5: `CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0033_file_upload_texts, 0035_alter_preprintaccess_options in repository).`

ajrbyers: Egads. Should be okay now.

src/templates/admin/repository/submit/files.html
```diff
@@ -27,12 +27,17 @@ <h2>Information</h2>
                 <div class="content">
                     <p>
                         {% trans 'You can use the form below to select and upload your manuscript file.' %}
+                        <a href="#" data-open="submission_information" aria-controls="submission_information" aria-haspopup="true" tabindex="0">Review submission requirements.</a>
                     </p>
                     {% if request.repository.limit_upload_to_pdf %}
-                        <p>{{ request.repository.name }} {% trans 'requires that all author uploads be PDF files.' %}</p>
+                        <p>{{ request.repository.name }} {{ request.repository.require_pdf_help }}</p>
                     {% else %}
                         <p>{% trans 'You can upload a file in any format though PDF is recommended.' %}</p>
```
joemull, line 35: The new flexibility for other parts of this page is great. I wish this line were removed though, since it could potentially conflict with the custom text when someone doesn't require PDFs but wants to give other instructions. Also I don't think users need to be told any file type is allowed, since the upload will be a smooth process in that case. What if we remove the line altogether?

joemull (commented):
Looks good, just one question about some text that is still hard-coded

src/templates/admin/repository/submit/files.html
ajrbyers: > The new flexibility for other parts of this page is great. I wish this line were removed though, since it could potentially conflict with the custom text when someone doesn't require PDFs but wants to give other instructions. Also I don't think users need to be told any file type is allowed, since the upload will be a smooth process in that case. What if we remove the line altogether?

Agreed. Commit incoming.

joemull (approved):

PR #3483, opened 2023-04-05
allow journal codes up to 40 characters

justingonder:
https://github.com/BirkbeckCTP/janeway/issues/3482


Bbkctp:

PR #3484, opened 2023-04-05
Updating code max_length from 25 to 40

justingonder:
https://github.com/BirkbeckCTP/janeway/issues/3482


Bbkctp:

ajrbyers:
@justingonder thanks but this will also need a database migration in order to have effect.

PR #3490, opened 2023-04-18
Peer review invite

ajrbyers:
- Adds a new view for editors to import large numbers of reviewers from CSV
- When a review round has more than 4 reviewers the reviewer display will collapse into a table
- Adds a new setting to disable review decisions, uses a new default called `DECISION_NO_RECOMMENDATION`, reviewers will now see this as an option in normal reviews
- Added a setting to make Open Review Permission opt-in by default


src/utils/testing/helpers.py
```diff
@@ -483,7 +483,7 @@ def create_editor_assignment(article, editor, **kwargs):
 
 def create_revision_request(article, editor, **kwargs):
     note = kwargs.get('note', 'Test note')
-    decision = kwargs.get('decision', review_models.review_decision()[0][0])
+    decision = kwargs.get('decision', review_models.ALL_REVIEW_DECISION_CHOICES[0][0])
```
mauromsl, line 486: Where is `ALL_REVIEW_DECISION` defined? I think you used a function with lower case naming

src/review/const.py
```diff
@@ -10,3 +10,12 @@ class EditorialDecisions(EnumContains):
...
     MINOR_REVISIONS = 'minor_revisions'
     MAJOR_REVISIONS = 'major_revisions'
+
+
+class ReviewerDecisions(EnumContains):
+    DECISION_ACCEPT = 'accept'
+    DECISION_MINOR = 'minor_revisions'
+    DECISION_MAJOR = 'major_revisions'
+    DECISION_REJECT = 'reject'
+    DECISION_NO_RECOMMENDATION = 'none'
```
mauromsl, line 20: This is just a Q, why do we need a `'none'` value as a choice as opposed to leaving the field empty (`None`)

mauromsl (changes requested):
A few comments inline.

I think we also need to document the new behaviour of these open review settings

PR #3511, opened 2023-05-02
Issue assignment clarity

### LINKED ISSUE #2750: Article and Issue can disagree about being friends
**Describe the bug**
It is possible for an Article to think that it's in an Issue (via Article.primary_issue or Article.projected_issue, both ForeignKey), but for the Issue to disagree (via Issue.articles, a ManyToManyField). See screenshots for an example.

Screenshot from 2022-02-07 14-25-40

Screenshot from 2022-02-07 14-26-17

**Janeway version**
1.4.0

**Expected behavior**
The model properties should be modified to make sure there's just one source of truth about this. Or, we should do lots of testing on the interfaces dealing with issue assignment to find the bug that lets users create this situation.

joemull:
The issue described in #2750 has likely been fixed, but in investigating it, I found a bug in the View Metadata modal that meant users could not change Projected Issue after it was set, because they were sent to the Issue Manager instead of the Projected Issue assignment form.

I changed the link to the correct form, but I did not want to get rid of the link between View Metadata and the Issue Manager, because even if it was the "wrong" link, some people might be using it to manage primary issue assignment before pre-publication. So I added a table for Primary Issue to View Metadata below Projected Issue, and I clarified the text around these forms so that new editors could more easily get up to speed on the difference between Projected and Primary and manage both flexibly.

image

Projected Issue Assignment

Closes #2750

### DIFF
--- a/src/templates/admin/elements/metadata.html
+++ b/src/templates/admin/elements/metadata.html
@@ -248,7 +248,9 @@
             <th>Volume</th>
             <th>Number</th>
             <th>Published</th>
-            <th></th>
+            {% if user_is_editor %}
+                <th></th>
+            {% endif %}
         </tr>
         {% if article.primary_issue %}
         <tr>
@@ -257,12 +259,21 @@
             <td>{{ article.primary_issue.volume }}</td>
             <td>{{ article.primary_issue.issue }}</td>
             <td>{{ article.primary_issue.date }}</td>
-            <td><a href="{% url 'manage_issues_id' article.primary_issue.pk %}">Issue Manager</a></td>
+            {% if user_is_editor %}
+                <td>
+                    <a href="{% url 'manage_issues_id' article.primary_issue.pk %}">
+                        Issue Manager
+                    </a>
+                </td>
+            {% endif %}
         </tr>
         {% else %}
         <tr>
             <td colspan="6">
-                <p>This article does not have a primary issue yet.{% if user_is_editor or request.user.is_staff %} You can assign one now using the <a target='_blank' href="{% url 'manage_issues' %}">issue manager</a>, or wait and assign one during the pre-publication checklist.{% endif %}</p>
+                <p>This article does not have a primary issue yet.</p>
+                {% if user_is_editor %}
+                    <p>You can assign one now using the <a target='_blank' href="{% url 'manage_issues' %}">issue manager</a>, or wait and assign one during the pre-publication checklist.</p>
+                {% endif %}
             </td>
         </tr>
         {% endif %}
--- a/src/templates/admin/elements/metadata.html
+++ b/src/templates/admin/elements/metadata.html
@@ -213,8 +213,8 @@
 <div class="content">
     <table class="scroll">
         <tr>
-            <th>Issue ID</th>
-            <th>Issue Title</th>
+            <th>ID</th>
+            <th>Title</th>
             <th>Volume</th>
             <th>Number</th>
             <th>Published</th>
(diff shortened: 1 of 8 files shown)


src/templates/admin/elements/summary_modal.html
```diff
@@ -16,7 +16,7 @@ <h4><i class="fa fa-file">&nbsp;</i>Article Metadata</h4>
             {% if user_is_editor %}
                 {% if article.ithenticate_id %}
                 <div class="title-area">
-                    <h2>Similarty Check</h2>
```
ajrbyers, line 19: S. M. R. T.

src/templates/admin/review/projected_issue.html
```diff
@@ -28,13 +28,13 @@ <h2>Help</h2>
 				</div>
 				<div class="content">
 					<p><strong>{% trans "What is this for?" %}</strong></p>
-					<p>{% trans "The projected issue setting helps you track which issue an article is intended to be published with."%}</p>
+					<p>{% trans "The projected issue field is for journals that add articles to issues on a rolling basis and publish them individually, rather than publishing them all at once. It helps you track internally which issue an article is intended to be published with, without displaying an unpublished article under a published issue."%}</p>
```
ajrbyers, line 31: It might be worth commenting that non-Issue based journals can use it to inform typesetters of the issue an article is going to be part of?

ajrbyers (changes requested):
I've added some comments but this looks good!

src/templates/admin/elements/metadata.html
```diff
@@ -237,6 +237,38 @@ <h2>Projected Issue</h2>
...
+            <th></th>
+        </tr>
+        {% if article.primary_issue %}
+        <tr>
+            <td>{{ article.primary_issue.id }}</td>
+            <td>{{ article.primary_issue.issue_title }}</td>
+            <td>{{ article.primary_issue.volume }}</td>
+            <td>{{ article.primary_issue.issue }}</td>
+            <td>{{ article.primary_issue.date }}</td>
+            <td><a href="{% url 'manage_issues_id' article.primary_issue.pk %}">Issue Manager</a></td>
```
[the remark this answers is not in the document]
joemull: Good catch

src/templates/admin/elements/metadata.html
```diff
@@ -237,6 +237,38 @@ <h2>Projected Issue</h2>
...
+        </tr>
+        {% else %}
+        <tr>
+            <td colspan="6">
+                <p>This article does not have a primary issue yet.{% if user_is_editor or request.user.is_staff %} You can assign one now using the <a target='_blank' href="{% url 'manage_issues' %}">issue manager</a>, or wait and assign one during the pre-publication checklist.{% endif %}</p>
```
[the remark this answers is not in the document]
joemull: No, we don't. This was already there but I'll remove it

src/templates/admin/review/projected_issue.html
joemull: I'm not sure what you mean by a non-issue-based journal that puts articles into issues. Does rolling publication in chronologically not cover the use case for this field?

joemull: Or are you saying, just mention communication with typesetters?

ajrbyers: Me neither. I _think_ it should be that Issue based journals _can_ also use the field...

src/templates/admin/production/assign_typesetter.html
```diff
@@ -18,11 +18,11 @@
                 {% include "elements/forms/errors.html" with form=form %}
                 {% csrf_token %}
             </div>
-            {% if not article.projected_issue %}
+            {% if not article.primary_issue and not article.projected_issue %}
```
mauromsl, line 21: if you want to be fully comprehensive about this, we should check for `article.issue` instead, since in some cases the article might have been added to an issue without declaring the `primary_issue` explicitly (I know, this part of the codebase has grown hairy..)

mauromsl (changes requested):
One more suggestion to make sure we cover all false positives

src/templates/admin/review/projected_issue.html
joemull: I've reworded it more openly

ajrbyers (approved):

mauromsl (approved):
Thanks @joemull, it all looks much clearer now

PR #3512, opened 2023-05-02
#3240 when a mailgun email is logged as failed or bounced it will now send an email to the actor (if they are staff, editor or repo manger) and log that it was sent

### LINKED ISSUE #3240: Email notifications about bounced emails are not logged
When an email bounces, Janeway tries to tell the sender that by sending them an email notification. However, this email is not logged in the way most emails are:

https://github.com/BirkbeckCTP/janeway/blob/e2baa44ee43ff3847a513f079a194432dd9843cc/src/utils/logic.py#L81-L90

We should log it.

ajrbyers:
- `check_mailgun_stat` now triggers actor message
- `attempt_actor_email` has been renamed `send_bounce_notification_to_event_actor`
- Notification now works for both journal articles and repository preprints
- Bounce notifications are logged
- Closes #3240

### DIFF
--- a/src/submission/views.py
+++ b/src/submission/views.py
@@ -383,7 +383,6 @@ def delete_funder(request, article_id, funder_id):
     return redirect(reverse('submit_funding', kwargs={'article_id': article_id}))
 
 
-
 @login_required
 @article_edit_user_required
 def delete_author(request, article_id, author_id):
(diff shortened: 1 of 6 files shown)


ajrbyers:
Marking this as a draft, I want to add a test for the two routes.

mauromsl (approved):

PR #3515, opened 2023-05-03
Fixes a number of issues with Janeway version missmatches

### LINKED ISSUE #3462: Plugins cannot be updated when Janeway is updated to a new version of Django
**Describe the bug**
When plugin upgrades introduce a breaking change that also depends on a change to a new version of django, it can be very difficult to upgrade them

**Janeway version**
v1.4.4

**To Reproduce**
If you upgrade your Janeway installation so that it uses a new version of Django introducing breaking changes, it is likely that you will need to also update your plugins. At that point Janeaway will be set to a new version x.y.z and the updated plugin will pin that version as their JANEWAY_VERSION. When running the update.sh script with these two updated codebases, Janeway will fail to load any management commands because the version the plugin requires (x.y.z) has not yet been installed. If you update Janeway's codebase first but not the plugin, errors might be raised due to the plugin not being updated

**Expected behavior**
The new Janeway version against which the plugin version is checked, needs to be checked against the codebase, rather than against the value stored in the database.

mauromsl:
Due to how Janeway versioning currently works, it is impossible to upgrade Janeway and plugins at the same time (see #3462 for context)

The version should live statically in the codebase rather than just in the database. This PR does still preserve those models so that version changes can still be tracked.

It also makes use of the semver library rather than parsing version numbers ad-hoc during startup checks

closes #3462

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -48,6 +48,7 @@ python-dateutil==2.8.1
 python-docx==0.8.6
 python-magic==0.4.13
 requests==2.25.1
+semver==3.0.0
 six==1.14.0
 sqlparse==0.2.3
 swapper==1.3.0
--- a/src/core/model_utils.py
+++ b/src/core/model_utils.py
@@ -513,14 +513,23 @@ class BaseSearchManagerMixin(Manager):
     def get_search_lookups(self):
         return self.search_lookups
 
-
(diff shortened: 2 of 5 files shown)


ajrbyers (changes requested):
Looks good. One comment inline.

PR #3523, opened 2023-05-04
Fix bug with section display on submissions page

### LINKED ISSUE #3481: Alignment problem with Sections table on journal Submissions page in headless mode
**Describe the bug**
The table near the bottom of the Submissions page, which indicates whether various sections of the journal are public submissions, peer reviewed, or indexed, has an alignment problem. It appears this alignment problem happens when the journal is in headless mode, but not when the journal uses Janeway's front-end.

**Janeway version**
1.4.3; also reproducible in latest 1.5 RC

**To Reproduce**
See 'Sections' portion of this page, using Janeway's front-end (table is in proper alignment):
https://www.glossa-journal.org/submissions/

See 'Sections' portion of these two pages, in headless mode. The Public Submissions and Peer Reviewed column headers are shoved to the left. This is the case regardless of which theme I choose:
~~https://journalpub.escholarship.org/aicrj/submissions/~~ (temporarily removed block from display)
https://journalpub.escholarship.org/glossapsycholinguistics/submissions/

example in 1.5: https://journalpub-dev.escholarship.org/docutest/submissions/

**Screenshots**

joemull:
Fixes #3481

### DIFF
--- a/src/themes/OLH/templates/elements/section_block.html
+++ b/src/themes/OLH/templates/elements/section_block.html
@@ -1,36 +1,54 @@
 <h3>{% trans 'Sections' %}</h3>
-<div class="row">
-    <div class="medium-6 columns">
-
-    </div>
-    <div class="medium-2 columns">
-        <p>{% trans "Public Submissions" %}</p>
-    </div>
-    <div class="medium-2 columns">
-        <p>{% trans "Peer Reviewed" %}</p>
-    </div>
-    <div class="medium-2 columns">
-        <p>{% trans "Indexed" %}</p>
-    </div>
-</div>
-<hr/>
-{% for section in sections %}
-    <div class="row">
-        <div class="medium-6 columns">
-            <p>{{ section.name }}</p>
-        </div>
-        <div class="medium-2 columns">
-            <strong>{% if section.public_submissions %}<i class="fa fa-check"></i>{% else %}
-                <i class="fa fa-times"></i>{% endif %}</strong>
-        </div>
-        <div class="medium-2 columns">
-            <strong>{% if section.number_of_reviewers > 0 %}<i class="fa fa-check"></i>{% else %}
-                <i class="fa fa-times"></i>{% endif %}</strong>
-        </div>
-        <div class="medium-2 columns">
-            <strong>{% if section.indexing %}<i class="fa fa-check"></i>{% else %}<i class="fa fa-times"></i>{% endif %}
-            </strong>
-        </div>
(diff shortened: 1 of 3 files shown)


ajrbyers (changes requested):
Two things that jump out at me:

1. Justin asked about headless Janeway so we should include the change in the admin submission template (https://github.com/BirkbeckCTP/janeway/blob/master/src/templates/admin/journal/submissions.html)
2. Other tha the class of the table these are pretty much identical, could the content of the table be a single included template to make maintanance easier?

joemull:
> Two things that jump out at me:
> 
>     1. Justin asked about headless Janeway so we should include the change in the admin submission template (https://github.com/BirkbeckCTP/janeway/blob/master/src/templates/admin/journal/submissions.html)
> 
>     2. Other than the class of the table these are pretty much identical, could the content of the table be a single included template to make maintanance easier?

1. Thanks, and that was my first thought too, but if I'm not mistaken, `elements/section_block` does not exist in the admin templates. The OLH theme is serving as a fallback in this case, right?
2. Materialize uses material icons, and the clean theme already had its own table-based block that had a few other differences, so my inclination is to keep them separate. OK with you?

ajrbyers:
>     1. Thanks, and that was my first thought too, but if I'm not mistaken, `elements/section_block` does not exist in the admin templates. The OLH theme is serving as a fallback in this case, right?
> 
>     2. Materialize uses material icons, and the clean theme already had its own table-based block that had a few other differences, so my inclination is to keep them separate. OK with you?

For 1. it would be good to add this so that admin has its own explicit template. If someone changed the base theme to, say, material, then it would break the headless version.

For 2. If you prefer to leave them individual thats fine alternatively you could pass in a variable to use Font Awesome or Material Icons.

mauromsl (approved):

mauromsl:
> > ```
> > 1. Thanks, and that was my first thought too, but if I'm not mistaken, `elements/section_block` does not exist in the admin templates. The OLH theme is serving as a fallback in this case, right?
> > 
> > 2. Materialize uses material icons, and the clean theme already had its own table-based block that had a few other differences, so my inclination is to keep them separate. OK with you?
> > ```
> 
> For 2. If you prefer to leave them individual thats fine alternatively you could pass in a variable to use Font Awesome or Material Icons.

I'm not a big fan of having theme specific execution logic on common templates. if anything we can use the templating system instead so that each theme can provide their own icon.

ajrbyers (approved):

PR #3526, opened 2023-05-05
Auto-generated table of contents for articles (TOCs) can now have HTML in them.

### LINKED ISSUE #3521: Footnote callouts displaying in "Jump To" article TOC
**Describe the bug**
If there happens to be a footnote callout on a section heading (not terribly common, but certainly not unheard of) then the number of the footnote will display as plain text in the "Jump To" list of section links. If you're reading the "Jump To" list, and you haven't checked the section heading itself for comparison, it _looks_ like a seemingly random number has been appended to the end of the section title.

**Janeway version**
1.4.3

**To Reproduce**
1. Go to https://journals.publishing.umich.edu/ergo/plugins/typesetting/preview_galley/article/1141/galley/1363/
2. Click on "[5.2. Ainslie42](https://journals.publishing.umich.edu/ergo/plugins/typesetting/preview_galley/article/1141/galley/1363/#heading7)"
3. Notice that the link to footnote 42 is (correctly) displayed as a superscript in the section heading, but in the "Jump To" list, the title of the section runs directly into the footnote number.

**Expected behavior**
Ideally, Janeway should be able to recognize when a footnote has been tagged within a section `<title>`, and _hide_ that portion of the `<title>` text when populating the "Jump To" list

**Screenshots**

ajrbyers:
Closes #3521

### DIFF
--- a/src/themes/OLH/assets/js/table_of_contents.js
+++ b/src/themes/OLH/assets/js/table_of_contents.js
@@ -12,6 +12,7 @@ $( document ).ready(function() {
     $("#main_article h2, #main_article h3").each(function () {
 
         link = $(this);
+        link.find('a').contents().unwrap();
         title = link.html();
 
         if (!link.attr("id")) {
--- a/src/themes/OLH/assets/js/table_of_contents.js
+++ b/src/themes/OLH/assets/js/table_of_contents.js
@@ -12,7 +12,7 @@ $( document ).ready(function() {
     $("#main_article h2, #main_article h3").each(function () {
 
         link = $(this);
-        title = link.text();
+        title = link.html();
 
         if (!link.attr("id")) {
             link.attr('id', 'heading' + iter.toString());
--- a/src/themes/material/assets/toc.js
+++ b/src/themes/material/assets/toc.js
@@ -12,6 +12,7 @@ $( document ).ready(function() {
     $("#main_article h2").each(function () {
 
         link = $(this);
+        link.find('a').contents().unwrap();
         title = link.html();
 
         if (!link.attr("id")) {
--- a/src/themes/material/assets/toc.js
+++ b/src/themes/material/assets/toc.js
@@ -12,7 +12,7 @@ $( document ).ready(function() {
     $("#main_article h2").each(function () {
 
         link = $(this);
-        title = link.text();
+        title = link.html();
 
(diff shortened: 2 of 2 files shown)


src/themes/OLH/assets/js/table_of_contents.js
```diff
@@ -12,7 +12,8 @@ $( document ).ready(function() {
     $("#main_article h2, #main_article h3").each(function () {
 
         link = $(this);
-        title = link.text();
+        link.find('a').contents().unwrap();
+        title = link.html();
```
mauromsl, line 16: Does this solve the problem? If I'm reading this correctly and following the example from Patrick, we are unwrapping (destroying) the anchor tag, but leaving in the <sup> tag inside the HTML?

I don't see any code removing the <sup> tag, so I think it will still be rendered, (though as an actual <sup> rather than as text)

mauromsl (changes requested):
More of a question than a request change at this point

src/themes/OLH/assets/js/table_of_contents.js
ajrbyers: The thing that was breaking the TOC was the <a> tag inside the <sup>.

At this point we have conflicting requests. Glossa want the title in the TOC represented exactly as it is in the text that includes the <sup> so I chose displaying the <sup> properly.

mauromsl: Where is the request from Glossa?

ajrbyers: > Where is the request from Glossa?

Via email. Personally I think the TOC should display the exact title even if it includes a `<sup>`.

mauromsl: > > Where is the request from Glossa?
> 
> Via email. Personally I think the TOC should display the exact title even if it includes a `<sup>`.

I think it is worth running it past both folks again. If we close this as resolved we won't actually resolve the original request

joemull: I agree a footnote number might look a little odd in the TOC but I'm OK with removing that as a separate request. The light touch on this PR is OK for now

joemull (approved):

mauromsl (approved):

Issue #3380, opened 2023-02-02
Improve documentation on subthemes


joemull:
We are working to improve the OLH website, including writing a subtheme for the press level, which will be tracked in its own repository. Project board here: https://github.com/orgs/BirkbeckCTP/projects/28

As part of that, we have discovered a few things about the subthemes that we can document for other people writing them.

joemull:
With the OLH site build in the rear-view mirror and conversations in June 2024 about Janeway's theme structure, I think the current documentation on subthemes is ok for now, so I am going to close this issue.

PR #3531, opened 2023-05-05
Press managers can customise some elements of journal footers

### LINKED ISSUE #3295: Editable footer at publisher level
**problem**:

Certain policies are too prominent when they needn’t be included in the navigation bar for each journal, e.g. ‘Ethics and Malpractice Statement’, but should still really feature on journal sites somewhere; conversely, publisher level navigation items that are really needed on the main navigation bar are also at risk of being removed by journal editors editing the nav bar (e.g. Research Integrity).

**solution**:

The ability for the publisher to configure the footer, so that aPublisher Policies pages can be created. These policies would be accessible but no longer unnecessarily prominent. Nav bar links to these pages should still be possible to make and include as needed. 

If there is a means of making such nav bar items permanent (non-removable) at publisher level then this would be appreciated but less important than being able to create nav bar links to policy pages. 

**extra context**:

For an example, see: see https://www.igi-global.com/).

The pages we currently would like for this to happen to are the ‘Publication Ethics and Malpractice Statement’ and ‘Research Integrity’. It has been discussed that this might be possible to implement as soon as Janeway v.1.5.1. We would like to be able to roll this out to our journals alongside the announcement for Janeway 1.5.1.

joemull:
## Feature overview
This work lets press managers customise journal footers in a few ways:

- you can add navigation links to journal footers that go to press-level pages (e.g. "Publisher Policies")
- you can add custom text to journal footers (e.g. a postal address)

footer-olh

footer-material

footer-clean

These elements are managed at the press level, so journal editors cannot remove or override them.

Links are managed in the press-level **Content Manager**:

cms-footer-links

Journal footer text is managed under the **Press Manager** with the field **Journal footer text**

This feature is due to come out in Janeway 1.5.1.

## Possible future work

Journal editors still won't be able to customise their footers in this way. I think it's open for discussion whether we'd still want to add that functionality, as written up in https://github.com/BirkbeckCTP/janeway/issues/3393. It may no longer be necessary.

## Technical notes

Closes #3295 

I've implemented this with as little re-engineering as possible, since we plan to rework the press-level settings framework with https://github.com/BirkbeckCTP/janeway/issues/3528, and we had limited time to roll out this feature.

I've built this branch on a few commits from #3380 that covered related work previously done for https://github.com/BirkbeckCTP/path/issues/4. Hopefully including these in this branch will make merging #3380 easier later.


src/cms/forms.py
```diff
@@ -46,6 +55,13 @@ def __init__(self, *args, **kwargs):
 
         self.fields['top_level_nav'].queryset = top_level_nav_items
 
+        if request.journal:
```
ajrbyers, line 58: Consider using self.fields.pop as we do in `core.forms`.
https://github.com/BirkbeckCTP/janeway/blob/d2b5ea7d31fe7d42c43795d208352c4e7c27f60d/src/core/forms.py#L139

ajrbyers (approved):
I'm marking this as approved as its generally really good. I've added one suggestion re popping `for_footer` and `extend_to_journals` from the field list when they aren't being used rather than making them `HiddenInput` fields.

pgoussy:
> 

This sounds good! I can certainly imagine scenarios where we would want the ability to customize these footers on a journal-by-journal basis, but I don't know if there are any specific journals requesting anything like that at the moment...

mauromsl (approved):

PR #3536, opened 2023-05-09
#3491 editors will no longer see the Review Form ID field and the slug field has been removed entirely

### LINKED ISSUE #3491: 2 UX tweaks for review forms (hide both slug and form id in UI)
1. When creating a new review form, Editors are asked to provide a 'slug' (required) and are confused because they are unfamiliar with that term and don't understand its purpose. Further, I can't actually see anywhere that the Slug is used in the UI after entering it, though it is referenced a few times in the code.

2. Similar to #3289 - the review form's ID number appears in the UI when an editor is assigning a reviewer and selecting a form for the reviewer to complete. The ID number is irrelevant to the user and causes confusion.

Ideally, the slug and ID number for review forms would be made mostly invisible to the user in order to reduce confusion.

If the slug has a purpose I've not caught, it could at least be given an easier to understand label, or some help text.

ajrbyers:
- Closes #3491


src/review/models.py
```diff
@@ -306,16 +307,27 @@ class ReviewForm(models.Model):
...
+    )
+    thanks = models.TextField(
+        help_text="Message displayed after the reviewer is finished.",
+    )
 
     elements = models.ManyToManyField('ReviewFormElement')
     deleted = models.BooleanField(default=False)
 
+    def save(self, *args, **kwargs):
+        self.slug = slugify(self.name)
```
mauromsl, line 326: This would make the `slug` change when the name changes, do you think it might be a problem? (e.g. if we use it for URLs that might then 404)

I had a quick look and I don't think we use the slug at all for review forms. If that is the case, I'd advocate for removing it entirely

mauromsl (commented):
Just a question inline

src/review/models.py
ajrbyers: I considered that. The only time we use this is to identify if there is or isn’t a default-form so it’s worth leaving it in.

joemull: What about:
```py
def save(self, *args, **kwargs):
    if not self.slug:
        self.slug = slugify(self.name)
```

mauromsl: > 

Where do we do this?

ajrbyers: https://github.com/search?q=repo%3ABirkbeckCTP%2Fimports%20default-form&type=code

joemull: It looks like it's checked here:
https://github.com/BirkbeckCTP/janeway/blob/c4a0ada031820fdd0a4a9904d22bb4fed1773445/src/journal/models.py#L1254-L1261

joemull: Why do you think it should update @ajrbyers? I'm not familiar with what you have in mind, because I don't know what the slug is for in this case. Why do we have it?

I'm also in favor of removing it, if that makes things simpler.

ajrbyers: The slug field is intended to be a slug representation of the name field, if you change the name the slug should also change. As I said, we can remove it, but that's not what this PR is about and would require work on other repositories.

joemull (approved):

mauromsl (approved):

PR #3538, opened 2023-05-09
3427 automated reminder help text has been update to make the function of the Review (Accepted) review type clearer

### LINKED ISSUE #3427: Reminders that seem correctly configured are not sent
**Describe the bug**
Editors sometimes configure review assignment reminders with a combination of fields that does not result in a reminder being sent.

For example, setting a reminder for X days *before* an assignment is accepted won't work because there is no acceptance due date, but it's represented with this Info text:

"Sent 14 days before Review (Accepted) is due."

**Janeway version**
1.4.3

**Expected behavior**
A couple of options:
1) The help text should give instructions about which combination of fields works
2) Better yet, we make this a confirmable form and warn about combinations that don't work

ajrbyers:
Closes #3427

### DIFF
--- a/src/templates/admin/cron/reminders.html
+++ b/src/templates/admin/cron/reminders.html
@@ -36,7 +36,7 @@
                         {% for reminder in reminders %}
                             <tr>
                                 <td>{{ reminder.pk }}</td>
-                                <td>{{ reminder.display_text }}</td>
+                                <td>Sent {{ reminder.days }} days {{ reminder.run_type }} {{ reminder.get_type_display }} is due.</td>
                                 <td>{{ reminder.template_name }} <a href="{% url 'cron_create_template' reminder.pk reminder.template_name %}"><i class="fa fa-edit">&nbsp;</i></a></td>
                                 <td>{{ reminder.subject }}</td>
                                 <td>
--- a/src/templates/admin/cron/reminders.html
+++ b/src/templates/admin/cron/reminders.html
@@ -36,7 +36,7 @@
                         {% for reminder in reminders %}
                             <tr>
                                 <td>{{ reminder.pk }}</td>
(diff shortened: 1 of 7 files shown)


mauromsl (approved):

src/templates/admin/cron/reminders.html
```diff
@@ -43,8 +36,7 @@ <h2>Active Reminders</h2>
                         {% for reminder in reminders %}
                             <tr>
                                 <td>{{ reminder.pk }}</td>
-                                <td>Sent {{ reminder.days }} days {{ reminder.run_type }} {{ reminder.get_type_display }} is
-                                    due.</td>
+                                <td>Sent {{ reminder.days }} days {{ reminder.run_type }} {{ reminder.get_type_display }} is due.</td>
```
joemull, line 39: I have a request about the way the choices and this sentence are formed, since I think it's key to the clarity problem.

I added some detail here:
https://github.com/BirkbeckCTP/janeway/issues/3427#issuecomment-1554220703

What if we make this line like this:
```html
<td>Sent {{ reminder.days }} days {{ reminder.run_type }} for {{ reminder.get_type_display }}.</td>
```

And change the choices to:
```py
RUN_TYPE_CHOICES = (
    # Before the event
    ('before', 'Before the due date'),
    # After the event
    ('after', 'After the due date'),
)
```

joemull (changes requested):
Love the new help text. Just one further change requested based on the original reporter's problem.

src/templates/admin/cron/reminders.html
ajrbyers: I think this would be a good change! 👍🏻

joemull (approved):

PR #3541, opened 2023-05-10
Peer review invite 

ajrbyers:


src/review/views.py
```diff
@@ -2681,6 +2681,10 @@ def upload_reviewers_from_csv(request, article_id):
                     messages.SUCCESS,
                     '{} Review assignments saved.'.format(len(reviewers)),
                 )
+
+            article.stage = submission_models.STAGE_UNDER_REVIEW
+            article.save()
+
```
mauromsl, line 2687: Should this still happen even if there is an import error above?

mauromsl (commented):
A question inline

mauromsl (approved):

joemull (approved):

PR #3543, opened 2023-05-12
Ensure article titles are marked as safe consistently

### LINKED ISSUE #971: Article Title
**Describe the bug**
Sometimes article titles with HTML aren't marked safe

**Janeway version**
1.3.3

**Expected behavior**
{{ article.title }} should always be safe

**Additional Context**
Alternatively: add a safe_title property that returns the title with `mark_safe` and then update templates in core and in plugins.

mauromsl:
closes #971 
Adds a new property to preprints/articles (`safe_title`).

Replaced all explicit `title|safe`
replaced all missing safe titles with `.safe_title`

### DIFF
--- a/src/core/migrations/0082_mark_safe_article_title.py
+++ b/src/core/migrations/0082_mark_safe_article_title.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+import re
+from django.conf import settings
+from django.db import migrations
+
+REGEX = re.compile("{{\ ?([a-z\._-]*)article.title\ ?}}")
+OUTPUT_FMT = "{{ %sarticle.safe_title }}"
+
+def replace_matches( match):
+    prefix = match.group(1)
+    return OUTPUT_FMT % prefix
+
+def replace_article_title(apps, schema_editor):
+    SettingValue = apps.get_model('core', 'SettingValue')
+    setting_values= SettingValue.objects.all()
+    for setting in setting_values:
+        for language in settings.LANGUAGES:
+            value_string = 'value_{}'.format(language[0])
+            if setting and getattr(setting, value_string, None):
+                setattr(
+                    setting, value_string,
+                    re.sub(REGEX, replace_matches, getattr(setting, value_string))
+                )
+                setting.save()
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0081_alter_account_preferred_timezone'),
+    ]
+
+    operations = [
+        migrations.RunPython(replace_article_title, reverse_code=migrations.RunPython.noop),
+    ]
--- a/src/repository/models.py
+++ b/src/repository/models.py
@@ -11,6 +11,7 @@ from django.db import models
 from django.db.models import Q
 from django.utils import timezone
 from django.conf import settings
+from django.utils.html import mark_safe
 from django.utils.translation import gettext_lazy as _
 from django.dispatch import receiver
 from django.shortcuts import reverse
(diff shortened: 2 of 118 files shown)


src/core/migrations/0082_mark_safe_article_title.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
+import re
+from django.conf import settings
+from django.db import migrations
+
+REGEX = re.compile("{{\ ?([a-z\._-]*)article.title\ ?}}")
```
ajrbyers, line 6: This REGEX would miss `{{ author_review.assignment.article.title|safe }}` in the `author_copyedit_deleted` email.

ajrbyers (changes requested):
This is great @mauromsl. I've added a comment about the regex perhaps missing matches where `article.title` is already marked as safe.

src/core/migrations/0082_mark_safe_article_title.py
mauromsl: > This REGEX would miss `{{ author_review.assignment.article.title|safe }}` in the `author_copyedit_deleted` email.

I thought of this, but since `**.article.title|safe` is already a valid value, I thought it'd be best to not overly-complicate the regex. I can extend it to catch and  remove the `|safe` though

ajrbyers: Nah it’s fine. You’re right!

src/repository/models.py
```diff
@@ -506,6 +507,13 @@ def authors(self):
 
         return [pa.account for pa in preprint_authors]
 
+    @property
+    def safe_title(self):
+        if self.title:
+            return mark_safe(self.title)
+        else:
+            return "<Untitled>"
```
joemull, line 515: I have a question about the use of angle brackets here: `<Untitled>`. 

Based on past experience in academia, I would expect a parenthetical like this to use square brackets like `[Untitled]` instead. That's what most of our academics would expect, to the best of my knowledge, though perhaps angle brackets would be used if you were writing in French or German or another language--I'm not sure.

I would also prefer that `<Untitled>` didn't look like an XML element in downstream data processing, where a parser would be looking for HTML or XML-like syntax. I know you haven't used `safe_title` in templates that generate XML, and Django won't think `<Untitled>` is HTML, because it's not marked safe, but is it still worth considering search crawlers, archival scrapers, and screen readers?

Let me know if angle brackets have another purpose I'm not thinking of.

src/utils/install/journal_defaults.json
```diff
@@ -4645,7 +4645,7 @@
             "type": "rich-text"
         },
         "value": {
-            "default": "<p>Dear {{ author_review.assignment.article.correspondence_author.full_name }},</p><p>The copyediting review request for \"{{ author_review.assignment.article.title|safe }}\" that we sent you has been deleted.</p><p>Regards,</p>{{ request.user.signature|safe }}"
+            "default": "<p>Dear {{ author_review.assignment.article.correspondence_author.full_name }},</p><p>The copyediting review request for \"{{ author_review.assignment.article.safe_title|safe }}\" that we sent you has been deleted.</p><p>Regards,</p>{{ request.user.signature|safe }}"
```
joemull, line 4648: This one has a `safe` dangling at the end still

joemull (changes requested):
This looks good to me, overall. I had one question and I also flagged one tiny change.

ajrbyers (approved):
I don't have a particular opinion on `<>` vs `[]` but once Joe's queries are resolved this is good to go for me.

ajrbyers:
I did ask Martin about his thoughts on this and he agreed that [Untitled] would be better.

PR #3544, opened 2023-05-12
Editors and Section Editors can now access the account API endpoint.

ajrbyers:


mauromsl (changes requested):
A  question inline that might require a change

ajrbyers:
@joemull both Mauro and I have commits on this PR so can you give it a quick once over?

mauromsl (approved):

PR #3578, opened 2023-05-24
Revert unwanted nullable file field on Galley

### LINKED ISSUE #3577: Galleys without a File make it impossible to access article archive or article workflow pages
**Describe the bug**
v1.5.0 introduced a change where the file associated with a `Galley` can now be null. In the past, editors would often delete a galley just by deleting the linked file from document manager. Since the file has been made nullable and no longer CASCADEs on deletion, editors lock themselves out of an article when they do this.

**Janeway version**
v1.5.0

**To Reproduce**
1. Add a Galley to an article
2. Delete the file from document manager
3. Visit a workflow page that lists galleys or the article page itself

**Expected behavior**
The behaviour of deleting a galley when its file is deleted from document manager should be reinstated

**Additional context**
An easy solution to this problem is to reinstate the on_delete=CASCADE parameter on the relationship between Galley and File, Since currently it doesn't make sense to have galleys without files associated to them

mauromsl:
When including the `on_delete` statements for Django 3.2 compatibility, we shifted the behaviour of Galley.file making it nullable and not ON DELETE CASCADE. This has unraveled a few issues on installations running  v1.5.0, since galleys without a linked file were never expected.

closes #3577


ajrbyers (approved):
Had a quick look at deleting a bunch of files to ensure there weren't any further cascading issues and it looks okay.

joemull (approved):

PR #3591, opened 2023-05-25
#2444 added a new setting that controls the appearance of a list of reviewers that have previously completed a review in a past round for the current article.

### LINKED ISSUE #2444: populate "add reviewer" with reviewers from previous rounds
It's very common for us to send out re-review requests to the previous reviewers when authors submit a revision.

It would be great if the "Add Reviewer" list would auto-populate with the reviewers who'd completed reviews in the previous round. Ideal would be that they appear in the list and we'd have the chance to edit the particulars (deadline, email contents etc.), and that we'd still have the chance to not send to a particular reviewer or also select additional reviewer(s) if desired.

ajrbyers:
- Moved the query in `get_reviewer_candidates` into its own function `get_reviewers` so that it can be reused.
- Added a new setting, `display_past_reviewers`. that determines whether a table is displayed with reviewers who have previously completed a review for the current article
- Reviewers only appear once, either in the first or second table
- No other elements in the review assignment workflow have changed
- Closes #2444 

image


src/review/views.py
```diff
@@ -1050,6 +1050,15 @@ def add_review_assignment(request, article_id):
     """
     article = get_object_or_404(submission_models.Article, pk=article_id)
     reviewers = logic.get_reviewer_candidates(article, request.user)
+
+    # if setting enabled, fetch reviewers who have completed a review
+    # in a past review round.
+    past_reviewers = []
```
mauromsl, line 1056: Should we pass `past_reviewers` to get_reviewer_candidates either to get them highlighted (e.g with an annotation) or at least filtered out from the second table?

mauromsl (changes requested):
Looks okay, just a general comment about the two tables. It'd be best to only have one table (with the past reviewers flagged up). If this is a bit out of scope for this PR at least have them removed from the second table?

PR #3607, opened 2023-06-16
Fix "make install" error on pristine clone

### LINKED ISSUE #2775: `.gitignore` in `db/postgres-data` prevents `initdb` from initializing postgres cluster in docker service `janeway-postgres`
If I run `make install` on a pristine clone of janeway, I get an error like the following: `nc: bad address 'janeway-postgres'`.

I think this happens when `docker-compose` tries to run the `janeway-postgres` service. The service bind-mounts the folder `db/postgres-data` into `/var/lib/postgresql/data`, however, the folder `db/postgres-data` contains one `.gitignore` file. This file prevents postgres' `initdb` script from running, the service does not start and the service `start_dependencies` continues to unsuccessfully waiting for the `janeway-postgres` and emits the error.

**To Reproduce**
To see the error from `initdb`, one could do:
```
git clone git@github.com:gamboz/janeway.git
cd janeway
cat <<EOX > .env
DB_NAME=janeway
DB_USER=janeway
DB_PORT=5432
DB_PASSWORD=pass
DB_VENDOR=postgres
JANEWAY_PORT=8000
PGADMIN_PORT=8001
EOX
docker-compose run janeway-postgres
```

You should see:
> ...
> initdb: directory "/var/lib/postgresql/data" exists but is not empty
> It contains a dot-prefixed/invisible file, perhaps due to it being a mount point.
> ...

I think that you could add a suggestion to remove the `.gitignore` file in the [documentation](https://janeway.readthedocs.io/en/latest/installation.html#running-janeway-with-docker-and-docker-compose), but I also see that you want to [separate Janeway's filesystem from its codebase](https://github.com/BirkbeckCTP/janeway/projects/15), so I'm not sure what to suggest 🙂

camilstaps:
This is my first PR, so I apologize if I'm doing something wrong here.

- The change to docker-compose.yml gives postgres a different data directory from the mount path, which is non-empty (.gitignore). This is what is suggested in https://serverfault.com/questions/1018377/postgres-mount-volume-error-in-k8s. Closes #2775.
- The change to Makefile makes sure that after deleting `db/*` the contents of this directory is reinitialized from the git index; otherwise you are left with a changeset that you don't want to commit. This is not really related to the previous point, but I thought it was too small to make a separate issue and PR. If you prefer I can remove this commit of course.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,6 +112,7 @@ db-load-backup: #Loads a previosuly captured backup in the db directory (e.g.: m
 	@sudo mv /tmp/$(DB_VOLUME) db/
 uninstall:	## Removes all janeway related docker containers, docker images and database volumes
 	@bash -c "rm -rf db/*"
+	@bash -c "git checkout db"
 	@bash -c "docker rm -f `docker ps --filter 'name=janeway*' -aq` >/dev/null 2>&1 | true"
 	@bash -c "docker rmi `docker images -q janeway*` >/dev/null 2>&1 | true"
 	@echo " Janeway has been uninstalled"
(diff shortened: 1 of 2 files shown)


Bbkctp:

PR #3609, opened 2023-06-21
Let HTML galley prettify be optional

gamboz:
The current behavior of `remove_css_from_html` is to "prettify" the source of HTML galleys, but this introduces spurious spaces in constructs such as the following:
```html
see [Kato, <a href="#">2014</a>] for...
```
⇨
```html
see [Kato, 
  <a href="#">2014</a>
] for...
```
(with an extra space between "2014" and the following "]")

The changes proposed here add an optional argument to `remove_css_from_html` (and to `save_galley` that invokes it) so that one can decide to avoid "prettification" when saving a galley.

The default value of the argument ensures that the current behavior does not change.

### DIFF
--- a/src/production/logic.py
+++ b/src/production/logic.py
@@ -92,7 +92,7 @@ def get_galley_label_and_type(file):
     return label, type_
 
 
-def save_galley(article, request, uploaded_file, is_galley, label=None, save_to_disk=True, public=True):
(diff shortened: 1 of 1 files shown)


ajrbyers (approved):

mauromsl:
closed by #4044 4044

PR #3612, opened 2023-06-27
Set NULL Account names (first, middle, last) to the empty string

gamboz:
Proof of concept :slightly_smiling_face: 

This PR was born to fix a problem in the crossref deposit XML where people without a given name ends as "Mr. None...", but might be a valid solution also elsewhere and follows django [recommendations](https://docs.djangoproject.com/en/4.2/ref/models/fields/#null)


src/core/models.py
```diff
@@ -208,9 +208,9 @@ class Account(AbstractBaseUser, PermissionsMixin):
     email = PGCaseInsensitiveEmailField(unique=True, verbose_name=_('Email'))
     username = models.CharField(max_length=254, unique=True, verbose_name=_('Username'))
 
-    first_name = models.CharField(max_length=300, null=True, blank=False, verbose_name=_('First name'))
-    middle_name = models.CharField(max_length=300, null=True, blank=True, verbose_name=_('Middle name'))
-    last_name = models.CharField(max_length=300, null=True, blank=False, verbose_name=_('Last name'))
+    first_name = models.CharField(max_length=300, null=False, default="", blank=False, verbose_name=_('First name'))
+    middle_name = models.CharField(max_length=300, null=False, default="", blank=True, verbose_name=_('Middle name'))
+    last_name = models.CharField(max_length=300, null=False, default="", blank=False, verbose_name=_('Last name'))
```
joemull, line 213: This is a good catch. i agree these fields should all be `null=False` because they are CharField. 

However, I don't think the `default=""` is necessary because Django's default empty storage is an empty string. See the [documentaiton on null](https://docs.djangoproject.com/en/4.2/ref/models/fields/#null). I could be wrong about this but I think that's how it works.

Also, if we are going to allow the `first_name` field to be empty, it needs to have `blank=True` or otherwise people will continue to get form validation errors.

mauromsl: We still need the `default` parameter, otherwise I think object creation will still fail when no value is provided and I'd argue we also need to still set  `blank=True` so that the django forms in the admin area don't declare empty values as invalid.

mauromsl (approved):

src/core/migrations/0083_auto_20230627_1400.py
```diff
@@ -0,0 +1,39 @@
...
+    for account in Account.objects.filter(
+        Q(first_name=None) | Q(middle_name=None) | Q(last_name=None),
+    ):
+        if account.first_name is None:
+            account.first_name = ""
+        if account.middle_name is None:
+            account.middle_name = ""
+        if account.last_name is None:
+            account.last_name = ""
+        account.save()
```
mauromsl, line 28: When autogenerating a migration that makes a field non-nullable, you can provide a default value to the `makemigrations` command, so this shouldn't be necessary

src/core/migrations/0083_auto_20230627_1400.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("core", "0082_galley_file_not_nullable"),
+    ]
+
+    operations = [
+        migrations.RunPython(set_null_names_to_empty_string),
```
mauromsl, line 38: We are missing the operation that alters the schema, which shouldn't require this `RunPython` statement according to my previous comment

mauromsl (changes requested):

src/core/models.py
joemull: As I understand it, `default` and `blank` work hand-in-hand, right? If we set `blank=True`, then Django will not care if nothing is entered in the field, and on save, it will set the field in the database to an empty string. I *believe* this will happen whether or not default is set, so long as `blank` is `True`.

mauromsl: > As I understand it, `default` and `blank` work hand-in-hand, right? If we set `blank=True`, then Django will not care if nothing is entered in the field, and on save, it will set the field in the database to an empty string. I _believe_ this will happen whether or not default is set, so long as `blank` is `True`.

Not quite, while `blank` is declared on the model, `ModelForm` and `AdminForm` will use it to determine if a field is required or not during form validation (`form.is_valid`) and before the instance of the model is created (before the `default` value can be used during `form.save()`). So even if there is a default provided, it is best to also be explicit and declare `blank=True` as to allow forms using this model to leave the field blank.

Here is the offending line on `ModelForm`: https://github.com/django/django/blob/07cc014cb30bc3c343a25c81aad6820bbc72c0d9/django/forms/models.py#L355

joemull: > Not quite: while blank is declared on the model, ModelForm and AdminForm will use it to determine if a field is required or not during form validation (form.is_valid) and before the instance of the model is created (before the default value can be used during form.save()). So even if there is a default provided, it is best to also be explicit and declare blank=True as to allow forms using this model to leave the field blank.

To be clear, you and I are saying the same thing about `blank=True`. We both think we should set this.

I was only saying that when we do that, we might not need `default=""`, but really I don't mind if we keep it.

gamboz:
Things to consider:
- people without a `last_name`
  - https://www.w3.org/International/questions/qa-personal-names
  - https://shinesolutions.com/2018/01/08/falsehoods-programmers-believe-about-names-with-examples/
- contributors that are not people:
  - collaborations (e.g. https://doi.org/10.22323/2.08020305)
  - organizations (as hinted by crossref schema)
- percolations 
  - for ORCID it is the "first name" that is the required field
  - same for OJS3 (mandatory "first name", optional family name)
  - crossref schema has mandatory `<surname>` (e.g. [schema 4.8.1](https://data.crossref.org/reports/help/schema_doc/4.8.1/crossref4_8_1_xsd.html#person_name))
  - same for JATS (mandatory `<surname>`; e.g. [JATS green 1.3](https://jats.nlm.nih.gov/archiving/tag-library/1.3/element/name.html))
  - citation generation: ideally, the family name field should be used, for citations to render the entire name.
  - when sharing data with other 3rd party services other than Crossref (e.g Google Scholar, DOAJ...) each upstream will probably make their own assumptions about how to handle a name without a last name/family name
  -  It will require changing how we generate certain name abbreviations (email signatures and citations) as well as migrating names where the last_name is present but the first_name is not

src/core/migrations/0083_auto_20230627_1400.py
gamboz: I think this is fixed in  10abe982354c11e29ceb656d0963df6e96b4c6a4

src/core/models.py
gamboz: In 10abe982354c11e29ceb656d0963df6e96b4c6a4 I've left only `first_name` with `blank=True` (AFAICT this go towards the comments collected in Discord).

WRT leaving `default=...` etc., I prefer to see them explicit, although redundant when they are the default values, so that I don't forget, but it's not a strong opinion :slightly_smiling_face:

joemull (approved):

ajrbyers:
@mauromsl whats the status of this PR? I think you approved it then requested changes.

joemull:
I think the idea of using empty strings rather than nulls has been superseded by https://github.com/openlibhums/janeway/pull/4153.

But Matteo's comments about making more name fields optional (i.e. last name) should be saved and put in a new issue about making our name fields more globally inclusive, I think. Those comments really relate to just this change:

```py
last_name = models.CharField(
    ...
    blank=True
    ...
)
```

That might have far-reaching implications but I think it needs to happen at some point.

mauromsl (changes requested):
We've had a second look at this PR, since it is probably the correct approach for string values. The problem we have is an odd behaviour with existing code:

When you provide `None` as an explicit value to ORM methods like `save` and `create`, the default value provided by the model is ignored `default=''` and this the database (which no longer allows a `NULL` value) will throw an error at commit time. We've already had this issue when applying similar changes to other string value fields.

We recommend adding a `pre_save` signal that handles these cases, just to avoid breaking the behaviour from existing plugins.

PR #3619, opened 2023-06-30
Fixes an issue were email settings were empty on Docker

mauromsl:
When using the docker development setup, and when the `JANEWAY_EMAIL_BACKEND` is present but not set, our `dev_settings.py` file would set the EMAIL_BACKEND to ''

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,11 +47,6 @@ ifdef VERBOSE
 endif
 
 # Email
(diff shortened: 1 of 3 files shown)


joemull (commented):
This looks OK to me but I say that without context as I'm not a Docker user

ajrbyers (approved):

PR #3644, opened 2023-07-26
Check field value before accessing an attribute

yakky:
While `ReviewAssignment.review_round` is usually set by logic functions, in case of customization of the workflow this might be `None`, it's best to check its value before accessing an attribute of the foreignkey


Bbkctp:

ajrbyers (approved):

ajrbyers:
test this please

joemull (approved):

PR #3650, opened 2023-07-28
3630 the Live Article link in LDM will now link to the remote url when is_remote and remote_url are set

### LINKED ISSUE #3630: Live Article link doesn't work for remote items
**Describe the bug**
Under 'Logs, Documents & More' for published items there is a link to 'Live Article' - the URL for this link goes to the Janeway frontend, which does not work for remote articles / journals that are 'headless'

**Janeway version**
1.5

**To Reproduce**
1. Go to the Archive page for any published article that is remote = true and has a remote url
2. Select Logs, Documents & More > Live Article - you are taken to the Janeway public URL for the item instead of the remote URL

**Expected behavior**
Note that on the Archive page itself there is already a Remote Article View link. I would expect the Live Article link described above to function the same as the Remote Article View link, when an article is remote.

*CDL can work on this if Janeway team approves of the logic

ajrbyers:
I think we need to consider whether article's `url` method should return the `remote_url` when it and `is_remote` are set, but it might have a large impact that needs to be considered.

Closes #3630

### DIFF
--- a/src/templates/admin/elements/article_jump.html
+++ b/src/templates/admin/elements/article_jump.html
@@ -24,7 +24,7 @@
         <li><a href="{% url 'manage_archive_article' article.pk %}">Archive Page</a></li>
         <li><a href="{% url 'manage_article_workflow' article.pk %}" target="_blank">Manage Workflow Stage</a></li>
         {% if article.is_published %}
-            <li><a href="{{ article.url }}">Live Article</a></li>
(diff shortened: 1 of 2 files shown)


mauromsl (changes requested):
I feel this  will get annoying fairly quickly, as I bet it also happens on OAI feed, JATS Stub, REST API etc... should we alter the behaviour of `article.url` to do this check instead of polluting the templates with this logic?

We can add a method like `article.local_url` so the behaviour of returning a local path (even for remote articles) can be preserved.

PR #3651, opened 2023-07-28
#3643 when editing a single setting using the edit_key view unchecking a boolean now works as expected.

### LINKED ISSUE #3643: Checkbox toggles in All Settings cannot be updated as of 1.5
**Describe the bug**
Demo'd for Mauro on 7/26/23: Some settings cannot be updated as of 1.5. This appears to be limited to checkbox toggles. For example, try to override the default value of use_crossref. After saving your change, it will revert back to the default. This also only seems to occur when accessing settings from the All Settings interface.

As a workaround, you can achieve your desired setting by changing in in the Django admin.

**Janeway version**
1.5

ajrbyers:
Closes #3643

### DIFF
--- a/src/core/forms.py
+++ b/src/core/forms.py
@@ -39,7 +39,7 @@ class EditKey(forms.Form):
         if self.key_type == 'rich-text':
             self.fields['value'].widget = SummernoteWidget()
         elif self.key_type == 'boolean':
-            self.fields['value'].widget = forms.CheckboxInput()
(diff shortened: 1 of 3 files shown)


joemull (changes requested):
This seems to now work but confusingly the word "Value" is displayed to the user beside the check box. Any way to avoid that?

image

ajrbyers:
> This seems to now work but confusingly the word "Value" is displayed to the user beside the check box. Any way to avoid that?
> 
> image

Good catch, should be removable.

PR #3654, opened 2023-08-03
#3652 when an editor views the submission form they will now see information about which sections are not open for public submission.

### LINKED ISSUE #3652: As an editor I'd like to see which sections are closed for submission
As an editor, if I turn off submissions for a section (submission.models.Section.public_submission), I'd like for the dropdown in the submission form (submission.forms.ArticleInfoSubmit) to reflect that.

For authors, it will be invisible, but for me, it's still visible, which provides me flexibility but doesn't give me confidence that the feature is working and that authors won't see it.

Add a parenthetical "public submission closed" or something similar in the dropdown option label.

ajrbyers:
- Adds a new method to section that returns a display name for editors that includes information on public submission status
- Adds a new field that displays the above method as the <option> label
- Adds help text explaining that editors will see all sections but that authors will not
- Adds two tests to check that what editors and authors see
- Closes #3652 

Editor view:
image
image

### DIFF
--- a/src/submission/fields.py
+++ b/src/submission/fields.py
@@ -1,6 +1,0 @@
-from django import forms
-
-
-class EditorSectionChoiceField(forms.ModelChoiceField):
-    def label_from_instance(self, obj):
-        return obj.editor_display_name()
--- a/src/submission/fields.py
+++ b/src/submission/fields.py
@@ -1,0 +1,6 @@
+from django import forms
+
+
+class EditorSectionChoiceField(forms.ModelChoiceField):
+    def label_from_instance(self, obj):
+        return obj.editor_display_name()
--- a/src/submission/models.py
+++ b/src/submission/models.py
@@ -2063,7 +2063,7 @@ class Section(AbstractLastModifiedModel):
             return self.name
         return f"Unnamed Section {self.pk}"
 
-    def editor_display_name(self):
+    def display_name_public_submission(self):
         """
         Returns a display name that informs the user if the section is
         close for submission.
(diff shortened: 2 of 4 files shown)


src/submission/models.py
```diff
@@ -2063,6 +2063,19 @@ def __str__(self):
             return self.name
         return f"Unnamed Section {self.pk}"
 
+    def editor_display_name(self):
```
mauromsl, line 2066: Why is this function called `editor_display_name` if it doesn't do anything regarding an editor role. How about `display_name_public_submission`

src/submission/fields.py
```diff
@@ -0,0 +1,6 @@
+from django import forms
+
+
+class EditorSectionChoiceField(forms.ModelChoiceField):
```
mauromsl, line 4: This class doesn't actually enforce any behaviour that warrants the "Editor" part of the name and since it is a such a niche use case, I wonder if it doesn't make more sense to just do this on `EditorArticleInfoForm.__init__()` instead, until the abstraction makes more sense

mauromsl (changes requested):
Just some nitpicking about unneeded abstraction and class/function names

src/submission/fields.py
ajrbyers: I thought of doing this in `EditorArticleInfoForm.__init__()` as well but chose to do it this way to avoid repeating the logic around `FILTER_PUBLIC_FIELDS` that exists in the `ArticleInfo` form.

PR #3663, opened 2023-08-16
#3372 the issue title is now cached in all available languages, or the default if none are enabled.

### LINKED ISSUE #3372: Cached Issue Display and Languages
**Describe the bug**
Sometimes an issue display title will get stuck displaying the wrong information. An example was reported by UGent on one of their Dutch language journals where a title was generated using the 'en' language code and so displayed "Issue" rather than the Dutch "Nummer".

**Janeway version**
v.1.4.4

**To Reproduce**
1. Create an issue with the wrong language set
2. There is no obvious way to force a change to the cached display
3. Switching languages may also not change wording due to the cached nature of the field

**Expected behavior**
This field should be translatable and be set for each active language.

ajrbyers:
- Closes #3372


mauromsl (changes requested):
Just one small comment on a potential bug that this could introduce

mauromsl (approved):

PR #3667, opened 2023-08-18
When editing repository settings repository managers can choose to save and stay on the current page or save and move to the next step.

ajrbyers:


src/repository/views.py
```diff
@@ -1537,16 +1537,24 @@ def repository_wizard(request, short_name=None, step='1'):
...
+                        kwargs=kwargs,
+                    )
+                )
+            else:
+                kwargs['step'] = step
+                return redirect(
+                    reverse(
+                        'repository_wizard_with_id',
+                        kwargs=kwargs,
+                    )
```
mauromsl, line 1556: minor nitpick: Seems like inly the `kwargs["step"]` definition should live in the if block, the URL is still the same in both cases

mauromsl (changes requested):

mauromsl (approved):

PR #3681, opened 2023-09-01
#3487 when requesting revisions the Editor Note field is now optional and is no longer inserted into the outgoing request email.

### LINKED ISSUE #3487: As an editor I don't know when to use the Editor note field on revision requests

When editors are sharing reviews with authors, they see a field called "Editor note", and then a few screens later, they get a field with a full templated email to the author. This raises questions about what Editor note should be used for.

Editor new to the system: "Is there a template mail for the notification to the authors? Or do we have to write a mail in the window 'Editor note'?"

Editor familiar with the system: "There is indeed a template mail for notification to authors, which you can tweak as you want to. It appears once you go past the window that has the editor note, the major/ minor revisions recommendation, and the revisions due date. I usually ignore the small window editor note (I put in abc, because you have to put a value); I select 'major revisions', and I select a date (10 weeks from the decision, indeed). Then I proceed to the notification mail template where I remove the text from that 'editor note window (abc), and then I modify the template informing the author how many reviews I received, what the tone of the reviews is, what they need to pay special attention to etc. Just a few lines, no need to write a synthesis of the reviews."

The editor note field is indeed required and does not have help text that would explain what other options editors have a few screens later:

https://github.com/BirkbeckCTP/janeway/blob/77bec6d75e40b34d64fbf4731b2a79613f12a02e/src/review/models.py#L498

The user story above tells me the editor note field

- [ ] needs help text
- [ ] should be optional

ajrbyers:
- Help text has also been updated
- Closes #3487


src/core/migrations/0083_update_revision_request_template.py
```diff
@@ -0,0 +1,55 @@
+# Generated by Django 3.2.19 on 2023-08-31 20:17
+
+from django.db import migrations
+from django.conf import settings
+from django.utils import translation
+
+SETTING_TEXTS_TO_REMOVE = [
+    ' Details of the changes required are below:<br><br>{{ revision.editor_note }}',
+    ' Details of the changes required are below:<br/><br/>{{ revision.editor_note }}'
+]
```
mauromsl, line 10: Do we really want to remove this? It is a change on behavour that some editors may not like/expect. It would cause less friction to tweak the template in the cases where editors prefer it not be shared in the outgoing email notification

mauromsl (changes requested):
More a question than a change request

src/core/migrations/0083_update_revision_request_template.py
ajrbyers: Personally I think removing it makes sense. No other email templates work in this manner so its a bit of an odd one out actually.

ajrbyers:
I've removed the migration for the time being. We can canvass editors at a later date.

PR #3702, opened 2023-09-13
Include GET parameters when redirecting after login #3701

### LINKED ISSUE #3701: GET parameters dropped during login
When I am redirected after logging in, I am taken back where I was before, but I lose the search parameters and (non-sensitive) form inputs I had typed before.

https://github.com/BirkbeckCTP/janeway/blob/8fd9bcda77e378e54a31029dc3bb86ee41c8cfda/src/security/decorators.py#L44-L46

We can avoid this by keeping GET parameters in addition to `request.path`

Use case: calculator form on new supporters page, which uses GET for initial calculation using non-sensitive data, so that it's easy for anyone, but POST for submission, so that it's secure and validated, etc.

**Janeway version**
https://github.com/BirkbeckCTP/janeway/commit/84d141b9165ca564f02e90641e476bd0da744c94

joemull:
Closes #3701.

### DIFF
--- a/src/security/decorators.py
+++ b/src/security/decorators.py
@@ -41,8 +41,8 @@ def base_check(request, login_redirect=False):
         or not request.user.is_active
     ):
         if login_redirect is True:
-            path = request.path + '?' + request.GET.urlencode()
-            params = urlencode({"next": path})
+            request_params = request.GET.urlencode()
+            params = urlencode({"next": f"{request.path}?{request_params}"})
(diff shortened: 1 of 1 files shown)


ajrbyers:
@joemull this PR has borked some tests, looks like it has to do with Mock() being used for some testing.

joemull:
> @joemull this PR has borked some tests, looks like it has to do with Mock() being used for some testing.

Should be OK now.

ajrbyers (approved):

PR #3708, opened 2023-09-13
Convenience methods on press object #3707

### LINKED ISSUE #3707: Active journals and published articles on press sites
The homepage elements provide one way of listing journals and articles on press sites, but the OLH site design is calling for a more flexible way to get these querysets for use in templates.

Property methods on the press object that get active journals and published articles.

joemull:
Closes #3707.

### DIFF
--- a/src/press/models.py
+++ b/src/press/models.py
@@ -11,10 +11,12 @@ import uuid
 from django.conf import settings
 from django.core.validators import MinValueValidator
 from django.db import models
+from django.utils import timezone
 
 from core import models as core_models
 from core.file_system import JanewayFileSystemStorage
 from core.model_utils import AbstractSiteModel, SVGImageField
+from submission import models as submission_models
 from utils import logic
 from utils.function_cache import cache
 from utils.logger import get_logger
@@ -343,6 +345,22 @@ class Press(AbstractSiteModel):
     @property
     def code(self):
         return 'press'
+
+    @property
+    def active_journals(self):
+        from journal import models as journal_models
+        # imported here to avoid circular imports
+        return journal_models.Journal.objects.filter(
+            hide_from_press=False,
+        )
+
+    @property
+    def published_articles(self):
+        return submission_models.Article.objects.filter(
+            journal__press=self,
+            stage=submission_models.STAGE_PUBLISHED,
+            date_published__lte=timezone.now(),
+        )
 
     class Meta:
         verbose_name_plural = 'presses'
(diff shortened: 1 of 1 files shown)


src/press/models.py
```diff
@@ -312,6 +314,22 @@ def preprint_dois_enabled(self):
...
+        from journal import models as journal_models
+        # imported here to avoid circular imports
+        return journal_models.Journal.objects.filter(
+            hide_from_press=False,
+        )
+
+    @property
+    def published_articles(self):
+        return submission_models.Article.objects.filter(
+            journal__press=self,
```
ajrbyers, line 328: `Journal` does not have an FK to `Press` so you can't filter on it here.

ajrbyers (changes requested):

src/press/models.py
joemull: Thanks for catching that. Just removed it.

ajrbyers (approved):

src/press/models.py
```diff
@@ -344,6 +346,21 @@ def navigation_items(self):
     def code(self):
         return 'press'
 
+    @property
+    def active_journals(self):
+        from journal import models as journal_models
+        # imported here to avoid circular imports
+        return journal_models.Journal.objects.filter(
+            hide_from_press=False,
+        )
```
mauromsl, line 355: Django provides a model loader to avoid circular imports (and runtime imports):
`https://docs.djangoproject.com/en/3.2/_modules/django/apps/config/#AppConfig.get_model`

I also find the term `active` overlaps with its publication status (currently publishing or archived), perhaps `public_journals` or `listed_journals` would be a better name.

src/press/models.py
```diff
@@ -344,6 +346,21 @@ def navigation_items(self):
...
+
+    @property
+    def published_articles(self):
+        return submission_models.Article.objects.filter(
+            stage=submission_models.STAGE_PUBLISHED,
+            date_published__lte=timezone.now(),
+        )
+
```
mauromsl, line 363: as above, best to get the model with https://docs.djangoproject.com/en/3.2/_modules/django/apps/config/#AppConfig.get_model to avoid inevitable future circular imports

mauromsl (changes requested):

src/press/models.py
joemull: Good suggestion--change made

mauromsl (approved):

src/press/models.py
joemull: For this one we still need submission_models.STAGE_PUBLISHED

mauromsl: > For this one we still need submission_models.STAGE_PUBLISHED

Yep, one day we will also move these to a consts.py module to avoid circularity

PR #3710, opened 2023-09-13
Use press-level contact template at press level #3709

### LINKED ISSUE #3709: Separate press-level contact template is not used
**Describe the bug**
The press-level contact template at `themes/*/templates/press/journal/contact.html` is not used by the contact view even if there is no journal:

https://github.com/BirkbeckCTP/janeway/blob/742cc1d6d4f908fa916de7b7380c2ea766bdb8c8/src/journal/views.py#L1875-L1878

**Janeway version**
https://github.com/BirkbeckCTP/janeway/commit/742cc1d6d4f908fa916de7b7380c2ea766bdb8c8

joemull:
Important: Please don't merge this in v1.5.1! It could introduce further bugs in the press-level templates that we haven't caught because Janeway was using the journal-level templates. Let's introduce it in v1.5.2 or later.

Closes #3709.

### DIFF
--- a/src/journal/views.py
+++ b/src/journal/views.py
@@ -1874,8 +1874,10 @@ def contact(request):
 
     if request.journal and request.journal.disable_front_end:
         template = 'admin/journal/contact.html'
-    else:
+    elif request.journal:
         template = 'journal/contact.html'
+    else:
+        template = 'press/journal/contact.html'
     context = {
         'contact_form': contact_form,
         'contacts': contacts,


ajrbyers (commented):
Looking at the themes I think only OLH has `press/journal/contact.html` (which is an iffy path anyway. It might be worth adding `press/contact.html` for each of the themes?

joemull:
> Looking at the themes I think only OLH has `press/journal/contact.html` (which is an iffy path anyway. It might be worth adding `press/contact.html` for each of the themes?

@ajrbyers Can do, but I thought the press-level clean and material themes weren't supported? Do we know anyone using them?

Agree about the iffy path. We had a conversation about this a while back and I think the decision was not to change any of these dodgy theme template paths until we can decide on a good convention to standardize them all. Does that still hold here?

mauromsl (approved):

PR #3713, opened 2023-09-13
Retrieve press-level email subjects from settings #3694

### LINKED ISSUE #3694: Notify hook does not handle press-level email subjects
**Describe the bug**
The `notify_hook` function that dynamically pulls email subjects from settings only works for journal-level emails:

https://github.com/BirkbeckCTP/janeway/blob/84d141b9165ca564f02e90641e476bd0da744c94/src/utils/notify_plugins/notify_email.py#L125-L127

**Janeway version**
https://github.com/BirkbeckCTP/janeway/commit/84d141b9165ca564f02e90641e476bd0da744c94

joemull:
Closes #3694.

### DIFF
--- a/src/core/logic.py
+++ b/src/core/logic.py
@@ -54,14 +54,14 @@ def send_reset_token(request, reset_token):
             request.press.password_reset_text,
             template_is_setting=True,
         )
-        subject = 'Password Reset'
     else:
         message = render_template.get_message_content(
             request,
             context,
             'password_reset',
         )
-        subject = 'subject_password_reset'
+
+    subject = 'subject_password_reset'
 
     notify_helpers.send_email_with_body_from_user(
         request,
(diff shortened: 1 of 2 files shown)


ajrbyers (changes requested):
I just had a quick look at the logic for some of the emails that send at the press level as well as the journal (Password Reset and Registration Confirmation) at the moment when no `request.journal` is present these send a hard-coded subject see:

https://github.com/BirkbeckCTP/janeway/blob/742cc1d6d4f908fa916de7b7380c2ea766bdb8c8/src/core/logic.py#L38-L72

The two functions in `core.logic` need a little tweak to make this PR complete.

ajrbyers (approved):

mauromsl (approved):

PR #3714, opened 2023-09-13
Improvements to meta tags #3692

### LINKED ISSUE #3692: Meta tag improvements
As a developer I'd like a few improvements relate to writing meta tags:

- [x] Django provides `request.build_absolute_uri()` for getting an absolute URI for any location. This is useful in templates when creating precise meta tags. So I'd like a template tag for this method, because so far as I know, we don't have another way to build these URLs in some cases. Example usage:

    ```django
    <meta property="og:image" content="{% build_absolute_uri request.press.default_carousel_image.url %}">
    ```

- [x] Use Django's built-in `truncatewords` rather than our custom `truncatesmart` in `templates/common/elements/social_meta.html`

joemull:
Closes #3692.

### DIFF
--- a/src/core/templatetags/fqdn.py
+++ b/src/core/templatetags/fqdn.py
@@ -57,3 +57,9 @@ def external_journal_url(journal, url_name=None, *args):
         path = None
 
     return journal.site_url(path=path)
+
+
+@register.simple_tag(takes_context=True)
+def build_absolute_uri(context, relative_url, *args):
+    request = context.get('request')
+    return request.build_absolute_uri(relative_url)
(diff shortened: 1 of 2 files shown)


src/core/templatetags/fqdn.py
```diff
@@ -57,3 +57,9 @@ def external_journal_url(journal, url_name=None, *args):
         path = None
 
     return journal.site_url(path=path)
+
+
+@register.simple_tag(takes_context=True)
+def build_absolute_uri(context, relative_url, *args):
```
ajrbyers, line 63: As far as I can tell this templatetag is never used.

ajrbyers (changes requested):
See comment inline. The new templatetag added isn't used, is this PR incomplete?

ajrbyers (approved):

mauromsl (approved):

PR #3715, opened 2023-09-13
Improvements to news

### LINKED ISSUE #3691: Press overview page for news
We have several kinds of "news" at the OLH: blog posts, announcements, journal news, etc. We'd like the ability to structure these in sets on one landing page.

Create some new urls and views that can filter both by tag and by content_type (e.g. press or journal) and return several querysets based on those parameters.

Deprecate the `comms` app and migrate to something more robust and flexible like https://github.com/saadmk11/django-newsfeed, though we'd need to discuss how to set policies around it to maintain our data protection responsibilities, as this app includes subscription capabilities.

joemull:
Closes #3691.


src/comms/models.py
```diff
@@ -14,6 +14,15 @@
 __maintainer__ = "Birkbeck Centre for Technology and Publishing"
 
 
+class ActiveNewsItemManager(models.Manager):
+    def get_queryset(self):
+        return super().get_queryset().filter(
```
ajrbyers, line 19: This could also be written using a single filter with Q:

```
return super().get_queryset().filter(
    (models.Q(start_display__lte=timezone.now()) | models.Q(start_display=None)) &
    (models.Q(end_display__gte=timezone.now()) | models.Q(end_display=None))
)
```

src/comms/models.py
```diff
@@ -28,8 +37,16 @@ class NewsItem(models.Model):
...
 
-    large_image_file = models.ForeignKey('core.File', null=True, blank=True, related_name='large_news_file',
-                                         on_delete=models.SET_NULL)
+    large_image_file = models.ForeignKey(
+        'core.File',
+        null=True,
+        blank=True,
+        related_name='large_news_file',
+        on_delete=models.SET_NULL,
+        help_text='An image for the top of the news item page and the'
```
ajrbyers, line 46: Spaces are missing from the help text (also in the migration above)

src/comms/views.py
```diff
@@ -188,6 +195,58 @@ def news_list(request, tag=None):
...
+
+    all_tags = models.Tag.objects.all().annotate(
+        Count('tags')
+    ).order_by('-tags__count', 'text')
+
+    tag_texts = [tag.text for tag in all_tags]
+    tag_sets = {}
+    object_sets = {}
+    for key in request.GET:
+        if key == 'tags__text':
```
ajrbyers, line 222: This might just be me but it feels odd to expose the django notation in the GET parameters. Could this just be 'tags' and below 'type'?

ajrbyers (changes requested):
This is generally really good. I've added a couple of notes and one change that needs to be made around the help_text on `large_image_file.`

src/comms/views.py
joemull: I think I did this to be more explicit and readable because I was dealing with bugs arising from the ambiguity of "tags"--is it an object, a string, etc.

ajrbyers (approved):

src/comms/views.py
```diff
@@ -148,29 +148,32 @@ def serve_news_file(request, identifier_type, identifier, file_id):
...
+            content_type=request.model_content_type,
+            object_id=request.site_type.id,
+        )
+
+    if tag:
+        news_objects = news_objects.filter(
+            tags__text=urllib.parse.unquote(tag),
+        )
+        tag = models.Tag.objects.get(
+            text=urllib.parse.unquote(tag)
```
mauromsl, line 173: Minor nitpick: in the new implementation we are unquoting the tag twice.

src/comms/views.py
```diff
@@ -188,6 +195,58 @@ def news_list(request, tag=None):
...
+                        content_type__model=value,
+                    )
+
+    pinned_items = models.NewsItem.active_objects.filter(
+        content_type=request.model_content_type,
+        object_id=request.site_type.id,
+        pinned=True,
+    )
+
+    template = 'press/news/overview.html'
```
mauromsl, line 243: I can't find this template neither in the codebase nor this PR, is it coming in another PR?

mauromsl (changes requested):
A couple of comments inline

src/comms/views.py
joemull: It's intended to pick this up: https://github.com/BirkbeckCTP/hourglass/blob/main/templates/press/news/overview.html

Should I add a comment that it is only supported on the hourglass theme?

mauromsl: It is worth adding a very basic template for OLH theme to avoid confusion

mauromsl: Either solution is better than having a view request a template that doesn't exist. If we don't see use for this view outside of hourglass, then it is fine to have it there.

Thoughts @ajrbyers

ajrbyers: EIther works for me. In the short term a component is fine but we should consider extending some of the Hourglass features onto presses in the future.

joemull: Great. Let's move it to a component for now. 

I will extricate it from this PR and then re-request review so we can keep the other improvements here.

joemull:
Jenkins: test this please

joemull:
Jenkins: test this please

joemull:
Jenkins: test this please

joemull:
Alright, the press overview has been removed and tests are passing, so this should be ready to merge, @mauromsl

mauromsl (approved):

PR #3721, opened 2023-09-18
#3486 DOI batch deposit timestamps are now generated in the view using datetime rather than the now templatetag which uses timezone.

### LINKED ISSUE #3486: Crossref deposits rejected by given lower version number generated from a localised datetime
**Describe the bug**
When a user makes a crossref deposit via the web interface, the time stamp used for the version number is localised to their own timezone. As are result, a consecutive deposit made from a different timezone can be flagged by crossref as outdated.

**Janeway version**
v1.5

**To Reproduce**

1. Make a crossref deposit
2. In the user profile, change your timezone to one at least one hour behind the currently configured one
3. Make a new crossref deposit
4. See error on response XML

**Expected behavior**
Janeway should always use the server configured timezone settings when generating the timestamp used for versioning

ajrbyers:
```
<timestamp>20230918114200656</timestamp>
```

- Output remains in the same format but uses server time rather than the user's timezone. Example output from the view above.
- Note this particular example has a timestamp suffix '656' tagged onto it.
- Closes #3486

### DIFF
--- a/src/identifiers/logic.py
+++ b/src/identifiers/logic.py
@@ -230,7 +230,7 @@ def create_crossref_doi_batch_context(journal, identifiers):
 
     template_context = {
         'batch_id': uuid4(),
-        'now': timezone.now(),
+        'now': datetime.datetime.now(),
         'timestamp': int(round((datetime.datetime.now() - datetime.datetime(1970, 1, 1)).total_seconds())),
         'timestamp_suffix': timestamp_suffix,
         'depositor_name': setting_handler.get_setting('Identifiers', 'crossref_name',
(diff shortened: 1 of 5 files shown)


joemull (changes requested):
This is great. But I think we might want to change the template tag "now" in the component template as well:

https://github.com/BirkbeckCTP/janeway/blob/4bf344ae453cdcccd2f6461fd0cedfd89cba17d6/src/templates/common/identifiers/crossref_component.xml#L6

May need to be added to context for that template as well.

mauromsl (approved):

PR #3725, opened 2023-09-19
Quick search for press-level websites

### LINKED ISSUE #3432: User-friendly site search
New visitors often look for a quick search function to find information about a press or journal on their website. Ideally this kind of search would...

- [ ] Contain all the substantive text on a press-level website, including CMS pages, news items, contacts, and journal descriptions. (Journal-level websites are TBD.)
- [ ] Provide results as you type
- [ ] Highlight search terms in results so you can evaluate before clicking

Build a search index on the server with lunr.py that can be searched by lunr.js in the browser.

- Use Google to power site search
- Use same SQL search tool as for full-text articles

joemull:
- Closes #3432.
- Closes BirkbeckCTP/hourglass#32.
- For how it will look and feel on the Hourglass theme, see https://github.com/BirkbeckCTP/hourglass/pull/72.
- Closes BirkbeckCTP/consortial_billing#19.

Note that this represent a change in course in early Sept 2023, when it became clear that lunr wasn't going to meet our needs because of an issue with the stemming algorithm that leads to unwanted results, including not picking up the word "Janeway"!

I think that MiniSearch is a very good alternative--I might have gone with it over lunr if I had found it earlier. It uses fuzzy searching rather than stemming, and it has a lot more adoption by later tools. One example is Vitepress, a Vue-ecosystem static site generator. It also offers autocomplete, which I have not implemented yet but can if time.

The only drawback of MiniSearch is there is no Python equivalent that can create the index server-side. So we are back to relying on the browser to index documents provided in a JSON file. I plan to test the performance of this using the OLH database but have not done so yet. MiniSearch promises to be respectful of browser resources, at least in terms of the index size, though computational resources immediately after page load are still a question mark for me. See the [creator's blog post for a more detailed comparison](https://lucaongaro.eu/blog/2019/01/30/minisearch-client-side-fulltext-search-engine.html).

A potential solution is to index it on the server using Node, which will already be a requirement from several other pieces of this theme, like Tailwind. Not an ideal solution, but an option.

Another potential solution is to disable it entirely, including inputs and indexing, on smaller screens. This would avoid most scenarios where computational resources are limited, though not all.

### DIFF
--- a/src/cms/logic.py
+++ b/src/cms/logic.py
@@ -110,6 +110,9 @@ def get_press_site_search_data():
         if response.status_code != 200:
             return
 
+        if 'text/html' not in response.headers['Content-Type']:
+            return
+
         html = BeautifulSoup(response.text, 'html.parser')
         body = html.find('body')
         if not isinstance(body, Tag):
--- a/src/cms/logic.py
+++ b/src/cms/logic.py
@@ -9,7 +9,6 @@ import glob
 import json
 import time
 import requests
-from typing import Tuple, List, Dict
 from urllib.parse import urljoin, urlparse
 from bs4 import BeautifulSoup, Comment, NavigableString, Tag
 
@@ -34,15 +33,17 @@ SITE_SEARCH_PATH = os.path.join(
 )
 
 
-def add_search_index_document(
-    docs: List[Dict],
-    url: str,
-    name: str,
-    text: str,
-) -> List[Dict]:
+def add_search_index_document(docs, url, name, text):
     """
     Adds the data for a part or page to the search index
+
+    :type docs: list[dict]
+    :type url: str
+    :type name: str
+    :type text: str
+    :rtype: list[dict]
     """
+
     data = {}
     data['id'] = len(docs)
     data['url'] = url
@@ -52,13 +53,14 @@ def add_search_index_document(
     return docs
 
 
-def gobble_sibling_text(
-    sibling: Tag | NavigableString,
-    original_part: Tag | NavigableString,
-) -> str:
+def gobble_sibling_text(sibling, original_part):
     """
     Recursively collects a series of parallel elements for
     indexing together, stopping before another heading
+
+    :type sibling: Tag | NavigableString
+    :type original_part: Tag | NavigableString
+    :rtype: str
     """
 
     # Handle no sibling
@@ -91,10 +93,13 @@ def gobble_sibling_text(
     return gobbled_text.strip()
 
 
-def get_text_for_parent(
-    parent: Tag | NavigableString,
-    original_part: Tag | NavigableString,
-) -> str:
+def get_text_for_parent(parent, original_part):
+    """Gets for a parent element and its siblings
+
+    :type parent: Tag | NavigableString
+    :type original_part: Tag | NavigableString
+    :rtype: str
+    """
     if not parent:
         return ''
 
@@ -105,10 +110,13 @@ def get_text_for_parent(
         return get_text_for_parent(parent.parent, original_part)
 
 
-def get_text_for_header(part: Tag | NavigableString) -> str:
+def get_text_for_header(part):
     """
     Gets the sibling text or parent text
     when given an h2, h3, or h4
+
+    :type part: Tag | NavigableString
+    :rtype: str
     """
 
     # Try gobbling first
@@ -120,12 +128,16 @@ def get_text_for_header(part: Tag | NavigableString) -> str:
     return get_text_for_parent(part.parent, part)
 
 
-def add_part_as_doc(
-    docs: List[Dict],
-    part_url: str,
-    part_name: str,
-    part_text: str,
-) -> List[Dict]:
+def add_part_as_doc(docs, part_url, part_name, part_text):
+    """
+    Adds a section of the page as an index document
+
+    :type docs: list[dict]
+    :type part_url: str
+    :type part_name: str
+    :type part_text: str
+    :rtype: list[dict]
+    """
     if not part_name or not part_text:
         # HTML tree with id has no heading
         # or no text content
@@ -133,12 +145,19 @@ def add_part_as_doc(
     return add_search_index_document(docs, part_url, part_name, part_text)
 
 
-def add_searchable_page_parts(
-        docs: List[Dict],
-        url: str,
-        body: Tag,
-        headings: List | None = None,
-    ) -> List[Dict]:
+def add_searchable_page_parts(docs, url, body, headings=None):
+    """
+    Finds the sections of the page that can stand
+    alone as index documents, adds them to the docs,
+    and decomposes them so that they are not also indexed
+    as part of their larger page.
+
+    :type docs: list[dict]
+    :type url: str
+    :type body: Tag
+    :type headings: list | None
+    :rtype: list[dict]
+    """
     if not headings:
         headings = ['h4', 'h3', 'h2']
 
@@ -158,13 +177,38 @@ def add_searchable_page_parts(
     return docs
 
 
-def get_page(
-    url: str,
-    fetched_urls: set[str]
-) -> Tuple[BeautifulSoup | None, set[str]]:
+def remove_fragment(url):
+    """
+    Removes any fragment found in the url
+
+    :type url: str
+    :rtype: str
+    """
+    fragment = urlparse(url).fragment
+    return url.replace('#' + fragment, '')
+
+
+def normalize_url(url):
+    """
+    Removes any fragment and trailing slash
+
+    :type url: str
+    :rtype: str
+    """
+    return remove_fragment(url).rstrip('/')
+
+
+def get_page(url, fetched_urls):
+    """
+    Gets a new page using requests
+
+    :type url: str
+    :type fetched_urls: set[str]
+    :rtype: tuple[BeautifulSoup | None, set[str]]
+    """
     time.sleep(.1)
     try:
-        fetched_urls.add(url)
+        fetched_urls.add(normalize_url(url))
         headers = {
             'Accept': 'text/html; charset=utf-8'
         }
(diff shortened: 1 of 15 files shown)


joemull:
Just to note this is not quite ready to be reviewed, as there is a bug https://github.com/BirkbeckCTP/janeway/issues/3432#issuecomment-1743070359

GitHub is not giving me the option to change this to a draft PR.

ajrbyers:
> Just to note this is not quite ready to be reviewed, as there is a bug [#3432 (comment)](https://github.com/BirkbeckCTP/janeway/issues/3432#issuecomment-1743070359)
> 
> GitHub is not giving me the option to change this to a draft PR.

Hey Joe - I’ve converted it to draft.

src/cms/logic.py
```diff
@@ -0,0 +1,323 @@
...
+
+from journal import models as journal_models
+from repository import models as repository_models
+from cms import models as models
+from bs4 import BeautifulSoup, NavigableString, Tag, Comment
+from django.core.files.base import ContentFile
+from press import models as press_models
+from utils.logger import get_logger
+
+import json
```
ajrbyers, line 22: Can this be updated to match the recommended import order?

https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#imports

src/cms/logic.py
```diff
@@ -0,0 +1,323 @@
...
+        return
+    for non_content_selector in ['script', '.djdt-hidden']:
+        for element in body.select(non_content_selector):
+            element.decompose()
+    for element in body.find_all(text=lambda text: isinstance(text, Comment)):
+        element.extract()
+    return body
+
+
+def get_base():
```
ajrbyers, line 181: Could this use the `site_url` method in `AbstractSiteModel`?

src/cms/management/commands/delete_site_search_data.py
```diff
@@ -0,0 +1,32 @@
...
+        parser.add_argument('--press_id', type=int)
+
+    def handle(self, *args, **options):
+        if options['press_id']:
+            files_deleted = cms_logic.delete_search_data(press_id=options['press_id'])
+        else:
+            files_deleted = cms_logic.delete_search_data()
+
+        if files_deleted:
+            logger.info(
```
ajrbyers, line 22: As we've seen recently `logger.info` wont output in prod.

src/core/middleware.py
```diff
@@ -117,6 +118,15 @@ def process_request(request):
             request.site_type = request.site_object = press
             request.model_content_type = ContentType.objects.get_for_model(press)
             request.press_base_url = press.site_url()
+
```
ajrbyers, line 121: If this doesn't change very often it might be worth caching `get_search_data_file` otherwise it will be hit on every request.

ajrbyers (approved):
This looks good. I've added some comments in line but nothing major.

ajrbyers:
Note also: failing tests here

src/cms/logic.py
```diff
@@ -0,0 +1,323 @@
...
+        # HTML tree with id has no heading
+        # or no text content
+        return
+    add_search_index_document(part_url, part_name, part_text)
+    part.decompose()
+
+
+def add_searchable_page_parts(url, body):
+
+    headings = ['h4', 'h3', 'h2']
```
mauromsl, line 121: This should not be defined inside the function

src/cms/logic.py
```diff
@@ -0,0 +1,323 @@
...
+
+    def get_part_url(part):
+        part_id = part.get('id', '')
+        if part_id:
+            return urljoin(url, '#' + part_id)
+        else:
+            return remove_fragment(url)
```
mauromsl, line 128: why is this function nested? it doesn't seem to make use of parent state as say a closure or decorator

src/cms/logic.py
```diff
@@ -0,0 +1,323 @@
...
+
+def update_search_data(press_id=1):
+    press = press_models.Press.objects.get(pk=press_id)
+    docs_filename = os.path.join(
+        settings.SITE_SEARCH_DIR,
+        f'_press_{ press.pk }_documents.json'
+    )
+    docs_file, created = models.MediaFile.objects.get_or_create(
+        label=docs_filename
+    )
```
mauromsl, line 276: It would be good to encapsulate this in a function, rather than repeat the logic in various functions: `get_or_create_search_file`

src/cms/logic.py
```diff
@@ -0,0 +1,323 @@
...
+                f'Left-over test files: {os.listdir(SITE_SEARCH_PATH)}'
+            )
+        else:
+            os.rmdir(SITE_SEARCH_PATH)
+
+    return files_deleted
+
+
+@cache(900)
+def get_search_data_file(press):
```
mauromsl, line 310: If we turn this into `get_search_data_file` into `get_search_data_url` and return instead `docs_file.url` we also save us from having to run the query on every request.

src/core/middleware.py
mauromsl: @joemull  With If we turn the function `get_search_data_file` into `get_search_data_url` and return instead `docs_file.url` we would only have to cache that string

src/cms/logic.py
```diff
@@ -0,0 +1,323 @@
...
+        }
+        response = requests.get(url, headers=headers)
+    except requests.exceptions.ConnectionError:
+        logger.warn(f'Could not access {url}')
+        logger.warn('Please run server to index site search')
+        return
+
+    if response.status_code != 200:
+        return
+
```
mauromsl, line 155: If feel like something should be logged here too in some circumstances: If we've determined that the URL should be indexable, it should let us know when a 500 or 404 is returned. I imagine you are also guarding against 401 codes..

src/cms/logic.py
```diff
@@ -0,0 +1,323 @@
...
+    ] + [
+        repo.site_url() for repo in repository_models.Repository.objects.all()
+    ]
+
+
+def url_in_scope(deeper_url):
+    base = get_base()
+    if urlparse(deeper_url).hostname != urlparse(base).hostname:
+        return False
+    for excluded_url in excluded_urls():
```
mauromsl, line 198: `excluded_urls()` will be called very often, it should be cached.

src/cms/logic.py
```diff
@@ -0,0 +1,323 @@
...
+            part_url = get_part_url(part)
+            part_name = part.get_text().strip()
+            part_text = get_text_for_header(part)
+            add_part_as_doc(part, part_url, part_name, part_text)
+
+
+def get_page(url):
+    time.sleep(.1)
+    try:
+        fetched_urls.add(url)
```
mauromsl, line 143: It may be done elsewhere and I haven't spotted it, but shouldn't get_page ensure we don't request the same page multiple times. (esentially an early return `if  url in fetched_urls: return`)

mauromsl (changes requested):
Good work @joemull

If added a few comments inline. I'm still on the fence about having to scrape our own site... and I can think of a number of problems ahead if we were to recommend this approach to others but for our initial run of the OLH site, it should be ok

src/cms/logic.py
```diff
@@ -0,0 +1,323 @@
...
+    fragment = urlparse(url).fragment
+    return url.replace('#' + fragment, '')
+
+
+def url_is_unique(deeper_url):
+    def normalize(url):
+        url = remove_fragment(url)
+        if url.endswith('/'):
+            url = url[:-1]
+        return url
```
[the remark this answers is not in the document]
joemull: I put it there so that the context in which it is meant to be used was clearer to me. But I've taken in out since you find it confusing.

src/cms/logic.py
joemull: For context, but I'll take it out

src/cms/logic.py
joemull: Yes this is defended against via `add_searchable_page`, which is only called recursively if `url_in_scope` and `url_is_unique` return True. I've now renamed `url_is_unique` to `url_is_new` to help with readability.

src/cms/logic.py
```diff
@@ -0,0 +1,323 @@
...
+logger = get_logger(__name__)
+
+SITE_SEARCH_PATH = os.path.join(
+    settings.MEDIA_ROOT,
+    'press',
+    settings.SITE_SEARCH_DIR,
+)
+
+search_documents = []
+fetched_urls = set()
```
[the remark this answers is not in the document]
joemull: Just to avoid having a lot of params for every function. I've gone ahead and passed them around, as well as adding type annotations to make it clearer what is being passed.

src/cms/logic.py
```diff
@@ -1,15 +1,367 @@
...
+        return ''
+
+    text = gobble_sibling_text(parent.next_sibling, original_part)
+    if len(text.strip()) > len(original_part.get_text().strip()):
+        return text
+    else:
+        return get_text_for_parent(parent.parent, original_part)
+
+
+def get_text_for_header(part: Tag | NavigableString) -> str:
```
mauromsl, line 108: While I'm a big fan of type hints, I don't think we are ready to include them in the Janeway repository just yet, since most of the codebase is still using Google style docstring annotations (*"consistency is key"*)
Also, this operand is only compatible with python 3.10+ and we still support 3.8+

src/cms/logic.py
```diff
@@ -1,15 +1,367 @@
...
+    if url.endswith('/'):
+        url = url[:-1]
+    return url
+
+
+def url_is_new(fetched_urls: set[str], deeper_url: str) -> bool:
+    if deeper_url in fetched_urls:
+        return False
+    for fetched_url in fetched_urls:
+        if normalize_url(fetched_url) == normalize_url(deeper_url):
```
mauromsl, line 247: minor improvement:
if we stored the result of `normalize_url` in `fetched_urls` we would avoid the *O(N \* M)* time complexity of having to iterate the whole set everytime `url_is_new` is called

mauromsl (changes requested):
Almost there, just a comment on type hints and a minor performance improvement

src/cms/logic.py
joemull: OK, I have used docstrings until we can upgrade to type annotations

mauromsl (approved):

PR #3747, opened 2023-10-03
The OLH theme article block now renders 12 columns across rather than 10 eliminating a large chunk of whitespace.

ajrbyers:
Before:
Screenshot 2023-10-03 at 15 45 27

After:
Screenshot 2023-10-03 at 15 45 09

### DIFF
--- a/src/themes/OLH/templates/journal/article.html
+++ b/src/themes/OLH/templates/journal/article.html
@@ -149,7 +149,7 @@
             <div class="row">
                 <div class="large-8 columns border-right">
                     <div class="row">
-                        <div id="article" class="large-10 columns">
+                        <div id="article" class="large-12 columns">
                 {% if journal_settings.article.disable_article_large_image %}
                     <small>{{ article.section.name }}</small>
                     <h1>{{ article.title|safe }}</h1>


joemull (changes requested):
@ajrbyers would you mind explaining the reason you want to make this change? Was it maybe a bug according to how the framework expects you to lay out columns? Or a problem with the whitespace?

I am not sure about the effect, which is to make the body of the article wider on large screens, for two reasons:

- Best practices for line length say [you should aim for 50-75 characters per line](https://baymard.com/blog/line-length-readability). By my count this change would allow about 110 characters per line, so it would cause more fatigue and less accessibility for readers.
- White space may be there for a reason on large screens, so as not to make the page too dense visually.

Happy to consider other reasons for making the change, though, as there may be a way to achieve them without widening the text.

ajrbyers:
> @ajrbyers would you mind explaining the reason you want to make this change? Was it maybe a bug according to how the framework expects you to lay out columns? Or a problem with the whitespace?
> 
> I am not sure about the effect, which is to make the body of the article wider on large screens, for two reasons:
> 
>     * Best practices for line length say [you should aim for 50-75 characters per line](https://baymard.com/blog/line-length-readability). By my count this change would allow about 110 characters per line, so it would cause more fatigue and less accessibility for readers.
> 
>     * White space may be there for a reason on large screens, so as not to make the page too dense visually.
> 
> 
> Happy to consider other reasons for making the change, though, as there may be a way to achieve them without widening the text.

UCL requested we remove the white space for their journals which are heavy on tables and they don't particularly like the scroll bar and don't consider the "view table" popup as an acceptable solution.

If we want to aim for 50-75 characters per line we will need to drastically reduce the size on material (100+), clean (80+) and OLH (80+) and come up with a better way to present tables.

I don't mind leaving the hack in place, or adding a sub theme to UCL as a short term solution, I'll consider an alternative approach.

Issue #3801, opened 2023-12-04
Accessible Drag and Drop


ajrbyers:
**Is your feature request related to a problem? Please describe.**
The current implementation of drag and drop using JQUI Sortable is not accessible, namely it cannot be controlled with keyboard and mouse only and is not screen reader friendly.

**Describe the solution you'd like**
Re-implement sorting using accessible methods.

**Additional context**
W3C advice: https://www.w3.org/wiki/PF/ARIA/BestPractices/DragDrop
Interesting article by Salesforce: https://medium.com/salesforce-ux/4-major-patterns-for-accessible-drag-and-drop-1d43f64ebf09

PR #3806, opened 2023-12-04
Press staff groups

### LINKED ISSUE #3795: Press team
As a press, we want to configure our staff page by department using Janeway profiles.

Add press models for teams and groups much like the editorial team and group for journals.

joemull:
Closes #3795.

This adds press models, and just a few forms and view changes, that let you set up press-level teams.

I did not re-use the manager views due to accessibility issues with JQUI (see #3801).

There are a few fields that belong with the Edit Account form that I've added to the edit_profile view. So far I've only implemented the template on Hourglass for these. If we extend this feature to other presses we can add that form at the same time as we solve #3801.


joemull:
Jenkins test this please

src/core/models.py
```diff
@@ -302,6 +303,7 @@ def string_id(self):
         return str(self.id)
 
     def get_full_name(self):
+        """Deprecated"""
```
ajrbyers, line 306: It might be worth adding a pointer to the `full_name` method.

src/core/views.py
```diff
@@ -417,6 +418,13 @@ def edit_profile(request):
             request.journal
         ).value
 
+    if user.staffgroupmember_set.first():
```
ajrbyers, line 421: I think this works for our current use case but if we ever wanted to have one profile split across two groups there would be no way to edit the second entry. I don't think a change is needed now as it will slow down development but we should consider opening a ticket to convert this into a formset in future?

https://docs.djangoproject.com/en/3.2/topics/forms/formsets/

ajrbyers (changes requested):

ajrbyers (approved):

src/core/views.py
```diff
@@ -516,16 +536,25 @@ def public_profile(request, uuid):
...
-    }
+        template = 'core/accounts/public_profile.html'
+        context = {
+            'user': user,
+            'roles': roles,
+        }
+        if not context['roles']:
+            raise Http404()
+    else:
+        template = 'press/staff_profile.html'
```
mauromsl, line 553: I can't find this template here and it doesn't exist on master. Is it in another PR?

mauromsl (commented):
Just a comment on a missing template

src/core/views.py
joemull: Thanks for catching that--I'm going to refactor this view as a part of [#3895](https://github.com/BirkbeckCTP/janeway/pull/3901), so if you would eventually merge that into this, or straight into main, that would be great.

I'm going to make the press and journal views for staff and editorial groups all point to the same profile template, for simplicity and consistency.

joemull: This has now been done on https://github.com/BirkbeckCTP/janeway/pull/3901

joemull:
Closing this in favor of https://github.com/BirkbeckCTP/janeway/pull/3901

PR #3808, opened 2023-12-06
Custom CMS templates

### LINKED ISSUE #3794: When creating a CMS page, I want to load a Django template
When my theme has custom templates (let's say in `themes/my_theme/templates/custom/`), I may want to use them in the CMS. 

But I should not be able to use `{% include %}` because that would let anyone with CMS access query lots of non-public data, including some that they should not have access to.

- Create a dropdown on the CMS page form that lets you select a template from a pre-approved list, such as all the templates in `templates/custom`. When selected, this template would be rendered as the body of the page, rather than the text in the body field.
- Provide some way of turning this feature on or off, or specifying the path to search for custom templates, like a Django setting

joemull:
Closes #3794


src/cms/forms.py
```diff
@@ -20,11 +23,32 @@ class PageForm(BleachableModelForm, JanewayTranslationModelForm):
...
+
         super(PageForm, self).__init__(*args, **kwargs)
 
+        if self.instance:
+            journal_type = ContentType.objects.get(app_label="journal", model="journal")
+            if self.instance.content_type == journal_type:
+                self.journal = journal_type.get_object_for_this_type(pk=self.instance.object_id)
+            press_type = ContentType.objects.get(app_label="press", model="press")
+            if self.instance.content_type == press_type:
+                self.press = press_type.get_object_for_this_type(pk=self.instance.object_id)
```
ajrbyers, line 41: These queries may run eeven when `self.journal` and `self.press` are set at the start of `__init__`.

ajrbyers (changes requested):
One comment added, otherwise very good!

src/cms/forms.py
joemull: Yeah that's by design. I can add a comment.

src/cms/views.py
```diff
@@ -106,7 +106,9 @@ def view_page(request, page_name):
                                      content_type=request.model_content_type,
                                      object_id=request.site_type.pk)
 
-    if request.journal:
+    if current_page.template:
+        template = current_page.template
```
mauromsl, line 110: State in the database and actual theme code can diverge now, we should check page exists before trying to render (when not present, return 404 to user and log an error)

mauromsl (changes requested):
Looks good, I added a minor comment and I think it would be good to add a TestCase of a CMS page using a custom template

src/cms/views.py
joemull: Good call. I've added a check for the template path in the view, and tests to make sure the view logic works.

mauromsl (approved):

PR #3812, opened 2023-12-07
Management command for sending emails

### LINKED ISSUE #3809: Management command for sending email
Administrators may need to send an email outside the workflow.

A management command that makes it easy to populate and send an email at the command line

joemull:
Closes #3809.

### DIFF
--- a/src/utils/management/commands/send_email.py
+++ b/src/utils/management/commands/send_email.py
@@ -13,7 +13,6 @@ from utils import models as utils_models, setting_handler
 from utils.notify_helpers import send_email_with_body_from_user
 from utils.logger import get_logger
 from cron.models import Request
-from argparse import BooleanOptionalAction
 
 logger = get_logger(__name__)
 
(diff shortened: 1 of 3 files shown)


src/utils/management/commands/send_email.py
```diff
@@ -0,0 +1,273 @@
...
+            request.site_type = journal
+        elif repository:
+            request.site_type = repository
+        else:
+            request.site_type = press
+
+        subject = options['subject']
+        to = options['to']
+
+        body_file = os.path.join(settings.BASE_DIR, options['body'])
```
mauromsl, line 146: This is a bit weird, usually with command line programs you would expect the relative path to be calculated from the current working directory (default behaviour of Python's `open()`)

src/utils/management/commands/send_email.py
```diff
@@ -0,0 +1,273 @@
...
+                    request,
+                    subject,
+                    to,
+                    body,
+                    log_dict=log_dict,
+                    cc=cc,
+                    bcc=bcc_chunk,
+                )
+
+        logger.info(self.style.SUCCESS('Email sent'))
```
mauromsl, line 260: can we add a short sleep timer (a fraction of a second) here in between batches? It is  particularly important when using email backends like Mailgun, to make sure we don't overrun limits

mauromsl (changes requested):
Thanks Joe, this will be very useful I'm sure. A couple of suggestions in line

ajrbyers (approved):

PR #3820, opened 2023-12-08
Breadcrumbs added to unassigned templates

### LINKED ISSUE #3797: missing breadcrumbs on unassigned articles page
**Describe the bug**
The breadcrumbs at the top of the unassigned articles page show "Dashboard" only instead of "Dashboard | Unassigned Articles".

**Janeway version**
Found using Playtest.  So 1.5.0

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to Dashboard
2. Click on Unassigned Articles
3. Look at second line from the top, just has a link to 'dashboard' instead of the full breadcrumb.

**Expected behaviour**
Dashboard [link] / Unassigned Articles [link]

## Front-end Issues

 - OS: MacOS
 - Browser: Same in both Safari and Chrome

**Additional context**
I think I've seen similar breadcrumb issues on other pages too, so the problem may be more widespread.  Also worth noting while handling this, that the current page is listed as a link in all breadcrumbs (so on Dashboard the breadcrumb is just Dashboard and clicking it reloads the pages) so that clicking the final part of the breadcrumb results in a refresh of the page.  I wonder whether it would be better for the current page name to be displayed as text only.

StephDriver:
created an unassigned_base 

assigned across unassigned article and the unassigned article \ projected issue pages.

Closes #3797

### DIFF
--- a/src/templates/admin/elements/breadcrumbs/unassigned_base.html
+++ b/src/templates/admin/elements/breadcrumbs/unassigned_base.html
@@ -1,0 +1,1 @@
+<li><a href="{% url 'review_unassigned' %}">Unassigned articles</a></li>
--- a/src/templates/admin/review/projected_issue.html
+++ b/src/templates/admin/review/projected_issue.html
@@ -7,6 +7,13 @@
 {% block title-section %}Projected Issue{% endblock %}
 {% block title-sub %}#{{ article.pk }} / {{ article.correspondence_author.last_name }} / {{ article.safe_title }}{% endblock %}
 
+{% block breadcrumbs %}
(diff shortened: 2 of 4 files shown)


Bbkctp:

joemull (approved):
This looks great to me. I tested it out and it much improves the breadcrumbs you see when viewing unassigned articles.

It's worth noting that you can access the projected issue pane from across the workflow (not just from unassigned) by viewing the article metadata and then following a link. But that's OK in my view because unassigned is the conceptual "home" of projected issues. The URL also always has "unassigned" as part of the path.

PR #3823, opened 2023-12-11
#3382 author dashboard update to link to article history

### LINKED ISSUE #3382: Allow authors to browse peer review reports after their articles are published
The author dashboard page presents a link to the article status page whilst the article is moving through the workflow. However, once the article is published, the button is no longer rendered and there are no links for the author to revisit this page. This rendered them unable to check the peer review reports and other editorial information after the article is published

Ensure the link to the author's article status page is always available from their dashboard

**Additional context**
image

StephDriver:
After publication, both the live article and the article history (e.g. peer reviews) should be accessible from the author dashboard.

Updated the author dashboard:
- column heading 'View Live' changed to 'View
- button 'View Live' changed to 'Live'
- button 'History' added before 'Live' in 'View' Column which links to same page as 'View Status' does before publication (copy/paste of code for View Status).

closes #3382

### DIFF
--- a/src/templates/admin/elements/core/author_dashboard.html
+++ b/src/templates/admin/elements/core/author_dashboard.html
@@ -99,7 +99,8 @@
                             <td>{{ article.get_stage_display }}</td>
                             <td>{{ article.date_published }}</td>
                             <td>
-                                <!--same link as "View Status" above for unpublished articles so history, peer reviews etc can be viewed -->
+                                {% comment "same link as "View Status" above for unpublished articles so history, peer reviews etc can be viewed" %}
+                                {% endcomment %}
                                 <a class="small button" href="{% url 'core_dashboard_article' article.pk %}">History</a> 
                                 {% for revision in article.active_revision_requests %}
                                     {% if not revision.date_completed %}
@@ -108,7 +109,8 @@
                                     {% endif %}
                                 {% endfor %}
                                             
-                                <!--links to published article-->
+                                {% comment "links to published article" %}
+                                {% endcomment %}
                                 <a class="small button" href="{{ article.url }}">Live</a>
                             </td>
                         </tr>
--- a/src/templates/admin/elements/core/author_dashboard.html
+++ b/src/templates/admin/elements/core/author_dashboard.html
@@ -89,7 +89,7 @@
                         <th>Title</th>
                         <th>Current Stage</th>
                         <th>Date Published</th>
-                        <th>View Live</th>
+                        <th>View </th>
                     </tr>
                     </thead>
                     <tbody>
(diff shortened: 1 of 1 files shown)


ajrbyers (changes requested):
I think in this instance the comments can be removed but this looks good otherwise.

StephDriver:
comments removed as requested

ajrbyers (approved):

mauromsl:
jenkins: add to whitelist

src/templates/admin/elements/core/author_dashboard.html
```diff
@@ -99,7 +99,14 @@ <h2>Published Articles</h2>
...
                             <td>{{ article.date_published }}</td>
                             <td>
-                                <a class="small button" href="{{ article.url }}">View Live</a>
+                                <a class="small button" href="{% url 'core_dashboard_article' article.pk %}">History</a> 
+                                {% for revision in article.active_revision_requests %}
+                                    {% if not revision.date_completed %}
+                                        <a class="small button" href="{% url 'do_revisions' article.pk revision.pk %}">Revision
+                                            Request</a>
+                                    {% endif %}
+                                {% endfor %}
```
joemull, line 108: I think this is not the right thing to check for and display here. It just checks for in-progress revision requests, but in most cases there won't be any of those by the time the article is published. 

It should get completed requests, i.e. `article.completed_revision_requests`.

I think it should also check for completed copyediting reviews, via a new function that doesn't yet exist called `article.completed_author_copyedits`, based on `article.active_author_copyedits`. Together these will mirror what is happening here for active requests:

https://github.com/BirkbeckCTP/janeway/blob/bc2635fab46f45873860c78a2324f241b0bfa13f/src/templates/admin/elements/core/author_dashboard.html#L59-L69

However, the views controlling those links expect that the author only ever views an active request. They raise a 404 if the request has been completed:

https://github.com/BirkbeckCTP/janeway/blob/3f49d4b36d19a7591452461837b498e232b3e2a6/src/review/views.py#L1942-L1948

https://github.com/BirkbeckCTP/janeway/blob/3f49d4b36d19a7591452461837b498e232b3e2a6/src/copyediting/views.py#L718-L723

So we need to modify those views and think through what the author should see when viewing a completed request.

The one other complication is there could be more than one revision request or more than one copyediting review, so we need to alter the labels so the user can tell them apart. Maybe add the ID of the review after the label?

image

joemull (changes requested):
The History and Live buttons are working but the review link is not: 

image

This gets to the larger question of what the author should be able to see, because so far we have only expected authors to view assignments in progress, not retrospectively. 

I think this feature may need to go back to backlog refinement so we can decide what the author should be able to see after assignments are closed. I think there's a case to be made that they should see the full history of assignments, but that will require editing each related view and template to work out the display of past information. More detail inline.

joemull:
> I think this feature may need to go back to backlog refinement so we can decide what the author should be able to see after assignments are closed. I think there's a case to be made that they should see the full history of assignments, but that will require editing each related view and template to work out the display of past information.

On second thought maybe one for design?

joemull:
@ajrbyers we discussed this in backlog refinement and want to send it back to you for another review. Do you agree it might be best to close this PR until we can redefine the issue?

ajrbyers:
> @ajrbyers we discussed this in backlog refinement and want to send it back to you for another review. Do you agree it might be best to close this PR until we can redefine the issue?

Sure. I've shifted the original Issue to Needs Design in Triage.

PR #3824, opened 2023-12-11
3491 moves the generation of the default review form to a method of the Journal object so that it can be easily called.

ajrbyers:


src/journal/models.py
```diff
@@ -511,6 +512,25 @@ def description_for_press(self):
...
+        main_element = review_models.ReviewFormElement.objects.create(
+            name='Review',
+            kind='textarea',
+            required=True,
+            order=1,
+            width='large-12 columns',
+            help_text=gettext('Please add as much detail as you can.'),
+        )
+
+        default_review_form.elements.add(main_element)
```
mauromsl, line 532: Should we use `.get_or_create()` to avoid duplicates and having to be overly defensive when calling this?

mauromsl (changes requested):

PR #3831, opened 2023-12-12
#2123 news item template tag filters updated for HTML

### LINKED ISSUE #2123: Hyperlinks inserted in news items are not rendered
**Describe the bug**
Announcements added through the news manager are being rendered as text rather than HTML, so hyperlinks are stripped out

**Janeway version**
v1.3.9

**To Reproduce**
1. Go to the news manager
2. Add a news item with a hyper link in the body
3. Go to the news page

**Expected behavior**
Since the input is  a rich-text box, hyperlinks and other HTML tags should be rendered on the news page and home page.

StephDriver:
identified templates as core/news/index.html across themes OLH, Material and Clean.

previously used striptag and truncatesmart

updated to truncatewords_html and safe

truncatesmart was set at 400 char, have used 50 words as an equivalent.

there are other templates which also use striptag and truncatesmart, these may need to considered for a similar update.

Closes #2123

### DIFF
--- a/src/themes/clean/templates/core/news/index.html
+++ b/src/themes/clean/templates/core/news/index.html
@@ -13,7 +13,7 @@
             <div class="col-md-12">
                 <h4>{{ item.title }}</h4>
                 <h6>{{ item.byline }} on {{ item.posted|date:"Y-m-d" }}</h6>
-                <p>{{ item.body|striptags|truncatesmart:400 }}</p>
+                <p>{{ item.body|truncatewords_html:50|safe }}</p>
                 <a href="{% url 'core_news_item' item.pk %}">{% trans 'Read More' %}</a>
                 <hr/>
             </div>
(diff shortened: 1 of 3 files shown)


ajrbyers (approved):

ajrbyers:
Thanks @StephDriver, this looks good. Over to @joemull for second review.

src/themes/clean/templates/core/news/index.html
```diff
@@ -13,7 +13,7 @@ <h1>{{ journal_settings.news.news_title }}</h1>
             <div class="col-md-12">
                 <h4>{{ item.title }}</h4>
                 <h6>{{ item.byline }} on {{ item.posted|date:"Y-m-d" }}</h6>
-                <p>{{ item.body|striptags|truncatesmart:400 }}</p>
+                <p>{{ item.body|striptags|truncatewords_html:50 }}</p>
```
ajrbyers, line 16: striptags should be dropped and safe put in its place as with others.

ajrbyers (changes requested):
See comment.

src/themes/clean/templates/core/news/index.html
StephDriver: good catch.  done & tested.

ajrbyers: Nice!

ajrbyers (approved):

joemull:
@StephDriver I think our preferred workflow would be to rebase this one on the command line as well, so perhaps we can handle it at the same time as #3862?

joemull (approved):

PR #3837, opened 2023-12-13
#3836 create accessiblity issue reporting template

### LINKED ISSUE #3836: create accessibility issue template
Accessibility issues should be reported and easily identified for inclusion in Accessibility Roadmap planning.

A new template, connecting to the a11y tag, for the reporting of accessibility issues

StephDriver:
new template for accessibility issue reporting.

using the a11y tag

based on bug report template

asks for missing alt-text not to be reported at present unless part of the alt text audit (to reduce likelihood of multiple reports - better to do this as one audit than accept these issue individually).

Closes #3836

### DIFF
--- a/.github/ISSUE_TEMPLATE/a11y_report.md
+++ b/.github/ISSUE_TEMPLATE/a11y_report.md
@@ -5,11 +5,11 @@ title: ''
 labels: a11y
 assignees: ''
 ---
-## Report Issue
-**Describe the inaccessible feature**
+## Report Accessibility Issue
+### Describe the inaccessible feature
 A clear and concise description of what cannot be accessed.
 
-**Which users are impacted?**
+### Which users are impacted?
  (delete as applicable)
  1. Everyone
  2. Readers (i.e. those accessing the published site but not logging in to Janeway)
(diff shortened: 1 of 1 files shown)


StephDriver:
future update of a11y template as issue #3839

.github/ISSUE_TEMPLATE/a11y_report.md
```diff
@@ -0,0 +1,64 @@
+---
+name: Accessibility Report
+about: Tell us about something that is not accessible.
+title: ''
+labels: a11y
+assignees: ''
+---
+## Missing Alt-text
+Please do not currently report missing alt-text for images unless you are part of the alt-text audit.
```
mauromsl, line 9: I would remove this for now. It would be good to have all known instances of alt-text missing from images logged in GitHub and users can use the search function to see if theirs has been reported already

mauromsl (commented):
Thanks @StephDriver , just one comment regarding the alt-text guidance

.github/ISSUE_TEMPLATE/a11y_report.md
StephDriver: This is only intended to be there a few weeks while I make an initial list of the missing alt-text.  It's to save *me* from having to double check every line of that list against any reports so that I can have a more efficient workflow while creating that first list.  After that, I absolutely agree this should be removed.

ajrbyers (approved):

mauromsl (approved):

PR #3852, opened 2023-12-19
PreprintFile download_url should point to the public url

myucekul:
repository_download_file is manager only view whereas repository_file_download is public view. Updating the download url so that correct item url in OAI and JATS metadata for repository items.

### DIFF
--- a/src/repository/models.py
+++ b/src/repository/models.py
@@ -817,7 +817,7 @@ class PreprintFile(models.Model):
 
     def download_url(self):
         return reverse(
-            'repository_download_file',
+            'repository_file_download',
             kwargs=self.reverse_kwargs(),
         )
 


Bbkctp:

hardyoyo (commented):
Obviously correct.

joemull (commented):
I wonder if it's worth changing the name of the function as well to make it more obviously distinct from `repository_download_file`?

ajrbyers:
> I wonder if it's worth changing the name of the function as well to make it more obviously distinct from `repository_download_file`?

I'ved opened up an issue to look at renaming one or both of the download views for preprint files (https://github.com/BirkbeckCTP/janeway/issues/4023).

joemull (approved):

ajrbyers (approved):

Issue #3857, opened 2024-01-03
many breadcrumbs displaying current page as link


StephDriver:
**Describe the bug**
Breadcrumb at top of page is showing current page as a clickable link, which then refreshes current page.

**To Reproduce**
see example at [journal]/cms/
sitewide issue

**Expected behaviour**
Breadcrumbs should be clickable for higher levels, but current page should be plain, making clear that it is the current page.

**Screenshots**
Example of correct breadcrumb at [journal]/publish/
<img width="531" alt="image" src="https://github.com/BirkbeckCTP/janeway/assets/5330770/36eca7a4-fdb0-4ce9-a84e-1fca0b60e7ec">

Example of incorrect breadcrumb at [journal]/cms/:
<img width="394" alt="image" src="https://github.com/BirkbeckCTP/janeway/assets/5330770/730272e0-0fef-4f2c-bd2c-cde3abfe28b4">

**Additional Context**

breadcrumb investigation grew out of issue #3797 
More pages affected than unaffected.  Details will be clear from the commit(s).

PR #3861, opened 2024-01-05
3857 breadcrumb remove currentpage href

StephDriver:
I have worked through the known areas of this issue one directory at a time, and each has its own commit on this branch.  Then there will be a final commit coming which tidies up any typos and trailing </a> tags that were missed.

The breadcrumbs have been implemented in three different ways - 
1. some were fully within the template itself, 
2. some used an include to a breadcrumb template within elements/breadcrumbs, 
3. and some used the include while listing the current page after it within the main template.

This fix was about making sure that final item in the list was not a link.

For 1.  there was a simple fix of removing the `<a>` tags from the final item.

For 2 and 3, Where an include was used, it needed to work for templates that only used the include, and those which had a final line in the template - i.e. sometimes return a list of links, and sometimes a list where the last item was not a link.  some of the time the problem was within the include, and then within the same section there might be some templates which used the include and then listed the current page with or without <a> tags in the main template.  Fixing this across each section required the some includes to be updated and some of the main templates, and some templates then worked without modification due to the updated includes.  Where there was a subpage listed within the main template, I have updated the includes with the suffix `with subpage="Yes"`so that the include can then display all items as links.

This is mostly done, but given how many changes were made, I will now review the changes on github so that I can check for any typos or trailing `</a>` that have not been caught already.

### DIFF
--- a/src/templates/admin/core/manager/settings/edit_setting.html
+++ b/src/templates/admin/core/manager/settings/edit_setting.html
@@ -8,12 +8,11 @@
 
 {% block breadcrumbs %}
     {{ block.super }}
+    <li><a href="{% url 'core_manager_index' %}">Manager</a></li>
     {% if request.GET.email_template %}
-        <li><a href="{% url 'core_manager_index' %}">Manager</a></li>
         <li><a href="{% url 'core_email_templates' %}">Email Templates</a></li>
     {% else %}
         {% if request.journal %}
-            <li><a href="{% url 'core_manager_index' %}">Manager</a></li>
             <li><a href="{% url 'core_settings_index' %}">Settings</a></li>
         {% else %}
             <li><a href="{% url 'core_default_settings_index' %}">Settings</a></li>
--- a/src/templates/admin/core/manager/settings/edit_setting.html
+++ b/src/templates/admin/core/manager/settings/edit_setting.html
@@ -11,21 +11,15 @@
     {% if request.GET.email_template %}
         <li><a href="{% url 'core_manager_index' %}">Manager</a></li>
         <li><a href="{% url 'core_email_templates' %}">Email Templates</a></li>
-        <li><a href="{% url 'core_manager_index' %}">Manager</a></li>
     {% else %}
         {% if request.journal %}
             <li><a href="{% url 'core_manager_index' %}">Manager</a></li>
             <li><a href="{% url 'core_settings_index' %}">Settings</a></li>
-            <li>
-                <a href="{% url 'core_edit_setting' setting.group.name setting.name %}">Edit {{ setting.pretty_name }}</a>
-            </li>
         {% else %}
             <li><a href="{% url 'core_default_settings_index' %}">Settings</a></li>
-            <li>
-                <a href="{% url 'core_edit_default_setting' setting.group.name setting.name %}">Edit {{ setting.pretty_name }}</a>
-            </li>
         {% endif %}
     {% endif %}
+    <li>Edit {{ setting.pretty_name }}</li>
 {% endblock %}
 
 {% block body %}
--- a/src/templates/admin/core/manager/users/edit.html
+++ b/src/templates/admin/core/manager/users/edit.html
@@ -12,7 +12,7 @@
     {{ block.super }}
     <li><a href="{% url 'core_manager_index' %}">Manager</a></li>
     <li><a href="{% url 'core_manager_users' %}">Users</a></li>
-    <li>{% if active == 'add' %}Add New User{% else %}Edit User{% endif %}</li>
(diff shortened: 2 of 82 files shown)


StephDriver:
4 typos caught and that's done.

Closes #3857

src/templates/admin/core/manager/users/history.html
```diff
@@ -6,7 +6,7 @@
     {{ block.super }}
     <li><a href="{% url 'core_manager_index' %}">Manager</a></li>
     <li><a href="{% url 'core_manager_users' %}">Users</a></li>
-    <li>History</li>
+    <li>History of {{user.first_name}} {{user.last_name}} </li>
```
joemull, line 9: This time `user` is the right variable because the default Django `user` is overwritten in the view's `context` with the appropriate account:

https://github.com/BirkbeckCTP/janeway/blob/dac0e5a703b5a68f67946b23cd385df90bf959e0/src/core/views.py#L1241-L1273

joemull (changes requested):
This is great--thank you for going through all these so carefully. I've checked about half of them manually and generally things are working really well. I think it's very close to ready to be merged. 

I noticed a few tiny things that must have been bugs before your changes, but they jumped out to me in reviewing your changes. Also there's one place that the template variables for user names need to be different, because of some inconsistency in how we set them in the views. All the details in comments inline.

joemull:
One other thing I forgot to mention: because we use the [typesetting plugin](https://github.com/BirkbeckCTP/typesetting) as a normal part of the standard workflow (and are planning to integrate it) we should at some point soon make the same changes there so that the workflow is consistent.

src/templates/admin/core/manager/settings/edit_setting.html
```diff
@@ -11,21 +11,15 @@
     {% if request.GET.email_template %}
         <li><a href="{% url 'core_manager_index' %}">Manager</a></li>
         <li><a href="{% url 'core_email_templates' %}">Email Templates</a></li>
-        <li><a href="{% url 'core_manager_index' %}">Manager</a></li>
     {% else %}
         {% if request.journal %}
             <li><a href="{% url 'core_manager_index' %}">Manager</a></li>
```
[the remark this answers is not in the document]
StephDriver: done

src/templates/admin/core/manager/users/edit.html
```diff
@@ -12,7 +12,7 @@
     {{ block.super }}
     <li><a href="{% url 'core_manager_index' %}">Manager</a></li>
     <li><a href="{% url 'core_manager_users' %}">Users</a></li>
-    <li>{% if active == 'add' %}Add New User{% else %}Edit User{% endif %}</li>
+    <li>{% if active == 'add' %}Add User{% else %}Edit {{user.first_name}} {{user.last_name}} {% endif %}</li>
```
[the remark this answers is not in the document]
StephDriver: great catch!  Updated to `user_to_edit.full_name` as suggested.

This then looked inconsistent with the body title areas which had been in the format `Edit <user email address>` so I have updated that to `Edit <user name> ( <user email address> )`.

image

StephDriver:
changes made - please re-review

src/templates/admin/journal/manage/issue_add_article.html
```diff
@@ -8,7 +8,8 @@
 
 {% block breadcrumbs %}
     {{ block.super }}
-    {% include "elements/breadcrumbs/issue_management.html" %}
+    {% include "elements/breadcrumbs/issue_management.html" with subpage="yes"%}
+    <li> Add Article </li>
```
ajrbyers, line 12: There is some whitespace in this `<li>`

src/templates/admin/submission/manager/delete_license.html
```diff
@@ -10,9 +10,12 @@
 {% block breadcrumbs %}
     {{ block.super }}
     <li><a href="{% url 'core_manager_index' %}">Manager</a></li>
-    <li><a href="{% url 'submission_licenses' %}">Licences</a></li>
     {% if license %}
```
ajrbyers, line 13: `license` will always be present on this template so we can remove the check here.

ajrbyers (changes requested):
This is great, I've added two comments for you to go over.

src/templates/admin/journal/manage/issue_add_article.html
StephDriver: fixed

src/templates/admin/submission/manager/delete_license.html
StephDriver: logic for if license removed.

ajrbyers (approved):

PR #3862, opened 2024-01-05
3858 missing breadcrumb fix

### LINKED ISSUE #3858: Missing Breadcrumbs
**Describe the bug**
Breadcrumbs are just displaying the default "dashboard" and not the actual path to the current page.

**To Reproduce**
Steps to reproduce the behaviour:
see example at dashboard/active/
sitewide issue

**Expected behaviour**
example of correct behaviour (proofing/requests/corrections/)
image

example of issue (dashboard/active):
image

**Additional context**
breadcrumb investigation grew out of issue #3797
More pages affected than unaffected. Details will be clear from the commit(s).

StephDriver:
added basic breadcrumbs to the templates as shown in the commits.  One commit per section.

Closes #3858

### DIFF
--- a/src/templates/admin/journal/article_log.html
+++ b/src/templates/admin/journal/article_log.html
@@ -7,7 +7,8 @@
 
 {% block breadcrumbs %}
     {{ block.super }}
-    <li> {{article.safe_title}} Log</li>
+    <li> Log </li>
+    <li> {{article.safe_title}}</li>
 {% endblock %}
 
 {% block body %}
--- a/src/templates/admin/journal/article_log.html
+++ b/src/templates/admin/journal/article_log.html
@@ -5,6 +5,11 @@
 {% block title-section %}Article {{ article.pk }} Log{% endblock %}
 {% block title-sub %}{{ article.safe_title }}{% endblock %}
 
+{% block breadcrumbs %}
+    {{ block.super }}
+    <li> {{article.safe_title}} Log</li>
+{% endblock %}
+
 {% block body %}
     <div class="box">
         <div class="title-area">
--- a/src/templates/admin/core/workflow.html
+++ b/src/templates/admin/core/workflow.html
@@ -4,6 +4,11 @@
 {% block title %}Journal Workflow{% endblock title %}
 {% block title-section %}Journal Workflow{% endblock title-section %}
 
+{% block breadcrumbs %}
+    {{ block.super }}
+    <li>Journal Workflow</li>
+{% endblock %}
+
 {% block body %}
     <div class="large-6 columns">
         <div class="box">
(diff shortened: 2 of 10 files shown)


ajrbyers:
Test this please

src/templates/admin/journal/article_log.html
```diff
@@ -5,6 +5,11 @@
 {% block title-section %}Article {{ article.pk }} Log{% endblock %}
 {% block title-sub %}{{ article.safe_title }}{% endblock %}
 
+{% block breadcrumbs %}
+    {{ block.super }}
+    <li> {{article.safe_title}} Log</li>
```
joemull, line 10: What if this breadcrumb were Dashboard / Logs / [Article title]? As it is it might read like "log" is a part of the article title.

joemull (approved):
Looks good, just one comment inline.

Also I noticed some of these changes are duplicates of changes in #3820, but we can handle it during merging. Whichever is merged later will just need to be worked on to resolve conflicts or rebase it before merging.

ajrbyers (approved):
I'll second Joe's suggestion for that breadcrumb. Otherwise looks good.

joemull:
@StephDriver this one now has conflicts due to the other one having been merged before it. Is that something you're comfortable with, or would you like to pair program fixing the conflicts? In any case I am going to put a status of Changes Requested on it.

src/templates/admin/journal/article_log.html
StephDriver: good point!

It can't be Dashboard / Logs / [Article title] because that messes with the expected breadcrumb format where each  part between the / is expected to be a link to a page.  And making 'logs' not a link would cause confusion - but I do agree it could cause confusion in current form.  I'll need to rethink and try a few alternatives.

StephDriver: This is a tricky one.  There is no good solution in current form - putting Log in as a previous breadcrumb when there is no Log page is a problem, likewise putting Log in as current could confuse the article title. 

But a full review of breadcrumbs and update is already being planned as part of a11y work.  So we need a 'good enough for now' solution that causes the least confusion.  

I am therefore going to put in a "Logs" breadcrumb that is not a hyperlink.  This means the article title is non distorted, and its clear there is no such page as it is not clickable so at least users won't be hunting for the page.  Its inelegant as this implies there should be a logs page in the structure when there isn't one - but seems the best solution for now given the pending breadcrumb review.

StephDriver: image

StephDriver:
> @StephDriver this one now has conflicts due to the other one having been merged before it. Is that something you're comfortable with, or would you like to pair program fixing the conflicts? In any case I am going to put a status of Changes Requested on it.

changes made - is this now suitable for merging? @joemull

joemull (approved):

PR #3891, opened 2024-01-12
#3785 bugfix - control user if statement contents updated

### LINKED ISSUE #3785: Control user button issue
Reported by Longleaf (ticket #7466)

"When I go into the Manager role and to the "Enrolled Users" area of any journal, the "Control User" button in the first row of the list of users appears to be functioning as the "Add Users to Role" button. When I clicked "Control User" in the row for a user (see screenshot below), I got the error message "No users selected," which is what happens if I click "Add Users to Role" without selecting a box. 

The "Control User" buttons for all other rows work as expected. It's just happening in the first row. Could you please look into this apparent bug?
image

StephDriver:
problem was with forms not communicating correctly with each other.  Updated with code from @ajrbyers to fix this.  Tested the fix.

Closes #3785

### DIFF
--- a/src/templates/admin/core/manager/users/index.html
+++ b/src/templates/admin/core/manager/users/index.html
@@ -64,13 +64,10 @@
                                        href="{% url 'core_user_edit' user.id %}">Edit</a></td>
                                 {% if settings.HIJACK_USERS_ENABLED and request.user.is_superuser %}
                                     <td>
-                                        <form action="{% url 'hijack:acquire' %}" method="POST">
-                                          {% csrf_token %}
-                                          <input type="hidden" name="user_pk" value="{{ user.pk }}">
-                                          <button type="submit" class="tiny button"> Control User </button>
-                                          <input type="hidden" name="next" value="{% url 'core_dashboard' %}">
-                                        </form>
-                                    </td>{% endif %}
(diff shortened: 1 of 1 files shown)


ajrbyers (approved):

mauromsl (commented):
@StephDriver : While I understand the changes, I'm struggling to grasp why it resolves the reported problem. Could you comment on how that's the case? the previous nested form had the same action as this button, with the same inputs being posted

StephDriver:
My notes on this solution are 'paste in Andy's code'- it was an issue I spent a lot of time investigating, could replicate but couldn't quite understand why it was happening, then I pair-programmed it with Andy and I'll admit I still don't understand what it was all about - but I ran the same tests after his changes and the problem had gone away.  So over to you to explain further @ajrbyers .

ajrbyers:
> @StephDriver : While I understand the changes, I'm struggling to grasp why it resolves the reported problem. Could you comment on how that's the case? the previous nested form had the same action as this button, with the same inputs being posted

You can revert it and take a look but for some reason we couldn't get to the bottom of the first button in the list wouldn't fire the hijack process.

Anyway - nested `<form>` tags are not allowed in HTML5 and this does resolve the issue we saw.

mauromsl (approved):

PR #3894, opened 2024-01-16
#3621 added a new get_reset_token form

### LINKED ISSUE #3621: Email address not validated in password reset form
## User story

If I want to reset my password, I'm first asked to enter my email address. If there is a match, I receive a reset token by email.

However, if I accidentally include a space before or after my email address, the string does not match, I do not get a reset token, and cannot reset my password.

Currently neither the view nor the template validates the email address:

https://github.com/BirkbeckCTP/janeway/blob/2fb70af7d19516df3136cafeb0f99516a566cb8b/src/core/views.py#L229-L234

https://github.com/BirkbeckCTP/janeway/blob/2fb70af7d19516df3136cafeb0f99516a566cb8b/src/themes/OLH/templates/core/accounts/get_reset_token.html#L27

https://github.com/BirkbeckCTP/janeway/blob/2fb70af7d19516df3136cafeb0f99516a566cb8b/src/themes/material/templates/core/accounts/get_reset_token.html#L16

https://github.com/BirkbeckCTP/janeway/blob/2fb70af7d19516df3136cafeb0f99516a566cb8b/src/themes/clean/templates/core/accounts/get_reset_token.html#L17

## Proposed solution

Validate the email address in the HTML using the "email" type:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email

## Other solutions

Create a Django form that also validates the email on the server.

StephDriver:
created a new GetResetToken form and linked through to each of the three themes.

Closes #3621


src/core/views.py
```diff
@@ -224,20 +224,30 @@ def get_reset_token(request):
...
-            return redirect(reverse('core_login'))
+        form = forms.GetResetToken(
+            request.POST,
+        )
+        if form.is_valid():
+            email_address = form.cleaned_data.get("email_address")
+            messages.add_message(
+                request, 
+                messages.INFO,
+                'If your account was found, an email has been sent to you.',
```
ajrbyers, line 238: We should mark this text for translation as we've done with the email label in forms.py.

ajrbyers (changes requested):

src/core/views.py
StephDriver: I've updated this, but am struggling to change the language in order to test it.  I can confirm the change I've made doesn't seem to affect the English version, but I can't work out whether it is functional in terms of translation.

ajrbyers: > I've updated this, but am struggling to change the language in order to test it. I can confirm the change I've made doesn't seem to affect the English version, but I can't work out whether it is functional in terms of translation.

In this case that is enough, it wont have a translation so will always show the English version until that is loaded in.

StephDriver:
superceded by 3621-password_reset_email_validation_rebased branch on PR #4177

PR #3901, opened 2024-01-23
Press editorial boards and staff groups

### LINKED ISSUE #3895: Press-level editorial board
Extend the editorial team model to the press level. This is different from the solution to #3795.

Add a ForeignKey to press (suggested by AB).

joemull:
Closes #3895 

I've based this branch on the one for #3806 so that the diffs make sense.


joemull:
Jenkins: test this please

joemull:
Jenkins: test this please

joemull:
Alright this one's ready @mauromsl. It includes things you requested in your review of #3806.

src/journal/views.py
```diff
@@ -1897,17 +1897,26 @@ def contact(request):
...
-        editorial_groups = core_models.EditorialGroup.objects.filter(journal=request.journal)
+        kwargs['pk'] = group_id
 
-    template = 'journal/editorial_team.html'
+    editorial_groups = core_models.EditorialGroup.objects.filter(**kwargs)
+
+    if request.journal:
+        template = 'journal/editorial_team.html'
+    else:
+        template = 'press/editorial_team.html'
```
mauromsl, line 1919: I don't think we have this template still, is it in hourglass? Can we make a copy of the journal/editorial_team.html templates for Janeway core?

mauromsl (changes requested):
Very cool addition for Janeway beyond the OLH rework. I added a couple comments inline

src/core/views.py
```diff
@@ -1617,6 +1658,7 @@ def edit_editorial_group(request, group_id=None):
                 else:
                     group = form.save(commit=False)
                     group.journal = request.journal
+                    group.press = request.press
```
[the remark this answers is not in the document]
joemull: Sorry, I meant to make the press field required. I've done that now. If you could check the migration and make sure you agree with how I've done it, that would be great.

As for whether the form is valid, it has the same behavior as before. The journal field was already excluded from the form in this view's logic, even when it was required, so I have excluded the press as well.

The view is a bit convoluted, but it does always add the journal and press, either here, via the instance that is passed to the form...

https://github.com/BirkbeckCTP/janeway/blob/921328e81b22b21938f5b01a54de759fae348a30/src/core/views.py#L1637-L1638

...or here, after the form is saved:

https://github.com/BirkbeckCTP/janeway/blob/921328e81b22b21938f5b01a54de759fae348a30/src/core/views.py#L1660-L1661

If that's a problem, it was pre-existing, but we can refactor further...

src/journal/views.py
joemull: Yes, I've added redirects and tested them manually to make sure there are no breaking template variables.

src/core/models.py
```diff
@@ -1420,9 +1424,17 @@ class Meta:
 class EditorialGroup(models.Model):
     name = models.CharField(max_length=500)
     description = models.TextField(blank=True, null=True)
+    press = models.ForeignKey(
+        'press.Press',
+        on_delete=models.CASCADE,
+        blank=True,
+        null=True,
+    )
```
mauromsl, line 1432: Shouldn't this be required too? We can provide it a default to ensure it is set correctly:
```
def default_press():
    return Press.objects.first()
[...]
press = models.ForeignKey(
        'press.Press',
        on_delete=models.CASCADE,
        blank=True,
        null=True,
       default=default_press,    
```

mauromsl (changes requested):
Cheers, just one final thing regarding default values

src/core/views.py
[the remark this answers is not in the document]
mauromsl: Right, I missed we are excluding `journal` there on the form fields. I'm not a big fan of the level of indirection here (it should be the form who sets `journal` and `press` rather than the view) but as you say, it is pre-existing code. I added one

joemull:
Jenkins: test this please

mauromsl (approved):

PR #3902, opened 2024-01-23
Disable bleach

mauromsl:
A number of problems have been reported when using the new bleach feature, some of which are causing data loss, so this PR "disables" the feature in the simplest way possible by commenting out some of the code. If we wan to re-instate the feature, we just have to revert this one commit


src/press/forms.py
```diff
@@ -80,7 +80,7 @@ def save(self, commit=True):
 
 class PressJournalDescription(forms.Form):
 
-    description = BleachField(required=False)
+    # description = BleachField(required=False)
```
ajrbyers, line 83: Need to add a TextField for description here.

ajrbyers (changes requested):

joemull (approved):

PR #3908, opened 2024-01-24
#3906 un-nest labels and inputs in themes

### LINKED ISSUE #3906: form field missing a label
A11y issue because form field appears to be missing a label.  This is due to the structure in the HTML where the input tag is included within the label tag:

Example found:
``` python 
src/themes/OLH/templates/elements/journal/article_filter_form.html:
            <label for="section-{{ section.id }}"><input id="section-{{ section.id }}"
                                                         value="{{ section.id }}" type="checkbox"
                                                         name="filter[]"
                                                         {% if section.id in filters %}checked="checked"{% endif %}>{{ section.name }}
            </label>
```

The label tag should be closed before the input tag opens.

StephDriver:
all file search for labels which nested their input and only two were found, both in theme templates marked as depreciated in 1.5.1.

Labels elsewhere looked un-nested, and hence avoiding this issue.

Closes #3906

### DIFF
--- a/src/themes/OLH/templates/elements/journal/article_filter_form.html
+++ b/src/themes/OLH/templates/elements/journal/article_filter_form.html
@@ -23,7 +23,6 @@
         </label>
     </div>
     <div class="section">
-        Hi there
         <label>{% trans "Filter" %}</label>
         {% for section in sections %}
             <input id="section-{{ section.id }}"
--- a/src/themes/OLH/templates/elements/journal/article_filter_form.html
+++ b/src/themes/OLH/templates/elements/journal/article_filter_form.html
@@ -23,13 +23,15 @@
         </label>
     </div>
     <div class="section">
+        Hi there
         <label>{% trans "Filter" %}</label>
         {% for section in sections %}
-            <label for="section-{{ section.id }}"><input id="section-{{ section.id }}"
-                                                         value="{{ section.id }}" type="checkbox"
-                                                         name="filter[]"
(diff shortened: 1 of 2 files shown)


src/themes/OLH/templates/elements/journal/article_filter_form.html
```diff
@@ -23,13 +23,15 @@
         </label>
     </div>
     <div class="section">
+        Hi there
```
ajrbyers, line 26: This needs to be removed.

ajrbyers (changes requested):
This is okay, does the display remain the same after the change?

src/themes/OLH/templates/elements/journal/article_filter_form.html
StephDriver: done

ajrbyers (approved):

joemull (commented):
Just thought I'd mention this template is no longer used, to my knowledge. I added this deprecation note recently when I switched this interface to using a class-based view:

https://github.com/BirkbeckCTP/janeway/blob/9fcbbe87b917bf3c4ce1d64b85d579f7c0634069/src/themes/OLH/templates/elements/journal/article_filter_form.html#L1

Maybe I should have just deleted the template? I wasn't confident doing so without more front-end tests in place.

joemull:
@ajrbyers Brilliant! I've added an issue for it.

joemull (approved):

Issue #3949, opened 2024-02-02
Empty tag URL should raise a 404


joemull:
**Describe the bug**

Following a URL to a non-existent news tag results in an error rather than a 404
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.10/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/vol/janeway/src/comms/views.py", line 173, in news_list
    tag = models.Tag.objects.get(
  File "/usr/local/lib/python3.10/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/django/db/models/query.py", line 435, in get
    raise self.model.DoesNotExist(
comms.models.Tag.DoesNotExist: Tag matching query does not exist.
```
**Janeway version**
1.5.2

**Expected behavior**
Raise a 404

PR #3950, opened 2024-02-02
#3949 bugfix - update getting tag object in news_list() func

HamidRezaSZ:
I updated the tag query to get the object or raise 404.


joemull:
Thanks @HamidRezaSZ !

PR #3954, opened 2024-02-05
Add template field to historical page

### LINKED ISSUE #3953: Template field not include on historical Page
**Describe the bug**
We created the `Page.template` field on a different branch from the `HistoricalPage` model.

As a result we get this error:

`django.db.utils.ProgrammingError: column "template" of relation "cms_historicalpage" does not exist`

**Janeway version**
1.5.2

joemull:
Fixes #3953


src/cms/migrations/0015_historicalpage.py
```diff
@@ -20,6 +20,7 @@ class Migration(migrations.Migration):
             fields=[
                 ('id', models.IntegerField(auto_created=True, blank=True, db_index=True, verbose_name='ID')),
                 ('support_copy_paste', models.BooleanField(default=True, help_text='Turn this on if copy-pasting content from a word processor, or using the toolbar to format text. It tells Janeway to clear out formatting that does not play nice. Turn it off and leave it off if anyone has added custom HTML or CSS using the code view, since it might remove custom code.')),
+                ('template', models.CharField(blank=True, help_text='The custom template to use instead of the content field.', max_length=100)),
```
mauromsl, line 23: This migration existed in 1.5.1. If we add a field here, existing installations won't have it installed (the migration has already run) and will see the reported error. I think we need a new migration to add the field.

Altering past migrations is not something we can do for schema altering migrations

mauromsl (changes requested):

src/cms/migrations/0015_historicalpage.py
joemull: Apologies, I thought it was new to 1.5.2, but yes you are right. I'll create a new migration.

mauromsl (approved):

PR #3981, opened 2024-02-20
Stops multiple galley image files with the same name from being loaded in the edit galley interface.

### LINKED ISSUE #3979: Multiple galley images with the same name results in an error
**Describe the bug**
If an editor/typesetter/production manager uploads two images with the same name this results in the image is not displayed. 

**Janeway version**
1.5.1-RC-5

**Pontential Solutions**
- [x] For legacy we should add a handle for this in the `identifier_figure` view
- [x] When an image is uploaded that already exists, warn the user and recommend they use the replace file option
- [x] If the replace file option isn't present on this interface, add it

ajrbyers:
This closes #3979 

The implementation includes two prongs:

`save_galley_image` now takes `check_for_existing_images` which defaults to `False`. When set to `True` the function will check if the image exists and if its found return `None`. Existing calls are protected.

In addition, I've updated the Typesetting Plugin view to use the new kwarg: https://github.com/BirkbeckCTP/typesetting/pull/204

### DIFF
--- a/src/production/logic.py
+++ b/src/production/logic.py
@@ -203,8 +203,8 @@ def save_galley_image(
             messages.add_message(
                 request,
                 messages.WARNING,
-                'The file you uploaded does not match the mime of the '
-                'file expected.',
+                f'The file you uploaded does not have the expected '
+                f'type: { expected_mime }.'
             )
 
     new_file = files.save_file_to_article(uploaded_file, galley.article, request.user)
--- a/src/production/logic.py
+++ b/src/production/logic.py
@@ -174,22 +174,39 @@ def replace_galley_file(article, request, galley, uploaded_file):
 
 
 def save_galley_image(galley, request, uploaded_file, label="Galley Image", fixed=False):
-
+    filename = uploaded_file.name
     if fixed:
-        filename = request.POST.get('file_name')
         uploaded_file_mime = files.check_in_memory_mime(uploaded_file)
         expected_mime = files.guess_mime(filename)
 
         if not uploaded_file_mime == expected_mime:
-            messages.add_message(request, messages.WARNING, 'The file you uploaded does not match the mime of the '
-                                                            'file expected.')
+            messages.add_message(
+                request,
+                messages.WARNING,
+                'The file you uploaded does not match the mime of the '
+                'file expected.',
+            )
+
+    # Check if an image with this name already exists for this galley.
+    # If found, return the found image and stop the save process.
+    check_image_exists = galley.images.filter(
+        original_filename=filename,
+    )
+    if check_image_exists:
+        messages.add_message(
+            request,
+            messages.WARNING,
+            f'An image called {filename} already exists. Use the replace '
+            f'file function to upload a new version.',
+        )
+        return check_image_exists
 
     new_file = files.save_file_to_article(uploaded_file, galley.article, request.user)
     new_file.is_galley = False
     new_file.label = label
 
     if fixed:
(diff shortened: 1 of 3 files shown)


ajrbyers:
Ignore this for now, I need to alter the comment, as I swapped from .first() to .latest()

src/production/logic.py
```diff
@@ -174,22 +174,39 @@ def replace_galley_file(article, request, galley, uploaded_file):
 
 
 def save_galley_image(galley, request, uploaded_file, label="Galley Image", fixed=False):
-
+    filename = uploaded_file.name
     if fixed:
```
mauromsl, line 178: Shouldn't there be a kwarg here called `check_for_existing images` as per https://github.com/BirkbeckCTP/typesetting/pull/204 ?

mauromsl (changes requested):
I think you forgot to add support for the lastest feature

src/production/logic.py
ajrbyers: I had the commit but hadn't pushed it :D

joemull:
@ajrbyers I will review this once that commit is added because I want to test out the interface.

joemull:
Or is it now ready?

ajrbyers:
Hey @joemull, it is ready now the commit is here: https://github.com/BirkbeckCTP/janeway/pull/3981/commits/726dd480b4c449313f9608243a9b7f63803d8eb0

src/production/logic.py
```diff
@@ -173,23 +173,46 @@ def replace_galley_file(article, request, galley, uploaded_file):
...
+    if check_for_existing_images and galley.images.filter(
+        original_filename=filename,
+    ).exists():
+        messages.add_message(
+            request,
+            messages.WARNING,
+            f'An image called {filename} already exists. Use the replace '
+            f'file function to upload a new version.',
+        )
+        return
```
mauromsl, line 196: I think also changing the default behaviour to override the matched figure by filename and galley ID with a new version is a bit tidier (As opposed to having duplicate files with idential names).
I understand that the patch to `article_figure` does resolve the reported problem, however we might see the problem reoccur on other views with similar logic.

mauromsl (commented):
It fixes the reported issue, but made a suggestion on implementation that I think will lead to a cleaner filesystem, especially for cases where images are bulk-loaded

src/production/logic.py
```diff
@@ -173,23 +173,46 @@ def replace_galley_file(article, request, galley, uploaded_file):
...
         expected_mime = files.guess_mime(filename)
 
         if not uploaded_file_mime == expected_mime:
-            messages.add_message(request, messages.WARNING, 'The file you uploaded does not match the mime of the '
-                                                            'file expected.')
+            messages.add_message(
+                request,
+                messages.WARNING,
+                'The file you uploaded does not match the mime of the '
+                'file expected.',
```
joemull, line 207: This is a good opportunity to make this text a little more user-friendly. Most people don't know what MIME types are, and we also aren't showing users potentially helpful information (e.g. what the expected type is). Maybe we could do something like `f'The file you uploaded does not have the expected type: { expected_mime }.'`

joemull (changes requested):
Tested it out and I like the behavior. Just one request to improve some help text.

src/production/logic.py
ajrbyers: Thanks Joe. Will update.

joemull:
This has been superseded by #4032.

PR #3988, opened 2024-02-22
#3987 when making an accept or decline review decision the "Cancel & go back" button now returns the editor to the correct pages

### LINKED ISSUE #3987: Cancelled desk-rejection leads to review screen, not unassigned.
If you click 'Cancel and go back' after selecting a desk rejection, it will lead to the 'Review' page, rather than the 'Unassigned' page. (Janeway does indicate the article is not in review yet, so it does not change the stage.) This will be confusing for users.

**Janeway version**
Janeway1.5.1.RC4

**To Reproduce**
1. Go to any unassigned article.
2. Click on click on 'Decline Article'
3. On /review/article/[ID]/move/decline/ click 'Cancel and go back'
4. This will take the user to /review/article/[ID]/
image
Rather than review/unassigned/article/[ID]/
image

Whilst users can navigate backwards using the timeline navbar at the top, this might be confusing for less experienced editors.

**Expected behavior**
After clicking 'cancel and go back' to return to 'review/unassigned/article/[ID]/'

ajrbyers:
- When an article is Unassigned, return to Unassigned
- All others should return to the decision helper page
- Closes #3987

### DIFF
--- a/src/templates/admin/review/decision.html
+++ b/src/templates/admin/review/decision.html
@@ -56,7 +56,11 @@
                             <h4>To {{ article.correspondence_author.full_name }}</h4>
                             <h5>From {{ request.user.full_name }}</h5>
                         </div>
-                        {% url 'review_in_review' article.pk as cancel_url %}
+                        {% if article.stage == 'Unassigned' %}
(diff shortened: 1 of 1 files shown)


StephDriver (approved):
does what it says.

PR #4000, opened 2024-03-01
Call get_setting with setting.name

gamboz:
Sometimes the function `utils.settings_handler.get_setting()` is called with a wrong argument.

The function expects a string as second argument (`setting_name`), but here and there it is called with a `core.Setting` object instead. I think that everything works fine anyway because `str(Setting)` is `Setting.name`, but I thought I should mention :slightly_smiling_face: 

Here I fix a couple of calls, and I add (temporary) code to report if someone is calling with non-str argument.

### DIFF
--- a/src/utils/setting_handler.py
+++ b/src/utils/setting_handler.py
@@ -91,6 +91,13 @@ def get_setting(
     :default: If True, returns the default SettingValue when no journal specific
         value is present
     """
+    if not isinstance(setting_name, str):
+        import inspect
+        callee = inspect.stack()[1]
+        logger.warning(
+            f"utils.get_setting called by {callee.function}::{callee.lineno}"
+            f" with setting_name {str(setting_name)}",
+        )
     try:
         setting = core_models.Setting.objects.get(
             name=setting_name,
@@ -120,7 +127,7 @@ def get_setting(
                     journal = None
                     return get_setting(
                         setting_group_name,
-                        setting,
+                        setting.name,
                         journal,
                         create,
                     )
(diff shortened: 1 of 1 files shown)


src/utils/setting_handler.py
```diff
@@ -91,6 +91,13 @@ def get_setting(
     :default: If True, returns the default SettingValue when no journal specific
         value is present
     """
+    if not isinstance(setting_name, str):
+        import inspect
```
ajrbyers, line 95: Can the import statement be shifted to the header?

ajrbyers (changes requested):
Thanks @gamboz, one comment in the review.

src/utils/setting_handler.py
gamboz: Moved.

I wasn't sure if you wanted to keep this sort of workaround, so I kept everything close for easier deletion :slightly_smiling_face:

joemull (commented):
Thanks gamboz--it is helpful to have the warning for sure. We can remove it when / if we put in Python type annotations, or if we add unit tests to the functions that call `get_setting`, with assertions like `Mock.assertCalledWith`.

If we want to make it clear it is temporary, maybe we could add a comment about that fact?

gamboz:
> Thanks gamboz--it is helpful to have the warning for sure. We can remove it when / if we put in Python type annotations, or if we add unit tests to the functions that call `get_setting`, with assertions like `Mock.assertCalledWith`.
> 
> If we want to make it clear it is temporary, maybe we could add a comment about that fact?

done

ajrbyers (approved):

PR #4014, opened 2024-03-12
Adds a new setting to conceal the editor details when an author is viewing their article's metadata.

### LINKED ISSUE #3738: Switch to conceal editors' names so authors cannot see who has been assigned to look after their submission
As things stand, authors are able to find out which member of the editorial team has been assigned to “look after” their submission (i.e. give it a preliminary assessment and, if it is deemed good enough to go out for review, administer this process).  This is undesirable because our research community is small and we need to protect the identity of the individual to avoid potential bad blood and to ensure that our editors can do as they see fit (e.g. reject an article) without fear of repercussions.
 
We’d like there to be a switch at the journal level that enables us to conceal the editors’ names that authors can currently see.  It would also be good to be able to switch this on or off at the level of an individual paper, in case we decide to change our policy in future but find it necessary to conceal editors’ identities from time to time.
 
N/A
 
**Additional context**

ajrbyers:
I believe this solution covers most of the request in #3738, as Phillipa is talking about the editors section of the article metadata page. However, in discussion with @mauromsl we noted that an editor may still reveal themselves when rejecting a paper as their email address will be used in the reply-to section.

There is some discussion to be had about how we approach this, do we want to enable the anonymous rejection of papers easily? The public display of an editorial board means that either way this is only partially anonymous as an author will be aware of who the editors are generally.

Personally I think this solution is enough, as editor should be accountable for article decisions, if editors _do_ want to circumvent the reply-to setup they could make use of a generic account to desk reject papers. Open to thoughts on this one. I've assigned everyone as a review so we can chip in our two cents.

Potentially closes #3738 

For clarification this page is accessed by authors from the Dashboard (Dashboard > Active Articles > View Status) screenshots with and without edutor details below.

Without editor details:
Screenshot 2024-03-12 at 17 17 05

With editor details:
Screenshot 2024-03-12 at 17 16 51

### DIFF
--- a/src/core/logic.py
+++ b/src/core/logic.py
@@ -317,6 +317,13 @@ def get_settings_to_edit(display_group, journal, user):
                     'general',
                     'data_figure_file_submission_instructions', journal
                 ),
+            },
+            {
+                'name': 'hide_editors_from_authors',
+                'object': setting_handler.get_setting(
+                    'general',
+                    'hide_editors_from_authors', journal
+                ),
             }
         ]
         setting_group = 'general'
--- a/src/templates/admin/elements/forms/group_submission.html
+++ b/src/templates/admin/elements/forms/group_submission.html
@@ -22,7 +22,7 @@
 </div>
 
 {% include "admin/elements/forms/table_multi_select_user.html" with field=edit_form.editors_for_notification queryset=request.journal.editors %}
-
+{% include "admin/elements/forms/field.html" with field=edit_form.hide_editors_from_authors %}
 <div class="title-area">
     <h2>System Settings</h2>
 </div>
--- a/src/templates/admin/elements/metadata.html
+++ b/src/templates/admin/elements/metadata.html
@@ -185,6 +185,8 @@
     </table>
 {% endif %}
 
+
+{% if user_is_editor or not journal_settings.general.hide_editors_from_authors %}
 <div class="title-area">
     <h2>Editors</h2>
 </div>
@@ -206,6 +208,7 @@
         </tr>
     {% endfor %}
 </table>
+{% endif %}
 
 <div class="title-area">
     <h2>Projected Issue</h2>
--- a/src/templates/admin/elements/metadata.html
+++ b/src/templates/admin/elements/metadata.html
@@ -191,7 +191,7 @@
     <h2>Editors</h2>
 </div>
 {% if user_is_editor and journal_settings.general.hide_editors_from_authors %}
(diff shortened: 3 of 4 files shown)


joemull:
@ajrbyers If you've got this one running with an example, would you mind throwing a screenshot on this PR so people can see what screen the change would be on?

joemull (changes requested):
OK, this is pretty good.

I am OK with the email address solution you propose if the editors are.

One visibility logic thing I'd rather see changed is how this includes `user_is_editor` as an override. If I am an editor viewing this page, I will see the editors listed, even if the hide setting is on. This will lead me to question whether the setting is working, because I have no way to verify that the author sees something different from me. 

So, I'd suggest adding some explanatory text explaining to editors that authors cannot see the list of editors. This template is re-used in several places, some author-facing and some internal. So the text would need to work in both contexts.

BTW, this was an interesting thing for me to think through, because it goes back to the theme that came out of the copyediting UX research: "seeing what others can see" is often a pickle in Janeway. A good future candidate for a re-usable design element explaining the role visibility of a given thing.

ajrbyers:
@joemull 

> So, I'd suggest adding some explanatory text explaining to editors that authors cannot see the list of editors. This template is re-used in several places, some author-facing and some internal. So the text would need to work in both contexts.

How about wrapping the explanitory note in `{% if user_is_editor %}` so that the author doesn't have to see it?

joemull:
> How about wrapping the explanitory note in `{% if user_is_editor %}` so that the author doesn't have to see it?

Yes that's what I was imagining!

joemull:
@ajrbyers errr, actually, the visibility info text would need to be based on `{% if journal_settings.general.hide_editors_from_authors %}`, not role-based. Authors then won't ever see it, because the outer check would hide the whole thing from them.

ajrbyers:
> @ajrbyers errr, actually, the visibility info text would need to be based on `{% if journal_settings.general.hide_editors_from_authors %}`, not role-based. Authors then won't ever see it, because the outer check would hide the whole thing from them.

As it turns out, its both. Anyway, commit is in.

S-Haime (approved):
I think this sounds good as a solution, as long as we (I) document it precisely enough and make clear when what information is visible so editors can make an informed decision. I see merit in being able to fully anonymously do desk rejects in certain fields (youch!) through a generic account, but I feel like this is not a common request.

joemull (approved):

mauromsl (approved):

PR #4018, opened 2024-03-12
Enhances the reviewer role list with additional, reviewer specific detail.

### LINKED ISSUE #3039: Peer Reviewer Management/Comms
Currently there is no interface for an editor to easily manage the pool of peer reviewers other than the roles page. We should extend this.

- [x] Update the Roles > Peer Reviewer page to include more information about reviews inc last completed, number of completed, average score, interests
- [x] Add edit button to this interface so that editors can easily alter interests
- [x] Add link to history

ajrbyers:
Closes #3039 

- Adds additional information when viewing the list of reviewers
- Also enhances the standard view with history and edit buttons
- This page also now loads much faster, with queries reduced from >2000 to <500.

image

### DIFF
--- a/src/core/views.py
+++ b/src/core/views.py
@@ -1258,6 +1258,11 @@ def user_edit(request, user_id):
             form.save()
             messages.add_message(request, messages.SUCCESS, 'Profile updated.')
 
+            if request.GET.get('return'):
+                return redirect(
+                    request.GET.get('return')
+                )
+
             return redirect(reverse('core_manager_users'))
 
     template = 'core/manager/users/edit.html'
--- a/src/core/views.py
+++ b/src/core/views.py
@@ -28,8 +28,7 @@ from django.views.decorators.csrf import ensure_csrf_cookie
 from django.contrib.contenttypes.models import ContentType
 from django.utils.translation import gettext_lazy as _
 from django.utils import translation
-from django.db.models import Q
-from django.utils.decorators import method_decorator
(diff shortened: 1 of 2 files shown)


joemull (changes requested):
This is great--I can imagine how useful it will be!

A couple small changes that might be nice:
- Put a default value, like an empty string, in the template on the cells for interests, number of assignments, last completed reviews, and average score, so they don't show up as "None"
- Adjust the widths of the first three columns so they don't take up so much space, or maybe so that they adjust automatically as needed?
- Clicking on the Edit User button and then saving the user results in being sent back to a different page, the Enrolled Users list. Is that by design? Would it be possible to return users back to the same page? Or is it not worth doing since we need to rethink how these user lists relate do one another?

ajrbyers:
> Adjust the widths of the first three columns so they don't take up so much space, or maybe so that they adjust automatically as needed?

I think the issue here is that the email address won’t break because it has no spaces, it makes the email column long.

joemull:
> I think the issue here is that the email address won’t break because it has no spaces, it makes the email column wide.

It doesn't have to do with the `width` attributes here?

https://github.com/BirkbeckCTP/janeway/blob/dcb366ef3f078a1c6b809856f984994bef6dad71/src/templates/admin/core/manager/roles/role.html#L27-L29

ajrbyers:
I've updated this PR with changes. The `edit_user` view now supports a return query string parameter, empty/none values now display `--`.

joemull (approved):

PR #4022, opened 2024-03-13
Fixes an issue where multiple jump links appear in an article's workflow nav when it has been moved through the workflow in a non linear fashion

### LINKED ISSUE #3593: Moving articles non-linearly through the workflow causes navigation dupes
**Describe the bug**

Andy says this happens when "Moving articles back and forward thru the workflow.
We can update this so it’s distinct."

ajrbyers:
As `workflow_stages` is used elsewhere I've opted to add a new method that explicitly returns distinct `WorkflowElements`. This will also reduce the number of queries run as the template will no longer be fetching related items.

Closes #3593

### DIFF
--- a/src/submission/models.py
+++ b/src/submission/models.py
@@ -1680,14 +1680,8 @@ class Article(AbstractLastModifiedModel):
         return core_models.WorkflowLog.objects.filter(article=self)
 
     def distinct_workflow_elements(self):
-        workflow_element_ids = core_models.WorkflowLog.objects.filter(
-            article=self,
-        ).values_list(
-            'element'
-        ).distinct()
-
         return core_models.WorkflowElement.objects.filter(
-            pk__in=[element_id[0] for element_id in workflow_element_ids]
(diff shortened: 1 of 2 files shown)


joemull (approved):

src/submission/models.py
```diff
@@ -1679,6 +1679,17 @@ def get_subject_area(self):
...
+    def distinct_workflow_elements(self):
+        workflow_element_ids = core_models.WorkflowLog.objects.filter(
+            article=self,
+        ).values_list(
+            'element'
+        ).distinct()
+
+        return core_models.WorkflowElement.objects.filter(
+            pk__in=[element_id[0] for element_id in workflow_element_ids]
+        )
```
mauromsl, line 1691: Super minor nitpick:

Can be a bit more compact with:
```
WorkflowElement.objects.filter(
    pk__in=article.workflowlog_set.values_list("element").distinct()
)
```

mauromsl:
Super minor optional comment inline but good to merge

PR #4035, opened 2024-03-18
Flexible resizing but no fullscreen for TinyMCE editor

### LINKED ISSUE #4034: Allow horizontal resizing but not full screen toggling of TinyMCE editor
## Problem

We have currently allowed the fullscreen feature on the TinyMCE editor. It has some accessibility issues when combined with our themes. The Escape key does not allow you to exit the full-screen view on Firefox on the OLH theme or the Hourglass theme. And on the Hourglass theme, the sticky nav bar covers up the visual interface for exiting full screen. As such it is a bit of a trap.

### OLH theme - no escape key

image

### Hourglass theme - no Escape key or button

image

## Proposed solution
Rather than enabling the fullscreen toggle, use the resize options to allow both horizontal and vertical resizing.

We currently have an invalid setting for "resize":

https://github.com/BirkbeckCTP/janeway/blob/d52b7719bf2bad26a9779246cf3305f0aa753ea2/src/core/janeway_global_settings.py#L680

image

This should be True, False, or 'both'. We can use both to allow resizing in both directions, like this:

image

## Alternatives
I tried debugging the Escape key event and the covered-up nav but could not do so in the time we have.

## Additional context
[TinyMCE docs: Controlling user editor resizing](https://www.tiny.cloud/docs/tinymce/6/editor-size-options/#controlling-user-editor-resizing)

joemull:
Fixes #4034

### DIFF
--- a/src/core/janeway_global_settings.py
+++ b/src/core/janeway_global_settings.py
@@ -677,19 +677,19 @@ TINYMCE_COMPRESSOR = False
 TINYMCE_DEFAULT_CONFIG = {
     "width": "100%",
     "min-height": "300px",
-    "resize": "height",
+    "resize": "both",
     "promotion": False,
     "branding": False,
     "menubar": "edit view insert format tools table help",
     "content_css": STATIC_URL + "/admin/css/admin.css",
     "plugins": "advlist autolink lists link image charmap preview anchor searchreplace visualblocks code"
-        " fullscreen insertdatetime media table code help wordcount spellchecker help",
+        " insertdatetime media table code help wordcount spellchecker help",
     "toolbar": "help removeformat | undo redo | bold italic underline strikethrough "
         "| fontsizeselect formatselect "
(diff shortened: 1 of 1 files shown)


mauromsl (commented):
I had decided against using "both" for the resize as it allows the iframe to flow outside its horizontal boundary and severely overlap with potential elements rendered side-by-side e.g:
image. I don't think we should enable this by default.

The keyboard shortcuts for toggling full screen are ctrl+shift+f and setting Esc as a shortcut has even more unintended consequences.

Another option is to use the `fullscreen_native` setting, which will use the browser native API for full screen and supports the Esc key. In fact I think I prefer this approach now that I've tried it. Here is an alternative PR: https://github.com/BirkbeckCTP/janeway/pull/4045

joemull:
We opted to do https://github.com/BirkbeckCTP/janeway/pull/4045.

PR #4057, opened 2024-03-26
use timezone when recording date_published at preprint acceptance

everreau:
When a publication date was selected during preprint acceptance the default date was shown in local time but recorded as if it were in UTC. Add the timezone info to the datetime object created at acceptance.

### DIFF
--- a/src/repository/views.py
+++ b/src/repository/views.py
@@ -6,6 +6,7 @@ __maintainer__ = "Birkbeck Centre for Technology and Publishing"
 import operator
 from dateutil.relativedelta import relativedelta
 from datetime import datetime
+from dateutil import tz
 
 from django.shortcuts import render, redirect, get_object_or_404
 from django.utils import timezone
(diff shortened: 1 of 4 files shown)


Bbkctp:

src/repository/views.py
```diff
@@ -879,9 +880,11 @@ def repository_manager_article(request, preprint_id):
                     'You must assign at least one galley file.',
                 )
             else:
+                d = datetime.fromisoformat(request.POST.get('datetime', timezone.now().strftime("%Y-%m-%d %H:%M")))
```
ajrbyers, line 883: What happens if a malformed date is passed in here?

ajrbyers (commented):
Thanks @everreau, just a question added inline about the view.

src/repository/views.py
everreau: @ajrbyers this scenario is unlikely because the form field does inline checking but I added a catch to handle this situation.

ajrbyers:
Dear Mr Jenkins: test this please

ajrbyers (approved):

mauromsl (approved):

PR #4061, opened 2024-03-28
Authors and editors can edit funder metadata

### LINKED ISSUE #3164: Give Funder object an admin representation
As an admin user I want to be able to search, see, create and edit Funder records in the admin area.

ajrbyers:
Closes #3164 (sort of)

- Adds a new funding statement field
- Updates the submission page for funding to have intro text, make use of a django form and validate it
- Updates the editor edit_metadata interface to do the same
- Adds a new edit view that is accessible to editors/staff at all times and to authors during submission
- Adds a spinner to inform the user when a search is underway (previously there was no way to know if a seach was underway).
- Additionally, after discussion, it was decided to rework the article/funder relationship to make it 1 Article to Many ArticleFunding entries. Two migrations take care of this process by looping over existing entries and adding the Article FK then renaming the model and removing the m2m. A method for Article called `funders` will return a queryset for backwards compatibility.

Screenshot 2024-03-28 at 10 18 31

Screen recording: https://github.com/BirkbeckCTP/janeway/assets/8321378/bdf5bc3c-d78a-439e-81a2-e98a3ad679c8


src/submission/models.py
```diff
@@ -322,14 +322,21 @@ class Meta:
...
     funding_id = models.CharField(
         max_length=500,
         blank=True,
         null=True,
         help_text="The grant ID (optional). Enter the ID by itself",
     )
+    funding_statement = models.TextField(
+        blank=True,
+        help_text=_("Additional information regarding this funding entry")
+    )
```
mauromsl, line 336: Funders have a m2m relationship with articles. Shouldn't this exist on the linking table, rather than here? Otherwise I can see two different articles competing for a Funder record and diverging funding statements competing against each other

mauromsl (changes requested):
Nice changes overall but I have a fundamental comment on the data model, that could imply further changes on the interface.

src/journal/views.py
```diff
@@ -137,18 +137,17 @@ def funder_articles(request, funder_id):
...
         return redirect
 
-    pinned_articles = [pin.article for pin in models.PinnedArticle.objects.filter(
-        journal=request.journal)]
-    pinned_article_pks = [article.pk for article in pinned_articles]
+    artile_funding_objects = submission_models.ArticleFunding.objects.filter(
+        fundref_id=funder_id,
+        article__journal=request.journal,
+        article__date_published__lte=timezone.now(),
+        article__section__pk__in=filters,
```
mauromsl, line 144: do we need to filter by `article__stage=sm_models.STAGE_PUBLISHED` as well? we always do it in tandem with `article__date_published`

src/submission/migrations/0076_auto_20240402_1301.py
```diff
@@ -0,0 +1,38 @@
...
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+def migrate_relationship_to_fk(apps, schema_editor):
+    Article = apps.get_model('submission', 'Article')
+    for article in Article.objects.all():
+        for funder in article.funders.all():
+            print(f"Adding FK to {funder.pk}")
```
mauromsl, line 11: Perhaps it would be best to `logger.debug` rather than polute stdout with these?

mauromsl (approved):
Nice, thanks for the refactor, less tech debt moving forward

I added a minor suggestion on the print statement you have on the migration

src/templates/admin/submission/submit_funding.html
```diff
@@ -4,113 +4,129 @@
...
+                </p>
+            </div>
+            <div class="title-area">
+                <h2>{% trans "Add Funding Source" %}</h2>
+            </div>
+            <div class="content submission-content">
+                <div class="row expanded">
+                    <div class="large-8 columns">
+                        <input type="text" name="funder_text"
+                               id="funder_text"/>
```
joemull, line 37: So I know this is just registered as a change because you changed the indentation, but it would be good add a label for accessibility.

src/templates/admin/elements/fundref/fundref.html
```diff
@@ -149,15 +150,37 @@
...
+                                    funder_statement_textarea.id = 'funding_statement';
+                                    funder_statement_textarea.name = 'funding_statement';
+                                    funder_statement_text.appendChild(funder_statement_textarea);
+
+                                    const funder_statement_label = document.createElement('p');
+                                    funder_statement_label.classList.add('help-text');
+                                    funder_statement_label.innerText = '{% trans funder_form.fields.funding_statement.help_text %}';
+                                    funder_statement_text.appendChild(funder_statement_label);
+
+                                    $("#funding_statement").jqte();
```
joemull, line 182: Why are we using JQTE here? I thought we just deprecated it

src/templates/admin/elements/fundref/fundref.html
```diff
@@ -149,15 +150,37 @@
...
+
+                                    const funder_statement_textarea = document.createElement('textarea');
+                                    funder_statement_textarea.id = 'funding_statement';
+                                    funder_statement_textarea.name = 'funding_statement';
+                                    funder_statement_text.appendChild(funder_statement_textarea);
+
+                                    const funder_statement_label = document.createElement('p');
+                                    funder_statement_label.classList.add('help-text');
+                                    funder_statement_label.innerText = '{% trans funder_form.fields.funding_statement.help_text %}';
+                                    funder_statement_text.appendChild(funder_statement_label);
```
joemull, line 180: These form inputs all [need labels in order to be accessible](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Text_labels_and_names?utm_source=devtools&utm_medium=a11y-panel-checks-text-label#form_elements_must_be_labeled). I know you're just adding one textarea and the other inputs are pre-existing, but you're changing quite a few things about how funders are edited, so it seems like the right time to fix this form.

Idea: Use the `foundation` template filter to generate empty HTML on this modal from an empty instance of the new Django form, and then use JS to populate the values in the bit above? Might also have the bonus of cutting down on this verbose JS code.

src/submission/views.py
```diff
@@ -372,7 +452,7 @@ def delete_funder(request, article_id, funder_id):
         journal=request.journal
     )
     funding = get_object_or_404(
-        models.Funder,
+        models.ArticleFunding,
         pk=funder_id
     )
```
joemull, line 458: I get an AttributeError error related to line 459 (below):

```
  File "/home/joe/.virtualenvs/janepr/lib/python3.10/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/home/joe/.virtualenvs/janepr/lib/python3.10/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/joe/.virtualenvs/janepr/lib/python3.10/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/home/joe/git/janepr/src/submission/decorators.py", line 60, in funding_is_enabled
  File "/home/joe/git/janepr/src/security/decorators.py", line 728, in wrapper
  File "/home/joe/git/janepr/src/submission/views.py", line 459, in delete_funder
    article.funders.remove(funding)

Exception Type: AttributeError at /submit/2/funders/1/delete/
Exception Value: 'QuerySet' object has no attribute 'remove'
```

src/submission/forms.py
```diff
@@ -490,11 +490,14 @@ class Meta:
...
+class ArticleFundingForm(forms.ModelForm):
 
     class Meta:
-        model = models.Funder
-        fields = ('name', 'fundref_id', 'funding_id')
+        model = models.ArticleFunding
+        fields = ('name', 'fundref_id', 'funding_id', 'funding_statement')
+        widgets = {
+            'funding_statement': forms.Textarea(),
+        }
```
joemull, line 500: Can this use TinyMCE for consistency? It is still rendering with JQTE for me.

joemull (changes requested):
Overall this is really good, and I think people will really appreciate the flexibility. I especially love the status indicator while the user is waiting for search results to return from the API.

As for changes that need to be made, I found one server error and a handful of accessibility and consistency fixes we can make.

ajrbyers:
> Overall this is really good, and I think people will really appreciate the flexibility. I especially love the status indicator while the user is waiting for search results to return from the API.
> 
> As for changes that need to be made, I found one server error and a handful of accessibility and consistency fixes we can make.

Good catch on the error, I totally missed that (as a result of making changes to the model after adding the view). I've made some updates and cut down on the JS used by cloning the modal then updating the ID and FOR attributes. I was going to suggest opening a new issue to refactor this as its probably beyond the scope of this issue but I gave it a go and got sucked down the rabbit hole.

ajrbyers:
FYI haven't re-requested review as I'm going to take a look at refactoring the first block of JS also. Might as well keep it consistent.

ajrbyers:
In the end I was able to cut the JS in half.

joemull (changes requested):
> In the end I was able to cut the JS in half.

Nice. I tried to test it out but I got a migration conflict:

```
CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0075_auto_20240312_0922, 0077_auto_20240402_1326 in submission).
To fix them run 'python manage.py makemigrations --merge'
```

ajrbyers:
> > In the end I was able to cut the JS in half.
> 
> Nice. I tried to test it out but I got a migration conflict:
> 
> ```
> CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0075_auto_20240312_0922, 0077_auto_20240402_1326 in submission).
> To fix them run 'python manage.py makemigrations --merge'
> ```

I've added a merge for this.

src/templates/admin/elements/fundref/fundref.html
```diff
@@ -1,184 +1,76 @@
...
+                    $('<td>').text(funder.name).appendTo(funder_tr);
+                    const add_funder_td = $('<td>').appendTo(funder_tr);
+                    const add_funder_button = $('<a>')
+                        .addClass('success button pull-right')
+                        .attr({
+                            'href': '#',
+                            'data-open': 'add_funder_' + index,
+                            'name': 'add_funder_' + index
+                        })
+                        .appendTo(add_funder_td);
```
joemull, line 41: This button is not working for me in the Edit Metadata pane. It works in submission.

This might be related to the `modal_clones_placeholder` only existing in submission, but I also inspected this in the browser and nothing gets added to that div, even when the feature works. They go to the end of the body.

If you find you need to refactor, I think I have an idea about how to do it without creating lots of modals. Just attach a click event listener to each Add button that modifies the original modal with the details for the selected funder. Define the function that modifies the modal outside the for loop that creates the table, so that you can more clearly specify and control inputs like you would with a Python function. Might this also be a way to avoid having to reinitialize TinyMCE and Foundation?

joemull (changes requested):
Nice to have all the forms loading the same way.

The buttons weren't working in one of the places I tested, unfortunately. More inline about that.

Also, can the following line be removed for clarity or is something depending on it, like the custom submission fields? I am asking mainly because I want to make sure I know where all the "Add funder" buttons are so I can test them with the new JS.

https://github.com/BirkbeckCTP/janeway/blob/d5b09cfb9d4a2d0542c4b197483e964715002a60/src/templates/admin/submission/submit_info.html#L130

src/templates/admin/elements/fundref/fundref.html
ajrbyers: The next time my brain says "this should be an issue of its own" I'm going to listen. I've rewritten this where the loop stores the details in data attributes on the button. When the button is clicked it fills the details in the modal and pops it open.

src/templates/admin/elements/fundref/fundref.html
```diff
@@ -1,184 +1,52 @@
...
+
+                add_funder_button.on('click', function() {
+                    $('#id_name').val($(this).data('name'));
+                    $('#id_fundref_id').val($(this).data('fundref-id'));
                 });
-            }
-    </script>
+            });
+
+            // Clear the HTML, display the table and reinitialize foundation
```
joemull, line 48: We can get rid of this comment as it's outdated, right?

joemull (approved):
It is now fully working for me! And the refactor is also much cleaner.

joemull:
> Also, can the following line be removed for clarity or is something depending on it, like the custom submission fields? I am asking mainly because I want to make sure I know where all the "Add funder" buttons are so I can test them with the new JS.
> 
> https://github.com/BirkbeckCTP/janeway/blob/d5b09cfb9d4a2d0542c4b197483e964715002a60/src/templates/admin/submission/submit_info.html#L130

Oh one last thing @ajrbyers . Did you see my question above?

ajrbyers:
> > Also, can the following line be removed for clarity or is something depending on it, like the custom submission fields? I am asking mainly because I want to make sure I know where all the "Add funder" buttons are so I can test them with the new JS.
> > https://github.com/BirkbeckCTP/janeway/blob/d5b09cfb9d4a2d0542c4b197483e964715002a60/src/templates/admin/submission/submit_info.html#L130
> 
> Oh one last thing @ajrbyers . Did you see my question above?

Thanks @joemull agreed, much cleaner. The include is perhaps a bit confusing given its extension is .html, its actually including this file: https://github.com/BirkbeckCTP/janeway/blob/deadd395261775624680530816ce05938d7d8988/src/templates/admin/elements/fundref/fundref.html

I think because he wanted to use Django template tags Martin gave it the .html extension. For clarity maybe we should change it to .js?

joemull:
@ajrbyers I'm not talking about the file extension but that I don't think it should be here at all. This is the `submit_info` page, which doesn't offer any funder interface, so far as I know:

https://github.com/BirkbeckCTP/janeway/blob/deadd395261775624680530816ce05938d7d8988/src/templates/admin/submission/submit_info.html#L130

Seems to me that the only places we need it are these two:

https://github.com/BirkbeckCTP/janeway/blob/deadd395261775624680530816ce05938d7d8988/src/templates/admin/submission/edit/metadata.html#L322

https://github.com/BirkbeckCTP/janeway/blob/deadd395261775624680530816ce05938d7d8988/src/templates/admin/submission/submit_funding.html#L150

Am I missing something?

ajrbyers:
> @ajrbyers I'm not talking about the file extension but that I don't think it should be here at all. This is the `submit_info` page, which doesn't offer any funder interface, so far as I know:
> 
> https://github.com/BirkbeckCTP/janeway/blob/deadd395261775624680530816ce05938d7d8988/src/templates/admin/submission/submit_info.html#L130
> 
> Seems to me that the only places we need it are these two:
> 
> https://github.com/BirkbeckCTP/janeway/blob/deadd395261775624680530816ce05938d7d8988/src/templates/admin/submission/edit/metadata.html#L322
> 
> https://github.com/BirkbeckCTP/janeway/blob/deadd395261775624680530816ce05938d7d8988/src/templates/admin/submission/submit_funding.html#L150
> 
> Am I missing something?

Nah you're not missing anything, I am though. Its not required on submit_info at all.

PR #4069, opened 2024-03-28
#3202 changed username verbose name to email so that Django builtins display it properly

### LINKED ISSUE #3202: Admin login page username -> email.
* After install user is directed to `/install` page where `username` is asked but valid field value is `email`.
* This is confusing because username has been created with email and password, so the login page should refer to valid account fields (i.e. email not the username).

* Change `username` to `email` in the `/install` page.

**Additional context**
* On the `/login` page there is a valid `Email` field label (not username) :-)

ajrbyers:
Closes #3202

### DIFF
--- a/src/core/models.py
+++ b/src/core/models.py
@@ -203,7 +203,7 @@ class AccountManager(BaseUserManager):
 
 class Account(AbstractBaseUser, PermissionsMixin):
     email = PGCaseInsensitiveEmailField(unique=True, verbose_name=_('Email'))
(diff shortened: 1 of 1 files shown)


mauromsl (changes requested):
A better approach to this is to override the form the `AdminSite` loads with our own version:

https://docs.djangoproject.com/en/3.2/ref/contrib/admin/#django.contrib.admin.AdminSite.login_form

ajrbyers:
Closed in favor of https://github.com/BirkbeckCTP/janeway/pull/4076

PR #4073, opened 2024-04-01
4072 django 4 2

### LINKED ISSUE #4072: Upgrade django version to 4.2
Django 3.2 extended support ends on April 2024 and thus we should upgrade to the next LTS of Django: 4.2

We tried this before and a few libraries had not yet been updated to support 4.2, but have been updated since (namely django_modeltranslations).

Amongst the benefits of running Django 4.2 is the new configuration for [storages](https://docs.djangoproject.com/en/5.0/releases/4.2/#custom-file-storages) which will allow us to have a cleaner implementation of multiple storage backends for #865

mauromsl:
closes #4072

### DIFF
--- a/src/core/models.py
+++ b/src/core/models.py
@@ -191,7 +191,7 @@ class AccountManager(BaseUserManager):
                 del kwargs["username"]
         try:
             validate_email(email)
-            self.normalize_email(email)
+            email = self.normalize_email(email)
         except(ValidationError, TypeError, ValueError):
             raise ValueError(f'{email} not a valid email address.')
 
--- a/src/core/models.py
+++ b/src/core/models.py
@@ -196,10 +196,12 @@ class AccountManager(BaseUserManager):
             raise ValueError(f'{email} not a valid email address.')
 
         account = self.model(
+            # The original case of the email is preserved
+            # in the email field
             email=email,
-            # Lowercasing the email into the username gets around duplicate
-            # accounts due to case-sensitivity. We still preserve the original
-            # casing for the email field and rely on username for CI constraint
+            # The email is lowercased in the username field
+            # so that we can perform case-insensitive checking
+            # and avoid creating duplicate accounts
             username=email.lower(),
         )
 
--- a/src/core/models.py
+++ b/src/core/models.py
@@ -25,6 +25,8 @@ from django.utils import timezone
 from django.contrib.contenttypes.fields import GenericForeignKey
 from django.contrib.contenttypes.models import ContentType
 from django.contrib.postgres.search import SearchVector, SearchVectorField
+from django.core.exceptions import ValidationError
+from django.core.validators import validate_email
 from django.utils.translation import gettext_lazy as _
 from django.db.models.signals import post_save
 from django.dispatch import receiver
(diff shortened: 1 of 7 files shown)


src/core/models.py
```diff
@@ -177,12 +179,27 @@ def create(self, **kwargs):
...
+        with 3rd party libraries, we allow a username argument, however only a
+        email address will be accepted as the username and email.
+        """
+        if not email and username:
+            email = username
+            if "username" in kwargs:
+                del kwargs["username"]
+        try:
+            validate_email(email)
+            self.normalize_email(email)
```
joemull, line 194: Should this be updating the value of `email`? The `normalize_email` method doesn't raise one of the errors you're checking for, it just returns the normalized email:

https://github.com/django/django/blob/1d85b416aa1382e7274e718c79b83dd050fb24ab/django/contrib/auth/base_user.py#L21-L34

Or if you want to keep `email` untransformed, then do we actually need this line?

src/core/models.py
```diff
@@ -177,12 +179,27 @@ def create(self, **kwargs):
...
+        except(ValidationError, TypeError, ValueError):
+            raise ValueError(f'{email} not a valid email address.')
 
         account = self.model(
-            email=self.normalize_email(email),
+            email=email,
+            # Lowercasing the email into the username gets around duplicate
+            # accounts due to case-sensitivity. We still preserve the original
+            # casing for the email field and rely on username for CI constraint
```
joemull, line 202: I don't entirely follow this comment, so others might have difficulty too. What does the last part mean?

joemull (changes requested):
I'm a bit unclear on the way the email is transformed or not transformed, and there's one line of code that I am a bit confused by, so just want to make it clear for others as well.

src/core/models.py
mauromsl: I tried to explain exisitng behaviour because I thought it wasn't clear, seems I didn't do that well. The basic idea is that we need a case-sensitive email field, but we need a case-insensitive  constraint to avoid accounts being created with variations in case e.g.: `m.SaNcHeZ@bbk.ac.uk`. The way this is achieved is by setting a unique constraint on `username` while making it case insensitive. Any ideas on how I can reword it to make it clear?

src/core/models.py
mauromsl: thanks! well spotted

src/core/models.py
joemull: OK, now I get it, thanks! Maybe something like this?

```
diff --git a/src/core/models.py b/src/core/models.py
index aba452e2..d480da59 100644
--- a/src/core/models.py
+++ b/src/core/models.py
@@ -196,10 +196,12 @@ class AccountManager(BaseUserManager):
             raise ValueError(f'{email} not a valid email address.')
 
         account = self.model(
+            # The original case of the email is preserved
+            # in the email field
             email=email,
-            # Lowercasing the email into the username gets around duplicate
-            # accounts due to case-sensitivity. We still preserve the original
-            # casing for the email field and rely on username for CI constraint
+            # The email is lowercased in the username field
+            # so that we can perform case-insensitive checking
+            # and avoid creating duplicate accounts
             username=email.lower(),
         )
```

ajrbyers (approved):

PR #4074, opened 2024-04-02
CMS pages can now display a table of contents

### LINKED ISSUE #3671: Add sidebar TOC to CMS pages
Currently some CMS pages are quite long. It can make it hard, as a reader, to get to the section I want to read.

- [ ] Add the ToC to CMS pages
- [ ] Consider making this optional via a setting

ajrbyers:
- Closes #3671 

Note: When a page has TOC enabled, the reading pane will be thinner than usual.


src/themes/material/templates/cms/page.html
```diff
@@ -1,19 +1,53 @@
...
+                + document.querySelector('.m3').offsetHeight
+            var y = $(this).scrollTop();
+            if (y > infobar_y) {
+                $('#toc-card').fadeIn();
+            } else {
+                $('#toc-card').fadeOut();
+            }
+        });
+    </script>
+{% endblock js %}
```
mauromsl, line 53: can't material use `{% static 'material/toc.js' %}` like clean does?

mauromsl (changes requested):
just a minor comment on the reuse of JS scripts

src/themes/material/templates/cms/page.html
mauromsl: oh right

src/themes/material/templates/cms/page.html
```diff
@@ -1,19 +1,53 @@
...
+                        <h4>
+                            {% trans "Table of Contents" %}
+                        </h4>
+                        <ul id="toc" class="section table-of-contents">
+                        </ul>
+                    </div>
+                </div>
+            </div>
+        {% endif %}
+    </div>
```
joemull, line 33: On material it covers up content:

image

src/themes/OLH/templates/cms/page.html
```diff
@@ -6,12 +6,27 @@
...
+                         data-anchor="content" data-sticky-on="large">
+                        <div class="section">
+                            <h3>{% trans "Table of Contents" %}</h3>
+                            <ul id="toc">
+
+                            </ul>
+                        </div>
+                    </div>
+                </aside>
+            {% endif %}
```
joemull, line 29: This doesn't work so well on mobile screens, because the TOC is put at the end, where it would not be expected:

image

joemull (changes requested):
Super feature, but it doesn't work well on narrow screens.

Why don't you turn it off entirely at the small breakpoint?

ajrbyers:
> Super feature, but it doesn't work well on narrow screens.
> 
> Why don't you turn it off entirely at the small breakpoint?

Cheers, @joemull, I've added a commit that will disable these elements on small screens.

joemull (changes requested):
Hmm, it is still displaying on small screens on clean:

image

joemull (approved):

PR #4082, opened 2024-04-08
Update psycopg2-binary version

joemull:
During a recent fresh installation I ran into an issue with pip install of psycopg2-binary. Version 2.8.6 uses an outdated installation method.

I am proposing bumping this to 2.9.8. I've read the changelog and don't believe there are any breaking changes for us between these versions, but please double-check. Version 2.9.9 drops support for python 3.6, so I don't think we should
go that far until we drop python 3.6.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -44,7 +44,7 @@ orcid==1.0.3
 packaging==23.2
 pdfkit==1.0.0
 pdfminer.six
-psycopg2-binary~=2.8.6
+psycopg2-binary~=2.9.8
 pyasn1==0.1.9
 pycparser==2.14
 PyJWT==2.4.0


ajrbyers:
@joemull why is this merging into [4072-django-4-2](https://github.com/BirkbeckCTP/janeway/tree/4072-django-4-2) ?

joemull:
> @joemull why is this merging into [4072-django-4-2](https://github.com/BirkbeckCTP/janeway/tree/4072-django-4-2) ?

Sorry, I guess that's a bit confusing. I saw Mauro was updating the requirements file, is why. I'll go ahead and change the branch to master.

mauromsl (changes requested):
Hey @joemull, I had a look at the changelog for psycopg2 and I think it is safe to do this upgrade. There is something wrong with your branch base though and it is bringing other commits with it. Can you rebase your branch first?

mauromsl (approved):

mauromsl:
blocked by #4073

Issue #3685, opened 2023-09-12
Provide choice of who is cc'd into author emails at prepublication stage


philippagrand:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated that I cannot simply cc in relevant editors/section editors/marketing people inhouse into the author email at prepublication stage alerting the relevant people when an article is due to publish.  At the moment I can choose peer reviewers or section editors but I don't want to alert either of those groups - I want to alert the journal editors and my marketing intern. 

**Describe the solution you'd like**
Give options to choose who to cc in to emails to authors regarding the publication of the article via a drop down list of everyone involved in the journal.  Or allow customisation somewhere on the site of who receives emails re publication of articles.

**Describe alternatives you've considered**
No alternatives tried.

**Additional context**
At present there seems to be no way to alert journal editors or others in the publisher team that an article is about to publish.  This function needs to be customisable to allow better choice about who receives emails when an article is due to publish (alongside the author).  It might also be useful to include ALL the authors of an article, not just one.

ajrbyers:
Hi @philippagrand we've got a similar feature with a ticket opened here: https://github.com/BirkbeckCTP/janeway/issues/3409

ajrbyers:
Marking this one as closed in favour of #3409

PR #4094, opened 2024-04-12
Editors have more control over publication notifications

### LINKED ISSUE #1186: Notify all authors when a prepublication email is sent.
At the pre-publication checklist stage, there is the option to email the author, section editor and peer reviewers. It would be great if all listed co-authors could be informed of the imminent publication.

### LINKED ISSUE #4089: Author Publication Notification Displaying HTML, Text Field Very Small
**Describe the bug**
When an Editor clicks the button on the Pre-Publication checklist to send a publication notification to the author, the text area for previewing the message is displaying HTML, and it's too small to read the contents of the box:

**Janeway version**
1.5.1 RC6 & Final

joemull:
Closes #4089. Also closes #1186.

This pull request improves the usability of the prepub notification modal, as well as upgrading a few underlying utilities and forms.

- Adds an email form that includes the To field, so that you can create a formset with different recipients for each form, passed to the formset in `initial`, a list of dicts. Without this, we'd have to modify each form in the formset based on its index value before calling `send_email`, which would be prone to bugs and not declarative enough.
- Upgrades the logging and display of To, CC, and BCC fields so that editors who check the logs can be sure they are not violating anonymous review when sending emails to peer reviewers.
- Moves the latest effective function for migrating translatable setting values into `migration_utils` and tests it.
- Lets editors view and modify the full emails that will be sent out to authors, section editors, and peer reviewers, and prefills the right email addresses based on new settings.

Screenshot from 2024-05-28 16-09-20
Screenshot from 2024-05-28 15-27-17
Screenshot from 2024-05-28 15-27-31

Screenshot from 2024-05-28 15-40-06
Screenshot from 2024-05-28 15-40-18
Screenshot from 2024-05-28 15-40-30


src/journal/views.py
```diff
@@ -1082,13 +1089,29 @@ def publish_article(request, article_id):
...
+                request.POST,
+                request.FILES,
+                setting_name='author_publication',
+                email_context={
+                    'article': article
+                },
+                request=request,
             )
+            if notify_author_form.is_valid():
+                logic.notify_author(request, article)
```
mauromsl, line 1102: This function has not been updated to handle the `SettingEmailForm` argument. it is instead reading off the body and handing it to the even framework.

As a result, the `CC` and `subject` parts of the form will be ignored, which might lead to frustration from the editor.

In order to resolve this `notify_author` and the events listening to `ON_AUTHOR_PUBLICATION` need to be refactored and make use of the result from `SettingEmailForm.as_dataclass()` method

mauromsl (changes requested):
Thanks Joe, this I think also provides a nice alternative for #3685 
I believe the current implementation won't cascade the form data properly onto the events framework, so that needs to be added.

A similar refactor was done in f8edb2801630d22e4fe0e1a97140325108354a67 and 8fc2ea1216d6300067ed9b1aaaceeef04b955f9f . You can follow the `"email_data"` argument to see how it is implemented.

src/journal/views.py
joemull: Thanks @mauromsl -- I have gone ahead and significantly refactored this to implement the usability upgrades we need more fully. See the updated PR description.

src/events/logic.py
```diff
@@ -224,8 +224,9 @@ class Events:
     ON_ARTICLE_PUBLISHED = 'on_article_published'
 
     # kwargs: request, article
-    # raised when an Editor notifies an author that publication is set
-    ON_AUTHOR_PUBLICATION = 'on_author_publication'
```
mauromsl, line 228: Even if we de-register the `send_author_publication_notification`, we should properly deprecate an event like this one before deletion as others might depend on it.

I'd recommend adding such warning on `events.logic.Event.register_for_event` perhaps with a set of deprecated events:
 ```python
 class Event():
 DEPRECATED_EVENTS = {
    ON_AUTHOR_PUBLICATION,
}
def register_for_event(event, *functions):
    if event in self.DEPRECATED_EVENTS:
        # raise DeprecationWarning
[...]

 ```

mauromsl (changes requested):
These are all very welcome changes!
I added one small comment inline.

More generally, I'm a bit scared by the fact this relatively large  refactor hasn't required updating tests.

Would it be possible to add a TestCase that confirms emails are being sent to the relevant parties according to configured settings?

src/events/logic.py
joemull: Good call, but if we are trying to add a deprecation warning for this event because we think people might be using it in plugins, we need to make it work with all the supporting functions.

I've added back the full set of functions to make this old event work the way it used to, and I've added the deprecation warning in `raise_event`.

joemull:
> More generally, I'm a bit scared by the fact this relatively large refactor hasn't required updating tests.

That would be because I didn't run the tests 😮 

> Would it be possible to add a TestCase that confirms emails are being sent to the relevant parties according to configured settings?

Yes, for sure.

joemull:
This now has 7 [failing tests](https://github.com/user-attachments/files/15535695/failing_tests.md), but I believe they are unrelated and some may be fixed in #4206.

mauromsl (approved):
good changes!

src/utils/migration_utils.py
```diff
@@ -36,3 +38,31 @@ def update_translated_settings(apps, setting_name, group_name, values_to_replace
                     setting_value.value = value
                     setting_value.save()
 
+
+def update_default_setting_values(apps, setting_name, group_name, values_to_replace, replacement_value):
```
ajrbyers, line 42: WWMD.

ajrbyers (approved):

PR #4101, opened 2024-04-16
#4081 reviewers from past review rounds are now valid review assignment choices.

### LINKED ISSUE #4081: "Display List of Reviewers from Past Review Rounds" breaks reviewer adding
**Describe the bug**
Whenever the feature "Display List of Reviewers from Past Review Rounds" is enabled in Manager, the reviewer selection table is unable to select reviewers. Any reviewer selected produces the error message "Select a valid choice. That choice is not one of the available choices."

**Janeway version**
Unknown! Michigan Press-managed installation and I'm not sure where to find the version number in the Dashboard or Manager. I believe they're running the current release.

**To Reproduce**
1. Enable the feature.
2. Visit a paper with previous reviewers.
3. Select "Add Reviewer" for a new round of review.
4. Select one of the previous reviewers, leaving everything else at defaults.
5. Click "Add Reviewer."
6. Error: "Select a valid choice..."

Behavior reproduced on Firefox, desktop, Linux and Chrome Android, so I don't think it has anything to do with the web browser in use.

**Expected behavior**
The reviewer should be added.

ajrbyers:
This PR tweaks the logic by no longer excluding the `past_reviewers` when calling `get_reviewer_candidates`. They are instead filtered out at the end when setting `reviewers` in the `context` argument. This means that past_reviewers remain valid reviewers but don't display twice in the review list.

Closes #4081


mauromsl (approved):

joemull (changes requested):
The logic fix is working but the display is still a bit confusing, because the table is missing a few cells on the empty row:

Screenshot from 2024-04-22 17-33-54

I got it to load properly by adding two trs:

```
diff --git a/src/templates/admin/review/add_review_assignment.html b/src/templates/admin/review/add_review_assignment.html
index 01652349..82d38cd4 100644
--- a/src/templates/admin/review/add_review_assignment.html
+++ b/src/templates/admin/review/add_review_assignment.html
@@ -75,6 +75,8 @@
                                 <td></td>
                                 <td></td>
                                 <td></td>
+                                <td></td>
+                                <td></td>
                             </tr>
                         {% endfor %}
                         </tbody>
```

image

But this still is still not quite accurate because the "empty" notice is displaying when the options are not empty. Can the set of reviewers be fixed a bit more comprehensively in the view using querysets and copied querysets if needed, so the template doesn't have to do so much logic? You could use queryset annotation to create the variables needed in the template that indicate whether the reviewer is suggested or past.

ajrbyers:
I've moved the setting of is_past_reviewer and is_suggested_reviewer over to optional annotations on the func that gets reviewer candidates.

Both settings off:
Screenshot 2024-04-23 at 16 22 53

Past Reviewers on:
Screenshot 2024-04-23 at 16 20 04

Both settings on:
Screenshot 2024-04-23 at 16 21 53

src/review/views.py
```diff
@@ -1175,21 +1166,26 @@ def add_review_assignment(request, article_id):
...
+        'general',
+        'display_past_reviewers')
+    enable_suggested_reviewers = request.journal.get_setting(
+        'general',
+        'enable_suggested_reviewers',
+    )
+    if display_past_reviewers and enable_suggested_reviewers:
+        table_sort = [7, 8]
+    elif display_past_reviewers or enable_suggested_reviewers:
+        table_sort = [7]
```
mauromsl, line 1179: What are these magic numbers? presumably numbered columns? Since the sorting happens on the client only, maybe the template should be doing this check so the logic is not split between the two. I feel like it will be easy to miss here if we later enhance that table.

Otherwise, have you considered sorting the queryset by the annotated fields instead?

src/review/logic.py
```diff
@@ -77,6 +81,31 @@ def get_reviewers(article, candidate_queryset, exclude_pks):
...
+
+    if article.journal.get_setting('general', 'display_past_reviewers'):
+        completed_reviewer_pks = article.completed_reviews_with_decision.values_list(
+            'reviewer__pk',
+            flat=True,
+        )
+        reviewers = reviewers.annotate(
+            is_past_reviewer=Case(
+                When(pk__in=completed_reviewer_pks, then=True),
+                default=False,
```
mauromsl, line 93: We could pass in a subquery or do an inner join to save us one roundtrip to the database but I'm not quite sure what the performance would be like, what have you found locally with this implementation?

src/review/views.py
```diff
@@ -1085,20 +1085,11 @@ def add_review_assignment(request, article_id):
...
-        past_reviewers = logic.get_previous_round_reviewers(
-            article,
-        )
     reviewers = logic.get_reviewer_candidates(
         article,
         user=request.user,
-        reviewers_to_exclude=past_reviewers,
     )
-
+    suggested_reviewers = past_reviewers = table_sort = list()
```
mauromsl, line 1092: This multiple assignment will set all variables to the same one list in memory. I don't imagine that is the intended behaviour.

Perhaps it hasn't led to any problems because I don't think you are using `suggested_reviewers` or `past_reviewers` anywhere

src/review/logic.py
```diff
@@ -77,6 +81,31 @@ def get_reviewers(article, candidate_queryset, exclude_pks):
     ).annotate(
         rating_average=Subquery(rating_average, output_field=IntegerField()),
     )
+
+    if article.journal.get_setting('general', 'display_past_reviewers'):
```
mauromsl, line 85: Do we even want this setting? It sounds that when it is off, all we do is remove the column and avoid the initial sorting. To me the name of the setting reads as if you would "hide" the past reviewers if I turned this off.

src/review/logic.py
```diff
@@ -77,6 +81,31 @@ def get_reviewers(article, candidate_queryset, exclude_pks):
...
+        article_keywords = [keyword.word for keyword in article.keywords.all()]
+        reviewers = reviewers.annotate(
+            is_suggested_reviewer=Case(
+                When(interest__name__in=article_keywords,
+                     then=Value(True)),
+                default=Value(False),
+                output_field=BooleanField(),
+            )
+        )
+
```
mauromsl, line 108: How's the performance here? sounds like it will run a full scan on a fairly large table.

src/templates/admin/review/add_review_assignment.html
```diff
@@ -30,9 +30,9 @@ <h2>1. Select Reviewer</h2>
                         {% endblocktrans %}
                     </p>
 
-                    {% if past_reviewers %}
+                    {% if journal_settings.general.display_past_reviewers %}
                         <div class="callout primary">
-                            <p><span class="fa fa-info-circle"></span> Reviewers who have compeleted a review for this article in a previous round will appear at the top of this list, they will show with a <strong>Yes</strong> in the <strong>Has Reviewed Article</strong> column.</p>
+                            <p><span class="fa fa-info-circle"></span> Reviewers who have completed a review for this article in a previous round will appear at the top of this list, they will show with a <strong>Yes</strong> in the <strong>Has Reviewed Article</strong> column.</p>
```
mauromsl, line 35: Do we even want this setting? It sounds that when it is off, all we do is remove the column and avoid the initial sorting. To me the name of the setting reads as if you would "hide" the past reviewers if I turned this off.

mauromsl (changes requested):
Some questions/comments inline, but certainly  much (c)leaner than it was!

mauromsl (commented):
Thanks for this! Added one last comment I think would be good to make the logic even simpler and also reduce config bloat

joemull (changes requested):
This is a thing of beauty! I tested it out and everything is working perfectly. The extra help text is nice too. The only thing I'd tweak are two comma splices:

...this process is silent, **so** they will not...

...will appear at the top of this **list. They** will show with...

joemull (approved):

PR #4107, opened 2024-04-18
Patch Notify Author email field with TinyMCE

joemull:
This uses a bare-bones Django form to put TinyMCE into an arbitrary template as a hotfix for #4089. It does not fully address #4089 because we would like to refactor it a bit more in a later minor version of Janeway--see #4094.

Screenshot from 2024-04-22 16-52-57

### DIFF
--- a/src/core/forms/forms.py
+++ b/src/core/forms/forms.py
@@ -13,7 +13,6 @@ from django.utils.translation import gettext_lazy as _
 from django.contrib.auth.forms import UserCreationForm
 from django.core.validators import validate_email, ValidationError
 
-from swapper import ImproperlyConfigured
 from tinymce.widgets import TinyMCE
 
 from core import email, models, validators
--- a/src/core/forms/forms.py
+++ b/src/core/forms/forms.py
@@ -13,6 +13,7 @@ from django.utils.translation import gettext_lazy as _
 from django.contrib.auth.forms import UserCreationForm
 from django.core.validators import validate_email, ValidationError
 
+from swapper import ImproperlyConfigured
 from tinymce.widgets import TinyMCE
 
 from core import email, models, validators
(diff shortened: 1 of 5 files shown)


src/templates/admin/elements/publish/author.html
```diff
@@ -32,4 +36,12 @@ <h4><i class="fa fa-envelope-o">&nbsp;</i>Notify Author</h4>
...
-</div>
\ No newline at end of file
+</div>
+<script type="module">
+    document.addEventListener('DOMContentLoaded', () => {
+        tinymce.init({
+            selector: '#notify_author_email',
+        })
+    })
+</script>
```
mauromsl, line 46: I suspect initialising TinyMCE this way will discard the config coming from settings.py. If so, can we use a simple Django Form with just this field? It is probably worth doing so regardless of this concern in any case

mauromsl (changes requested):

src/templates/admin/elements/publish/author.html
joemull: > this way will discard the config coming from settings.py

Yeah, it would not pick up the config, but it was the simplest way to patch it because of the view's entanglement with logic and the notify sequence. I've changed it to a Django form as discussed offline. Is this what you had in mind?

mauromsl (approved):
yes, that does the trick for me

src/core/forms/forms.py
```diff
@@ -13,6 +13,7 @@
 from django.contrib.auth.forms import UserCreationForm
 from django.core.validators import validate_email, ValidationError
 
+from swapper import ImproperlyConfigured
```
ajrbyers, line 16: Not sure what this is for, it doesn't seem to get used anywhere.

ajrbyers (changes requested):
Looks good. There is an import that may not belong on this branch though.

PR #4122, opened 2024-04-26
Repository managers using the manager interface can now preview versions and files without downloading them

### LINKED ISSUE #2540: Preprints: Moderator should be able to preview PDF in-line

As an preprints moderator, I want to view the latest version of the PDF of a preprint that is pending approval in-line in the browser window so I do not need to download the article. 

* "Pending" PDF is viewable in the browser (perhaps as a "preview" page) 
* PDF displayed is the latest PDF uploaded by submitter 

**Additional context**

Similar to [janeway feature request #1431: Preview for articles in pre-publication](https://github.com/BirkbeckCTP/janeway/issues/1431)?

ajrbyers:
Closes #2540

### DIFF
--- a/src/templates/admin/elements/repository/version_detail.html
+++ b/src/templates/admin/elements/repository/version_detail.html
@@ -25,6 +25,11 @@
                     <p><strong>File</strong></p>
                     {% if version.preprint.current_version.file %}
                     <p><a href="{% url 'repository_file_download' version.preprint.id version.preprint.current_version.file.id %}"><i class="fa fa-download"></i> Download</a></p>
+                    <p>
+                        <a aria-label="(opens in a new tab)" target="_blank" href="{% url 'repository_pdf' version.preprint.pk %}?file={% url 'repository_download_file' version.preprint.pk version.preprint.current_version.file.id %}">
+                            <span class="fa fa-eye"></span> Preview File
+                        </a>
+                    </p>
                     {% endif %}
                 </div>
             </div>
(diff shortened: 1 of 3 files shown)


src/templates/admin/elements/repository/version_detail.html
```diff
@@ -41,7 +46,16 @@ <h2>New Metadata</h2>
                     <p><strong>Published DOI{% if version.published_doi and version.published_doi != version.preprint.doi %} (changed){% endif %}</strong></p>
                     <p>{% if version.published_doi == version.preprint.doi %}Unchanged{% else %}{{ version.published_doi }}{% endif %}</p>
                     <p><strong>File</strong></p>
-                    <p>{% if version.file %}<a href="{{ version.file.download_url }}"><i class="fa fa-download"></i> Download</a>{% else %}No File{% endif %}</p>
+                    {% if version.file %}
+                    <p><a href="{{ version.file.download_url }}"><i class="fa fa-download"></i> Download</a></p>
```
StephDriver, line 50: I think the icon here should be decorative not semantic.  see https://docs.fontawesome.com/web/dig-deeper/accessibility

StephDriver (approved):
the only comments I have on this are about icon use.  And I'm not sure whether they matter at this stage or not.  We do need to using icons more accessibly, and there is an issue to find and update all current icon use to the accessible syntax.  But at some point we will also need to start writing accessible icons each time we add new ones.  

I have not marked all of them as the code repeated with similar issues re icons and assuming this gets merged before the icon update (likely as that's not even triaged yet) then they'll get fixed later.  But it raises the question of when and how to move over to specifying icons accessibly for new lines of code.

ajrbyers:
> the only comments I have on this are about icon use. And I'm not sure whether they matter at this stage or not. We do need to using icons more accessibly, and there is an issue to find and update all current icon use to the accessible syntax. But at some point we will also need to start writing accessible icons each time we add new ones.
> 
> I have not marked all of them as the code repeated with similar issues re icons and assuming this gets merged before the icon update (likely as that's not even triaged yet) then they'll get fixed later. But it raises the question of when and how to move over to specifying icons accessibly for new lines of code.

I think auto accessibility is probably on a newer version of FA so we should consider upgrading.

PR #4143, opened 2024-04-30
Ensures that only staff are allowed to assign the journal manager role

### LINKED ISSUE #4141: Non-staff able to assign Journal Manager role
**Describe the bug**
[Only users with Is Staff permission](https://github.com/BirkbeckCTP/janeway/pull/3464) are supposed to be able to assign the Journal Manager role to other users, but in a recent test with a non-staff Editor role account I was able to assign the Journal Manager role to other users. We've also seen other users of our system mass-enrolling Journal Managers when they do not have the Is Staff flag set.

**Janeway version**
1.5.1

**To Reproduce**
1. Log in to a journal with an account that is not staff, admin or superuser, but has the Editor role
2. Navigate to Staff > Manager > Enrol Users
3. Assign Journal Manager role to any other account. You will be able to do so successfully even though you should not be able to.

ajrbyers:
Two issues here:

1. A bed merge had moved `AccountRoleSerializer`'s `validate` method into `PreprintSerializer`.
2. The `enrol_users` view was not filtering out the journal manager role properly.

2 would not have been a serious issue without 1, as the API would have refused to assign the role and let the non-staff user know.

Additionally, I've renamed the roles variable as it shadows a an outer scope definition.

Closes #4141

### DIFF
--- a/src/api/serializers.py
+++ b/src/api/serializers.py
@@ -144,14 +144,21 @@ class AccountRoleSerializer(serializers.ModelSerializer):
         model = core_models.AccountRole
         fields = ('pk', 'journal', 'user', 'role')
 
-    def create(self, validated_data):
-        role = validated_data.get("role")
+    def validate(self, data):
+        request = self.context.get('request', None)
+        role = data.get("role")
 
-        if role.slug == 'reader':
-            raise serializers.ValidationError({"role": "You cannot add a user as a reader via the API."})
-        else:
-            account_role = core_models.AccountRole.objects.create(**validated_data)
-            return account_role
+        excluded_roles = ['reader']
+
+        # if the current user is not staff add the journal-manager role to
+        # the list of excluded roles.
+        if not request or not request.user.is_staff:
+            excluded_roles.append('journal-manager')
+
+        if role.slug in excluded_roles:
+            raise serializers.ValidationError({"error": "You cannot add a user to that role via the API."})
+
+        return data
 
 class RepositoryFieldAnswerSerializer(serializers.ModelSerializer):
     class Meta:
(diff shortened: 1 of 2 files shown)


mauromsl (changes requested):
As discussed, we need a test case as with other security issues. Otherwise looks good

ajrbyers:
@mauromsl test now in place. I've added an explicit kwarg to `helpers.create_user` for `is_staff` as setting this in the attr loop doesn't seem to work.

src/api/tests/test_api.py
```diff
@@ -0,0 +1,98 @@
...
+                'role': journal_manager_role.pk,
+                'journal': self.journal.pk,
+            },
+            SERVER_NAME=self.journal.domain,
+        )
+        self.assertFalse(
+            self.average_user.check_role(
+                self.journal,
+                journal_manager_role,
+            )
```
mauromsl, line 97: I'd probably check the status of the response and the message, but your check is actually more thorough

src/utils/testing/helpers.py
```diff
@@ -30,7 +30,7 @@
 from cms import models as cms_models
 
 
-def create_user(username, roles=None, journal=None, **attrs):
+def create_user(username, roles=None, journal=None, is_staff=False, **attrs):
```
mauromsl, line 33: I think this should work as part of `**attrs` unless I'm missing something

mauromsl (approved):
Added some thoughts, not even comments so I'm green ticking

src/utils/testing/helpers.py
ajrbyers: As mentioned in the message above, it doesn't.

ajrbyers: I'll double check this in a mo.

joemull (approved):

PR #4147, opened 2024-04-30
Fixed an issue where galleys generated throug transformation where getting their metrics recorded

### LINKED ISSUE #4146: Converting manuscript file into Galley leads to a galley with no type
**Describe the bug**
When you use the "Convert" option on a file to turn it from a manuscript into a galley, the galley instance is not given a type and thus metrics against the galley will be incorrectly labelled.

**Janeway version**
v1.5.2

**To Reproduce**
1. Create an article with a PDF file as its manuscript
2. Move the file into production/typesetting
3. Click on 'convert' and select the first option
4. click "edit on admin" on the galley
5. See error

**Expected behavior**
The galley type should be set correctly, as with other galley creation methods

**Additional context**
The implementation should make sure that `production.logic.save_galley` is called to ensure the galley is setup correctly

mauromsl:
closes #4146 

Adds a migration to fix existing malformed records


src/core/migrations/0088_fix_untyped_galleys.py
```diff
@@ -0,0 +1,32 @@
...
+        ('core', '0087_merge_20240315_1649'),
+        ('metrics', '0010_auto_20230317_1534'),
+    ]
+
+def fix_untyped_galleys(apps, schema_editor):
+    """ Fixes galleys that haven't had a type set and linked metrics records"""
+
+    Galley = apps.get_model('core', 'Galley')
+    ArticleAccess = apps.get_model('metrics','ArticleAccess')
+    for galley in Galley.objects.filter(type="",file__isnull=True, article__isnull=True):
```
ajrbyers, line 19: Is this supposed to exclude Galleys that have files/articles?

ajrbyers (commented):
Added a comment.

src/core/migrations/0088_fix_untyped_galleys.py
joemull: OK yes I have this question too.

ajrbyers:
Joe suggested over in https://github.com/BirkbeckCTP/typesetting/pull/208 that `create_galley_from_file` should be marked as deprecated.

src/journal/views.py
```diff
@@ -899,9 +900,19 @@ def article_file_make_galley(request, article_id, file_id):
...
+
 
-    logic.create_galley_from_file(file_object, article_object, owner=request.user)
+    # Avoid circular import.
+    from production import logic as production_logic
+
+    production_logic.save_galley(
+        article_object, request, content_file,
+        is_galley=True,
+    )
```
joemull, line 915: I tested this through the interface by uploading a manuscript file, converting it to a galley, and then checking the attributes in admin. It seems that there is no galley type being assigned. That's a problem, right? Is it possible there is still a logic error here?

joemull (changes requested):
I think the view should be saving a galley type when it converts files with the new logic, right? It doesn't seem to be doing so.

src/journal/views.py
joemull: I tested with a PDF and a JATS file.

mauromsl: was it a file with a valid mime that can be linked to a correct galley type? (PDF/HTML/XML)

joemull (approved):

src/journal/views.py
joemull: Just a note to record that this was an issue with my testing environment, and we confirmed the type is set properly.

ajrbyers (approved):

PR #4153, opened 2024-05-01
Allow italics in a variety of content fields

### LINKED ISSUE #4097: Italicisation for text fields in Janeway that do not currently allow this, e.g. news item titles
Currently there are certain fields in Janeway that do not support italicisation, such as titles for news items, news item preview text, and certain journal items such as the 'About' text block title, articles/items on the homepage carousel, and the automatic text above the license information on the submissions page.

Either the ability to add italics with HTML italics tag or where instances of the journal's name is displayed that users cannot edit, that the journal's name displays in italics by default (and that this might be easy to revert if the journal doesn't want this - which is unlikely).

**Additional context**
Journal titles are displayed in italics by the majority of citation styles and so this would not in my view be contributing to the proliferation of styles. The inability to be able to italicise certain titles leads to inconsistency in editorial formatting which is not ideal. This also discourages using the journal's name in places where it should be used, leading to potential accessibility problems (e.g. journal's logo being used as image instead of journal title used as text).

joemull:
This branch adds support for italics in a variety of content fields, especially those that might include journal titles or other titles of works that need italicization.

We can do this more easily and safely now thanks to safer storage of HTML in the database introduced in 1.5.1.

Screenshot from 2024-05-01 16-01-25
Screenshot from 2024-05-01 15-59-57
Screenshot from 2024-05-01 15-59-30
Screenshot from 2024-05-01 15-58-55
Screenshot from 2024-05-01 15-58-44

For the copyright text, I have opted to avoid putting the journal name altogether. This will require a retranslation of that string but I believe it is worth it because the existing translation is not ideal, because it's a partial sentence (excludes the journal name) and does not include punctuation as part of the translated string (the colon).
image

Closes #4097.


joemull:
Oh. Not quite ready. Problem with migrations.

joemull:
@ajrbyers I just wanted to make sure you saw this one is ready for review now.

ajrbyers (commented):
I've just given this a once over and before I test it out I wondered if there was a reason why the Frozen Author records had all been convered to bleach fields?

joemull:
> I've just given this a once over and before I test it out I wondered if there was a reason why the Frozen Author records had all been convered to bleach fields?

Ah, yes it's because they get used in the carousel subtitle. We said we wanted to support italics in carousel text.

src/core/homepage_elements/html/views.py
```diff
@@ -15,27 +16,32 @@
...
-                                            html_block_content,
-                                            request.journal)
-        messages.add_message(request, messages.INFO, 'HTML Block updated.')
+        form = forms.HTMLForm(
+            request.POST,
+        )
+        if form.is_valid():
+            form.save(request.journal)
+            messages.add_message(request, messages.INFO, 'HTML Block updated.')
         return redirect(reverse('home_settings_index'))
```
ajrbyers, line 40: Should this redirect fire even if the form is not valid?

src/journal/migrations/0066_issue_type_bleach_20240507_1359.py
ajrbyers, line 1: Is it possible to run these migrations in 0065 or do they need to be separate?

src/submission/migrations/0080_frozen_author_bleach_20240507_1350.py
ajrbyers, line 1: Same Q as before.

ajrbyers (approved):
I've added a couple of comments/questions but this looks great.

src/core/homepage_elements/html/views.py
joemull: Thanks--fixed it.

src/journal/migrations/0066_issue_type_bleach_20240507_1359.py
joemull: It is generally hard because Django has to do two things at once, and there is an error that is raised when doing both operations in one migration. I tried to solve it earlier in a press migration by setting a default value, but this did not fix it and Mauro had to patch it here:

https://github.com/BirkbeckCTP/janeway/blame/4097-additional_html_display/src/press/migrations/0032_auto_20231116_1057.py

I have tried using Mauro's solution on this branch, but it did not work for me and I figure my time was better spent creating a re-usable Python solution than try to delve into the SQL behind his patch.

So this time I have gone for what I hope is a simpler and more declarative solution of using two migrations. It's a common solution that people talk about on forums, and it seems to work well here, especially because we can abstract away the repetitive parts in `migration_utils`.

ajrbyers: Thanks!

ajrbyers:
@joemull just FYI this PR now has a conflict.

joemull:
@ajrbyers After rebasing I ran tests and picked up several things that needed to be addressed:

1) Utilities like importers trying to save None to affected fields on FrozenAuthor and IssueType
2) Fields from the Account model that allow null=True being directly saved to FrozenAuthor

I think this calls for standardizing the corresponding Account model fields, so I've done that. What do you think? Re-requesting review since it's a significant enough change.

ajrbyers (approved):
Thanks Joe, these changes look good. We should also consider how this will effect importers in the import plugin. Perhaps open an issue over there to check?

ajrbyers:
@mauromsl assigning this to you for a second review as the changes are significant.

mauromsl (changes requested):
One minor comment @joemull

joemull:
Rebased to resolve conflict in migration_utils.

src/comms/models.py
```diff
@@ -151,9 +152,16 @@ def best_image_url(self):
 
     def __str__(self):
         if self.posted_by:
-            return '{0} posted by {1} on {2}'.format(self.title, self.posted_by.full_name(), self.posted)
+            return '{0} posted by {1} on {2}'.format(
+                mark_safe(self.title),
```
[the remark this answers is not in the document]
ajrbyers: @joemull it should be fine to remove it.

mauromsl (approved):

ajrbyers (approved):
🎖️

PR #4158, opened 2024-05-01
Jenkins xml runner

mauromsl:


joemull (commented):
@mauromsl Looks like the test results got committed to Git because of a hyphen rather than underscore in the .gitignore line.

PR #4160, opened 2024-05-02
Adds a new column to the copyediting dashboard showing assigned copyeditors.

### LINKED ISSUE #4159: Janeway feature - Assigned copyeditor column in Copyediting tab
[Reported through FD #8137](https://janeway.freshdesk.com/a/tickets/8137) 

**Issue reported**
Some months ago we added a second copyeditor to AH's team, who is handling the copyediting of reviews and some of our research articles. Since this new person came on board, we have found it challenging to distribute copyediting assignments, as we do not have (that we know) a Janeway tab showing which manuscripts are assigned to which copyeditor and/or how many active tasks a given copyeditor has at a given time.

In the enrolled users section I can check the whole record of assignments for a copyeditor, but it is not that easy to identify quickly which ones are active.

In the kanban menu I can scroll down copyediting tasks and check who is in charge of them, but the list cannot be reordered by the assigned copyeditor because of how it is formatted. Also, The format makes it hard to count how many exactly each copyeditor has. 

Just having a "copyeditor" column in the copyediting section would solve the issue, allowing us to reorder the list or just search for the name of the copyeditor, so we get only the submissions that person has been assigned to.

Is this possible to do in Janeway in its current version? Perhaps it is something that could be implemented in the future?

Could this be rolled into existing thoughts and work around copyediting? Or the manager interface for copyediting / the dashboard overview?

ajrbyers:
- Added a new column that displays a list of copyeditors with links to the editor review page
- Moved the link to open an article's copyediting page from the "view" column (now deleted) to the title column
- Closes #4159 

image

### DIFF
--- a/src/templates/admin/copyediting/copyediting.html
+++ b/src/templates/admin/copyediting/copyediting.html
@@ -40,7 +40,7 @@
                         <td>{{ article.section.name }}</td>
                         <td>{% if article.projected_issue %}{{ article.projected_issue.display_title }}{% else %}None{% endif %}</td>
                         <td>{{ article.stage }}</td>
-                        <td>{% for assignment in article.copyeditassignment_set.all %}<a href="{% url 'editor_review' article.pk assignment.pk %}">{{ assignment.copyeditor.full_name }}</a>{% if not forloop.last %}, {% endif %}{% endfor %}</td>
+                        <td>{% for assignment in article.copyeditassignment_set.all %}<a href="{% url 'editor_review' article.pk assignment.pk %}">{{ assignment.copyeditor.full_name }}</a>{% if not forloop.last %}, {% endif %}{% empty %}No assignments{% endfor %}</td>
                     </tr>
                 {% empty %}
                     <tr>
--- a/src/templates/admin/copyediting/copyediting.html
+++ b/src/templates/admin/copyediting/copyediting.html
@@ -1,6 +1,7 @@
 {% extends "admin/core/base.html" %}
 
 {% block page_title %}Copyediting{% endblock %}
+{% block title %}Copyediting{% endblock %}
 {% block admin-header %}Articles in Copyediting{% endblock %}
 
 {% block breadcrumbs %}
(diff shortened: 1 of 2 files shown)


src/templates/admin/copyediting/copyediting.html
```diff
@@ -25,21 +26,21 @@ <h2>Articles in Copyediting</h2>
...
-                        <td>{{ article.safe_title }}</td>
+                        <td><a href="{% url 'article_copyediting' article.pk %}">{{ article.safe_title }}</a></td>
                         <td>{{ article.date_submitted }}</td>
                         <td>{{ article.date_accepted }}</td>
                         <td>{{ article.correspondence_author.full_name }}</td>
                         <td>{{ article.section.name }}</td>
                         <td>{% if article.projected_issue %}{{ article.projected_issue.display_title }}{% else %}None{% endif %}</td>
                         <td>{{ article.stage }}</td>
-                        <td><a href="{% url 'article_copyediting' article.pk %}">View</a></td>
+                        <td>{% for assignment in article.copyeditassignment_set.all %}<a href="{% url 'editor_review' article.pk assignment.pk %}">{{ assignment.copyeditor.full_name }}</a>{% if not forloop.last %}, {% endif %}{% endfor %}</td>
```
StephDriver, line 43: This will leave a blank cell if there are no copyeditors assigned.

For tables to be accessible there should not be any empty cells.

Please provide a default entry for when there are no copyeditors assigned.  This does not have to be visible, it could be just for screen readers.

StephDriver (changes requested):

src/templates/admin/copyediting/copyediting.html
ajrbyers: Good catch, added an `{% empty %}` statement.

StephDriver: Ah, good catch back at you - I think that is a different problem to the one I saw.

The `{% empty %}` statement looks like one for a row.  I was concerned about an empty cell.  For the case where an article had yet to have any copyeditors assigned.

StephDriver (approved):
looks good.

mauromsl (approved):

PR #4161, opened 2024-05-02
Fixes a bad typo on the editor's revision interface.

ajrbyers:
Kinda self explanatory this one.

### DIFF
--- a/src/templates/admin/review/revision/view_revision.html
+++ b/src/templates/admin/review/revision/view_revision.html
@@ -53,7 +53,7 @@
                     <p><strong>Covering Letter to Editor:</strong></p>
                     {{ revision_request.author_note|safe }}
                 {% else %}
-                    <p>The author has not writen a covering letter.</p>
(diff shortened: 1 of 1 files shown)


S-Haime (approved):
Good one (and welp that one is a bit awkward indeed!)

src/templates/admin/review/revision/view_revision.html
```diff
@@ -53,7 +53,7 @@ <h2>Author Responses</h2>
                     <p><strong>Covering Letter to Editor:</strong></p>
                     {{ revision_request.author_note|safe }}
                 {% else %}
-                    <p>The author not write a covering letter.</p>
+                    <p>The author has not writen a covering letter.</p>
```
joemull, line 56: "written" needs two Ts.

Also let's fix line 64:
```html
<p>The author write a response letter.</p>
```

joemull (changes requested):

joemull (approved):

PR #4173, opened 2024-05-08
Updates to class-based view for filterable lists

joemull:
These changes were made to provide functionality for BirkbeckCTP/consortial_billing#65.

They update the filterable class-based view in these ways:
- Make the methods essentially generic by checking to see if it is a set of articles before doing article-specific things
- Add filter types integer, search, and boolean
- Add a simple way to search using Django admin's search functionality

### DIFF
--- a/src/core/views.py
+++ b/src/core/views.py
@@ -1198,9 +1198,10 @@ def add_user(request):
         if registration_form.is_valid():
             new_user = registration_form.save()
             # Every new user is given the author role
-            new_user.add_account_role('author', request.journal)
+            if request.journal:
+                new_user.add_account_role('author', request.journal)
 
-            if role:
+            if role and request.journal:
                 new_user.add_account_role(role, request.journal)
 
             form = forms.EditAccountForm(
--- a/src/core/views.py
+++ b/src/core/views.py
@@ -2442,7 +2442,7 @@ class FilteredArticlesListView(generic.ListView):
         initial = dict(params_querydict.lists())
         for keyword, value in initial.items():
             if keyword in facets:
-                if facets[keyword]['type'] in ['date_time', 'date', 'integer', 'search']:
+                if facets[keyword]['type'] in ['date_time', 'date', 'integer', 'search', 'boolean']:
                     initial[keyword] = value[0]
 
         context['facet_form'] = forms.CBVFacetForm(
@@ -2493,6 +2493,11 @@ class FilteredArticlesListView(generic.ListView):
                         value_list[0],
                     )
                     predicates = []
+                elif facets[keyword]['type'] == 'boolean':
+                    if value_list[0]:
+                        predicates = [(keyword, True)]
+                    else:
+                        predicates = [(keyword, False)]
                 elif value_list[0]:
                     predicates = [(keyword, value) for value in value_list]
                 elif facets[keyword]['type'] not in ['date_time', 'date', 'integer', 'search']:
--- a/src/core/views.py
+++ b/src/core/views.py
@@ -2419,7 +2419,7 @@ class GenericFacetedListView(generic.ListView):
     facets = {}
 
     # These fields will receive a single initial value, not a list
-    single_value_fields = ['date_time', 'date', 'integer', 'search', 'boolean']
+    single_value_fields = {'date_time', 'date', 'integer', 'search', 'boolean'}
 
     # None or integer
     action_queryset_chunk_size = None
--- a/src/core/views.py
+++ b/src/core/views.py
@@ -2442,7 +2442,7 @@ class FilteredArticlesListView(generic.ListView):
         initial = dict(params_querydict.lists())
         for keyword, value in initial.items():
             if keyword in facets:
-                if facets[keyword]['type'] in ['date_time', 'date', 'integer']:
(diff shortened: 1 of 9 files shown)


src/core/views.py
```diff
@@ -2433,13 +2454,14 @@ def get_queryset(self, params_querydict=None):
...
-            )
-        ).exclude(
-            stage=submission_models.STAGE_UNSUBMITTED,
+        self.queryset = self.filter_queryset_if_journal(
+            self.queryset.filter(*q_stack)
         )
+        if hasattr(self.model, 'stage'):
+            self.queryset = self.queryset.exclude(
+                stage=submission_models.STAGE_UNSUBMITTED
+            )
```
mauromsl, line 2463: At a minimum we should be more explicit here and say what we mean:
`if self.model is submission_models.Article:`

But ideally,to achieve proper polymorphism, since this class is now meant to be used with different models, we need to turn it into an abstract class (`GenericFacetedListView`?) which implements this method in the lines of:

```python
class FacetedArticleListView(GenericFacetedListView):
    model = submission_models.Article
    def get_queryset(self):
        qs = super().get_queryset()
        return qs.filter(
            stage=submission_models.STAGE_UNSUBMITTED
        )

File: src/core/views.py [hunk 2]
```diff
@@ -2468,14 +2490,16 @@ def get_facets(self):
...
         # return None
         queryset = self.filter_queryset_if_journal(
             super().get_queryset()
-        ).exclude(
-            stage=submission_models.STAGE_UNSUBMITTED
         )
+        if hasattr(self.model, 'stage'):
+            queryset = queryset.exclude(
+                stage=submission_models.STAGE_UNSUBMITTED
+            )
```
mauromsl, line 2502: as above, either check class identity on `self.model` or implement a refactor

mauromsl (changes requested):
These look to be some useful new facets.

I've found a couple minor areas were we can improve the implementation, particularly one around polymorphism

src/core/views.py
```diff
@@ -2420,9 +2428,22 @@ def get_queryset(self, params_querydict=None):
...
+                    predicates = []
+                elif facets[keyword]['type'] == 'boolean':
+                    if value_list[0]:
+                        predicates = [(keyword, True)]
+                    else:
+                        predicates = [(keyword, False)]
+                elif value_list[0]:
                     predicates = [(keyword, value) for value in value_list]
-                elif facets[keyword]['type'] not in ['date_time', 'date']:
+                elif facets[keyword]['type'] not in ['date_time', 'date', 'integer', 'search']:
```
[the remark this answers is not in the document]
joemull: Thanks for pushing on this piece. I did some digging to try to remember *exactly* why it's there, and I think it is actually not needed, because we used a different method in the end. I'll document this in the commit message body removing this if/else check.

src/core/views.py
joemull: Yes, you are right. I didn't change the inheritance structure because I wanted to make sure you and Andy were good with expanding the view past articles. I will go ahead and implement it properly.

src/core/views.py
```diff
@@ -2346,20 +2346,20 @@ def manage_access_requests(request):
...
+    template_name = NotImplementedField
 
-    model = submission_models.Article
-    template_name = 'core/manager/article_list.html'
     paginate_by = '25'
     facets = {}
 
+    # These fields will receive a single initial value, not a list
+    single_value_fields = ['date_time', 'date', 'integer', 'search', 'boolean']
+
```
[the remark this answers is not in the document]
joemull: Ah right, good call, thanks.

joemull:
@ajrbyers rebasing done.

ajrbyers (approved):
Just one comment, it doesn't block this PR from being merged but we will need to address it before deployment:

This PR removes `FilteredArticlesListView` which is in use in the isolinear plugin.

https://github.com/BirkbeckCTP/isolinear/blob/11775ad04b83a36af4f8967536a0ba19e359e6f1/views.py#L248-L253

mauromsl:
> Just one comment, it doesn't block this PR from being merged but we will need to address it before deployment:
> 
> This PR removes `FilteredArticlesListView` which is in use in the isolinear plugin.
> 
> https://github.com/BirkbeckCTP/isolinear/blob/11775ad04b83a36af4f8967536a0ba19e359e6f1/views.py#L248-L253

Good Catch, we'll need to still support this with a deprecation warning

> Just one comment, it doesn't block this PR from being merged but we will need to address it before deployment:
> 
> This PR removes `FilteredArticlesListView` which is in use in the isolinear plugin.
> 
> https://github.com/BirkbeckCTP/isolinear/blob/11775ad04b83a36af4f8967536a0ba19e359e6f1/views.py#L248-L253


Good catch! @joemull can we preserve the name `FilteredArticlesListView` with a deprecation warning raised on its initialization? We want to avoid betraying semantic versioning

src/core/views.py
```diff
@@ -2405,19 +2407,20 @@ def manage_access_requests(request):
     )
 
 
-class FilteredArticlesListView(generic.ListView):
```
mauromsl, line 2408: We need to preserve this name as to not introduce a breaking change

mauromsl (changes requested):

mauromsl (approved):

PR #4176, opened 2024-05-10
#4174 fixes a bug that stopped metrics from displaying for repositor authors.

### LINKED ISSUE #4174: Preprints: Author dashboard metrics display doesn't load: 'repository.logic' lacks 'metrics_summary' attribute
**Describe the bug**
As an author, I expect to be able to access the preprint's dashboard page to review and update its metadata and PDF. 

**Janeway version**
1.5.3

**To Reproduce**
1. Submit a preprint as an author and publish it. 
2. View and download the preprint to generate some stats
3. Try to view the preprint in your author dashboard 

**Sentry error**
```
AttributeError

module 'repository.logic' has no attribute 'metrics_summary'

```

ajrbyers:
Closes #4174 

This PR also changes the way in which metrics are displayed for authors. They will now see a total count rather than two charts for this and last month. This change also allows the files block lower on the page to be full width.

Screenshot 2024-05-10 at 11 06 36


mauromsl (approved):

src/templates/admin/repository/author_article.html
```diff
@@ -146,6 +146,27 @@ <h2>Dates</h2>
                         {% endfor %}
                     </table>
                 </div>
+                <div class="title-area">
+                    <h2>Metrics</h2>
+                </div>
+                <div class="content">
+                    <p>Views track how often the article page has been visited, while Downloads record the number of times a
```
StephDriver, line 153: Curious as to why its a generic `the article page` but then `{{ request.repository.object_name }}'s ` for downloads.  

If that's because this is being displaced on the article page, then maybe use **this** article page rather than **the**.

StephDriver (approved):

PR #4177, opened 2024-05-13
3621 password reset email validation rebased

StephDriver:
created a new GetResetToken form and linked through to each of the three themes.

Closes https://github.com/BirkbeckCTP/janeway/issues/3621


ajrbyers (changes requested):
This is a very welcome change. Now that this page is using a Django form we should also display any form errors returned when the form is not valid. This can be done by adding the following include to each of the templates:

`{% include "elements/forms/errors.html" with form=your_form_name_here %}`

StephDriver:
error validation added and tested.

ajrbyers (approved):

src/core/forms/forms.py
```diff
@@ -116,6 +116,16 @@ def clean_password_2(self):
         return password_2
 
 
+class GetResetToken(forms.Form):
```
mauromsl, line 119: Small nitpick, can we add the "Form" suffix to this class name for consistency with the rest of the forms in this module.

mauromsl (approved):

PR #4178, opened 2024-05-13
3936 add alt text to social media buttons

### LINKED ISSUE #3936: menu item missing a text alternative
The facebook, twitter, linked in sharing menu items are shows as an icon only without any alt-text.

Image

https://www.iastatedigitalpress.com/farmreports/article/id/12984/

(and Twitter one probably needs to be updated!)
this also applies to other icons being used, such as the cite and print options.

Image

The images are missing alt-text, and as they are links, this also means the links have no text description as to where they are going (which was a separate site improve report on link missing a text alternative).

this includes the expanding menu icon for narrower screens:

Image

StephDriver:
- Share icons now have aria-labels on OLH and Material themes (no share icons found on Clean).
- Dyslexia, text size, print icons now have aria-labels on OLH (not found on Material or Clean).
- How to Cite icon now has Aria label and is no longer invisible to screen reader or keyboard navigation (tested with Mac Voice-over).  The dropdown inside it for different citation format also now works with tab and screen reader.
- The options hamburger button for narrow windows how has an aria label for 'options'.

All tested with Mac Voice Over.  Related bugs noticed (such as difference in order between visual and tab) which will feed into the current A11y audit, but as for what was reported in this issue, that has now been fixed.

Closes #3936

### DIFF
--- a/src/themes/OLH/templates/journal/article.html
+++ b/src/themes/OLH/templates/journal/article.html
@@ -123,13 +123,13 @@
                             <ul class="menu vertical medium-horizontal align-right"
                                 data-responsive-menu="dropdown">
                                 {% hook 'article_buttons' %}
-                                <li><a href="{{ article.local_url }}print/"><i class="fa fa-print">&nbsp;</i></a></li>
-                                <li><a href="javascript:resizeText(-1)">A-</a></li>
-                                <li><a href="javascript:resizeText(1)">A+</a></li>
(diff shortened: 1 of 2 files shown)


ajrbyers (changes requested):
This is great but as they're on the front of house we should mark the aria-labels as translatable!

ajrbyers (approved):

joemull (approved):

PR #4179, opened 2024-05-13
3933 article menu focus tab order bugfix

### LINKED ISSUE #3933: is it clear which page element has focus from the keyboard?
focus tab order is does not follow the same order as the visual layout.

example:
https://www.iastatedigitalpress.com/mmb/article/id/15762/

visual order:
- ...
- share to linked in
- citation style dropdown
- Dyslexia
- A+
- A-
- Print icon
- DOI link (lower down the article)
- ...

Tab order:
- ... 
- share to linked in
- print icon
- A-
- A+
- Dyslexia
- DOI link (lower down the article)

StephDriver:
- items not showing were corrected in #4178 so this branches off that.
- With all items now present, but in wrong order two further changes required:
    - remove 'align-right' which was reversing the visual order of the menu items.
    - adjust column widths within the Foundation Flex layout so that the menu itself remained 'right aligned' (while contents was not reordered).  Small is covered by the hamburger menu, so this required settings for medium and large.  

Closes #3933

### DIFF
--- a/src/themes/OLH/templates/journal/article.html
+++ b/src/themes/OLH/templates/journal/article.html
@@ -123,13 +123,13 @@
                             <ul class="menu vertical medium-horizontal align-right"
                                 data-responsive-menu="dropdown">
                                 {% hook 'article_buttons' %}
-                                <li><a href="{{ article.local_url }}print/"><i class="fa fa-print">&nbsp;</i></a></li>
-                                <li><a href="javascript:resizeText(-1)">A-</a></li>
-                                <li><a href="javascript:resizeText(1)">A+</a></li>
+                                <li><a href="{{ article.local_url }}print/" aria-label="print article"><i class="fa fa-print">&nbsp;</i></a></li>
+                                <li><a href="javascript:resizeText(-1)" aria-label="decrease text size">A-</a></li>
+                                <li><a href="javascript:resizeText(1)" aria-label="increase text size">A+</a></li>
                                 {% if author.display_email and not journal_settings.article.hide_author_email_links %}<li><a href="mailto:{{ article.correspondence_author.email }}"><i class="fa fa-envelope"></i></a></li>{% endif %}
-                                <li><a href="#" id="dyslexia-mode">{% trans "Dyslexia" %}</a></li>
+                                <li><a href="#" id="dyslexia-mode" aria-label="Dyslexia mode">{% trans "Dyslexia" %}</a></li>
                                 <li>
-                                    <a><i class="fa fa-comments"></i></a>
+                                    <a aria-label="Cite article"><i class="fa fa-comments"></i></a>
                                     <ul class="menu vertical">
                                         <li><a data-open="HarvardModal">{% trans "Harvard Citation Style" %}</a></li>
                                         <li><a data-open="VancouverModal">{% trans "Vancouver Citation Style" %}</a></li>
--- a/src/themes/OLH/templates/journal/article.html
+++ b/src/themes/OLH/templates/journal/article.html
@@ -104,7 +104,7 @@
             <div class="mini-bar sticky" data-sticky data-margin-top="0" data-sticky data-anchor="content">
                 <div class="row">
                     <div class="title-bar" data-responsive-toggle="options-menu" data-hide-for="medium">
-                        <button class="menu-icon" type="button" data-toggle></button>
+                        <button class="menu-icon" type="button" data-toggle aria-label="Options"></button>
                         <div class="title-bar-title">{% trans 'Options' %}</div>
                     </div>
                     <div id="options-menu">
--- a/src/themes/OLH/templates/journal/article.html
+++ b/src/themes/OLH/templates/journal/article.html
@@ -108,7 +108,7 @@
                         <div class="title-bar-title">{% trans 'Options' %}</div>
                     </div>
                     <div id="options-menu">
-                        <div class="large-6 columns">
+                        <div class="large-3 medium-4 columns">
                             <ul class="menu vertical medium-horizontal pad-left-15"
                                 data-responsive-menu="drilldown medium-dropdown">
                               {% with article.get_doi_url as doi_url %}
@@ -119,8 +119,8 @@
                               {% endwith %}
                             </ul>
                         </div>
-                        <div class="large-6 columns">
-                            <ul class="menu vertical medium-horizontal align-right"
+                        <div class="large-4 medium-5 columns">
+                            <ul class="menu vertical medium-horizontal"
                                 data-responsive-menu="dropdown">
                                 {% hook 'article_buttons' %}
                                 <li><a href="{{ article.local_url }}print/" aria-label="print article"><i class="fa fa-print">&nbsp;</i></a></li>
--- a/src/themes/OLH/templates/journal/article.html
+++ b/src/themes/OLH/templates/journal/article.html
@@ -129,7 +129,7 @@
                                 {% if author.display_email and not journal_settings.article.hide_author_email_links %}<li><a href="mailto:{{ article.correspondence_author.email }}"><i class="fa fa-envelope"></i></a></li>{% endif %}
                                 <li><a href="#" id="dyslexia-mode" aria-label="Dyslexia mode">{% trans "Dyslexia" %}</a></li>
                                 <li>
-                                    <a aria-label="Cite article"><i class="fa fa-comments"></i></a>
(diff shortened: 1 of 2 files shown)


src/themes/OLH/templates/journal/article.html
```diff
@@ -104,36 +104,36 @@ <h3>{{ article.title|safe }}</h3>
...
-                                <li><a href="https://twitter.com/intent/tweet?text={{ article.title }} {% if doi_url %}{{ doi_url }}{% else %}{{ article.url }}{% endif %}" target="_blank"><i class="fa fa-twitter"></i></a></li>
-                                <li><a href="https://www.linkedin.com/sharing/share-offsite?url={% if doi_url %}{{ doi_url }}{% else %}{{ article.url }}{% endif %}" target="_blank"><i class="fa fa-linkedin"></i></a></li>
+                                <li><a href="https://www.facebook.com/share.php?p[url]={% if doi_url %}{{ doi_url }}{% else %}{{ article.url }}{% endif %}" target="_blank" aria-label="Share on Facebook"><i class="fa fa-facebook"></i></a></li>
+                                <li><a href="https://twitter.com/intent/tweet?text={{ article.title }} {% if doi_url %}{{ doi_url }}{% else %}{{ article.url }}{% endif %}" target="_blank" aria-label="Share on X"><i class="fa fa-twitter"></i></a></li>
+                                <li><a href="https://www.linkedin.com/sharing/share-offsite?url={% if doi_url %}{{ doi_url }}{% else %}{{ article.url }}{% endif %}" target="_blank" aria-label="Share on Linked In"><i class="fa fa-linkedin"></i></a></li>
```
mauromsl, line 118: We should flag these strings for translation (e.g `aria-label={% trans "Share on Facebook" %}`)

src/themes/OLH/templates/journal/article.html
```diff
@@ -104,36 +104,36 @@ <h3>{{ article.title|safe }}</h3>
...
                               {% endwith %}
                             </ul>
                         </div>
-                        <div class="large-6 columns">
-                            <ul class="menu vertical medium-horizontal align-right"
+                        <div class="large-4 medium-5 columns">
```
mauromsl, line 122: unless I'm missing something, the values no longer add up to 12 columns on these divs (`large-3` + `large-4`). If we want to leave a gap I'd suggest adding another div (`<div class="large-5 medium-3"`)

src/themes/OLH/templates/journal/article.html
```diff
@@ -104,36 +104,36 @@ <h3>{{ article.title|safe }}</h3>
...
+                            <ul class="menu vertical medium-horizontal"
                                 data-responsive-menu="dropdown">
                                 {% hook 'article_buttons' %}
-                                <li><a href="{{ article.local_url }}print/"><i class="fa fa-print">&nbsp;</i></a></li>
-                                <li><a href="javascript:resizeText(-1)">A-</a></li>
-                                <li><a href="javascript:resizeText(1)">A+</a></li>
+                                <li><a href="{{ article.local_url }}print/" aria-label="print article"><i class="fa fa-print">&nbsp;</i></a></li>
+                                <li><a href="javascript:resizeText(-1)" aria-label="decrease text size">A-</a></li>
+                                <li><a href="javascript:resizeText(1)" aria-label="increase text size">A+</a></li>
```
mauromsl, line 128: as above, we need to flag text on these aria-labels as translatable strings

src/themes/OLH/templates/journal/article.html
```diff
@@ -104,36 +104,36 @@ <h3>{{ article.title|safe }}</h3>
...
                                 {% if author.display_email and not journal_settings.article.hide_author_email_links %}<li><a href="mailto:{{ article.correspondence_author.email }}"><i class="fa fa-envelope"></i></a></li>{% endif %}
-                                <li><a href="#" id="dyslexia-mode">{% trans "Dyslexia" %}</a></li>
+                                <li><a href="#" id="dyslexia-mode" aria-label="Dyslexia mode">{% trans "Dyslexia" %}</a></li>
```
mauromsl, line 130: as above, we need to flag text on these aria-labels as translatable strings

src/themes/OLH/templates/journal/article.html
```diff
@@ -104,36 +104,36 @@ <h3>{{ article.title|safe }}</h3>
...
                                 <li>
-                                    <a><i class="fa fa-comments"></i></a>
+                                    <a href="#" aria-label="Cite article"><i class="fa fa-comments">&nbsp;</i></a>
```
mauromsl, line 132: as above, we need to flag text on these aria-labels as translatable strings

src/themes/material/templates/journal/article.html
```diff
@@ -254,13 +254,13 @@ <h4>
...
-                            target="_blank"><i class="fa fa-facebook"></i></a>
+                            target="_blank" aria-label="Share on Facebook"><i class="fa fa-facebook"></i></a>
                         <a class="waves-effect waves-light btn btn-small twitter-bg" 
                             href="https://twitter.com/intent/tweet?text={{ article.title }} {% if doi_url %}{{ doi_url }}{% else %}{{ article.url }}{% endif %}"
-                            target="_blank"><i class="fa fa-twitter"></i></a>
+                            target="_blank" aria-label="Share on X"><i class="fa fa-twitter"></i></a>
                         <a class="waves-effect waves-light btn btn-small linkedin-bg"
                             href="https://www.linkedin.com/sharing/share-offsite?url={% if doi_url %}{{ doi_url }}{% else %}{{ article.url }}{% endif %}"
-                            target="_blank"><i class="fa fa-linkedin"></i></a></button>
+                            target="_blank" aria-label="Share on Linked In"><i class="fa fa-linkedin"></i></a></button>
```
mauromsl, line 263: as above, we need to flag text on these aria-labels as translatable strings

mauromsl (changes requested):
Thanks @StephDriver, I've added a few comments regarding the need to flag text in HTML for translation using [Django's `translate` tag](https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#translate-template-tag)

There is also a comment regarding a set of row columns that no longer fill the entire viewport width

src/themes/OLH/templates/journal/article.html
StephDriver: I did look into that - both in the framework docs and then it took a bit of trial and error to find values that worked when tested.  As I understood it values didn't need to add up to 12 - rather they set the minimum space, and when the sum was less than twelve the remaining space was managed dynamically so that the it balanced as the window resized.

Since there are the other changes to make I'll look at this again too, but from memory it wasn't needed.

StephDriver: > Incomplete Rows
In order to work around browsers' different rounding behaviors, Foundation will float the last column in a row to the right so the edge aligns. If your row doesn't have a count that adds up to 12 columns, you can tag the last column with a class of end in order to override that behavior.

https://get.foundation/sites/docs-v5/components/grid.html#:~:text=You%20can%20use%20column%20or,devices%20will%20inherit%20those%20styles.

mauromsl: I think they describe it further down:
image


Since the framework will float the last column, the default behaviour will be up to the browser implementation. On Firefox, I have seen gaps whenever columns don't add up to 12. So I believe it is worth ensuring that here, even if just for consistency.

mauromsl (changes requested):

src/themes/OLH/templates/journal/article.html
StephDriver: ok

StephDriver:
Mauro I've updated this, but I found what I think is a separate issue (#4340) - could you take a quick look when reviewing this to confirm it is a separate matter and not something I should be addressing in this PR?

For the rest, I've updated as you requested and tested across Safari, Chrome and Firefox.

mauromsl:
> Mauro I've updated this, but I found what I think is a separate issue (#4340) - could you take a quick look when reviewing this to confirm it is a separate matter and not something I should be addressing in this PR?
> 
> For the rest, I've updated as you requested and tested across Safari, Chrome and Firefox.

Yes, I believe it is worth resolving #4340 here, since I think it is this PR that makes it worse

mauromsl (changes requested):
As mentioned in comments, worth anticipating #4340 here, since it is this PR that changes the order of elements in the mini-bar

StephDriver:
Closed in favour of #4355

PR #4183, opened 2024-05-16
#3909 table popups updated to have unique ids

### LINKED ISSUE #3909: element IDs are not unique
example page is https://www.iastatedigitalpress.com/mmb/article/id/13894/

tables and <a> tags have duplicate element IDs, possibly more than those.

StephDriver:
Table modal originally was a copy of table, so all ids were same as original ids.  Not all elements in table have id.

- prefix `copy-of-`
- added to table id
- added to all child ids of that table

paired programming with @mauromsl 

Closes #3909

### DIFF
--- a/src/journal/logic.py
+++ b/src/journal/logic.py
@@ -595,9 +595,19 @@ def get_all_tables_from_html(content):
     tables = []
 
     for table in soup.findAll('div', attrs={'class': 'table-expansion'}):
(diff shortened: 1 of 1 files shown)


mauromsl (approved):
Approved

ajrbyers (approved):
Nice!

PR #4193, opened 2024-05-21
Drop unused libraries

frafra:
This is just some minor refactoring, so I didn't bother to open an issue just for that. I hope this is not a problem.
The CI does not show any regression (6 failures, 1 error - that happens with master as well).

I automatically dropped the unused imports with `ruff check src --select F401 --per-file-ignores '*/migrations/*.py:F401' --fix` and then detected the unused libraries with `deptry src --requirements-txt requirements.txt --requirements-files-dev dev-requirements.txt --json-output /dev/stdout 2>/dev/null | jq -r '.[] | select(.error.code == "DEP002") | .module' | grep -v ^django | grep -v ^psycopg`. I tried not to touch automatically generated code, like migrations.

It would be good to have a look to double-check, because:
- The CI could have missed some paths (if so, it would be nice to write such a missing test)
- I ignored the Django dependencies (should any of them be dropped?)
- There could be some corner case scenario, such as some dependencies might be optional and be there just to improve performances (it would be nice to add a comment on the side if that is the case)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,5 @@
-beautifulsoup4==4.9.3
 boto==2.46.1
 bs4==0.0.1
-colorlog==3.1.4
-chardet==3.0.4
-crossrefapi==1.5.0
 Django==4.2.11
 django-bleach==3.0.1
 django-bootstrap4==23.2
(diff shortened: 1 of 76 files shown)


Bbkctp:

joemull (changes requested):
Hi @frafra, thanks for this effort but there are some issues with the results in `requirements.txt`. 

One example: the `deptry` tool doesn't seem to be properly recognizing libraries that are installed under a different name than they are imported, like `beautifulsoup4` in `requirements.txt` vs. `from bs4 import BeautifulSoup` in Python files. In fact `bs4` needs to be removed from `requirements.txt`, because it was [likely only added by mistake](https://pypi.org/project/bs4/).

Other examples include `Pillow` and `python-docx`.

You're right that these are not caught by our tests--we are still working on getting our test coverage up, so any help writing tests would be much appreciated.

As for imports, these could use cleaning up, but I'd defer to others on whether this is the right approach.

PR #4201, opened 2024-05-24
An experimental app that lets users edit HTML and XML galleys in the browser.

ajrbyers:
This is a draft PR that adds an app to Janeway that allows editors to directly edit HTML and XML galley files in the browser. Things still to do:

- [x] Consider adding a history element (how, when file based?)
- [x] Link to Production or, if present, Typesetting for image management.
- [x] Add help on adding images to Galleys using this interface (this will slightly differ for XML vs HTML)
- [x] Integrate this with the Archive page
- [x] Integrate this with Prod/Typesetting 
- [x] As the CodeEditor widget is now in core, we should update the LiquidXSLT and CustomStyling plugins to use it and DRY up the code a bit.

### DIFF
--- a/src/core/include_urls.py
+++ b/src/core/include_urls.py
@@ -46,6 +46,7 @@ urlpatterns = [
     path('workflow/', include('workflow.urls')),
     path('discussion/', include('discussion.urls')),
     path('oidc/', include('mozilla_django_oidc.urls')),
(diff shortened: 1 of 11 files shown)


ajrbyers:
This has been moved over to its own app that could be merged into core in the future: https://github.com/BirkbeckCTP/file_editor

PR #4202, opened 2024-05-24
Date and time fields use browser-default selection widgets

### LINKED ISSUE #4095: JQUI datepicker interferes with TinyMCE toolbar
## Describe the bug
If there is a JQUI datepicker above a TinyMCE text field, the datepicker dropdown is hidden beneath the text field toolbar:

## Janeway version
v1.5.1-RC-6+

## Proposed solution
- Replace all instances of JQUI datepicker with the standard HTML picker, provisionally, until we have a back-office design system for these.
- Replace all instances of JQUI datepicker and timepicker with a new DateTimeField modification that uses `datetime-local`.

### In scope

- [x] src/templates/admin/core/manager/news/index.html:    {% include "elements/datepicker.html" with target="#id_start_display" %}
- [x] src/templates/admin/core/manager/news/index.html:    {% include "elements/datepicker.html" with target="#id_end_display" %}
- [x] src/templates/admin/review/add_review_assignment.html:    {% include "elements/datepicker.html" with target="#id_date_due" %}
- [x] src/templates/admin/review/revision/request_revisions_notification.html:{% include "elements/datepicker.html" with target="#id_date_due" %}
- [x] src/templates/admin/review/revision/edit_revision_request.html:    {% include "elements/datepicker.html" with target="#id_date_due" %}
- [x] src/templates/admin/review/revision/request_revisions.html:    {% include "elements/datepicker.html" with target="#id_date_due" %}
- [x] src/templates/admin/review/edit_review.html:    {% include "elements/datepicker.html" with target="#id_date_due" %}
- [x] src/templates/admin/copyediting/add_copyeditor_assignment.html:    {% include "elements/datepicker.html" with target="#id_due" %}
- [x] src/templates/admin/copyediting/editor_review.html:    {% include "elements/datepicker.html" with target='#due' %}
- [x] src/templates/admin/copyediting/edit_assignment.html:    {% include "elements/datepicker.html" with target="#id_due" %}
- [x] src/templates/admin/journal/publish_article.html:    {% include "elements/datepicker.html" with target="#date" %}
- [x] src/templates/admin/preprints/article.html:    {% include "elements/datepicker.html" with target="#date" %}
- [x] src/templates/admin/journal/publish_article.html:    {% include "elements/timepicker.html" with target="#time" %}
- [x] src/templates/admin/preprints/article.html:    {% include "elements/timepicker.html" with target="#time" %}

### Not in scope

- [ ] src/templates/admin/production/edit_typesetter_assignment.html:    {% include "elements/datepicker.html" with target="#id_due" %}
- [ ] src/templates/admin/proofing/proofing_article.html:    {% include "elements/datepicker.html" with target="#id_due" %}
- [ ] src/templates/admin/proofing/request_typesetting_changes.html:    {% include "elements/datepicker.html" with target="#id_due" %}
- [ ] src/templates/admin/proofing/edit_proofing_assignment.html:    {% include "elements/datepicker.html" with target="#id_due" %}

joemull:
This closes #4095 as documented in that issue. Also closes https://github.com/BirkbeckCTP/janeway/issues/4205

image

The pub date selector was the only date+time combination:

Screenshot from 2024-05-24 18-10-28

In adding the form for selecting a pub date, I fixed up the confirmation message a bit so that it is crystal clear even with a timezone difference:

Screenshot from 2024-05-24 18-10-44

To make the datetime picker work, I used [a custom model field as documented on the Django forum](https://forum.djangoproject.com/t/possible-datetime-local-bug/9477/7) as discussed.

But as a side effect, an error is thrown when saving the model in the admin. (Change the Article admin back to `admin.ModelAdmin` to see the error.) To handle this I created an override to use the same widget in the admin. It is a bit ugly:

Screenshot from 2024-05-24 18-39-57

If we don't like this we can simply drop the ModelField and use the custom FormField or widget independently.

### DIFF
--- a/src/templates/admin/elements/publish/pubdate.html
+++ b/src/templates/admin/elements/publish/pubdate.html
@@ -1,3 +1,5 @@
+{% load foundation %}
+
 <div class="reveal small" id="pubdate" data-reveal data-animation-in="slide-in-up"
      data-animation-out="slide-out-down">
     <div class="card">
@@ -17,7 +19,7 @@
             {% endif %}
             <form method="POST">
                 {% csrf_token %}
-                {{ pub_date_form }}
+                {{ pub_date_form|foundation }}
                 <button type="submit" name="pubdate" class="small success button">Set Publication Date</button>
             </form>
             <button class="close-button" data-close aria-label="Close reveal" type="button">


src/journal/views.py
```diff
@@ -1016,6 +1017,7 @@ def publish_article(request, article_id):
     :param article_id: Article PK
     :return: contextualised django template
     """
+    from submission import forms as submission_forms
```
ajrbyers, line 1020: Does having this import in the head cause a circular import?

src/journal/views.py
```diff
@@ -1083,19 +1085,33 @@ def publish_article(request, article_id):
...
+                        )
+                    )
+                else:
+                    messages.add_message(
+                        request, messages.SUCCESS,
+                        _('Publication date unset')
                     )
-                )
             else:
-                modal = 'pubdate'
```
ajrbyers, line 1098: Is this unset on purpose? When the form is invalid generally we'd want to open the modal again.

src/submission/forms.py
```diff
@@ -524,3 +524,19 @@ def utility_clean_orcid(orcid):
 
     # ORCID is None.
     return orcid
+
+
+class PubDateForm(forms.ModelForm):
```
ajrbyers, line 529: This is great.

src/templates/admin/elements/publish/pubdate.html
```diff
@@ -18,21 +17,8 @@ <h5>Publication Date/Time</h5>
...
-                        <div class="large-4 columns">
-                            <label for="time">Time</label>
-                            <input type="text" name="time" id="time" required {% if article.date_published %}value="{{ article.date_published|date:'H:i' }}"{% endif %}>
-                        </div>
-                        <div class="large-12 columns">
-                            <button type="submit" name="pubdate" class="small success button">Set Publication Date</button>
-                        </div>
-                    </div>
-                </div>
+                {{ pub_date_form }}
```
ajrbyers, line 20: This is being output in table format. Perhaps use as_p or the foundation tag?
Screenshot 2024-05-29 at 13 26 41

ajrbyers (commented):
Generally this is a very welcome change. I've added a couple of questions inline though.

src/journal/views.py
joemull: Yes I believe so. There is another function in this file that has the same import.

src/templates/admin/elements/publish/pubdate.html
joemull: Huh, that is wild. Firefox clears it in my inspector pane (after parsing the HTML) but I can see it when I inspect the source. Is this a Django default feature, or something we have configured? In any case, I've added the foundation filter.

ajrbyers (approved):

mauromsl (approved):

PR #4219, opened 2024-05-31
A warning will show in prepublication when an article has no license set.

### LINKED ISSUE #2270: Add a warning at prepub when an article has no license or a non CC BY license
It is possible for an article to end up in the prepublication stage without a license. A couple of cases that might lead to this are:
 - Imported articles
 - License field disabled at submission (a bit of an edge case)

Add a warning at prepub indicating that the article doesn't have a license. Similarly alert when an article has no license, a license with no URL (for malformed imported licenses) or it will be published as "All rights reserved".

ajrbyers:
- Adds two warnings, one when there is no set license and another when the set license has no URL.

Closes #2270 

image image


mauromsl (approved):

ajrbyers:
Note: discussed with @mauromsl and decided that adding a warning for non CC licenses was a too opinionated.

ajrbyers:
> One of these doesn't have `target="_blank"`. Is that intentional?
> 
> The projected issue one (not changed in this PR) doesn't have one either.

Thanks Joe, I've added that in. The projected issue link opens a modal so doesn't require one.

joemull (changes requested):
"Opens in new tab" is nice, but this syntax `sr-only` is not used in foundation, is it?

https://sites.get.foundation/visibility.html#accessibility

It looks like it's included in `font-awesome`, but that's not what we want to draw on here, right?

joemull (approved):

PR #4267, opened 2024-06-10
Fix bug that caused `&amp;` to display and improve one-line rich-text fields

### LINKED ISSUE #4258: HTML entities in bleached settings are not all rendered properly
## User story
I might use an ampersand in the title of my journal:
Screenshot from 2024-06-06 14-34-49

But it is converted to an entity and not displayed properly on save:

Screenshot from 2024-06-06 14-35-09

Screenshot from 2024-06-06 14-37-40
Screenshot from 2024-06-06 14-36-29

Angle brackets are also transformed:
Screenshot from 2024-06-06 14-46-09

image

But other [commonly escaped characters](https://developer.mozilla.org/en-US/docs/Glossary/Entity) are left alone:

Screenshot from 2024-06-06 14-48-32
Screenshot from 2024-06-06 14-48-49

Downstream, ampersands may also be an issue, but many systems could probably handle them:
https://community.crossref.org/t/html-entities-in-journal-titles/3061/2

## Janeway version
1.5.5

joemull:
Closes #4258.

This fixes a bug with ampersands reported in the issue above.

It also improves usability and handling of a number of related things:
- Adds a Janeway custom bleach form field (previously we just had a custom bleach model field)
- Changes the bleaching of setting to depend on setting type, meaning that charfields, text fields, numbers etc. are left alone--not even escaping of ampersands is applied
- Adds a new setting type for one-line HTML fields, called "mini HTML"
- Users can now edit one-line rich-text fields like article title with a small TinyMCE widget
- Documents the four main text-related field types in Janeway fields and settings
- Standardizes the setting types for settings that should use `text`, `char`, or `mini-html` to the appropriate type


src/core/forms/forms.py
```diff
@@ -13,10 +13,11 @@
 from django.contrib.auth.forms import UserCreationForm
 from django.core.validators import validate_email, ValidationError
 
-from tinymce.widgets import TinyMCE
+from tinymce.widgets import TinyMCE, mark_safe
```
ajrbyers, line 16: `mark_safe` is imported here but doesn't seem to be used.

ajrbyers (changes requested):
Just one comment, looks good.

src/core/migrations/0089_update_setting_types.py
```diff
@@ -0,0 +1,95 @@
...
+    ('general', 'matromo_tracking_code', 'text', 'char'),
+    ('email_subject', 'subject_notify_correction_cancelled', 'text', 'char'),
+    ('general', 'copyright_submission_label', 'char', 'mini-html'),
+    ('email_subject', 'subject_accepted_review_reminder', 'text', 'char'),
+    ('email_subject', 'subject_default_review_reminder', 'text', 'char'),
+    ('email_subject', 'subject_notify_se_draft_declined', 'rich-text', 'char'),
+    ('crossref', 'crossref_date_suffix', 'text', 'char'),
+    ('general', 'default_journal_language', 'text', 'char'),
+    ('email_subject', 'subject_revisions_complete_receipt', 'text', 'char'),
+    ('email_subject', 'subject_author_copyedit_deleted', 'text', 'char'),
```
mauromsl, line 73: is it worth running a single query here instead against the `'email_subject'` group?

Just so that it picks up custom emails (reminders) and those coming from plugins. We will also need to update the plugins that come with emails (I believe typesetting is the only one)

src/core/migrations/0089_update_setting_types.py
```diff
@@ -0,0 +1,95 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0088_fix_untyped_galleys'),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            update_text_and_char_setting_types,
+            reverse_code=migrations.RunPython.noop
```
mauromsl, line 93: Can we have another migration that replaces `"&amp;"` for `"&"` on journal titles? Just so that we self-heal any journals that have been edited since.

src/core/model_utils.py
```diff
@@ -592,10 +599,60 @@ def pre_save(self, model_instance, *args, **kwargs):
...
+    A form field to hold limited HTML phrasing content,
+    generally for use inline or on one line.
+    It uses a much smaller bleach allowlist
+    and loads by default with a minimal TinyMCE widget.
+    It is the default formfield used by JanewayBleachCharField.
+    """
+
+    def __init__(self, *args, **kwargs):
+        kwargs['allowed_tags'] = get_allowed_html_tags_minimal()
+        kwargs['allowed_attributes'] = get_allowed_attributes_minimal()
```
mauromsl, line 628: We should only do this if these `kwargs` are not passed by the caller. Overriding the provided arguments does violate [Liskov's substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle). Beyond software engineering principles, it could be useful to be able to dynamically configure an instance of this  field that supports different attributes/tags

src/core/model_utils.py
```diff
@@ -592,10 +599,60 @@ def pre_save(self, model_instance, *args, **kwargs):
...
-    """ An override of BleachField to use a TextInput but get sanitization"""
-    def formfield(self, **kwargs):
-        kwargs["widget"] = forms.TextInput()
+    """
+    An override of JanewayBleachField to use a minimal form field
+    and widget but get sanitization.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
mauromsl, line 652: Seems we do not need to override `__init__`

src/core/model_utils.py
```diff
@@ -592,10 +599,60 @@ def pre_save(self, model_instance, *args, **kwargs):
...
+            mce_attrs={
+                'plugins': 'help code',
+                'menubar': '',
+                'forced_root_block': 'div',
+                'toolbar': 'help removeformat | undo redo | ' \
+                           'bold italic superscript subscript',
+                'height': '8rem',
+                'resize': True,
+                'elementpath': False,
+            }
```
mauromsl, line 641: as above, we'd lose the benefits of composition here by ignoring the provided `widget` in `kwargs`

mauromsl (changes requested):
Very happy with this implementation. Well done @joemull.

As mentioned inline, it'd be good to fix any potentially corrupt journal title due to entity encoding (it is probably safe to only replace "&" but wouldn't harm to also fix encoded "<" and ">")

Couple minor comments regarding some class initialisers

src/core/model_utils.py
joemull: So I thought about this and tried to design for greater flexibility, but we are hindered by the `formfield` definition in Django Beach, which itself violates the substitution principle:

```
https://github.com/marksweb/django-bleach/blob/d675d09423ddb440b4c83c8a82bd8b853f4603c7/django_bleach/models.py#L42-L61
```

The received kwargs are updated by defaults in the opposite way recommended by Django's docs:

https://docs.djangoproject.com/en/5.0/howto/custom-model-fields/#specifying-the-form-field-for-a-model-field

I don't think we can get around this because whatever custom version of formfield we write will need to call super, which will include Django Bleach's definition.

src/core/migrations/0089_update_setting_types.py
joemull: I will put in a PR for typesetting.

src/core/model_utils.py
joemull: This field is affected in the same way as the others, so I made it consistent as a direct override of `kwargs` so that at least it is consistent with the other two.

ajrbyers (approved):

mauromsl (approved):

PR #4273, opened 2024-06-11
Removed visible template tags from all theme registration pages

### LINKED ISSUE #4272: Django tags visible on the registration page
Screenshot 2024-06-11 at 14 09 04

ajrbyers:
Closes #4272

Screenshot 2024-06-11 at 14 13 02
Screenshot 2024-06-11 at 14 14 09
Screenshot 2024-06-11 at 14 15 21

### DIFF
--- a/src/templates/common/elements/password_rules.html
+++ b/src/templates/common/elements/password_rules.html
@@ -1,0 +1,14 @@
+<li>
+    {% blocktrans trimmed with password_length=request.press.password_length %}
+        Your password should be {{ password_length }} characters long.
+    {% endblocktrans %}</li>
+{% if request.press.password_upper %}
+    <li>{% blocktrans trimmed %}
+        Your password should contain at least one upper case letter.
+    {% endblocktrans %}</li>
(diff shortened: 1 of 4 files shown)


joemull (changes requested):
I think using {% blocktrans trimmed %} and introducing some linebreaks would help a lot with readability here:

```html
<li>{% blocktrans trimmed with password_length=request.press.password_length %}
  Your password should be {{ password_length }} characters long.
{% endblocktrans %}</li>
{% if request.press.password_upper %}
  <li>{% blocktrans trimmed %}
    Your password should contain at least one upper case letter.
  {% endblocktrans %}</li>
{% endif %}
{% if request.press.password_number %}
  <li>{% blocktrans trimmed %}
    Your password should contain at least one number.
  {% endblocktrans %}</li>
{% endif %}
```

Could we also just move these three blocks to one common template use `include`?

joemull (approved):

mauromsl (approved):

Issue #3899, opened 2024-01-19
Context lost during account registration


joemull:
## User story
As a prospective library supporter, I might fill out the subsidy fee calculator to check my fee, then click start signup, where I will then be asked to log in (for email verification and spam blocking).

But after I finish the account signup process, I am not brought back to the supporter signup page, and I'm not given an explanation for why. This means I will have to realize what has happened, relocate the signup page and refill in my info from the calculator page that would otherwise be carried through with GET params.

## Janeway version
1.5.1

## Expected behavior

Scenario: I try to access a page that requires a login and I am redirected.

- [x] As a logged-out account holder, I am brought back to the target page after logging in
- [x] As a logged-out account holder who has forgotten my password, I am brought back to the target page after resetting my password
- [x] As someone with no account, I am brought back to the target page after creating and activating an account
- [x] As a logged-out ORCID holder, I am brought back to the target page after signing in via the ORCID website
- [x] The new templates created for https://github.com/openlibhums/janeway/issues/4380 are edited

## Context

We do not fully implement the Django authentication system. If we did, managing this would be a lot easier. But we want to maintain our own sets of views so that we can support ORCID authentication.

joemull:
@ajrbyers you mentioned this week that OIDC requires a next URL, which is why one is provided here:

https://github.com/BirkbeckCTP/janeway/blob/15a708d337dce9679df488a5e2753aefb9858d1c/src/themes/OLH/templates/core/login.html#L31

Can you say why it sends people to the site index, rather than the dashboard, which is where the other login pathways end?

ajrbyers:
> @ajrbyers you mentioned this week that OIDC requires a next URL, which is why one is provided here:
> 
> https://github.com/BirkbeckCTP/janeway/blob/15a708d337dce9679df488a5e2753aefb9858d1c/src/themes/OLH/templates/core/login.html#L31
> 
> Can you say why it sends people to the site index, rather than the dashboard, which is where the other login pathways end?

It can be altered to go to the dashboard if prefered.

PR #4279, opened 2024-06-14
ORCiD info is prefilled when registering

### LINKED ISSUE #3337: Allow users to prefill registration form with ORCID record data
As a Janeway user, I'd like to use ORCID to register a new Janeway account and reduce the amount of time I spend filling in forms. 

After connecting my ORCID iD to Janeway, Janeway should use the resulting access token to obtain data from my ORCID record and fill it in the Registration form. My expectation is that data required to register a Janeway account and which is not visible to Janeway's ORCID API client can be filled in be me when I complete the registration form with my password, etc. 

---- 

#### Disclaimer 

To the best of our ability, ORCID Janeway functionality should be built with the understanding that a) current ORCID support is for the Public API only, and b) eventually there will be a toggle setting at the press/system level which allows the admin to choose Public API features or Member API features. The Member API would involve slightly calls (``pub.orcid.org`` vs ``api.orcid.org``) and a different scope in the API (``/authenticate`` vs ``/read-limited`` & ``/update``). 

### What 

1. A user should have the option to "Register with ORCID": 

   a. During registration, the user should be prompted at the top of the registration form to "Register with ORCID". 

   b. When a user tries to sign in with an ORCID iD that is not already in the system, they should be prompted to sign into an existing account or "register with ORCID", i.e. using information from their ORCID record. 

2. When the user chooses to "Register with ORCID", either send a call to read their ``/record`` from the ORCID API and look for relevant details, or send a call to read specific parts of their record: 

    a. Name: read: ``/personal-details``: ``personal-details:given-names`` (First Name) & ``personal-details:family-name`` (Last Name) 

    b. Email: read: ``/email``: ``email:email``

    c. Affiliation: read ``/educations`` & ``/employments``, or ``/activities``, or ``/record``: ``activities:affiliation-group`` If there are multiple listed affiliations, then prompt the User to choose their desired affiliation. 

    d. Country: Determine from chosen affiliation's data 

3. The user's authenticated ORCID iD should be displayed in the registration "form", to confirm that it is actively connected to their account. There should be an option to remove the ORCID iD, which should a) remove the iD from the registration form/account, and b) simultaneously [revoke any received access tokens](https://info.orcid.org/ufaqs/how-can-i-revoke-tokens/)

image

Example: registration form pre-filled with ORCID record data from https://sandbox.orcid.org/0000-0003-3009-4952 

image

everreau:
Closes #3337.

- Add functionality to prefill registration field from orcid info
- Allow user to prefill form from orcid but remove orcid (for some reason, it was in the spec from Alainna)
- Display orcid to their display standards
- Test it.


Bbkctp:

ajrbyers:
Jenkins: Whitelist

ajrbyers:
Jenkins: test this please

ajrbyers:
For clarification it allows the user to remove the ORCID but doesn’t remove it automatically?

everreau:
@alainna 1. hm.. if that's the use case for the remove orcid link I can probably just make it simpler.  This remove the orcid removes the id but keeps the form auto-filled.  (This seemed unnecessarily complicated to me but that was how I interpreted the spec).  If it's just for the case that it's the wrong orcid we could just link back to the original form with no token.
2. I don't know where you're refering to HTTP v HTTPS There is some http in the tests but that's not really related to what will show in real life.  In the real implementation it grabs the orcid url from the orcid response.

everreau:
@alainna Leaving the form auto-filled is not simpler and it seems like a weird thing to do.  "I want to use my orcid data to fill the form but I do NOT want to save that orcid". However, I already implemented this functionality so I can leave it as-is.  I was just asking to make it simpler. If we don't simplify it then they can just click the "Register" link again to get a clean form.

src/utils/tests.py
```diff
@@ -1103,3 +1104,37 @@ def test_mailgun_webhook(self):
...
+        self.assertIsNone(details["last_name"])
+        self.assertEqual(len(details["emails"]), 0)
+        self.assertIsNone(details["affiliation"])
+        self.assertIsNone(details["country"])
+
+    def test_redirect_uri(self):
+        press= helpers.create_press()
+        repo = helpers.create_repository(press, [], [])
+        self.assertEqual(build_redirect_uri(repo), "http://localhost/login/orcid/?action=login")
+        self.assertEqual(build_redirect_uri(repo, action="register"), "http://localhost/login/orcid/?action=register")
```
joemull, line 1140: This test is failing for me because it's missing some data around the request. But once that data is provided and the function runs, I think it will still not be a great idea to pass query parameters in the `path` argument of `site_url` and `build_url` functions. I understand why you are putting it in the path and not a query dict or something, because [we don't provide a way to pass query parameters](https://github.com/BirkbeckCTP/janeway/issues/4321) into that function and even if you could, the [URL encoding is redundant](https://github.com/BirkbeckCTP/janeway/issues/4291) so the URL would be garbled on its way back from ORCiD.

I discovered these things while working on a separate feature, #3899, whereby users will be carried through the login and registration process with a bit more care. It's still a work in progress but I'm thinking of expanding simplifying and unifying our token models so that we can retrieve next URLs and query parameters like this, even when we send things out to ORCiD.

If we were to fix up the tokens, we could make use of ORCiD's [`state` parameter](https://info.orcid.org/documentation/integration-guide/customizing-the-sign-in-register-screen/#h-identify-the-researcher-by-a-custom-state-parameter), which would let us link up the appropriate "next" url, whether login or registration (your feature) or something else (my feature).

I know this is a lot but I am curious what your thoughts are. If you want to glance over my still-messy WIP, it's here: https://github.com/BirkbeckCTP/janeway/pull/4322

joemull (changes requested):
Thanks @everreau--I think this is good but it is pulling on some pre-existing spaghetti we have around URL formation and carrying users through redirects that I think we need to address a bit more holistically. I happen to be working on another feature that touches the same code, so I am sharing some information related to that below. See comments inline.

As for the feature to remove the ORCID, I think we should take the additional context from Alainna and simplify it like you suggest. I think it would be counter-intuitive to offer just the deletion of the ORCID, without removing the rest of the form details, because the user will wonder what they are actually doing when they submit the form after that. Are they still connecting their orcid account? Maybe some kind of "Start over" or "Retry" button would be clearer?

src/utils/tests.py
everreau: hi @joemull -- sorry for my delayed response.  I got waylaid by other tasks and some time off. I've checked in corrections to the more minor issues you've pointed out.

I like your idea for fixing the return url business.  I'm also thinking we might be able to use this if fix something else Alainna requested. She doesn't want to have to add each new journal/repo url to the redirect urls but instead have a press or installation wide return url that is able to redirect appropriately.

everreau: @joemull How do you want to proceed with this work?  We'd like to get this and a few other orcid features that would depend on this rewrite into the next release.

joemull: @everreau I think unifying the various `Token` models to a single one and passing the token ID in `state` to the ORCID API is the right direction. I could potentially separate that work from the rest of my branch and put it through review with higher priority. You could then rebase this branch on that and finish up this work.

I'd have to run it by @mauromsl and @ajrbyers to confirm they like the solution and figure out a time frame. It will be tight, because we are hoping to finish up 1.7 in the next few weeks.

Shall I do that and let you know?

Or if you have time to handle it ASAP, would you want to copy-paste and fix up my WIP code that unifies the Token objects and sends and receives `state` as a part of this PR?

everreau: @joemull I can probably do it. I think this is my highest priority task at the moment. I'll check with Chad to be sure. Thank you!

joemull: That would be super! Let me know if I can help with context.

joemull: @everreau I just had a chat with Andy and Mauro, and can we actually ask you to hold off on this? We need a chance to discuss it properly.

everreau: @joemull no problem.  Chad has indeed deemed this my most important task so we'd like to get going as soon as we can.  Thanks!

everreau:
@joemull updated code to use state.  If the test above is still failing for you please let me know what is missing.  It's passing for me.

joemull (approved):
This looks good to me. I also tested it out manually to make sure I understood it.

As for the whole `state` thing, it's not quite what I had in mind but that's OK because your solution works. Just a heads up, I may refactor it a bit when I get to the next URL branch. OK If I add you as a reviewer on that PR when it is ready, to make sure I don't mess this up?

The test that was failing for me was only failing when I ran the orcid tests in isolation, and I also might have a fix for it in my refactor, so I am OK with merging this. I want to make sure we don't hold you up any more than we have already.

joemull:
Oh, there are merge conflicts on this branch. Normally we ask the branch owner do a git rebase locally onto a later version and then force push. Is that alright with you?

ajrbyers (approved):
I'm happy with this. @mauromsl is also going to look over it as he requested some of the changes.

mauromsl (approved):

mauromsl:
Thanks @everreau!

As Joe mentioned, there are a few conflicts that need resolving, would you mind performing a rebase of `master` as per our remote, ahead of merging to master?

everreau:
no problem @mauromsl , I'll have it done by end of day. @joemull I'm happy to review your PR when ready.  As I'm sure you're aware this whole login page could use some rewriting but it sounded like you all wanted to hold off on that.

PR #4281, opened 2024-06-17
Unified user list

### LINKED ISSUE #4185: General-purpose user list
## User need

The manager tools for managing users and roles are single-purpose and don't provide a wide or flexible enough view for many use cases at the press level.

## Proposed solution

Add a general purpose view at the press level called "All Users" or similar that allows searching by a range of text fields (perhaps the same ones as the admin view search for the Account model) and filtering by facets such as journal and role, to begin with. The email address should be visible in the results list.

I think the filterable CBV should be used with the new admin-based search function I introduced recently.

We could potentially sunset a few of the existing user management tools if their functionality is served by the new list.

joemull:
Closes #4185.

This introduces a unified list of accounts that users with different permissions can search and filter at both the journal and press levels.

## Journal Users
All the journal's reviewers:
image

Adding a role:
image

## All Users
All the staff accounts:
image

Adding a role from the press level:
image

## Deprecations and further work
I was able to unify many of the existing features of the previous user lists, with these exceptions:

### Reviewer interest management

@ajrbyers did some work on this recently. To keep this work, I've added a new item to the reviewer settings called "Reviewer Interests" that loads the old reviewer list.

image

In the future, I think we could expand on this work by creating a CBV view for user interests, with similar functionality to the new user+role list. 

### Cross-journal management of roles by editors

An affordance of the old Enrol Users list was the ability for an editor to search outside their journal to find people that should have a role in their journal, so they can add them. I could not implement that feature in this list without introducing a lot of confusion about privileges, press vs. journal level, and hard-coded URLs. We should probably reconsider the feature and determine the best way to offer it going forward.


ajrbyers (changes requested):
I pulled this and had a play around with this and this is a **massive** step in the right direction. The search and filter options work really well and will help us eliminate numerous pages.


**General comment:**
As we discussed in the 18th June standup this page will benefit a lot from the introduction of HTMX. Perhaps the list of roles could be converted into labels (similar to active/inactive & staff/superuser) and have an HTMX rendered modal/form that would allow the user to select or unselect roles from a multi-select list rather than having to add/remove them one at a time.

**Requested changes:**
If we are to incorporate this into v1.7 I think we need to leave the current pages as they are and introduce this as a beta page.

- [x] Remove deprecation warning (not sure if when out of debug this is ignored but locally it presents a Django page
- [x] Remove `(old)` labels
- [x] Remove link to Reviewer Interests
- [ ] Mark the new feature as new

src/core/views.py
```diff
@@ -2629,3 +2635,115 @@ def __init__(self, *args, **kwargs):
...
+            choices = [(None, '---------')]
+            choices.extend([(n.journal.pk, n.value) for n in journal_names])
+            accountrole_form.fields['journal'].widget.choices = choices
+        context['accountrole_form'] = accountrole_form
+
+        return context
+
+    def post(self, request, *args, **kwargs):
+
+        post_data = request.POST.copy()
```
ajrbyers, line 2727: Rather than copy post data so that journal can be inserted into it would be good to handle this in the same way we have before by passing `journal` into the form as a kwarg and handle it in the `__init__` and `save` methods.

ajrbyers (changes requested):
Thanks for those changes. This is looking great. I've added one request around the `AccountRoleForm`.

src/core/views.py
joemull: Thanks @ajrbyers -- turns out those lines were only needed by an earlier draft of this, so I've cleaned them up.

ajrbyers (approved):
🎖️

src/templates/admin/elements/removable_role_chips.html
```diff
@@ -0,0 +1,28 @@
...
+  <div class="flex">
+    {% csrf_token %}
+    <input hidden name="params_string" value="{{ params_string }}">
+    {% for accountrole in accountroles %}
+      {% if accountrole.journal == journal %}
+        <div class="small button-group">
+          <span class="small button secondary chip">
+            {{ accountrole.role.name }}
+          </span>
+          <button
```
mauromsl, line 11: Interesting design choice. I find the <span> elements styled as buttons a bit confusing, do we really need the split instead of having a single button? It reminds me of the accesibility issues that often arise around [checkboxes and their labels](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role)

mauromsl (changes requested):
thanks @joemull, I quite like this single view. As mentioned by @ajrbyers, this view would benefit from some added interactivity with HTMX, since we are loosing some dynamic behaviour we currently have in search (datatables) and role management (ajax). Hopefully one step back and two steps forward in that case.

joemull:
Thanks @mauromsl -- I've taken these comments on board.

It now looks a little different:

Editor's journal-level view
Image

Super user's press-level view
image

Mobile
image
image

mauromsl (approved):
Awesome! love the new changes!

PR #4284, opened 2024-06-18
Fixes unclosed footer link in clean theme

### LINKED ISSUE #4285: Unclosed <a> tag in clean theme footer
**Describe the bug**
There is an unclosed <a> tag in the clean theme's footer.

**Janeway version**
1.5.x & 1.6.x

ajrbyers:
Closes #4285


joemull (changes requested):
Would it be worth just breaking this line up a bit so that it is easier to read the nesting? I dislike making nitpick linting comments but I do believe that long HTML lines are partially to blame for bugs like this, because they make it harder to see when tags and template blocks like are not closed properly.

joemull (approved):

StephDriver (approved):
Tested and works.

PR #3847, opened 2023-12-18
initial solution, but found related bugs

StephDriver:
addresses #3746 but further issues found, and when tested, same behaviour was seen on master.

- production files are marked as manuscript files
- proofing files cause an apparently infinite loop of uploads, when test-server is halted, the file is found to have been uploaded multiple times.

to replicate issues:
1. go to document manager
2. upload a proofing file or a production file.

PR #4292, opened 2024-06-24
3746 document manager upload improvements

### LINKED ISSUE #3746: Document manager file upload error
**Describe the bug**
When a user attempts to upload a file  but clicks an upload button that does not correspond to the document field it causes a server error

**Janeway version**
1.5.0

**To Reproduce**
1. Go to document management for any article
2. Click "Upload File"
3. Fill in file label
4. Select a document under "New Manuscript File"
5. Click the "Upload" button in the "Figure/Data File" section

**Expected behavior**
Probably this should cause a form error indicating there is no file to upload but also the interface could be changed so there aren't a bunch of different upload buttons.

## Front-end Issues

 - MacOS 13.2.1
 - Firefox 118.0.1

StephDriver:
### Work done
Taking into account the changes requested on #3847 I have started again and addressed this issue.

The request was for a select for the file type, rather than a popup for each file type.  On careful consideration, I implemented this with a radio group instead, so the file type can be selected and there is only one popup.  The change to radio group rather than select is because each file type has both name and description, and the user needs to choose which file - this is more possible when all file types are onscreen at the same time (i.e. with a radio group) than when the description is within a select so that it is only visible when viewing the list or for the selected item. 

I did consider having the file descriptions separate from the file selection, but this again felt less helpful to the user.  If they have to choose between file types, far better for them to mark it adjacent to the description, than have to look elsewhere for that description.

As I was substantially rewriting  this modal, I have also made minor accessibility improvements to the
- heading structure (modal starts with an H2 and other headings now consistent), 
- the icons (marked as decorative, including the * for required field)
- the placeholder text now serves as an example rather than an instruction.

### Before:
image

### After:
image

Closes #3746

### DIFF
--- a/src/journal/views.py
+++ b/src/journal/views.py
@@ -2443,22 +2443,23 @@ def document_management(request, article_id):
     if request.POST and request.FILES:
 
         label = request.POST.get('label') if request.POST.get('label') else 'File'
+        file_type = request.POST.get('file-type-chooser', None)
 
-        if 'manu' in request.POST:
+        if file_type == 'manu':
             from core import files as core_files
-            file = request.FILES.get('manu-file')
+            file = request.FILES.get('new-file')
             new_file = core_files.save_file_to_article(file, document_article,
                                                        request.user, label=label, is_galley=False)
             document_article.manuscript_files.add(new_file)
             messages.add_message(
                 request,
                 messages.SUCCESS,
-                _('Production file uploaded.'),
+                _('Manuscript file uploaded.'),  
             )
 
-        if 'fig' in request.POST:
+        if file_type =='fig':
             from core import files as core_files
-            file = request.FILES.get('fig-file')
+            file = request.FILES.get('new-file')
             new_file = core_files.save_file_to_article(
                 file,
                 document_article,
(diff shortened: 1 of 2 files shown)


ajrbyers (changes requested):
Sorry, Steph, you're going to hate me... but... we should probably have this as a Django form to reduce some of the view logic.

If you need a hand with that process let me know and we can slot in somre pair programming time.

StephDriver:
image

for reference when converting this into a form.

StephDriver:
> Sorry, Steph, you're going to hate me... but... we should probably have this as a Django form to reduce some of the view logic.
> 
> If you need a hand with that process let me know and we can slot in somre pair programming time.

@ajrbyers I've made a start on this while going through training videos on forms (see WIP commit) but I would appreciate pair programming to finish it off.  

I have moved this part of the document_management.html template into a new document_upload.html template, and likewise split its view and then created a stub for forms.py and added the new DocumentUploadForm to the __init__,py but I am not sure how to tie it all together.

StephDriver:
Work before next pair prog:
- [ ] update form aria labels

Items remaining for next pair prog:

- [ ] Create a urls.py entry for the new view
- [ ] Set the document upload form to post to the new view
- [ ] Add security decorators to the new view
- [ ] Move save file logic to form.save()

StephDriver:
> Work before next pair prog:
> 
> * [ ]  update form aria labels

removed aria label from Radio select, because 1, its redundant as radio buttons have semantic meaning anyway, and 2. when added that same label is added to each and every choice, which isn't helpful.

StephDriver:
pair prog 2.

next 
- [ ] write the save override function in forms.py using views.py file_type if logic as basis, and the print(data) line to see what is being passed in.

StephDriver:
- [ ] before review, will need to rebase and particularly fix conflicts with  #4327

StephDriver:
based on above, to summarise 
 - write the save override function in forms.py using views.py file_type if logic as basis, and the print(data) line to see what is being passed in.
 - Create a urls.py entry for the new view
 - Set the document upload form to post to the new view
 - Add security decorators to the new view
 - Move save file logic to form.save()

and then finally handle the rebase carefully re 4327 as above.

Issue #3911, opened 2024-01-24
container element is empty


StephDriver:
Some lists where there are no list elements are being put out as empty containers, rather than being skipped when there are no list elements to include.

example at https://www.iastatedigitalpress.com/iscore/article/id/12400/
with empty download and jump to lists. 

More generally, we need to review how our lists are populated such that if there are no list items we don't insert an empty list.  While this may not be visible, it will be picked up by screen-readers.

ajrbyers:
AB and SD: Likely caused by a forloop without a {% if %} check.

StephDriver:
was closed by #4293

PR #4293, opened 2024-06-24
3911 do not render empty lists on article page

StephDriver:
Updated django templates to check that there are galleys before creating download and checksum lists, across all three themes. Material already checked for galleys for download lists, but not for checksums.

For empty autogenerated tables of contents in OLH theme, updated the JS to remove the `toc` element if no headings were found, i.e. if list not populated.

Marked as draft as it is the end of the day, and I have yet to test this, so that must be done before sending for review.

### DIFF
--- a/src/themes/OLH/templates/journal/article.html
+++ b/src/themes/OLH/templates/journal/article.html
@@ -372,12 +372,12 @@
                                                 <i class="fa fa-info"></i>
                                                 Note to proofreader: Download links on this page begin to work when the article is published.
                                             </p>
-                                        {% else %}
-                                            <p> {% trans 'Downloads are not available until this article is published ' %}</p>
                                         {% endif %} 
                                     {% else %}
                                         <p> {% trans 'Downloads are not available for this article.' %}</p>
                                     {% endif %}
+                                {% else %}
+                                    <p> {% trans 'Downloads are not available until this article is published ' %}</p>
                                 {% endif %}
                             </div>
                             <div class="section">
--- a/src/themes/OLH/templates/journal/article.html
+++ b/src/themes/OLH/templates/journal/article.html
@@ -351,19 +351,24 @@
                             <div class="section">
                                 {% if article.is_published or proofing %}
                                     <h3>{% trans "Download" %}</h3>
-                                    <ul>
-                                        {% for galley in galleys %}
-                                            <li>
-                                                <a href="{% url 'article_download_galley' article.id galley.id %}">{% trans "Download" %} {{ galley.label }}</a>
-                                                {% if galley.file.mime_type == 'application/pdf' and journal.view_pdf_button %}
-                                                    </li>
-                                                    <li>
-                                                    <a target="_blank" href="{% url 'article_view_galley' article.id galley.id %}">{% trans 'View PDF' %}</a>
-                                                    </li>
-                                                {% endif %}
-                                            </li>
-                                        {% endfor %}
-                                    </ul>
+                                    {% if galleys %}
+                                        <ul>
+                                            {% for galley in galleys %}
+                                                <li>
+                                                    <a href="{% url 'article_download_galley' article.id galley.id %}">{% trans "Download" %} {{ galley.label }}</a>
+                                                    {% if galley.file.mime_type == 'application/pdf' and journal.view_pdf_button %}
+                                                        </li>
+                                                        <li>
+                                                        <a target="_blank" href="{% url 'article_view_galley' article.id galley.id %}">{% trans 'View PDF' %}</a>
+                                                        </li>
+                                                    {% endif %}
+                                                </li>
+                                            {% endfor %}
+                                        </ul>
+                                    {% else %}
+                                        <p> {% trans 'Downloads are not available for this article.' %}</p>
+                                    {% endif %}
+
                                     {% if proofing %}
                                         <p id="note_to_proofreader_2">
                                             <i class="fa fa-info"></i>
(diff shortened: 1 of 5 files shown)


StephDriver:
### Download and Checksum were addressed on the template.

Before:
image

```html
<div class="section">
   <h3>Download</h3>
    <ul> 
    </ul>
</div>
```
image

After:

```html
<div class="section">                             
    <h3>Download</h3>
     <p> Downloads are not available for this article.</p>
</div>
```

### Autogenerated TOC addressed in the JS
This removes the "jump to" section completely.

StephDriver:
### To Remove Empty Sections or Not to Remove...
For Checksum and Download, I have left the section there but added a statement that there are none available.  This is because these are standard sections that a reader might 'look for' and so they should not be hidden but state clearly that they are empty.

But for the automated table of contents, I have removed the section entirely when it is empty.  This is because the table of contents listing the sections does not make any sense in the context of an article without contents - and a reader would not look for it when there is no contents to navigate. To follow the pattern above with downloads and checksums would add to user confusion not prevent it.

StephDriver:
Closes #3911

src/themes/OLH/templates/journal/article.html
```diff
@@ -351,19 +351,24 @@ <h4>{% trans "Authors" %}</h4>
...
+                                                </li>
+                                            {% endfor %}
+                                        </ul>
+                                    {% else %}
+                                        <p> {% trans 'Downloads are not available for this article.' %}</p>
+                                    {% endif %}
+
                                     {% if proofing %}
                                         <p id="note_to_proofreader_2">
                                             <i class="fa fa-info"></i>
```
joemull, line 374: This note to the proofreader should show up only if there are galleys, as otherwise it may be taken to mean that there may be galleys that are not showing, with the new "Downloads are not available" message. Could we put it inside `if galleys` above?

joemull (changes requested):
Nice one! I like the design decision to consistently display sections that are standard and often expected, like downloads and checksums.

I noticed a few things that we probably want to fix up for consistency between themes and screen sizes.

src/themes/OLH/templates/journal/article.html
StephDriver: done

src/themes/clean/templates/journal/article.html
```diff
@@ -100,15 +100,19 @@ <h2>{% trans "Abstract" %}</h2>
                     <div class="d-lg-none d-xl-none d-md-none">
```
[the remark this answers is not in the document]
StephDriver: yes, ooops missed that!

src/themes/material/templates/journal/article.html
```diff
@@ -254,13 +254,13 @@ <h4>
...
-                            target="_blank"><i class="fa fa-facebook"></i></a>
+                            target="_blank" aria-label= {% trans "Share on Facebook" %}><i class="fa fa-facebook"></i></a>
                         <a class="waves-effect waves-light btn btn-small twitter-bg" 
                             href="https://twitter.com/intent/tweet?text={{ article.title }} {% if doi_url %}{{ doi_url }}{% else %}{{ article.url }}{% endif %}"
-                            target="_blank"><i class="fa fa-twitter"></i></a>
+                            target="_blank" aria-label={% trans "Share on X" %}><i class="fa fa-twitter"></i></a>
                         <a class="waves-effect waves-light btn btn-small linkedin-bg"
                             href="https://www.linkedin.com/sharing/share-offsite?url={% if doi_url %}{{ doi_url }}{% else %}{{ article.url }}{% endif %}"
-                            target="_blank"><i class="fa fa-linkedin"></i></a></button>
+                            target="_blank" aria-label={% trans "Share on Linked In" %}><i class="fa fa-linkedin"></i></a></button>
```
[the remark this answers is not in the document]
StephDriver: This has made me really think - the aria-label is to be read aloud, so I was writing it as it would be said, so I've had to check weather a screen reader will read LinkedIn intelligibly - and you are right, it reads it fine!

src/themes/material/templates/journal/article.html
```diff
@@ -473,13 +473,17 @@ <h4>{% trans field.field.name %}</h4>
                         <h4>
                             {% trans "File Checksums" %} (MD5)
                         </h4>
```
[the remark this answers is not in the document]
StephDriver: I'm searching each of the article files for Download and Checksum areas and ensuring consistency of this behaviour across all of them.

NB as far as I can tell, we only show Checksums at large screen-width.  Is that intentional?

StephDriver:
Summary of Changes:

- I have made the changes requested.
- I retested the new functionality (as described in previous comment) and updated where needed.
- I found that the Material and Clean themes use the same toc.js asset, which is in the Material directory, this seems to be a carbon copy of the toc.js in the OLH assets.  I updated that toc.js and added the 'toc-section' id to the article.html files so that the toc-section can be hidden when no headings have been found while generating the toc. 
- The material theme tested fine, because it has other functionality which suppresses the TOC for short articles, such as the test example. The TOC card is not visible until one scrolls further down the page, and the page is too short to reach that point which is why it appeared to work correctly before the material toc.js update.  I decided that for consistency, I would update the material article.html to be consistent with the other two themes, despite this... which means if later updates are made to the material theme which change when the TOC card is visible, such that it is visible on shorter pages, the purpose of this issue PR, to hide the TOC when there is no content for it, will then function consistently with the other two themes.   But I am unable to directly test it working - I can note that it is exactly the same as in the Clean theme and that does work.

src/themes/clean/templates/journal/article.html
```diff
@@ -118,7 +121,6 @@ <h2>{% trans "Abstract" %}</h2>
                         {% endif %}
```
joemull, line 121: This is no longer inside the if galleys check.

joemull (changes requested):
Amazingly thorough! I just noticed one proofreader note that got left out of one of the galley checks.

src/themes/clean/templates/journal/article.html
StephDriver: ta, its been a bit of a monster to find all the relevant bits, so I'm glad you spotted that.

joemull (approved):

ajrbyers (changes requested):
Thanks for this @StephDriver this looks awesome. However, there is a conflict being flagged. Looks like an easy one though. If you need assistance with it let me know.

PR #4311, opened 2024-07-03
Crossref deposit improvements

### LINKED ISSUE #3131: Crossref deposit adjustments
The below changes are requested by a publishing librarian:

- [x] Add our `first_page` and `last_page` fields to the XML. 

- [x] The proceedings title should include exactly what is in the journal name and not end with a final space hyphen space. 

    ```xml
    <proceedings_title>Proceedings of the 7th International Digital Human Modeling Symposium (DHM 2022) and Iowa Virtual Human Summit 2022 - </proceedings_title>
    ```

- [x] The publisher should come from the journal publisher field not from the press as a whole

    ```xml
    <publisher>
        <publisher_name>University of Iowa Libraries Publishing</publisher_name>
    </publisher>
    ```

- [ ] The article number is important for citation. There is not an article number field in crossref. However, looking at PLoS One (looking at their schema tags and also looking at how it imports into Zotero) I think the article number should map to the first page number. JM question: Can you clarify what you mean here? Are you saying the crossref deposit should use Janeway's article number in Crossref's first page field if there's no first page in Janeway? 

- [ ] Map our `page_numbers` field to Crossref's field `other_pages`. JM note: It looks like Crossref's field `other_pages` is used for items with non-contiguous page information. That's different from our page_numbers field, which is for a custom page range, like with roman numerals or something. I think we should not populate it with our custom page range field because the systems ingesting this info will not expect that field to duplicate info in first_page and last_page, just provide extra info. Something I'm not considering?

joemull:
This completes 3 of 5 requests in #3131:

- map Janeway's first page and last page fields to Crossref
- output clean proceedings title
- respect custom publisher names set at journal level

Depending on if the reporter is happy with these fixes alone, closes #3131.

### DIFF
--- a/src/templates/common/identifiers/crossref_conference.xml
+++ b/src/templates/common/identifiers/crossref_conference.xml
@@ -4,17 +4,19 @@
                 {% if crossref_issue.issue.issue_title %}
                 <conference_name>{{ crossref_issue.issue.issue_title }}</conference_name>
                 {% else %}
-                <conference_name>{{ crossref_issue.journal.title }} - {{ crossref_issue.issue.year }}</conference_name>
+                <conference_name>{{ crossref_issue.journal.title }}</conference_name>
                 {% endif %}
             </event_metadata>
             <proceedings_metadata>
                 {% if crossref_issue.issue.issue_title %}
                 <proceedings_title>{{ crossref_issue.issue.issue_title }}</proceedings_title>
                 {% else %}
-                <proceedings_title>{{ crossref_issue.journal.title }} - {{ crossref_issue.issue.year }}</proceedings_title>
+                <proceedings_title>{{ crossref_issue.journal.title }}</proceedings_title>
                 {% endif %}
                 <publisher>
-                  <publisher_name>{{ crossref_issue.issue.journal.press.name }}</publisher_name>
+                    <publisher_name>
+                        {{ crossref_issue.issue.publisher }}
+                    </publisher_name>
                 </publisher>
                 <publication_date media_type="online">
                     <month>{{ crossref_issue.issue.date.month }}</month>
--- a/src/identifiers/logic.py
+++ b/src/identifiers/logic.py
@@ -339,7 +339,8 @@ def create_crossref_article_context(article, identifier=None):
         'date_accepted': article.date_accepted,
         'date_published': article.date_published,
         'license': article.license.url if article.license else '',
-        'pages': article.page_numbers,
+        'first_page': article.first_page,
+        'last_page': article.last_page,
         'scheduled': article.scheduled_for_publication,
     }
 
--- a/src/identifiers/logic.py
+++ b/src/identifiers/logic.py
@@ -341,6 +341,7 @@ def create_crossref_article_context(article, identifier=None):
         'license': article.license.url if article.license else '',
         'first_page': article.first_page,
         'last_page': article.last_page,
+        'other_pages': article.page_numbers,
         'scheduled': article.scheduled_for_publication,
     }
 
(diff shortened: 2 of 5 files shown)


ajrbyers (changes requested):
This looks fine, though perhaps we should include the Janeway field `pages` as `other_pages` here when present? There is also a failing test.

joemull:
I've now rebased it onto master to avoid failing tests, but it has the same number of failing tests as master: FAILED (failures=3, errors=54) -- some of these are plugin-related.

ajrbyers (approved):

src/templates/common/identifiers/crossref_conference.xml
```diff
@@ -4,17 +4,19 @@
...
-                <conference_name>{{ crossref_issue.journal.title }} - {{ crossref_issue.issue.year }}</conference_name>
+                <conference_name>{{ crossref_issue.journal.title }}</conference_name>
                 {% endif %}
             </event_metadata>
             <proceedings_metadata>
                 {% if crossref_issue.issue.issue_title %}
                 <proceedings_title>{{ crossref_issue.issue.issue_title }}</proceedings_title>
                 {% else %}
-                <proceedings_title>{{ crossref_issue.journal.title }} - {{ crossref_issue.issue.year }}</proceedings_title>
+                <proceedings_title>{{ crossref_issue.journal.title }}</proceedings_title>
```
mauromsl, line 14: I think this can be a breaking change for a lot of folks. When a conference proceedings "journal" has articles for multiple years, they can't set a single journal title that works. I need the logic has to be altered on `inderifiers.logic` to cover the following cases:

 - When an article has a publication_title, that always takes precedece.
 - When a conference title is used and the issue has a year, preserve the existing functionality
 - when a conference title is used but no year is available in the issue, render just the title.

mauromsl (changes requested):

mauromsl (changes requested):
@joemull I think the reporter is right about this comment

>     * [ ]  Map our `page_numbers` field to Crossref's field `other_pages`. JM note: It looks like Crossref's field `other_pages` is used for items with non-contiguous page information. That's different from our page_numbers field, which is for a custom page range, like with roman numerals or something. I think we should not populate it with our custom page range field because the systems ingesting this info will not expect that field to duplicate info in first_page and last_page, just provide extra info. Something I'm not considering?

Yes, `page_numbers` in Janeway doesn't have an explicit use case, but its main use case it is still for non-contiguous page ranges. There is also no reason why roman numerals can't be part of this field according to Crossref schema, so I think that use case is safe too. Then I'd be happy to close #3131

src/templates/common/identifiers/crossref_conference.xml
joemull: I don't think so, because `issue.year` has never worked. There is no `journal.models.Issue.year` property or method.

This is the original commit that introduced it:
https://github.com/openlibhums/janeway/blob/eda3e9cf58b0142d218c21b789462f13949e707b/src/templates/common/identifiers/crossref_conference.xml#L32

This is a more recent commit that changed the line but preserved the bug:
https://github.com/openlibhums/janeway/blob/36e06831d72f5ba586e580faf567a7a3dc6e0705/src/templates/common/identifiers/crossref_conference.xml#L14

To chart a way forward, I think we should omit the year argument for now, and longer term create some of the metadata fields that would make it possible to structure this deposit in the way Crossref recommends:
 https://www.crossref.org/documentation/schema-library/markup-guide-record-types/conference-proceedings/

joemull:
> There is also no reason why roman numerals can't be part of this field according to Crossref schema, so I think that use case is safe too.

Alright, that's fine with me--I've made that change.

mauromsl (approved):

PR #4316, opened 2024-07-06
Fix migration order error

yakky:
There is a migration order mismatch on `master`


Bbkctp:

ajrbyers:
Thanks @yakky. This is happening more and more where branches are merged and we’ve been discussing how to mitigate. 

Jenkins: test this please

yakky:
@ajrbyers probably adding `python src/manage.py makemigrations --check` at the start of `jenkins/entrypoint.sh` would allow to detect such errors when one rebases an open PR before merge and update migrations accordingly

yakky:
Closing this because I pushed an extra commit that shouldn't be there. I will reopen using a clean commit

Issue #785, opened 2019-01-23
Collection on Submission


ajrbyers:
Editors may want authors to submit articles for a collection so we should:

- [ ] Add a boolean that marks a collection as open for submission
- [ ] Add a setting to control if the field appears or not
- [ ] List available collections for the author to submit to
- [ ] Display this in the metadata popup/unassigned page

yakky:
@ajrbyers this is a proposal to fulfil the above requirements. We (at sissa medialab) we are already working on something along the lines of what is outlined below and we can take care of the implementation 

# Changes to support special issues / Collections

The goal of this change is to enrich the basic Issue model in order to implement more flexibility in the submission process in order to create issues with "special" policies / automation which might be useful for some journals, but are at the same time broad and flexible enough to cover a lot of different use cases.

## Enabled features
- Allow users to select an issue to submit to during the submission process based on some restrictions defined for each issue (available section / open-close submission dates, people allowed to post article on the issue)
- Automatically assign articles for review to a number of editors
- Provide richer content for such issues, which add more content / context to render them using customised templates

### Model changes

Model changes are necessary to store additional information to implement special issues / collection advanced behaviour.

- Update `Issue` to add:
	- `short_name`: Name for internal usage, in manager listings etc
	- `description = models.TextField`: special issue may have requirements for a presentation / introduction to be used in listings or elsewhere
	- `open_date = models.DateTimeField`: Use to determine when submission to this special issue is allowed or not
	- `close_date = models.DateTimeField`
	- `invitees = models.ManyToManyField(Account)` one may want to restrict submission to the special issue to a subset of authors. This many to many can be used to restrict the lists of available special issues in the submission process (if empty all users will be allowed to submit to the issue); restriction only applies to currently logged-in (submitting) author; co-authors will not be subject to such restriction; This is useful for example for conference proceedings
	- `allowed_sections = models.ManyToManyField(Section)`: special issues may have a subset of sections attached, this will restrict selectable sections on submission
	- `editors = models.ManyToManyField(Account)` used to automate assigned of submitted articles
    -  `documents` (m2m to `core.File`) to accommodate files needed for administrative or internal uses
    - `expected articles`: to check if the submitted articles is aligned with expectations and postpone deadline
    - `internal notes`

### Changes to Submission process

- The issue selection must be made an extra step in the submission wizard to allow
	- checking the current date against open / close date
	- checking the current user against the `Issue.invitees` section
[...]

PR #4318, opened 2024-07-06
Add automation / metadata for Collections

yakky:
**Disclaimer**

This is just a preliminary PoC to validate the major implementation concepts as discussed in #785 

**Description**

Add additional metadata for the issues to enable some (optional) automatic behavior during the submission process

A new (optional) step is added to allow user to select the issue to submit to if this is enabled by the journal on an issue by issue basis.

If an issue is marked as "Collection" additional fields can be used to "guide" the submission process.

Each case below is optional and they can be used together to implement different submission models

- date based selection: issues are available for selection only if the submission is within the submission timeframe set in the issue (fields: date_open / date_close)
- invitation based selection: issues are available for selection only if the user is among a list on invited users
- sections restriction: sections available in the submit_info step are filtered according to the (optional) issue selected
- automatic editor selection: articles can be assigned automatically to a set of editors assigned to the issue (not implemented yet)

Add additional fields to issue to handle presentation and internal data


Bbkctp:

yakky:
@ajrbyers I would like an early feedback on this approach which implements the core of the change. The submission process is functionally complete, while manager views, editor assignment and other minor features (and tests!) must still be implemented.

Before proceeding I would like to know if it's the right approach according to you.

PR #4327, opened 2024-07-15
Improve usability of file types

### LINKED ISSUE #4326: Usability of file types
## User story
As an editor I know that Janeway understands files in terms of their "type," which is different from the file ending (e.g. DOCX, PDF), and has to do with the purpose of the file in a given workflow. Files could have a "type" of manuscript, figure, data, supplementary, production, galley, proofing, etc. I am often expected to know what the differences between these types are, and how each type of file will behave, where it will be displayed, who it will be visible by, etc. If I want to get files into Janeway so that I can work with them, I have to know which file type I need, and where to upload it.

I am sometimes confused by file types in Janeway. Sometimes it's because of inconsistencies in the language used to describe types:
- I go to upload a Proofing file, but then I get a confirmation saying "Galley uploaded"
- I am expected to upload a "Production" file, but actually what I need in production is manuscript / copyedited files, but I'm not sure what the difference is

Sometimes it's because of inconsistencies between interfaces:
- I can enter a label in some places but not others
- I can upload some types of files through the document manager, but not others

## Short-term solutions
- [ ] Upgrade the document manager to include more (if not all) file types. (An issue is already open to make the document management form clearer -- see #3746.)
- [ ] Bring some basic standardization to the language we use to refer to file types and consistently define the types in file management interfaces across workflows.

## Long-term solution
Longer term, perhaps after https://github.com/BirkbeckCTP/janeway/issues/865, I think we need to do a bit more:
- formalize our file types as constants or an enumeration so that they can be used as a programmatic basis for workflow logic in Python code
- document them with guidance on workflows that are commonly used with each type
- make consistent create, update, and delete forms and interfaces
- make consistent display components that can be customized per user (e.g. always sort latest file first, always sort by alpha file name, etc.)
- standardize validation (e.g. preventing creation of duplicates via checksum)
- standardize settings like which mime types should be allowed for each Janeway file type

joemull:
Closes #4326, or at least the first part of it.

### DIFF
--- a/src/core/templatetags/files.py
+++ b/src/core/templatetags/files.py
@@ -1,6 +1,7 @@
 import os
 
 from django import template
+from django.core.exceptions import FieldError
 from django.template.defaultfilters import filesizeformat
 
 from bs4 import BeautifulSoup
@@ -59,6 +60,15 @@ def file_type(article, file):
     galley_files = [galley.file for galley in article.galley_set.all()]
     galley_sub_files = list()
     review_files = [review.review_file for review in article.reviewassignment_set.all()]
+    try:
+        # GalleyProofing belongs to the typesetting plugin
+        annotated_files = models.File.objects.filter(
+            galleyproofing__round__article=article,
+        )
+    except FieldError:
+        annotated_files = models.File.objects.none(
+            proofingtask__round__article=article,
+        )
 
     for galley in article.galley_set.all():
         for image in galley.images.all():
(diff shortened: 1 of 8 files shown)


ajrbyers (approved):

joemull:
I've rebased this to make it easier to test as it was dependent on #4328.

mauromsl (approved):

PR #4328, opened 2024-07-17
Buttons and labels for complicated forms

joemull:
This adds a few template tags, widgets, and model properties that I used in fixing BirkbeckCTP/typesetting#166.

- A set of buttons that select or deselect all the checkbox inputs in a fieldset (an upgrade of of the jquery-based check_all.js, which could only be used once per page and was not accessible)
- A `detail` method on the galley object to make it possible to provide information about the linked file in a form choice
image

- A set of buttons that make it easier to quickly infill a date
image

- Support for initiating multiple datatables on a single page

### DIFF
--- a/src/core/templatetags/due_dates.py
+++ b/src/core/templatetags/due_dates.py
@@ -1,0 +1,17 @@
+from django import template
+from django.utils import timezone
+
+
+register = template.Library()
+
+@register.simple_tag
+def due_date(days=0, input_type="date"):
+    """
+    Get a string representing today's date or the now datetime,
+    with an offset of X days.
+    """
+    due = timezone.now() + timezone.timedelta(days=days)
+    if input_type == "date":
+        return due.strftime("%Y-%m-%d")
+    elif input_type == "datetime-local":
+        return due.strftime("%Y-%m-%dT%H:%M")
--- a/src/core/templatetags/uuid.py
+++ b/src/core/templatetags/uuid.py
@@ -1,0 +1,8 @@
+from django import template
+from uuid import uuid4
+
+register = template.Library()
+
+@register.simple_tag
+def short_uuid4():
+    return f'u{str(uuid4())[:8]}'
--- a/src/core/templatetags/uuid.py
+++ b/src/core/templatetags/uuid.py
@@ -4,5 +4,5 @@ from uuid import uuid4
 register = template.Library()
 
 @register.simple_tag
-def short_uuid4():
-    return f'u{str(uuid4())[:8]}'
+def get_uuid4():
+    return f'u{str(uuid4())}'
--- a/src/templates/admin/core/widgets/select_all.html
+++ b/src/templates/admin/core/widgets/select_all.html
@@ -16,7 +16,7 @@ Usage:
 
 {% load uuid %}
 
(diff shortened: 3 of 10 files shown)


src/core/templatetags/uuid.py
```diff
@@ -0,0 +1,8 @@
+from django import template
+from uuid import uuid4
+
+register = template.Library()
+
+@register.simple_tag
+def short_uuid4():
```
ajrbyers, line 7: When truncating a UUID there is no way to guarantee it will be unique although a collision is super unlikely. Is there any need to shorten the UUID realistically?

src/core/templatetags/due_dates.py
```diff
@@ -0,0 +1,17 @@
+from django import template
+from django.utils import timezone
+
+
+register = template.Library()
+
+@register.simple_tag
+def due_date(days=0, input_type="date"):
```
ajrbyers, line 8: This is a useful template tag and I can think of a few places where we might want to use it but it might be worth consdering a more generic name for the tag. `offset_date` or something maybe?

src/core/templatetags/due_dates.py
```diff
@@ -0,0 +1,17 @@
...
+    """
+    Get a string representing today's date or the now datetime,
+    with an offset of X days.
```
ajrbyers, line 11: Adding a usage section in here to show folks how to use the tag would also be good.

src/templates/admin/core/widgets/select_all.html
ajrbyers, line 1: This is great!

src/templates/admin/core/widgets/select_all.html
```diff
@@ -0,0 +1,44 @@
...
+    {% include "admin/core/widgets/select_all.html" %}
+    <input id="red" name="red" type="checkbox"><label for="red">Red</label>
+    <input id="blue" name="blue" type="checkbox"><label for="blue">Blue</label>
+</fieldset>
+
+{% endcomment %}
+
+{% load uuid %}
+
+{% short_uuid4 as pid %}
```
ajrbyers, line 19: See comment on UUID truncation.

src/templates/admin/core/widgets/soon_date_buttons.html
ajrbyers, line 1: I like this but is there a realistic situation where the due date will be 0-3 days?

ajrbyers (changes requested):
Some cool utilities in here, I've added some comments to think about.

src/core/templatetags/uuid.py
joemull: It was for readability of the rendered code, keeping in mind the low stakes of a clash for a convenience button, but I can expand it, so that it is safe to use in even high-stakes interfaces.

src/templates/admin/core/widgets/soon_date_buttons.html
joemull: I was thinking of the pub date form in prepub. That's why I made a `datetime-local` option.

ajrbyers (approved):

mauromsl (approved):
Love this!

PR #4332, opened 2024-07-22
Adds a skip to content link to all themes that is tab navigable

### LINKED ISSUE #3932: is it possible to skip straight to the main content?
New A11y feature: 

make sure article templates (especially those with large headers before the article) have a 'skip to content' link at the top of the page.

ajrbyers:
- Adds a new common template
- Replaces existing skip to content for clean
- Adds ID to `<main>`
- Closes #3932 
Screenshot 2024-07-22 at 12 20 43
Screenshot 2024-07-22 at 12 20 32
Screenshot 2024-07-22 at 12 20 16

### DIFF
--- a/src/static/common/css/common.css
+++ b/src/static/common/css/common.css
@@ -298,3 +298,26 @@ article .list-romanupper {
     padding: 15px;
     margin-top: 15px;
 }
+
+.skip-container {
+    position: absolute;
+    top: -100px;
+    left: 50%;
+    transform: translateX(-50%);
+    z-index: 100;
+    transition: top 0.3s;
+}
+
+.skip-to-content {
+    background: #f0f0f0;
+    color: #000;
+    padding: 8px 16px;
+    text-decoration: none;
+    display: inline-block;
+}
+
+.skip-to-content:focus,
+.skip-to-content:hover,
+.skip-container:focus-within {
+    top: 10px;
+}
--- a/src/static/common/css/common.css
+++ b/src/static/common/css/common.css
@@ -299,25 +299,31 @@ article .list-romanupper {
     margin-top: 15px;
 }
 
+/* Styling for skip to main content link */
 .skip-container {
-    position: absolute;
-    top: -100px;
-    left: 50%;
-    transform: translateX(-50%);
-    z-index: 100;
-    transition: top 0.3s;
+    position: relative;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    height: 0;
+    overflow: hidden;
 }
 
-.skip-to-content {
-    background: #f0f0f0;
-    color: #000;
-    padding: 8px 16px;
-    text-decoration: none;
-    display: inline-block;
+.skip-to-content,
+.skip-to-nav {
+    position: absolute;
+    left: -999px;
 }
 
 .skip-to-content:focus,
-.skip-to-content:hover,
+.skip-to-nav:focus {
+    position: relative;
+    left: 0;
+    margin-block-start: 10px;
+    margin-block-end: 10px;
+}
+
 .skip-container:focus-within {
-    top: 10px;
+    height: auto;
+    overflow: visible;
 }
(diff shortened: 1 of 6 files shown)


StephDriver:
Andy, I'm not too keen on this.  Pretty much everywhere I've seen "skip to content" implemented, it doesn't appear over the top of existing content.  I've seen some places do it by having strategically positioned whitespace, others by dynamically scrolling the whole page down or sideways to make space for new accessible menu items like this.   Here's an example of the functionality on the Access to Work site from .gov.uk - i.e. aimed at disabled users. (tabs are highlighting in yellow)

https://github.com/user-attachments/assets/eb8b004d-6f4a-400a-b3d9-36803bf4e9d0

StephDriver (changes requested):
see comment on Conversation page - I don't think having this over the top of other content is an accessible implementation.  It won't just be people who can't see the page who use this functionality. It shouldn't be obscuring other content.

ajrbyers:
Screenshot 2024-07-24 at 10 05 23
Screenshot 2024-07-24 at 10 05 17
Screenshot 2024-07-24 at 10 05 57
Screenshot 2024-07-24 at 10 05 52
Screenshot 2024-07-24 at 10 05 32
Screenshot 2024-07-24 at 10 05 42

StephDriver (changes requested):
I don't think the behaviour here is quite right yet. 

When I tested this, while the "skip to content" link did move the page (✅ ) if I then tab again, I'm back at the top at the next item after skip to content.  So basically having had the skip, I can't continue from that point ( :x: ).  After skipping the next tab item should be the next one after the start of hte main content, not taking me back to the top of the page.

I've tested this on Chrome with a press site and journal site - finding the same behaviour on both.

ajrbyers:
> I don't think the behaviour here is quite right yet.
> 
> When I tested this, while the "skip to content" link did move the page (✅ ) if I then tab again, I'm back at the top at the next item after skip to content. So basically having had the skip, I can't continue from that point ( ❌ ). After skipping the next tab item should be the next one after the start of hte main content, not taking me back to the top of the page.
> 
> I've tested this on Chrome with a press site and journal site - finding the same behaviour on both.

Which theme did you test this with?

StephDriver:
> Which theme did you test this with?

OLH.

I've just tried Clean and Material - and those are both fine.  So I think just an OLH issue?

ajrbyers:
> > Which theme did you test this with?
> 
> OLH.
> 
> I've just tried Clean and Material - and those are both fine. So I think just an OLH issue?

Yeah I just tested it there too. I think its intercepting the click so the anchor link isn't set and the focus doesn't move.

StephDriver:
I'll leave it with you then - I think that's the only problem remaining - from what I can see the rest of the requirements are there:


✅ link is the first focusable control on the Web page.
✅  description of the link communicates that it links to the main content.
✅ (the link is either always visible - No) or visible when it has keyboard focus (Yes).
✅  activating the link moves the focus to the main content.
( :x: OLH, ✅ Clean & Material) after activating the link, the keyboard focus has moved to the main content.

from [WCAG G1](https://www.w3.org/TR/WCAG20-TECHS/G1.html)

ajrbyers:
@StephDriver I've updated OLH's JS to stop it intercepting the anchor hash being added to the URL bar. All looks okay now. To test this locally you will need to run the `build_assets` command.

StephDriver:
I have been testing it with random pages, and this time that ended up being a homepage, where it skipped to a search bar rather than the start of the main content.  But it works fine on an article page, and on the list of articles too.  I wonder whether at this point it would be best to implement this solution, since it seems fine for 99% of the site, and then raise a new issue for reviewing its functionality on the landing page? 

I am suggesting that because I think the problem is actually with the landing page structure not this issue.  The landing page's main content starts with a search bar, before having a heading. Rather than change this new functionality to look for the heading which isn't where it should be on that page, I'd rather improve the structure of the landing page.

StephDriver (approved):
I have left a comment on the behaviour on landing pages, which I think should be addressed separately.  Otherwise I'm happy to approve this.

joemull (approved):

PR #4344, opened 2024-08-06
When serving an article's PDF via the identifier URL, don't try to read an attribute from a None object

gamboz:
URLs in the form `/article/<identifier_type>/<identifier>/download/pdf/` call a view that tries to get the article using the identifier.

If the article does not exist (maybe because there was a typo in the identifier), or if the article has not yet been published, the view should return a 404.

This MR tries to fix a small bug that makes the clause above to fail badly when the article does not exists, because, ATM, the code checks if the non-existing article has been published :slightly_smiling_face: 

Unfortunately I wasn't able to replicate the issue on any online system where I tried. For instance, both the following URLs return a neat 404, but AFAICT the first should work and the second should give 500:
- https://olh.openlibhums.org/article/doi/10.16995%2Folh.16937/download/pdf
- https://olh.openlibhums.org/article/doi/10.16995%2Folh.16937xxxxxxxxxxxxxxxx/download/pdf

So, it's very possible I'm missing something here (i.e. please feel free to trash this MR).

### DIFF
--- a/src/journal/views.py
+++ b/src/journal/views.py
@@ -2646,7 +2646,7 @@ def serve_article_pdf(request, identifier_type, identifier):
         identifier,
     )
 
-    if not article_object and not article_object.is_published:
+    if not article_object or not article_object.is_published:
         raise Http404
 
     pdf = article_object.pdfs.first()


ajrbyers:
@gamboz in this instane if `article_object` is none the second part of the `if` statement is not evaluated so it immediately raises the 404. You can test this by doing something like this in a python shell:

```python
In [1]: article = None

In [2]: if not article or test:
   ...:     print('Article is none so even though test is never set it wont break.')
   ...: 
Article is none so even though test is never set it wont break.

In [3]: 
```

gamboz:
@ajrbyers  I agree, but the original code has `if not article and...` (with `and`, not `or`).

mauromsl (approved):

ajrbyers (approved):

ajrbyers:
> @ajrbyers I agree, but the original code has `if not article and...` (with `and`, not `or`).

Haha, of course. Stupid me, I was looking at your new version XD

mauromsl:
> URLs in the form `/article/<identifier_type>/<identifier>/download/pdf/` call a view that tries to get the article using the identifier.
> 
> If the article does not exist (maybe because there was a typo in the identifier), or if the article has not yet been published, the view should return a 404.
> 
> This MR tries to fix a small bug that makes the clause above to fail badly when the article does not exists, because, ATM, the code checks if the non-existing article has been published 🙂
> 
> Unfortunately I wasn't able to replicate the issue on any online system where I tried. For instance, both the following URLs return a neat 404, but AFAICT the first should work and the second should give 500:
> 
>     * https://olh.openlibhums.org/article/doi/10.16995%2Folh.16937/download/pdf
> 
>     * https://olh.openlibhums.org/article/doi/10.16995%2Folh.16937xxxxxxxxxxxxxxxx/download/pdf
> 
> 
> So, it's very possible I'm missing something here (i.e. please feel free to trash this MR).

Provided examples do not reproduce it because they are missing a trailing slash, I can reproduce it like this: https://olh.openlibhums.org/article/id/999999999/download/pdf/

PR #4345, opened 2024-08-06
Updates clean to use frozen authors for citation modals.

### LINKED ISSUE #2422: Generated citations use article author accounts rather than frozen records
**Describe the bug**
The generated citations sometimes differ from the "how to cite" block. This is because the generated citations use the underlying account details rather than the frozen metadata linked to the article.

**Janeway version**
v1.3.10

ajrbyers:
Clean now uses `frozenauthor_set` like OLH already does.

Closes #2422


StephDriver (approved):
ronseal.

joemull (approved):

PR #4347, opened 2024-08-07
Standard way of bringing users back to starting point after completing side tasks

### LINKED ISSUE #3899: Context lost during account registration

### LINKED ISSUE #4042: Incorrect login redirect in some rare cases
## Problem
I was working on #3899 and realised that this function has a bug, though I don't think this code is ever run because we don't pass a string as `login_redirect`:

https://github.com/BirkbeckCTP/janeway/blob/4218c13fc3b8b6d453b8fc5b4e946f0ea2362c69/src/security/decorators.py#L48-L50

Line 49 should have `login_redirect`, probably?

## To reproduce

1. In dev, set `login_redirect` to a string inside `base_check`.
2. Log out and try to access the manager page.
3. Inspect the URL.

Here's the pdb output:

```
> /home/joe/git/janepr/src/security/decorators.py(51)base_check()
-> params = urlencode({"next": redirect})
(Pdb) n
> /home/joe/git/janepr/src/security/decorators.py(52)base_check()
-> return redirect('{0}?{1}'.format(reverse('core_login'), params))
(Pdb) params
'next=%3Cfunction+redirect+at+0x7ff153f1b010%3E'
```

### LINKED ISSUE #4291: Redundant percent-encoding when building url query string
## Problem
The dictionary that is passed into `utils.logic.build_url` is sent through `urllib.parse.urlencode` and then also through `urllib.parse.quote_plus`.

https://github.com/BirkbeckCTP/janeway/blob/92517963a92a173173a58244520be0cfe019bef8/src/utils/logic.py#L153-L154

This is redundant and results in unusable query strings. Compare:

```py
In [3]: quote_plus(urlencode({'a': 'b', 'x': 'y', 'next': '/target/page/?d=e&f=g'}))
Out[3]: 'a%3Db%26x%3Dy%26next%3D%252Ftarget%252Fpage%252F%253Fd%253De%2526f%253Dg'
```

```py
In [4]: urlencode({'a': 'b', 'x': 'y', 'next': '/target/page/?d=e&f=g'})
Out[4]: 'a=b&x=y&next=%2Ftarget%2Fpage%2F%3Fd%3De%26f%3Dg'
```

I believe we just want `urlencode` here.

If a URL is needed for inclusion in another URL, as a redirect callback, like with the ORCID API, then the function that builds the ORCID URL (`utils.orcid.build_redirect_uri`) should separately quote or escape the full URL that is returned from `build_url`.

## Proposed solution
Allow a Django `QueryDict`, a plain dict, or a query string that already has percent-encoded values to be passed to the function. URL-encode the first two with '/' marked safe, to match the default behavior of the `|urlencode` template filter. For query strings, do nothing except pass it on to `SplitResult`.

### LINKED ISSUE #4321: Cannot pass query parameters to site_url functions
## Describe the bug
The `site_url` functions for journal, repo, and press do not pass any query string on to `build_url`.

## Janeway version
1.6

## Expected behavior
There is a way to pass query parameters to this chain of functions.

joemull:
Closes #3899.

Fixes #4291.
Fixes #4042.
Fixes #4321.

Users often have a main task (e.g. submit an article) and a side task (e.g. log in / register) that they need to complete during the main task.

This PR introduces a standard pattern for keeping track of main tasks through a sequence of side task views and templates.

It includes:

* half a dozen new template tags and logic functions

* numerous updates to views and templates, especially to avoid hard-coding unpredictable URL elements

* new documentation both where it was missing and for new code 

* ~40 new tests (see note below on middleware)

* fixing four bugs encountered in doing this work

* migrations for the removal of two deprecated fields that held outdated help text

* refactoring the `user_login_orcid` function (see note below)

* migrations for updated setting values related to password resets and account confirmation

## Note on one change to the middleware

Testing the templates across themes is very difficult when using the Django test Client, because of the way we load theme directories. Initially I wrote a helper that would allow me to patch `template_override_middleware.Loader.get_theme_dirs`. It looked like this:

```py
def get_theme_dirs(journal):
    """
    Together with the mock library,
    this lets you test public-facing views, once for each theme.

    Example usage:

    @patch('utils.template_override_middleware.Loader.get_theme_dirs')
    def test_a_view_for_each_theme(self, get_theme_dirs):
        for theme_dirs in testing_helpers.get_theme_dirs():
            get_theme_dirs.response_value = theme_dirs
            response = self.client.get('/my/path/', {})
            self.assertIn(
                'The text I am looking for',
                response.content.decode(),
            )
    """
    request = Request()
    request.journal = journal
    with request_context(request):
        engine = Engine()
        loader = template_override_middleware.Loader(engine)
        dirs = loader.engine.dirs
        return [
            [os.path.join(settings.BASE_DIR, 'themes', 'clean', 'templates')] + dirs,
            [os.path.join(settings.BASE_DIR, 'themes', 'OLH', 'templates')] + dirs,
            [os.path.join(settings.BASE_DIR, 'themes', 'material', 'templates')] + dirs,
        ]
```

However this did not work, because the Django test client handles the creation of the request independently. It gives you no way to modify the request when using `Client.get`. So the request used in my helper was different than the one used in the test, and the theme dirs were missing a lot of template locations that are important for includes in the templates I am testing. Basically, the Django test client’s request is artificially stale in comparison to Janeway’s typical requests.

To make the test run in an environment that more closely resembles production, I decided to make use of the simple query parameter `theme=clean` and `theme=OLH`, which we use for debugging. I expanded the if clause to include `settings.IN_TEST_RUNNER`.

If there is another way of mocking the themes comprehensively which can be used effectively with the Django test Client, I’m happy to consider it.

### Note on integration with ORCiD

There is some tricky juggling of view logic and context in between this branch and the ORCiD work that Esther has been putting through.

Esther and I successfully merged some of our work in Esther’s PR about [prefilling the registration form with info from ORCiD](https://github.com/openlibhums/janeway/pull/4279).

But there is [another PR that has been open since Sept 2024](https://github.com/openlibhums/janeway/pull/4399) that this work will interfere with. @ajrbyers please assign me as a reviewer if you would like me to work with Esther to resolve the differences.

In terms of the ORCiD work on this branch, the `user_login_orcid` view was getting really complicated and hard to read, so I refactored the view logic and documented the behavior with comments. Hopefully it is now clear how `next` and `action` work independently but are both needed in the `state` parameter that is sent to ORCiD authentication and received back.

## Future work
If we like this implementation, we can eventually refactor existing return URLs using the same pattern, so that we do not have to reason about code like this: 

https://github.com/openlibhums/janeway/blob/d9c1ea56fcb2a0e12c5a8798e0518320e36e3da7/src/templates/admin/journal/document_management.html#L57-L62


src/core/logic.py
```diff
@@ -37,79 +38,101 @@
 
 def send_reset_token(request, reset_token):
     core_reset_password_url = request.site_type.site_url(
-        reverse(
+        reverse_with_next(
             'core_reset_password',
             kwargs={'token': reset_token.token},
+            request=request,
         )
     )
```
mauromsl, line 46: As discussed, since `request.site_type.site_url` now takes in an optional `query` parameter, we want to avoid encoding the querystring onto the `path` parameter ahead of time. Instead, we can pass in the query elements to be encoded separately and delegete URL encoding to `site_url`. We could either use a regular dictionary or a mutable [`Querydict`](https://docs.djangoproject.com/en/4.2/_modules/django/http/request/#QueryDict)

joemull:
@mauromsl this is now ready for another review. Note the updated description with lots of explanation of choices.

src/press/tests/test_views.py
```diff
@@ -0,0 +1,28 @@
...
+class URLWithReturnTests(PressViewTestsWithData):
+
+    def test_press_account_links_have_return(self):
+        request = helpers.get_request(press=self.press)
+        response = press_views.index(request)
+        content = response.content.decode()
+        self.assertIn('/login/?next=', content)
+        self.assertNotIn('"/login/"', content)
+        self.assertIn('/register/step/1/?next=', content)
+        self.assertNotIn('"/register/step/1/"', content)
```
mauromsl, line 28: I don't understand what is being tested here: If I visit the press index page, why does the login page require a `?next` url parameter at all times? A better test would be to invoke a view where we expect the user to be redirected to the press view with a specific URL with a specific `next`.

src/core/tests/test_views.py
```diff
@@ -97,3 +125,360 @@ def test_edit_profile(self):
...
+            self.assertNotIn('"/register/step/1/"', content)
+
+    def test_journal_submissions_account_links_have_return(self):
+        for theme in self.themes:
+            response = self.client.get('/submissions/', data={'theme': theme})
+            content = response.content.decode()
+            self.assertIn('/login/?next=/submissions/', content)
+            self.assertNotIn('"/login/"', content)
+            self.assertIn('/register/step/1/?next=/submissions/', content)
+            self.assertNotIn('"/register/step/1/"', content)
```
mauromsl, line 484: What are we testing here? that the current template injects its own path onto the login link of the submissions page?

We are not setting a `SERVER_NAME` on the `self.client.get` call, so the result of this test is at the mercy of the state defined in the `settings.py` file. In order to test the journal page explicitly, we need to either use the journal domain as `SERVER_NAME` or run the request in path mode.

src/utils/logic.py
```diff
@@ -344,3 +343,21 @@ def get_aware_datetime(unparsed_string, use_noon_if_no_time=True):
...
+    Then re-encode them into a query string and generate the final URL.
+    :param original_url: the raw URL to be updated
+    :param new_params: the dict or QueryDict of new query parameters to add
+    """
+    parsed_url = urlparse(original_url) # ParseResult
+    parsed_query = QueryDict(parsed_url.query, mutable=True) # mutable QueryDict
+    parsed_query.update(new_params)
+
+    # Treat / as safe to match the default behavior of
+    # tempalte filters such as |urlencode
```
mauromsl, line 360: tempalte -> template

src/utils/logic.py
```diff
@@ -344,3 +343,21 @@ def get_aware_datetime(unparsed_string, use_noon_if_no_time=True):
...
+
+def add_query_parameters_to_url(original_url, new_params):
+    """
+    Parse a URL string and then safely update the query parameters.
+    Then re-encode them into a query string and generate the final URL.
+    :param original_url: the raw URL to be updated
+    :param new_params: the dict or QueryDict of new query parameters to add
+    """
+    parsed_url = urlparse(original_url) # ParseResult
+    parsed_query = QueryDict(parsed_url.query, mutable=True) # mutable QueryDict
```
mauromsl, line 356: do we need these comments

src/utils/orcid.py
```diff
@@ -113,3 +113,34 @@ def get_orcid_record_details(orcid):
...
+
+    :param next_url: a raw URL (can include its own query parameters)
+    :param orcid_action: 'login' or 'register'
+    :return: a query string with percent-encoded paramater values
+        that can be passed to ORCiD in a URL as a callback
+    """
+    querydict = QueryDict(mutable=True)
+    if next_url:
+        querydict['next'] = next_url
+    if orcid_action:
```
mauromsl, line 133: This code is fairly defensive for what seem to be non-nullable arguments, why is this?

mauromsl (changes requested):
Thanks Joe, this looks really good. Reviewing the ORCID changes was a bit more challenging than I'd like, however the tests did help the process a lot. makes me realise that we need integration level tests here as we did with DOAJ

I've added what I think are mostly minor comments in line.

As you mentioned in the description there is a fair bit of refactoring now needed on #4399, although I do think it is already conflicting with other changes that made it to master.

A more general question I have has to do with the login links and their use of next URL. It seems like we are going to always redirect the user to the page they were on when they hit login. I think this makes sense in most contexts, but for the landing/homepage, wouldn't it make sense to send the user to the dashboard? The way I would implement this is by not usin the `url_with_next` in the homepage and let the Janeway backend view determine the default routing to the dashboard

src/core/tests/test_views.py
joemull: Yes, that's what we're testing. Without the next parameter, the user would lose their submission context.

Good point nabout the settings--I've updated that.

I've also removed the negative assertions, because I've also reverted the login and registration links in the nav to preserve the old dashboard redirect.

src/press/tests/test_views.py
joemull: I've gone ahead and reverted the login and registration links in the nav, so I will update this test to assert that the next param is *not* in those links, to make the dashboard redirect behavior explicit in the context of the new standard way of establishing next URLs.

joemull: As for the suggestion "to invoke a view where we expect the user to be redirected to the press view with a specific URL with a specific next", I have done that pretty extensively for the journal level. Testing it here is very difficult because of my inability to use the Django test client at the press level. The view expects too many things to be loaded onto the view because of our middleware customizations. For example, it errored out on `django.contrib.messages.api.MessageFailure: You cannot add messages without installing django.contrib.messages.middleware.MessageMiddleware`. I think there are diminishing returns for time spent on this before we have more ways of referencing press URLs in Janeway (related: https://github.com/openlibhums/janeway/issues/4577 ). But suggestions welcome if you see low-hanging fruit.

joemull: Here's the test I tried, for reference:

In set up test data:

```py
        cls.user_email = 'w5gq1pa6s9g3vhzovpfh@example.org'
        cls.user_password = 'mruoeociqsgxdtd6eot0'
        cls.user = core_models.Account.objects.create_user(
            cls.user_email,
            password=cls.user_password,
        )
        # The raw unicode string of a 'next' URL
        cls.request = helpers.get_request(
            press=cls.press,
            user=cls.user,
        )
        cls.next_url_raw = '/target/page/?a=b&x=y'
```

The test:
```py
    @override_settings(URL_CONFIG='path')
    @patch('core.views.authenticate')
    def test_login_success_redirects_to_next(self, authenticate):
        authenticate.return_value = self.user
        self.request.POST.update({
            'user_name': self.user_email,
            'user_pass': self.user_password,
        })
        self.request.GET.update({
            'next': self.next_url_raw,
        })
        response = core_views.user_login(self.request)
        self.assertEqual(
            response,
            redirect(reverse(self.next_url_raw)),
        )
```

src/core/logic.py
```diff
@@ -29,87 +30,112 @@
...
 
+def reverse_with_next(url_name, request, next_url='', args=None, kwargs=None):
+    """
+    Reverse a URL but keep the 'next' parameter that exists on the request
+    or that the caller wants to introduce.
+
+    The value of 'next' on the request, or the value of 'next_url',
+    should be in decoded form.
+
+    :param request: HttpRequest, optionally containing 'next' in GET
```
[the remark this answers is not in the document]
joemull: Yeah that occurred to me earlier but I didn't get around to it. Will do it now.

src/utils/orcid.py
joemull: I was trying to avoid empty values, since that would result in `next=&action=`, which would then populate query dicts like `{'next': None, 'action': None}` which are a bit more bug prone to use in the views.

src/core/templatetags/next_url.py
```diff
@@ -0,0 +1,45 @@
...
+    A template tag to use instead of 'url' when you want
+    the reversed URL to include a new 'next' parameter that
+    contains the full and query string path of the current request.
+    """
+    request = context.get('request')
+    if request.journal or request.repository:
+        next_url = request.get_full_path()
+    else:
+        # This is the only way to refer to press URLs
+        next_url = request.build_absolute_uri()
```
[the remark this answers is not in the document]
joemull: Yeah I know the build_absolute_uri method uses the same function for the path, but it also includes the domain, and I don't know of another way to refer to press URLs accurately.

I've made a demo of the problem I'm facing: https://github.com/openlibhums/janeway/commit/39eebbce901b9d5c6eadfe6173768a7b54584ab3

Check this branch out and run the test on press to see the failing test. If you then uncomment my code including the `build_absolute_uri` call for press requests, the test passes. Any alternative suggestions welcome.

This issue is also documented in #4577 , where I name it as a path issue, but it might be an issue for domain mode as well.

src/core/logic.py
```diff
@@ -1037,3 +1031,30 @@ def filter_articles_to_editor_assigned(request, articles):
...
+    if next_url:
+        return next_url
+    elif request.journal:
+        return reverse('core_dashboard')
+    elif request.repository:
+        return reverse('repository_dashboard')
+    elif request.press:
+        return reverse('website_index')
+    else:
+        return reverse('website_index')
```
mauromsl, line 1060: Similar to my other comment, it would be best to avoid requiring `request` as an argument `site_type` would be a better argument and easier to switch over.

In fact, this function would feel better and more performant if it was statically calculated. One option is with a map:

```
SITE_TYPE_REDIRECT = {
    Journal: "core_dashboard",
    Press: "website_index"
[...]
}

def get_auth_redirect_url(site_type, next_url=None):
    return next_url or reverse(SITE_TYPE_REDIRECT[site_type])

```

or, more elegantly, expand `AbstractSiteModel`


```
class AbstractSiteModel(models.Model):
    AUTH_REDIRECT_NAME = "website_index"

    def auth_redirect_url(self, next_url=None):
        return next_url or reverse(self.AUTH_REDIRECT_NAME)

class Journal(AbstractSiteModel, [...]):
    CORE_DASHBOARD = "website_index"
```

In fact, in the latter implementation I would ditch the `next_url` argument and handle that in the caller for a cleaner, single responsibility approach

src/core/views.py
```diff
@@ -69,10 +69,8 @@ def user_login(request):
     next_url = request.GET.get('next', '')
     if request.user.is_authenticated:
         messages.info(request, 'You are already logged in.')
-        if next_url:
-            return redirect(next_url)
-        else:
-            return redirect(reverse('website_index'))
+        return redirect(logic.get_post_auth_url(request))
+
```
mauromsl, line 73: Consider my suggestions above, this could be `redirect(request.site_type.auth_redirect_url())`

mauromsl (changes requested):
Thanks Joe, I appreciate you bringing in my suggestions. I have one more and I'll be off your case here :P

src/press/tests/test_views.py
mauromsl: I wasn't aware of this problem when testing press URLs, I'll see if I can make this work and maybe close #4577

src/core/logic.py
joemull: Nice--I like the idea of moving it to the site model and slimming it down using `site_type`.

I do think this function needs to take `next_url` because sometimes, the request doesn't know about the next URL, only the view can figure it out (e.g. when someone's coming to Janeway from a successful ORCiD authentication).

src/core/templatetags/next_url.py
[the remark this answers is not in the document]
mauromsl: The problem exhibited in https://github.com/openlibhums/janeway/commit/39eebbce901b9d5c6eadfe6173768a7b54584ab3 is related to how the test is setup. It uses an incomplete request object mockup rather than a real request, so there are no headers from where to borrow a URL path. I've changed it to use Django's test client and the test is passing now. 
Does this fully resolve the concerns?

mauromsl (changes requested):
a comment regarding the issues around press URLs and tests

joemull:
Share your code @mauromsl ? You can just put it as a commit on that demo branch.

mauromsl:
> Share your code @mauromsl ? You can just put it as a commit on that demo branch.

I pushed a commit here last night https://github.com/openlibhums/janeway/compare/b-3899-press-url-slipperiness-demo

joemull:
> I pushed a commit here last night https://github.com/openlibhums/janeway/compare/b-3899-press-url-slipperiness-demo

Super, thanks. I've updated the tests to use `SERVER_NAME` which is the missing piece here. I've also standardized them to domain mode since otherwise we'd need to hardcode journal codes in URLs.

mauromsl (approved):

docs/source/dev/url-routing.rst
```diff
@@ -0,0 +1,23 @@
...
+
+In all views related to the side task, make sure to check the GET request parameters for `next`, and pass them along in one of two ways:
+
+  * If redirecting to another step in the side task: use `logic.reverse_with_next` when reversing the URL.
+
+  * If the side task is concluded, redirect to the value of `next`.
+
+Note that the `next` state is always transmitted as GET request parameter, and so it will always be visible in the query string. This is possible because Django makes GET parameters from the query string in the HTTP header available even when the request method is POST. Successive requests to the same endpoint will not by remove or change the query string, even if the request method changes; only redirects in the view, or reformations of an `href` or form `action` in the template, can change it. While it is not often desirable to make use of both GET and POST data, we do it intentionally here for a few reasons: the `next` pattern is consistent across views, and we avoid interfering with the payload of actual form data sent via POST.
```
ajrbyers, line 23: > will not by remove or change

Typo?

src/utils/migrations/0039_update_password_reset_links_in_settings.py
ajrbyers, line 1: These were left as plain links because summernote used to remove the {{}} tags. Has this been tested with tinymce?

ajrbyers (commented):
Looks good, I've added two questions that should be easy enough to sort then this can be merged.

src/utils/migrations/0039_update_password_reset_links_in_settings.py
joemull: Good call. I just tried this out by making a change via the TinyMCE editor and saving, and then inspecting the HTML. All looks good:

image

ajrbyers (approved):

PR #4355, opened 2024-08-09
4353 article page sharelinks a11y update

### LINKED ISSUE #1863: Add share links to all themes
Currently the clean theme does not have the share links as OLH and Material have. 

We should place these on all themes.

### LINKED ISSUE #3933: is it clear which page element has focus from the keyboard?

### LINKED ISSUE #4340: OLH theme mini-bar is wider than the main content
In the OLH Theme, the minibar sticky div which contains the share links is wider (on both sides) than the rest of. hte content.  This means the edges of it are very tight against the sides of the window and it "looks a bit odd".

This issue was identified while working on the tab order of those share buttons ( #3933 ).

the three divs below are the problem (around line 103)

```html
<div data-sticky-container class="row">
            <div class="mini-bar sticky" data-sticky data-margin-top="0" data-sticky data-anchor="content">
                <div class="row">
```

Considering these as div1, div2 and div3, the expectation is for div1 to contain div 2 which in turn contains div3.  But instead:
- div2 overflows the bottom of div1
- div3 overflows to left and right of div 2.

image

### LINKED ISSUE #4353: Create, update and/or make accessible article share links across all themes
This issue combines the following three related issues into one piece of work.

- #3933 
- #4340 
- #1863 

The plan is to:
- [x] update the share links on the OLH theme to
    - [x] make the tab order the same as the visual order
    - \[redundant\] include aria labels
    - [x] be positioned correctly within their containing div (see #4340)
- [x] Update the Material and Clean themes to match this update, tweaking it to fit their styles.
- [x] update Twitter to X within these share buttons

StephDriver:
Updated the share links across all themes for a11y.

- OLH - tab order of share bar now matches visual order, and a small alignment fix.
- Clean - now has share links
- All themes: now have X instead of Twitter (and updates to css to make X in black instead of the blue bird).

NB the share links for Clean are buttons because of the need to meet WCAG minimum target size. More comments on this [with screenshots on issue itself.](https://github.com/BirkbeckCTP/janeway/issues/4353#issuecomment-2277758130)

Closes #4353 

which encompassed these, so it also:

Closes #3933 
Closes #4340
Closes #1863

### DIFF
--- a/src/themes/material/assets/mat.css
+++ b/src/themes/material/assets/mat.css
@@ -562,8 +562,8 @@ nav a.sidenav-trigger i {
 :root {
     --facebook: #3b5998;
     --facebook-hover: #3b5998;
-    --twitter: #00acee;
-    --twitter-hover: #008abe;
+    --twitter: black;
+    --twitter-hover: white;
     --linkedin: #0e76a8;
     --linkedin-hover: #0b5e87;
 }
@@ -584,10 +584,12 @@ nav a.sidenav-trigger i {
 
 .twitter-bg {
     background-color: var(--twitter);
+    color: var(--twitter-hover) !important;
 }
 
 .twitter-bg:hover {
     background-color: var(--twitter-hover);
+    color: var(--twitter) !important;
 }
 
 .linkedin-bg {
--- a/src/static/common/css/common.css
+++ b/src/static/common/css/common.css
@@ -327,3 +327,11 @@ article .list-romanupper {
     height: auto;
     overflow: visible;
 }
+
+a.twitter-x{
+  color: black;
+  text-decoration: none;
+  font-size: large;
+  font-weight: bold;
+  line-height: 1.25;
+}
--- a/src/themes/OLH/templates/journal/article.html
+++ b/src/themes/OLH/templates/journal/article.html
@@ -101,44 +101,47 @@
     {% endif %}
 
         <div data-sticky-container class="row">
-            <div class="mini-bar sticky" data-sticky data-margin-top="0" data-sticky data-anchor="content">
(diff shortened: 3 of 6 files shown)


src/themes/material/assets/mat.css
```diff
@@ -584,10 +584,12 @@ nav a.sidenav-trigger i {
 
 .twitter-bg {
     background-color: var(--twitter);
+    color: var(--twitter-hover) !important;
 }
 
 .twitter-bg:hover {
     background-color: var(--twitter-hover);
+    color: var(--twitter) !important;
 }
```
mauromsl, line 593: If we are adding explicit `!important` colours to the `𝕏` button we should probably do the same for other ones. If not, we will introduce inconsistency wherever someone has overriden `btn` colours.

e.g see https://www.comicsgrid.com/article/id/15400/

Alternatively, do not use the `!important` property, though I bet you are using because the changes wouldn't take effect on the base theme styling otherwise.

mauromsl (commented):
Thanks @StephDriver, a small comment on the use of `!important` and how it might impact theme styling overrides.

mauromsl (changes requested):
.

joemull (approved):

ajrbyers (approved):

PR #4368, opened 2024-08-19
Fix repository migration conflict

### LINKED ISSUE #4356: Review handling of social media share links
# New Feature Request

We currently have three social media share links for articles: 
- Facebook
- X
- LinkedIn

There are other social media channels, and with the growth of AI for summarising articles, there may also be other desirable 'share' destinations.   Some journal editors may have strong views on which platforms they want to support, and so this may not be a one size fits all situation.

## Option 1:
provide a selection of social media share links (which?) from which a custom list can be defined as a setting at journal level.

## Option 2: 
Reconsider the current three share options and extend this to include other platforms (which ones?) e.g. Bluesky

## Option 3
unify the share option so that articles are shared in a standard way compatible with any destination 

## Context
While working on  
- https://github.com/BirkbeckCTP/janeway/issues/4353
there was a discussion about updating Twitter to X, and then whether we should also have Bluesky or other social media share buttons.

This lead to a larger discussion on whether which social media share optoins appear should be a customisable (i.e. a setting). and how this might link in with updating user profiles to allow for a choice of social media handles.

joemull:
Fixes #4356.

### DIFF
--- a/src/repository/migrations/0044_historicalrepository_review_submission_text_and_more.py
+++ b/src/repository/migrations/0044_historicalrepository_review_submission_text_and_more.py
@@ -7,7 +7,7 @@ from django.db import migrations, models
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('repository', '0043_auto_20240402_1256'),
+        ('repository', '0044_alter_repositoryfield_help_text'),
     ]
 
     operations = [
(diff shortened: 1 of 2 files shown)


ajrbyers (approved):

ajrbyers:
Once this is merged I'll rebase https://github.com/BirkbeckCTP/janeway/pull/4365

mauromsl (approved):

PR #4370, opened 2024-08-20
Splits back content archive into two views, removes query round trips to improve load time.

### LINKED ISSUE #3952: Author display at rejected & archived articles
We noticed the coloumn "ID" and "Identifier" are redundant in the rejected & archived articles display.

There should be another coloumn using that space, one journal asked for there being authors displayed, so they have a better overview over who already submitted a paper and got declined. It makes it easier to check for double submissions.

Replace the "Identifier" or "ID" coloumn with an "Author" coloumn.

### LINKED ISSUE #4362: Journal article archive page loads slowly
**Describe the bug**
In a large journal the archive page (/manage/archive/) can load very slowly initially and sometimes timeout. Generally the page is much faster after the first load. I have done some experimentation on the page and I believe displaying the DOIs/Identifiers significantly slows this page down since each is a DB query on the first load before they are cached.

**Janeway version**
1.5.6

**To Reproduce**
Load archive page in a larger journal (our largest journal has > 10,000 articles listed in this page)

ajrbyers:
This PR:

- Closes #4362 
- Closes #3952

This PR introduces a custom Aggregate function, GroupConcat, to concatenate query results into a single string with support for PostgreSQL, MySQL/MariaDB, SQLite, and MSSQL

This PR splits the archive page into two views. One for published articles and one for archived articles.

Additionally the use of subqueries massively reduces the number of database round trips and speeds the page up exponentially for installs with large numbers of published or archived articles.

### DIFF
--- a/src/submission/models.py
+++ b/src/submission/models.py
@@ -1154,6 +1154,22 @@ class Article(AbstractLastModifiedModel):
         """Method replaces the funders m2m model for backwards compat."""
         return ArticleFunding.objects.filter(article=self)
 
+    @property
+    def date_archived(self):
+        # Return early if the article is not actually archived.
+        if self.stage != STAGE_ARCHIVED:
+            return None
+
+        log_entry = ArticleStageLog.objects.filter(
+            article=self,
+            stage_to='Archived',
+        ).first()
+
+        if log_entry:
+            return log_entry.date_time
+
+        return None
+
     def __str__(self):
         return u'%s - %s' % (self.pk, truncatesmart(self.title))
 
--- a/src/submission/models.py
+++ b/src/submission/models.py
@@ -1154,22 +1154,6 @@ class Article(AbstractLastModifiedModel):
         """Method replaces the funders m2m model for backwards compat."""
         return ArticleFunding.objects.filter(article=self)
 
-    @property
-    def date_archived(self):
-        # Return early if the article is not actually archived.
-        if self.stage != STAGE_ARCHIVED:
-            return None
-
-        log_entry = ArticleStageLog.objects.filter(
-            article=self,
-            stage_to='Archived',
-        ).first()
-
-        if log_entry:
-            return log_entry.date_time
-
-        return None
-
     def __str__(self):
         return u'%s - %s' % (self.pk, truncatesmart(self.title))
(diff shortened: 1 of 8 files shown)


joemull (changes requested):
This seems to be working great technically, once I rebased it on master (14ec665).

Just needs some touch-ups to the copy:

The big heading "Article Archive" needs to be different on the two pages:
image

image

Just as an observation: it seems we are creating a lot of pages that have redundant headings with the big text and the small caps text. I think this comes from there being too many levels to the theme options for headings, and not a whole lot of consistency in how we use them. Something to clean up in the design system for sure.

src/submission/models.py
```diff
@@ -1154,6 +1154,22 @@ def funders(self):
...
+        log_entry = ArticleStageLog.objects.filter(
+            article=self,
+            stage_to='Archived',
+        ).first()
+
+        if log_entry:
+            return log_entry.date_time
+
+        return None
+
```
mauromsl, line 1172: Is it worth also annotating this date to the new `rejected_and_archived_articles` query to avoid another `n` times roundtrip?

src/utils/db_functions.py
```diff
@@ -0,0 +1,52 @@
...
+            self.template = "%(function)s(%(distinct)s%(expressions)s, '%(separator)s')"
+        elif self.vendor == 'mysql':
+            self.function = 'GROUP_CONCAT'
+            self.template = "%(function)s(%(distinct)s%(expressions)s SEPARATOR '%(separator)s')"
+        elif self.vendor == 'sqlite':
+            self.function = 'GROUP_CONCAT'
+            self.template = "%(function)s(%(expressions)s, '%(separator)s')"
+        elif self.vendor == 'mssql':
+            self.function = 'STRING_AGG'
+            self.template = "%(function)s(%(expressions)s, '%(separator)s')"
```
mauromsl, line 48: nice

mauromsl (approved):
added an optional comment but otherwise good to merge

src/submission/models.py
ajrbyers: Yeah - not sure why it didn't occur to me.

PR #4371, opened 2024-08-23
improving accessibility of tables

### LINKED ISSUE #4055: Table A11y updates
- [x] review use of tables across platform

From #4194 the following issues have been identified:
- Tables do not consistently have `<th>` tags for both rows and columns.  This occurs in both tables that are part of the main template for the page, e.g. article dates shown on the material theme, and for tables within articles themselves.
- footnotes within the tables that are to be written under the table, rather than as endnotes are not working well.

Part of the problem here is that JATS allows for inaccessible tables to be specified, thus user content will not always produce accessible tables. The row/column headings should be specified in the JATS and then picked up so if it is not correct there, it won't be correct on the page either.

Therefore we should - 
- [x] ensure non-article (i.e. non user content) tables are accessible.  Each data cell is expected to be understood in the context of one column header and one row header - so a single row of headers and a single column of headers for each table, using the `<th>` tag.
- [x] add user content/jats issue to Phase 2 of A11y project.

StephDriver:
This will take some time!  

Each table has a comment above it, so that it can be identified for work, and I can track which ones I've reviewed at any point.

right at the end of this, before merging, the comments will be removed.

closes #4055

### DIFF
--- a/src/core/homepage_elements/featured/templates/featured_setup.html
+++ b/src/core/homepage_elements/featured/templates/featured_setup.html
@@ -38,7 +38,7 @@
 
                 <div class="large-6 columns">
                     <h4>Add New Featured Article</h4>
-                    {# TableA11y #}
+                    {# TableA11y, fail, empty header cell #}
                     <table id="featured" class="small">
                         <thead>
                         <tr>
--- a/src/core/homepage_elements/featured/templates/featured_setup.html
+++ b/src/core/homepage_elements/featured/templates/featured_setup.html
@@ -38,6 +38,7 @@
 
                 <div class="large-6 columns">
                     <h4>Add New Featured Article</h4>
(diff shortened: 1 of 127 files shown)


StephDriver:
Process:

- [x] find tables and add comment above each.
- [ ] first review - expand comment to note obvious issues, and if none, which type of table (see below)
- [ ] fix obvious issues, and update comment to type of table.
- [ ] for each type of table, review in more detail, and fix any a11y markup required.
- [ ] Review all as to whether captions and summaries are required (probably not for our use, but check).
- [ ] remove the comments

Types of table (as per [WCAG guidelines](https://www.w3.org/WAI/tutorials/tables/)):
- V1 - One header (either row or column)
- V2 - header row and column row
- V3 - Irregular headers
- V4 - Multilevel headers

PR #4385, opened 2024-09-03
Fixed an issue with middle names and added support for page ranges in RIS and BIB exports (zotero, endnote, etc)

### LINKED ISSUE #3571: BibTeX citation export does not produce valid BibTeX because of a stray space
**Describe the bug**

**Janeway version**

1.4.3 deployed

**To Reproduce**

1. Select `Download BibTeX` from talk-bubble menu
2. Save `.bib` file locally.
3. Attempt to import `.bib` file to any BibTeX-aware citation manager (e.g., Zotero, Bibdesk) or process with Biber.
4. failure

This is failing because Janeway's current BibTeX template does not produce valid BibTeX. It's not valid because it produces a citekey with a space between the journalcode and the articleID. In BibTeX a citekey cannot spaces. So, when Janeway outputs "journalcode articleID" as the citekey, the parser treats "journalcode" as the citekey and then breaks because it doesn't know what to do with "articleID" after the space. Removing this space should generate parsable citekeys.

Current example output: `@article{journalcode 4321, author = {Rosalind Franklin}, …`

The space causing the error is the one between `journalcode` and `4321`. 
 
**Expected behavior**

In BibTeX output, the first element after `@article{` should be the citekey. If the problematic space is removed, the string consisting of the journalcode and the articleID can become the citekey, as I think the developer intended.

Proposed output: `@article{journalcode4321, author = {Rosalind Franklin}, …`

**Progress**

I submitted a pull request that I thought would solve the problem (https://github.com/BirkbeckCTP/janeway/pull/3565), but my Python is so far very limited, and it failed.

### LINKED ISSUE #3771: Citation export
[Reported by Harrison - Iowa]

An editor has reported the following happens when exporting a citation to Endnote:

"Katie R. Lybarger, E. S. B., K. J. F., L. A. E., L. N. D., M. C. H., J. L. V., M. D. Z, M. D. C., T. G. O. (2023). Determination of Consumer Color and Discoloration Thresholds for Purchase of Representative Retail Ground Beef. Meat and Muscle Biology, 7(1)."
 
When doing this using Zotero:

"Katie R. Lybarger, E. S. B., Kaylee J. Farmer, Lane A. Egger, Lindsey N. Drey, Melvin C. Hunt, Jessie L. Vipham, Morgan D. Zumbaugh, Michael D. Chao, Travis G. O’Quinn. (2023). Determination of Consumer Color and Discoloration Thresholds for Purchase of Representative Retail Ground Beef. Meat and Muscle Biology, 7(1)."
 
Added comment: 
_"In looking at the RIS files from Janeway and comparing it to one from Elsevier, I’m thinking it might be caused by listing all the authors as one entry, causing EndNote to see only one author with a lot of first and middle names. Elsevier has each author as a separate entry.
 
The editors would also like to have the article number and page range in the citation. I believe we’ve discussed adding this to the pages field."_

ajrbyers:
Updates RIS and BIBTEX with additional information and 
- Closes #3771 
- Closes #3571

### DIFF
--- a/src/submission/models.py
+++ b/src/submission/models.py
@@ -1400,7 +1400,7 @@ class Article(AbstractLastModifiedModel):
             return ", ".join([author.full_name() for author in self.authors.all()])
 
     def bibtex_author_list(self):
-        return " and ".join(
+        return " AND ".join(
             [author.full_name() for author in self.frozen_authors()],
         )
 
--- a/src/submission/models.py
+++ b/src/submission/models.py
@@ -1399,6 +1399,11 @@ class Article(AbstractLastModifiedModel):
         else:
             return ", ".join([author.full_name() for author in self.authors.all()])
 
+    def bibtex_author_list(self):
+        return " and ".join(
+            [author.full_name() for author in self.frozen_authors()],
+        )
+
     def keyword_list_str(self, separator=","):
         if self.keywords.exists():
             return separator.join(kw.word for kw in self.keywords.all())
--- a/src/templates/common/encoding/article_bibtex.bib
+++ b/src/templates/common/encoding/article_bibtex.bib
@@ -1,7 +1,7 @@
 {% load settings %}
 {% load encoding %}
-{% if article.journal.is_conference %}@conference{% else %}@article{% endif %}{% templatetag openbrace %}{{ article.journal.code }} {{ article.id }},
-	author = {{ article.author_list|latex_conform }},
+{% if article.journal.is_conference %}@conference{% else %}@article{% endif %}{% templatetag openbrace %}{{ article.journal.code }}{{ article.id }},
+	author = {{ article.bibtex_author_list|latex_conform }},
 	title = {{ article.title|escape|latex_conform }},
 	volume = {{ article.issue.volume|latex_conform }},
 	year = {{ article.date_published.year|latex_conform }},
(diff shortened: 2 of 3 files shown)


mauromsl (approved):

src/submission/models.py
```diff
@@ -1399,6 +1399,11 @@ def author_list(self):
         else:
             return ", ".join([author.full_name() for author in self.authors.all()])
 
+    def bibtex_author_list(self):
+        return " and ".join(
+            [author.full_name() for author in self.frozen_authors()],
+        )
```
mauromsl, line 1405: I believe the spec recommends `"AND"` (all uppercase) as the separator : https://bibtex.eu/fields/author/

mauromsl (changes requested):

mauromsl (approved):

src/templates/common/encoding/article_bibtex.bib
```diff
@@ -1,7 +1,7 @@
 {% load settings %}
 {% load encoding %}
-{% if article.journal.is_conference %}@conference{% else %}@article{% endif %}{% templatetag openbrace %}{{ article.journal.code }} {{ article.id }},
-	author = {{ article.author_list|latex_conform }},
```
StephDriver, line 4: Line 3 has me puzzled, could well be my lack of understanding, but here's my observation.  Please ignore if it a misunderstanding or not relevant:

I read this as a change from an output described by `templatetag openbrace journal.article.code articleid,` to `templatetag openbrace journalarticle.codearticleid,` -> i.e. that the last two terms have been concatenated because the space between them has been removed.

Was that intended?

StephDriver: marking as changes requested only in as far as asking you to consider this observation before merging.

ajrbyers: Yes, the space makes thie field invalid.

PR #4393, opened 2024-09-09
RFC: Use setting to get submit_info form

yakky:
I would like to be able to customize submission forms (and templates as a consequence) to modify the widgets and the behaviour of the form, without having to override the views or doing something tricky as that.

Replacing direct import of forms with import_string function + settings would make the overriding effortless, with any significant complexity in the janeway code and in a completely backward compatible code.

I attached an example about an implementation for ArticleInfoSubmit to get your feedback about the general idea

### DIFF
--- a/src/submission/forms.py
+++ b/src/submission/forms.py
@@ -6,6 +6,7 @@ __maintainer__ = "Birkbeck Centre for Technology and Publishing"
 import re
 
 from django import forms
+from django.utils.module_loading import import_string
 from django.utils.translation import gettext, gettext_lazy as _
 
 from submission import models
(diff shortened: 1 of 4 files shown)


ajrbyers:
@yakky  just an FYI I've added this to our planning call for discussion.

mauromsl (changes requested):
Hi @yakky ,

Sorry it has taken some time to get to this draft. We like the idea of being able to easily override the form class, but there are some concerns about being able to load arbitrary code using the Janeway settings interface - more comments inline.

As for overriding the templates, there is a feature (well, more of a happy accident) where you can use the [Django template engine and settings to load custom backoffice templates](https://docs.djangoproject.com/en/5.1/howto/overriding-templates/#overriding-from-the-project-s-templates-directory). It can be helpful when you also need to alter how the form is laid out, since Django's formfield templates will not allow you to control the outer HTML.

PR #4394, opened 2024-09-10
Bump version v1.7.0

mauromsl:


ajrbyers:
Can we bump the verison of sphinx so these will build?

mauromsl:
> Can we bump the verison of sphinx so docs will build?

done

Issue #2610, opened 2021-11-16
ORCID iDs should not be inputted manually


alainna:
**Is your feature request related to a problem? Please describe.**

Currently it is possible for unverified ORCID iDs to enter author metadata/account data through manual entry. This results in various user errors -- forgetting characters, inputting in incorrect format, inputting someone else's ORCID iD, etc. The only way that one can confirm that an individual holds an ORCID iD is by asking them to authenticate their iD by signing into ORCID and granting Janeway permission to obtain their ORCID iD and perhaps other data on their ORCID record. 

**Describe the solution you'd like**

Authentication status should be determined by a valid access token. 

1. **User profile authentication**: Account holders should be prompted to authenticated their ORCID iD and should not be able to input manually: 
![image](https://user-images.githubusercontent.com/15632391/142080121-3830981d-8308-4ed8-abbf-8247c77b790d.png)

   When loading a user’s personal account profile, the system should use the stored access token to check whether the token is valid (eg GET the user’s ORCID iD. If the token is no longer valid (or if there is no token), display a grey or red X version of the ORCID iD icon and ask the user to reconnect their ORCID iD (or something similar). 
![image](https://user-images.githubusercontent.com/15632391/142080755-be0f7d9e-defc-4f85-9cfd-2d6f3dae3dad.png)

2. **Co-author ORCID iD request**: Submitting authors and moderators/editors should not be able to manually input an ORCID iD to search for a co-author.  If the submitter searches by email address and finds the co-author's existing account, the co-author's authenticated ORCID iD should display if a valid access token is there. 

   Moderators/editors (and submitting co-authors?) should be able to send an email from the item view page to ask that the co-author connect their ORCID iD to Janeway.
![image](https://user-images.githubusercontent.com/15632391/142082169-50ee6af9-304d-4fa4-9d0e-86f11ca4c1fd.png)

S-Haime:
@mauromsl this is listed as in review, is that right?

PR #4399, opened 2024-09-11
save orcid auth tokens, only allow users to add orcid via authentication

everreau:
Addresses: https://github.com/openlibhums/janeway/issues/2610

When orcid is enabled:
- Save orcid auth token and expiration when orcid is authenticated
- Allow users to add orcid via authentication
- When orcid is removed by admin or user revoke token
- Notify users when token is not valid in profile and edit user views
- Don't allow users to manually edit orcids.


ajrbyers:
@everreau can you confirm if this removes the option to edit ORCID's manually?

everreau:
@ajrbyers yes, if orcid is enabled it does not allow manual editing of orcids.

ajrbyers:
@everreau we’ve moved all of the login and registration templates from themes to admin (https://github.com/openlibhums/janeway/tree/master/src/templates/admin/core/accounts). Could you update this PR?

the only major issue I can see is without the option to collect ORCIDs manually we need to include the option to collect them from co-authors post submission.

everreau:
@ajrbyers sorry for the delay responding.  I'll look at this work next week.

everreau:
@ajrbyers my apologies, I probably won't be able to work on this for a bit.  We're fully rebuilding one of our other major services by June.  Re: requesting orcids for co-authors, I remember at the time we had a discussion with Alainna about ORCIDs with FrozenAuthors and I think we came to a conclusion but I'd have to dig up some info to recall.

PR #4402, opened 2024-09-11
Add events for upload / delete files during the submission process

yakky:
Raising an event when user uploads / deletes files allows to handle files transformation during the submission process

### DIFF
--- a/src/submission/views.py
+++ b/src/submission/views.py
@@ -561,6 +561,17 @@ def submit_files(request, article_id):
                 messages.WARNING,
                 _('File deleted'),
             )
+
+            event_logic.Events.raise_event(
+                event_logic.Events.ON_ARTICLE_FILE_DELETE,
+                **{
+                    'request': request,
+                    'file_id': file_id,
+                    'original_filename': file.original_filename,
+                    'article': article
+                }
+            )
+            
             return redirect(reverse('submit_files', kwargs={'article_id': article_id}))
 
         if 'manuscript' in request.POST:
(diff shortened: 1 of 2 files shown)


src/submission/views.py
```diff
@@ -553,33 +553,64 @@ def submit_files(request, article_id):
...
+                                    'file_type': 'manuscript',
+                                    'article': article
+                                }
+                            )
+
+                        return redirect(
+                            reverse('submit_files', kwargs={'article_id': article_id}),
+                        )
+                    except Exception as e:
+                        ms_form.add_error(None, str(e))
```
mauromsl, line 610: A couple of notes here:
1. It is often not very useful to end users to see a system level exception message.
2. The broad `Exception` handler here could swallow errors that should be logged and potentially alert sysadmins.

I think one neat way of solving this would be to pass the form error handler as an optional argument to the event handlers (`ms_form.add_error`) and then tweak the exception handler to something like:

```py
except Exception as e:
    logger.error(e)
    logger.exception(e)
    if not ms_form.errors:
        ms_form.add_error(None,
            f"An unexpected error occurred and has been logged, you can contact support at {support_address}"
        )
```

src/submission/views.py
```diff
@@ -589,19 +620,34 @@ def submit_files(request, article_id):
...
+                                'file_id': new_file,
+                                'original_filename': new_file.original_filename,
+                                'file_type': 'data',
+                                'article': article
+                            }
+                        )
+
+                    return redirect(reverse('submit_files', kwargs={'article_id': article_id}))
+                except Exception as e:
+                    data_form.add_error(None, str(e))
```
mauromsl, line 648: The comment above applies here

mauromsl (changes requested):
Thanks @yakky, seems like a useful set of events to be able to hook to.

I've added a comment about the exception handler

PR #4420, opened 2024-09-24
Fixes a bug where submit_info would clear CI.

### LINKED ISSUE #4421: Competing Interests are wiped by during submission
**Describe the bug**
When submitting an article any competing interests added are dropped.

**Janeway version**
1.5.x

**To Reproduce**
1. Start a submission
2. Add CI
3. Complete submit info
4. See CI has been wiped

**Expected behavior**
CI should be saved

ajrbyers:
Closes #4421

- Removes `competing_interests` from ArticleInfo base class
- Adds new child class `EditArticleMetadata` which inherits from `ArticleInfo`
- `EditArticleMetadata` adds `competing_interests` as one of its fields

### DIFF
--- a/src/submission/forms.py
+++ b/src/submission/forms.py
@@ -33,8 +33,12 @@ class ArticleStart(forms.ModelForm):
 
     class Meta:
         model = models.Article
-        fields = ('publication_fees', 'submission_requirements', 'copyright_notice',
-                  'competing_interests')
+        fields = (
+            'publication_fees',
+            'submission_requirements',
+            'copyright_notice',
+            'competing_interests',
+        )
 
     def __init__(self, *args, **kwargs):
         journal = kwargs.pop('journal', False)
@@ -208,7 +212,6 @@ class ArticleInfo(KeywordModelForm, JanewayTranslationModelForm):
                     if editor_view:
                         self.fields[element.name].required = False
 
-
     def save(self, commit=True, request=None):
         article = super(ArticleInfo, self).save(commit=False)
 
@@ -238,6 +241,10 @@ class ArticleInfoSubmit(ArticleInfo):
     # Filter licenses and sections to publicly available only
     FILTER_PUBLIC_FIELDS = True
 
+    def __init__(self, *args, **kwargs):
+        super(ArticleInfoSubmit, self).__init__(*args, **kwargs)
+        self.fields.pop('competing_interests')
+
 
 class EditorArticleInfoSubmit(ArticleInfo):
     # Used when an editor is making a submission.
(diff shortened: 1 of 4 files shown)


src/submission/forms.py
```diff
@@ -238,13 +241,18 @@ class ArticleInfoSubmit(ArticleInfo):
...
+        self.fields.pop('competing_interests')
+
 
 class EditorArticleInfoSubmit(ArticleInfo):
     # Used when an editor is making a submission.
     FILTER_PUBLIC_FIELDS = False
 
     def __init__(self, *args, **kwargs):
         super(EditorArticleInfoSubmit, self).__init__(*args, **kwargs)
+        self.fields.pop('competing_interests')
```
mauromsl, line 255: Barbara Liskov wouldn't be proud of this, can we instead not have `competing_interests` on the base form and have a different child when it is needed?

I think what we meant to do here is to create a new `EditArticleMetadata` from that adds whichever fields are needed (in this case `competing_interests`)

```mermaid
classDiagram
    class ArticleInfo {
        <<abstract>>
    }

    class ArticleInfoSubmit {
        
    }
    class EditorArticleInfoSubmit {
        
    }
    class EditArticleMetadata {
competing_interests: String
}

    ArticleInfo <|-- ArticleInfoSubmit
    ArticleInfo <|-- EditorArticleInfoSubmit
    ArticleInfo <|-- EditArticleMetadata
    ```

mauromsl (changes requested):

mauromsl (approved):

joemull (changes requested):
These changes seem to make `ArticleInfo` into an abstract base, but we do not explicitly say that anywhere that I can tell. Can we document it as that, so we don't use it by itself anywhere later?

We should also change this test to use the new class, right?

https://github.com/openlibhums/janeway/blob/9051b5f8f3cac907954f0510bafd439e7d397618/src/submission/tests.py#L406-L419

mauromsl:
> These changes seem to make `ArticleInfo` into an abstract base, but we do not explicitly say that anywhere that I can tell. Can we document it as that, so we don't use it by itself anywhere later?
> 
> We should also change this test to use the new class, right?
> 
> https://github.com/openlibhums/janeway/blob/9051b5f8f3cac907954f0510bafd439e7d397618/src/submission/tests.py#L406-L419
FYI: I don't think we should change the tests, the behaviour is still implemented in `ArticleInfo` and thus, it is the correct class member to run the test on.

The current abstraction is a bit leaky (why does `ArticleInfo` take a special argument called `editor_view` if we already have `EditorArticleInfoSubmit`)

@ajrbyers my suggestion would be to raise a separate issue to fix the abstraction here but for the sake of this feature, don't refactor these forms just yet.

ajrbyers:
@joemull are you happy with @mauromsl's suggestion above?

joemull:
> @joemull are you happy with @mauromsl's suggestion above?

Yes, as long as we document it by putting a docstring on the class or opening a new issue.

joemull (approved):

Issue #4027, opened 2024-03-14
Mysql error during native install on Ubuntu 22.04


bjarman:
**Describe the bug**
Applying core.0083_press_editorial_team_20240116...Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "/usr/local/lib/python3.10/dist-packages/django/db/backends/mysql/base.py", line 73, in execute
    return self.cursor.execute(query, args)
  File "/usr/local/lib/python3.10/dist-packages/MySQLdb/cursors.py", line 206, in execute
    res = self._query(query)
  File "/usr/local/lib/python3.10/dist-packages/MySQLdb/cursors.py", line 319, in _query
    db.query(q)
  File "/usr/local/lib/python3.10/dist-packages/MySQLdb/connections.py", line 254, in query
    _mysql.connection.query(self, query)
MySQLdb.OperationalError: (1054, "Unknown column 'press_press.journal_footer_text' in 'field list'")

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/janeway/src/manage.py", line 16, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.10/dist-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.10/dist-packages/django/core/management/__init__.py", line 413, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.10/dist-packages/django/core/management/base.py", line 354, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.10/dist-packages/django/core/management/base.py", line 398, in execute
    output = self.handle(*args, **options)
  File "/janeway/src/utils/management/commands/install_janeway.py", line 99, in handle
    call_command('migrate')
  File "/usr/local/lib/python3.10/dist-packages/django/core/management/__init__.py", line 181, in call_command
    return command.execute(*args, **defaults)
  File "/usr/local/lib/python3.10/dist-packages/django/core/management/base.py", line 398, in execute
    output = self.handle(*args, **options)
  File "/usr/local/lib/python3.10/dist-packages/django/core/management/base.py", line 89, in wrapped
    res = handle_func(*args, **kwargs)
[...]

PR #4425, opened 2024-09-26
Fix issue #4027.

rtoi:
Ensure that column 'press_press.journal_footer_text' is created during the  clean install.


joemull (approved):

mauromsl (commented):
cheers @rtoi 

I've also included this patch on or v1.7.0 release branch

PR #4429, opened 2024-09-27
HTML author comments to editors now display properly

### LINKED ISSUE #4428: Comments to the Editor does not output correctly
**Describe the bug**
Comments to the Editor display in Unassigned is illegible when pasted from word without stripping styling.

**Janeway version**
1.5.x, 1.6.x

**To Reproduce**
1. Submit an article with HTML in the comments to the editor field.

**Expected behavior**
This should render properly.

ajrbyers:
Marks comments to the editor as safe. Adds a migration to fix old content. Adds change to plain_text_validator to remove carriage returns.

- Closes #4428


src/submission/migrations/0081_auto_20240927_1021.py
```diff
@@ -0,0 +1,44 @@
...
+                    article.comments_editor
+                )
+                article.comments_editor = linebreaksbr(article.comments_editor)
+            except ValidationError:
+                # these fields may have harmful content, clean it with bleach
+                article.comments_editor = bleach_value(
+                    article.comments_editor,
+                )
+                print(article.pk)
+                print(article.comments_editor)
```
mauromsl, line 29: can we remove these

src/utils/forms.py
```diff
@@ -144,8 +144,12 @@ def text_sanitizer(text_value, tags=None, attrs=None, excl=ENTITIES_MAP):
 
 def plain_text_validator(value):
     """ A field validator that ensures a textual input has no harmful code"""
-    sanitized = text_sanitizer(value)
-    if value != sanitized:
+
+    # bleach strips carriage returns, convert them to newlines.
+    cleaned = value.replace("\r", "")
```
mauromsl, line 149: The comment is inaccurate, we are not converting them to new lines. In this case it doesn't matter what we convert them to, all we want to do is for them to be ignored for validation.

mauromsl (changes requested):
Awesome, couple of minor things in line

mauromsl (approved):

src/utils/forms.py
joemull: I may be missing something but I believe this will not work for most cases, because it is not handling POSIX `\n`.

https://en.wikipedia.org/wiki/Newline#Representation

https://github.com/django/django/blob/52116774549e27ac5d1ba9423e2fe61c5503a4a4/django/template/defaultfilters.py#L485-L494

Would be great if for clarity we can also use a variable name like `string_with_no_newlines` rather than `cleaned`. The same name `cleaned` is also used above in `text_sanitizer` to mean something different, and `'cleaned' == 'sanitized'` is hard to read anyway.

joemull (changes requested):
Looks good other than `\r` vs. `\n`--see comment on the thread Mauro started.

src/utils/forms.py
joemull: Ah, that's what I thought it was initially and then second-guessed. So yeah, a better variable name would be great.

joemull (approved):

PR #4433, opened 2024-09-30
Let automatic registration of DOIs be optional.

gamboz:
Add a setting to make the automatic registrations of DOIs at acceptance be optional, while still allowing the manual registration.

Rationale: ATM that the same setting `use_crossref` is used
- for automatic registration at acceptance [`Article.accept_article()`](https://github.com/openlibhums/janeway/blob/dcc8f7c25e1e66b8cfda89d18add2a08aa1412ee/src/submission/models.py#L1516) 
- for manual registration after publication [`identifiers.views.issue_doi()`](https://github.com/openlibhums/janeway/blob/dcc8f7c25e1e66b8cfda89d18add2a08aa1412ee/src/identifiers/views.py#L272)

so that it's not possible to disable automatic registration at acceptance while retaining the possibility of manual registration at publication (for instance).

### DIFF
--- a/src/core/logic.py
+++ b/src/core/logic.py
@@ -429,7 +429,7 @@ def get_settings_to_edit(display_group, journal, user):
 
     elif display_group == 'crossref':
         xref_settings = [
-            'use_crossref', 'crossref_test', 'crossref_username', 'crossref_password', 'crossref_email',
+            'use_crossref', 'register_doi_at_acceptance', 'crossref_test', 'crossref_username', 'crossref_password', 'crossref_email',
             'crossref_name', 'crossref_prefix', 'crossref_registrant', 'doi_display_prefix', 'doi_display_suffix',
             'doi_pattern', 'doi_manager_action_maximum_size', 'title_doi', 'issue_doi_pattern', 'register_issue_dois'
         ]
@@ -452,7 +452,7 @@ def get_settings_to_edit(display_group, journal, user):
             'publisher_url', 'contact_info', 'privacy_policy_url', 'auto_signature',
             'slack_logging', 'slack_webhook', 'twitter_handle',
             'switch_language', 'enable_language_text', 'google_analytics_code',
-            'use_ga_four', 'display_login_page_notice', 'login_page_notice', 
+            'use_ga_four', 'display_login_page_notice', 'login_page_notice',
             'display_register_page_notice', 'register_page_notice',
             'support_email', 'support_contact_message_for_staff',
             'from_address', 'replyto_address',
--- a/src/journal/models.py
+++ b/src/journal/models.py
@@ -331,6 +331,11 @@ class Journal(AbstractSiteModel):
     def use_crossref(self):
         return setting_handler.get_setting('Identifiers', 'use_crossref', self, default=True).processed_value
 
+    @property
+    @cache(120)
+    def register_doi_at_acceptance(self):
+        return setting_handler.get_setting('Identifiers', 'register_doi_at_acceptance', self, default=True).processed_value
+
     @issn.setter
     def issn(self, value):
         setting_handler.save_setting('general', 'journal_issn', self, value)
--- a/src/submission/models.py
+++ b/src/submission/models.py
@@ -1513,7 +1513,8 @@ class Article(AbstractLastModifiedModel):
 
         if self.journal.use_crossref:
             id = id_logic.generate_crossref_doi_with_pattern(self)
-            id.register()
+            if self.journal.register_doi_at_acceptance:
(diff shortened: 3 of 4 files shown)


gamboz:
:thinking: I have two failures in the tests, but I can't pin-point them to something related to these changes:

Output of `...manage test...` (edited for clarity)
```
FAIL: test_orcid_registration (src.core.tests.test_app.CoreTests.test_orcid_registration)
...
Couldn't find '<input type="hidden" name="orcid" value="0000-0000-0000-0000" id="id_orcid">' in response

FAIL: test_reviewer_decision_form_no_decision (src.review.tests.ReviewTests.test_reviewer_decision_form_no_decision)
...
'<option value="" selected>...' not found in '<label for="id_decision">Recommendation:...

----------------------------------------------------------------------
Ran 639 tests in 118.368s
FAILED (failures=2)
```

It's been a while since I've run Janeway's tests, I hope I'm not doing something wrong :slightly_smiling_face: 

I'm setting this MR as ready, but please feel free to set it back to "draft".

gamboz:
oops pieces missing...

joemull:
Hi @gamboz, can you say a bit more about why you'd like to have the editor control DOI registration manually? In the past we have found editors are helped most when Janeway handles this for them, but we want to understand the user story you have in mind to make sure. Thanks!

gamboz:
> say a bit more about why you'd like to have the editor control DOI registration manually...

Yes, sure.

It's actually more about _not_ registering at acceptance; we plan to have the system register the DOI automatically at publication (in the near future).

Depending on the journal, we can have these situations:
- we don't want to let readers access accepted but non-published
  papers, so we can't provide a web page to be linked to the DOI at
  acceptance
- the form of the DOI depends the article's primary issue at
  publication and how many other articles of the same type (section)
  have been published in that issue (this is for "historical"
  reasons[*] and will probably review this decision); we cannot know
  this number at acceptance because the publication order depends also
  on how fast the production process goes for every paper
- we have had (rarely) papers withdrawn after acceptance

Consider also that most of our papers came from pre-print archives and
our typesetters are relatively fast (1~2weeks from acceptance to
publication), so the added bonus of registering the DOI at acceptance
is minimal for us.

[*] We had persistent identifiers before DOIs and used that kind of
logic to build these PIDs, then we brought the same logic into the
format of the DOI, but we see no reason to continue this solution ATM :slightly_smiling_face:

joemull:
OK, that helps explain the context. Small follow-up question: For most Janeway journals, registering the DOI at acceptance ensures that it can be used in typesetting, when inserting article metadata into publication-ready files. In your workflow, how can you be sure in typesetting what the DOI will be? Do you have a process for reserving the DOIs before typesetting that does not involve creating it in Janeway?

joemull:
As for including the issue in the DOI, yes I would encourage you to move away from that.

gamboz:
> ... registering the DOI at acceptance ensures that it can be used in typesetting, when inserting article metadata into publication-ready files. In your workflow, how can you be sure in typesetting what the DOI will be? Do you have a process for reserving the DOIs before typesetting that does not involve creating it in Janeway?

Our process is roughly this:
- we typeset in latex and add "placeholders" for DOI and publication date in the typesetted source files
- during production, galleys show the placeholders
- at publication ([here](https://github.com/sissamedialab/wjs-profile-project/blob/c52dfa96571ea2190e9ec27e282bf9ae9b6d4963/wjs/plugins/wjs_review/logic__production.py#L1394)), the system computes the DOI and sets it and the publication date in place of the placeholders
- galleys are re-generated

joemull:
OK, I'll take this back to the team for discussion.

ajrbyers:
> > ... registering the DOI at acceptance ensures that it can be used in typesetting, when inserting article metadata into publication-ready files. In your workflow, how can you be sure in typesetting what the DOI will be? Do you have a process for reserving the DOIs before typesetting that does not involve creating it in Janeway?
> 
> Our process is roughly this:
> 
>     * we typeset in latex and add "placeholders" for DOI and publication date in the typesetted source files
> 
>     * during production, galleys show the placeholders
> 
>     * at publication ([here](https://github.com/sissamedialab/wjs-profile-project/blob/c52dfa96571ea2190e9ec27e282bf9ae9b6d4963/wjs/plugins/wjs_review/logic__production.py#L1394)), the system computes the DOI and sets it and the publication date in place of the placeholders
> 
>     * galleys are re-generated

I'd love a demo of this process, we have other jourals using TEX!

mauromsl (approved):

ajrbyers (approved):

PR #4434, opened 2024-10-01
Added documentation for the triple anonymous peer review system

### LINKED ISSUE #4426: Write documentation for triple-anonymous review
**Describe the area of the documentation you would like updated/added.**
with the implementation of #4416, we need to update the documentation regarding
- Configuration
- What is considered PII

ajrbyers:
Closes #4426

### DIFF
--- a/docs/source/dev.rst
+++ b/docs/source/dev.rst
@@ -10,3 +10,4 @@ Information for developers
    dev/robotsandsitemaps
    dev/api
    dev/plugins
+   dev/pii
--- a/docs/source/dev/pii.rst
+++ b/docs/source/dev/pii.rst
@@ -1,0 +1,65 @@
+Personally Identifiable Information
+===================================
+
+Janeway collects personally identifiable information (PII) from users during registration and the assignment of authors to articles. This information is stored in `Account` and `FrozenAuthor` records.
+
+FrozenAuthor Model
+------------------
+The `FrozenAuthor` model is designed to capture a snapshot of author details at the time of submission and publication. The following fields are considered PII:
+
+- name_prefix: Does not directly identify an individual but could be combined with other information to identify an individual.
+- first_name: Can be used to directly identify an individual.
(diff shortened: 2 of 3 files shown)


S-Haime (approved):
Made some changes to language, otherwise no immediate issues. Will review and edit in more depth as part of Documentation project.

S-Haime:
Cannae find the original PR, but we may also need a 'quiet' notification to Editors when a Section-Editor assigns the Author as Reviewers in triple anonymous. The Author is ofc also blocked from doing the review (I checked) and if 1-click review is not enabled, they cannot decline. If they do have 1-click access, they could potentially review their own paper. (Uncertain as I get a server error when I test). Nevertheless, the Editor ought to be notified to keep an eye on this and ensure the author declines to review without informing the section-editor of the reason why,

S-Haime:
Actually we DEFINITELY need an alert and a half if Authors do ever review their own work (as it would be easy to overlook in a journal with a large number of articles in review - on the overview you'd only see initials, so you'd have to open the article review page to check whilst having the correct permissions).

ajrbyers:
> Actually we DEFINITELY need an alert and a half if Authors do ever review their own work (as it would be easy to overlook in a journal with a large number of articles in review - on the overview you'd only see initials, so you'd have to open the article review page to check whilst having the correct permissions).

When triple anon is enabled perhaps we should just filter out the authors from the reviewer list?

mauromsl (approved):

S-Haime:
> > Actually we DEFINITELY need an alert and a half if Authors do ever review their own work (as it would be easy to overlook in a journal with a large number of articles in review - on the overview you'd only see initials, so you'd have to open the article review page to check whilst having the correct permissions).
> 
> When triple anon is enabled perhaps we should just filter out the authors from the reviewer list?

I considered that, but it would give the game away to the section editor.

PR #4439, opened 2024-10-02
When attempting to fetch an image for a NewsItem that does not exist a 404 not found error is now returned

### LINKED ISSUE #4438: Unhandled ORM get
**Describe the bug**
When a user attempts to get a news item image with an ID that does not exist django returns `NewsItem.DoesNotExist`

**Janeway version**
1.7.x

**To Reproduce**
1. Create a url with a news item that does not exist eg: https://orbit.openlibhums.org/news/id/384/image/9898/
2. See error

**Expected behavior**
This sould return a 404.

**Additional context**
https://github.com/openlibhums/janeway/blob/dcc8f7c25e1e66b8cfda89d18add2a08aa1412ee/src/comms/views.py#L132-L148

ajrbyers:
- Added a try/except block that returns a 404 instead of a 500 when attempting to fetch an image for a `NewsItem` that does not exist.
- Closes #4438

### DIFF
--- a/src/comms/views.py
+++ b/src/comms/views.py
@@ -6,6 +6,7 @@ from django.utils import timezone
 from django.contrib import messages
 from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
 from django.db.models import Count
+from django.http import Http404
 
 from comms import models, forms
 from core import files, logic as core_logic, models as core_models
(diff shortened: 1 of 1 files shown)


StephDriver (approved):
I've tested this, and compared to master I can see the change and it works but I have found some strange behaviour:

- This will return a 404 if the news ID doesn't exist.  
- This will return a 404 if the news ID exists and there is no image for that item.
- But if there is an image, and you type in a different image number to the one associated with the image in the source of the News Item page, it will still access the image and not 404. Master behaves the same, so this is not introducing a new issue.

So - if the PR is just about non-existent news ID, then I think it does what it is supposed to do, but if it is about non-existent images, then it doesn't because it allows any image number as if it were the correct one as long as the news ID is correct.

e.g. 
if image is at: `news/id/699/image/166081/`

Then `news/id/69/image/166081/` will 404 (assuming no other item 69 exists)
But `news/id/699/image/166/` will still access the image as if it were `news/id/699/image/166081/`

I am approving this but leaving it up to the originator to consider the above and decide whether it is working as intended or not, or whether that is a separate issue etc.

ajrbyers:
@StephDriver I'm going to take this one back. It looks like the view has changed the way it operates at some point so there are some superflous arguments.

ajrbyers:
@StephDriver I've gone over this and these superflous attributes are there to work with the generic carousel URL pattern (one that can work for a news item, issue or article) so I think this is fine.

ajrbyers:
Added @mauromsl for second review.

mauromsl (approved):

PR #4450, opened 2024-10-08
Ensures that active reviews count is accurate

### LINKED ISSUE #4392: Active reviews count is incorrect
**Describe the bug**
When viewing a list the number of active reviews a reviewer has doesn't look right.

Its possible that reviews that are withdrawn are not removed from the count.

**Janeway version**
1.7.0

**Expected behavior**
This count should only show the number of open reviews a user has.

https://github.com/openlibhums/janeway/blob/e0552c100397f7a9a3adbe2bc6be83553d08fa7d/src/review/logic.py#L51-L58

ajrbyers:
This PR adds a new completed reviews manager and tidies up the display of review information when assigning reviewers or looking a their histories.

Example:

image

Previously this user would have shown multiple active reviews and the review that was withdrawn would be the latest completed date/time.

Closes #4392

### DIFF
--- a/src/review/logic.py
+++ b/src/review/logic.py
@@ -22,6 +22,7 @@ from django.db.models import (
     When,
     BooleanField,
     Value,
+    Q,
 )
 from django.shortcuts import redirect, reverse
 from django.utils import timezone
@@ -49,7 +50,8 @@ def get_reviewers(article, candidate_queryset, exclude_pks):
         ).exclude(date_complete__isnull=True).order_by("-date_complete")
     )
     active_reviews_count = models.ReviewAssignment.objects.filter(
-        is_complete=False,
+        Q(date_complete__isnull=True) | Q(is_complete=True),
+        article__journal=article.journal,
         reviewer=OuterRef("id"),
     ).values(
         "reviewer_id",
--- a/src/review/logic.py
+++ b/src/review/logic.py
@@ -45,12 +45,8 @@ from submission import models as submission_models
 def get_reviewers(article, candidate_queryset, exclude_pks):
     prefetch_review_assignment = Prefetch(
         "reviewer",
-        queryset=models.ReviewAssignment.objects.filter(
+        queryset=models.ReviewAssignment.completed_reviews.filter(
             article__journal=article.journal,
-            date_accepted__isnull=False,
-            date_complete__isnull=False,
-        ).exclude(
-            decision="withdrawn",
         ).order_by("-date_complete"),
     )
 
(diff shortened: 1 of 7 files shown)


src/review/logic.py
```diff
@@ -49,7 +50,8 @@ def get_reviewers(article, candidate_queryset, exclude_pks):
         ).exclude(date_complete__isnull=True).order_by("-date_complete")
     )
     active_reviews_count = models.ReviewAssignment.objects.filter(
-        is_complete=False,
+        Q(date_complete__isnull=True) | Q(is_complete=True),
```
mauromsl, line 53: Did you mean to use an XOR operator? 
>[Q](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#django.db.models.Q) objects and querysets can now be combined using ^ as the exclusive or (XOR) operator. XOR is natively supported on MariaDB and MySQL. For databases that do not support XOR, the query will be converted to an equivalent using AND, OR, and NOT.

https://docs.djangoproject.com/en/4.2/releases/4.1/#models

Also, it would be good to use a comment here explaining the implicitness of the query (we are trying to consider  declined reviews) or maybe even use a variable/custom filter, e.g

`INCOMPLETE_AND_NOT_DECLINED_FILTER = (Q(date_complete__isnull=True)^ Q(is_complete=True))`

mauromsl (changes requested):

ajrbyers:
I did some investigating here. We need to ensure that assignments where the reviewers decision was not accept are filtered out of the subquery.

StephDriver (approved):

src/review/logic.py
```diff
@@ -43,14 +44,25 @@
...
-        'reviewer',
+        "reviewer",
         queryset=models.ReviewAssignment.objects.filter(
-            article__journal=article.journal
-        ).exclude(date_complete__isnull=True).order_by("-date_complete")
+            article__journal=article.journal,
+            date_accepted__isnull=False,
+            date_complete__isnull=False,
+        ).exclude(
+            decision="withdrawn",
```
mauromsl, line 53: Looking at the above query and this one, it sounds like the business rules for a reviewassignment to be considered as "A user completed a review assignment for this paper using Janeway":

-   It has a `date_accepted`
- AND it has a `date_complete`
- AND the decision was not `"withdrawn"`

The `is_complete` being irrelevant.

If this is the case, can we do either of 

A. Store these filters into a variable (e.g `completed_reviews_filter`)
B. Use a custom manager that applies the filtering rules to the default queryset (e.g. `CompletedReviewsManager`)

mauromsl (changes requested):

mauromsl (approved):

PR #4453, opened 2024-10-09
Fix the item counts on the filters on the article search page

### LINKED ISSUE #4449: Filters need testing around item counts
On Architectural Histories journal there are lots of correctly tagged reviews but the filter now appears to only recognise one when selected. This is a recent issue as this worked fine previously, if it might be possible to get a fix please? Thank you!

Spotted in JW version 1.7.0 at https://journal.eahn.org/articles/?order_by=-date_published&page=1&paginate_by=25&date_published__date__gte=&date_published__date__lte=&section__pk=300

 
image

joemull:
Fixes #4449. 

Note that installations running the isolinear plugin will also need https://github.com/openlibhums/isolinear/pull/4

### DIFF
--- a/src/core/forms/forms.py
+++ b/src/core/forms/forms.py
@@ -589,18 +589,14 @@ class AccessRequestForm(forms.ModelForm):
 class CBVFacetForm(forms.Form):
 
     def __init__(self, *args, **kwargs):
-        # This form populates the facets that users can filter results on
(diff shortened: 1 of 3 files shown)


ajrbyers (approved):
Works like a charm.

mauromsl (approved):

PR #4456, opened 2024-10-11
contact forms validate on server not browser

### LINKED ISSUE #4268: All themes contact forms A11y updates
The contact us form is the only part of the front end which requires user input.  It does not currently comply with [WCAG 3.3 Input Assistance.](https://www.w3.org/WAI/WCAG22/quickref/?showtechniques=321%2C323%2C324%2C331#input-assistance), among other areas of WCAG.

Non-exhaustive list of issues found so far:
- [Material](https://www.glossa-journal.org/contact/) 
    - invalid entries identified by colour change (see WCAG 1.3.3)
    - errors are not described to the user in text.
- [Clean](https://politicalphilosophyjournal.org/contact/)
    - Invalid entries are not reported to the user (try entering random text in the email field, it doesn't react).  They are reported to the screen reader user, but only if they navigate back having left the item.  They are reported on landing, not on exiting.
    - the bracketed email address to evade scraping is awkward when read aloud by screen reader - consider how to handle that. without putting the email address into the HTML.
- [ ] #4482
    - example at [OLH](https://ijwwe.uwp.co.uk/contact/)
    - Page starts with an H4 not an H1
    - Contact form then starts with an H3
    - Invalid entries are not reported to the user (try entering random text in the email field, it doesn't react)
    - Required fields are marked by `*` which gets read out as 'star' - this should either be marked as decorative and the field marked-up as required, or it should be described.
    - Send message button changes colour on hover, but no non-colour indicator of this.

Note: the Clean contact form works well with voiceover, but doesn't indicate required fields visually (does tell the screen reader!).

StephDriver:
closes #4268 

Note from @ajrbyers:

- Steph and I discussed the approach at length and did some reasearch around validation. We decided to disable HTML5 validation as recommended by the UK Gov Frontend team: https://design-system.service.gov.uk/patterns/validation/

### DIFF
--- a/src/templates/admin/journal/contact.html
+++ b/src/templates/admin/journal/contact.html
@@ -30,7 +30,7 @@
                 </div>
                 <div class="large-8 columns">
                     <h3>{% trans "Contact" %}</h3>
-                    <form method="POST">
+                    <form method="POST" novalidate>
                         {% include "elements/forms/errors.html" with form=contact_form %}
                         {% csrf_token %}
                         <div class="row expanded">
--- a/src/themes/OLH/templates/journal/contact.html
+++ b/src/themes/OLH/templates/journal/contact.html
@@ -24,7 +24,7 @@
             </div>
             <div class="large-8 columns">
                 <h3>{% trans "Contact" %}</h3>
-                <form method="POST">
+                <form method="POST" novalidate>
                     {% include "elements/forms/errors.html" with form=contact_form %}
                     {% csrf_token %}
                     <label for="id_recipient">{% trans "Who would you like to contact?" %}</label>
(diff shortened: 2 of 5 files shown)


ajrbyers (changes requested):
https://github.com/openlibhums/janeway/issues/4268 talks about page structure as well as the form itself. Can we reduce the scope of 4268 assuming these are covered elsehwere or update the PR to cover these issues?

StephDriver:
> #4268 talks about page structure as well as the form itself. Can we reduce the scope of 4268 assuming these are covered elsehwere or update the PR to cover these issues?

Looking back, the issue had been reduced to only include Clean theme as the priority work.  I have reviewed it and this fix also addresses material, so I've moved the OLH items (which were the structure ones) to a new issue for fixing later as per the VPAT priorities.

ajrbyers (approved):
I've added a note about this change in devops (https://github.com/openlibhums/devops/issues/60).

joemull (approved):

PR #4464, opened 2024-10-25
Make display of funding info clearer on article page

### LINKED ISSUE #4461: ROR link treated as internal CMS page link
It appears that the field for external ROR funder links is treated as if the input is an internal CMS page. E.g.:
`https://journal.org/funder_articles/https://ror.org/[RORCODE]` instead of `https://ror.org/[RORCODE]`. The link cannot resolve and does not function. This is suboptimal. It would be good to resolve this ASAP as funding information is important.

Version
1.7

To reproduce / for example in the wild:
https://journal.digitalmedievalist.org/article/id/15119/
Scroll down to 'Funding' 
image
image
image

Reported to FD #9304. Please update editor when resolved.

joemull:
Fixes #4461.

See my comment on the original issue for a diagnosis: https://github.com/openlibhums/janeway/issues/4461#issuecomment-2437595887

To make the behavior clear and still display the various IDs for metadata purposes, I have laid out the data using a definition list. I have opted to use the `key_value_above` template from the back office as it is agnostic of styling. If crossing between the themes in this way is undesirable, I can move `key_value_above` to common templates.

image

image

image

### DIFF
--- a/src/journal/views.py
+++ b/src/journal/views.py
@@ -133,16 +133,15 @@ def serve_journal_cover(request):
 
 @has_journal
 def funder_articles(request, funder_id):
-    # NOTE: This view is not usable currently because
-    # it is visually indistinguishable from the main Article search page.
-    # There is a filter applied to the queryset by the funder,
-    # but that is not represented in the filters, and the main heading
-    # just says "Articles".
-    """ Renders the list of articles in the journal.
+    """ Deprecated. Renders the list of articles in the journal.
 
         :param request: the request associated with this call
         :return: a rendered template of all articles
         """
+    raise DeprecationWarning(
+        'This view is deprecated.'
+    )
+
     if request.POST and 'clear' in request.POST:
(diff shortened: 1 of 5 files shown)


src/journal/views.py
```diff
@@ -133,6 +133,11 @@ def serve_journal_cover(request):
 
 @has_journal
 def funder_articles(request, funder_id):
```
ajrbyers, line 135: Consider: Should we mark this view as deprecated? I don't think its particularly useful.

src/templates/common/elements/funder_info_for_readers.html
ajrbyers, line 1: Noice.

ajrbyers (approved):
See comment on view. Otherwise this is a good approach.

ajrbyers (approved):

StephDriver (approved):

PR #4466, opened 2024-10-28
The clean theme carousel now includes a link bringing it in line with other themes

ajrbyers:
Adds a link to the clean theme's carousel.

Screenshot 2024-10-28 at 10 25 51

### DIFF
--- a/src/themes/clean/assets/css/clean.css
+++ b/src/themes/clean/assets/css/clean.css
@@ -245,7 +245,6 @@ main a:hover {
 
 .carousel-caption h2 a {
     color: white;
-    text-decoration: none;
 }
 
 .row-eq-height {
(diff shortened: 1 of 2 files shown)


mauromsl (approved):

src/themes/clean/assets/css/clean.css
```diff
@@ -243,6 +243,10 @@ main a:hover {
     background-color: rgba(55, 55, 55, .8);
 }
 
+.carousel-caption h2 a {
```
StephDriver, line 246: When I tested this, the css here setting the link to white is overridden by both `_reboot.scss` at line 187 and `clean.css` at line 46.  I have had some trouble building the assets, so that might be part of it, but the `.carousel-caption` class does show up further down the list and the colour set within it is overridden.

StephDriver (changes requested):
links yes, CSS - can't get it working.  Could be my end. But needs checking before this is merged.

PR #4472, opened 2024-10-29
#3965 adds text to the move_to_next modal.

### LINKED ISSUE #3965: Edit 'skip stage' pop-up to include current stage and proposed stage.
Currently, when Janeway 'thinks' an article may need to be in a different stage, it does not show the current stage and the stage it would move it to. (See screenshot below for example.) This may confuse editors (especially less experienced ones) as they may not realise an article has in fact changed stage when they see this pop-up. 

image

Displaying the current stage and the stage it would move into on the pop-up would clarify this for editors. This would limit instances where articles are accidentally moved to the wrong stage. (When a stage has been skipped, they can't move it back themselves either from the workflow manager and it has to be done by a member of staff.)

(Based on working with new editors after a migration.)

ajrbyers:
- Adds text to explain the current stage and the stage the article will move to.

Closes #3965 

Screenshot 2025-01-15 at 09 18 28

### DIFF
--- a/src/templates/admin/elements/move_to_next_modal.html
+++ b/src/templates/admin/elements/move_to_next_modal.html
@@ -25,10 +25,10 @@
                 <input type="hidden" name="current_stage" value="{{ article.stage }}">
                 <div class="button-group expanded">
                     <button class="success button" type="submit">
-                        {% trans "Move to next stage" %}
+                        {% trans "Move to" %} {{ article.next_workflow_element|capfirst }}
                     </button>
                     <button class="alert button" data-close type="button">
-                        {% trans "Not right now" %}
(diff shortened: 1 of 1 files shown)


Astronaut-resting (approved):
Better!

StephDriver (approved):
really like this - far clearer.

StephDriver (changes requested):
much clearer, but I think the button text should also update to reflect this.  Instead of move to next stage, a move to `name of next stage` and likewise keep in `name of current stage`

also link to which issue is being closed.

StephDriver (approved):

joemull:
@ajrbyers FYI I was planning to review this one since Steph asked me to. She wasn't able to test it out and there is no screenshot here so I was going to test it today.

PR #4475, opened 2024-10-31
Fixed an issue where html tags where being displayed on news item titles

### LINKED ISSUE #4467: Italics and ampersand in News title not rendering properly
**Describe the bug**
I created a News item whose title field contained both italics and an ampersand character. On the journal's homepage, the italics was displayed as visible `<em>` tags, and the ampersand was displayed as an `&amp;` entity. 

**Janeway version**
1.7.0

**To Reproduce**
1. Go to https://journals.publishing.umich.edu/mp/
2. Scroll down to 'Latest News Posts'
3. See error (however, I've already revised the title of the News item to avoid the use of italics or ampersand, so that they don't have visible errors live on their homepage. See screenshots below

**Screenshots**
image
image

mauromsl:
closes #4467

### DIFF
--- a/src/themes/OLH/templates/journal/homepage_elements/news.html
+++ b/src/themes/OLH/templates/journal/homepage_elements/news.html
@@ -1,5 +1,6 @@
 {% load svg from press_url %}
 {% load truncate %}
+{% load escaping %}
 {% load i18n %}
 
 <section>
@@ -16,7 +17,7 @@
                             <a href="{% url 'core_news_item' item.pk %}"><img
                                     class="news-image"
                                     src="{{ item.best_image_url }}"
-                                    alt="{{ item.title|striptags }}"/></a>
+                                    alt="{{ item.title|striptags|unescape }}"/></a>
                         </div>
                     {% endif %}
                 {% endif %}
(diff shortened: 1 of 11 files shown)


src/themes/OLH/templates/journal/homepage_elements/news.html
```diff
@@ -16,14 +16,14 @@ <h2 class="title">{% trans 'Latest' %} {{ journal_settings.news.news_title }} {%
                             <a href="{% url 'core_news_item' item.pk %}"><img
                                     class="news-image"
                                     src="{{ item.best_image_url }}"
-                                    alt="{{ item.title }}"/></a>
+                                    alt="{{ item.title|safe }}"/></a>
```
ajrbyers, line 19: Should we use striptags here? Tags are not valid inside an alt attribute.

src/themes/material/templates/journal/homepage_elements/news.html
```diff
@@ -15,7 +15,7 @@
                         {% if item.large_image_file or request.journal and request.journal.default_large_image or request.press.default_carousel_image %}
                             <div class="card-image">
                                 <img src="{{ item.best_image_url }}"
-                                     alt="{{ item.title }}"/>
+                                     alt="{{ item.title|safe }}"/>
```
ajrbyers, line 18: Should we use striptags here? Tags are not valid inside an alt attribute.

ajrbyers (changes requested):

ajrbyers (approved):

StephDriver (changes requested):
The italics seem fine, but I don't think strip tags is sorting out the ampersand issue within the alt

image

image

ajrbyers (approved):

PR #4478, opened 2024-11-04
Fixes two issues related to prepublication author and reviewer notifications

### LINKED ISSUE #4477: Staff cannot send publication notification emails in certain circumstances
**Describe the bug**
A production manager was attempting to send a publication notification but received an error.

**Janeway version**
1.7.

**To Reproduce**
1. The circumstances are hard to replicate, recommend pulling a copy of the Michigan database
2. Can be replicated by having an article with no reviewers but notifying reviewers is enabled

**Expected behavior**
The email should be sent without error assuming the form is okay

ajrbyers:
- If no corresp author is present adds this as a warning
- When no corresp author is found set `author_initial['to']` to None, form validation will handle it being empty
- The change above surfaced a bug with in `article.non_correspondence_authors` also fixed by this PR
- The notification template had additional protection as it checked whether the given article had peer reviews, this was not present in the logic file so posting this form would error due to the second form not being present. This check is now also present in the logic.
- Shifts prepub warnings into their own file and inseted with an include.
- Closes #4477

### DIFF
--- a/src/submission/models.py
+++ b/src/submission/models.py
@@ -1121,7 +1121,7 @@ class Article(AbstractLastModifiedModel):
         if self.correspondence_author:
             return self.authors.exclude(pk=self.correspondence_author.pk)
         else:
-            return self.authors
+            return self.authors.all()
 
     def is_accepted(self):
         if self.date_published:
--- a/src/templates/admin/submission/edit/metadata.html
+++ b/src/templates/admin/submission/edit/metadata.html
@@ -21,7 +21,7 @@
         <div class="row expanded">
             <form method="POST"{% if journal_settings.general.abstract_required %} novalidate{% endif %}>
                 <div class="title-area">
-                    <h2>Edit Metadata</h2>
+                    <h2 id="edit-metadata">Edit Metadata</h2>
                     <a class="button" href="{{ return }}"><i class="fa fa-arrow-left"></i>Back</a>
                 </div>
                 <div class="content">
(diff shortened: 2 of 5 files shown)


src/submission/models.py
```diff
@@ -1121,7 +1121,7 @@ def non_correspondence_authors(self):
         if self.correspondence_author:
             return self.authors.exclude(pk=self.correspondence_author.pk)
         else:
-            return self.authors
+            return self.authors.all()
```
mauromsl, line 1124: a bit of a minor thing:
Now that `get_initial_for_prepub_notifications` I think we could get to a situation where non correspondence authors will be added both to the `To` and `cc` because of 

> `author_initial['to'] = article.correspondence_author.email if article.correspondence_author else None`

src/templates/admin/submission/edit/metadata.html
```diff
@@ -221,7 +221,7 @@ <h2>Funding</h2>
         </div>
 
         <div class="title-area">
-            <h2>Current Funders</h2>
+            <h2 id="edit-metadata-currnt-funders">Current Funders</h2>
```
mauromsl, line 224: I don't see this id used anywhere, but there is a potential typo in `currnt` -> `current` that might be worth fixing.

mauromsl (changes requested):
@ajrbyers Couple minor comments inline, feel free to pass on to second reviewer if you think they are a non-issue

mauromsl (approved):

StephDriver (approved):
review of code only as per discussion. I have worked through the new logic and that makes sense to me.

PR #4480, opened 2024-11-04
Fixes the display of date_submitted and date_accepted on OLH and Material themes

### LINKED ISSUE #4479: Date Submitted and Date Accepted are not displaying
**Describe the bug**
Date accepted and date submitted are not displaying on the material theme, may also be an issue on material.

**Janeway version**
1.7.x

**To Reproduce**
1. Enable display of either date accepted or date submitted
2. Check article

**Expected behavior**
Dates should be present

ajrbyers:
Closes #4479

### DIFF
--- a/src/submission/models.py
+++ b/src/submission/models.py
@@ -906,13 +906,14 @@ class Article(AbstractLastModifiedModel):
     def has_galley(self):
         return self.galley_set.all().exists()
 
+    @staticmethod
     @cache(600)
-    def publication_detail_settings(self):
-        display_date_accepted = self.journal.get_setting(
+    def publication_detail_settings(journal):
+        display_date_accepted = journal.get_setting(
             group_name='article',
             setting_name='display_date_accepted',
         )
-        display_date_submitted = self.journal.get_setting(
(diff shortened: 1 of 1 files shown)


src/submission/models.py
```diff
@@ -906,15 +906,30 @@ def metrics(self):
...
+        display_date_accepted = self.journal.get_setting(
+            group_name='article',
+            setting_name='display_date_accepted',
+        )
+        display_date_submitted = self.journal.get_setting(
+            group_name='article',
+            setting_name='display_date_submitted',
+        )
+        return display_date_submitted, display_date_accepted
+
```
mauromsl, line 920: This is an instance method, so the cache won't behave across article instances (A view with 20 articles would still miss the cache 20 times). Consider either adding it to the `journal_settings` we store on the `request` object OR turn this method into a classmethod/staticmethod.

mauromsl (changes requested):
Small comment on the cache

mauromsl (approved):

StephDriver (approved):
tested with accepted, and with submitted setting too, works.

PR #4481, opened 2024-11-04
Removes missed PII from review and revision interfaces

### LINKED ISSUE #4476: PPI visibility - 3A review
Reported by user:
"I have noticed a couple of places where the author data is not completely anonymised.  One occurs when I read reviews about the paper--the author's name is shown in greyscale at the top of the page.  The other occurs when I make a major revisions decision on a paper--the author's name is shown in the "To" box when I write the email.  I have attached screenshots of both."

Version
1.7

**To Reproduce**
See above and see screenshots.
image
image

Notes: Contact AS from F&E upon completion.

ajrbyers:
Closes #4476

### DIFF
--- a/src/security/test_security.py
+++ b/src/security/test_security.py
@@ -4001,6 +4001,9 @@ class TestSecurity(TestCase):
                 'review_in_review',
                 'review_decision',
                 'decision_helper',
+                'review_request_revisions',
+                'request_revisions_notification',
+                'review_view_review'
             ]
             general_views = [
                 'core_dashboard',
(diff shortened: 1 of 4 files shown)


mauromsl (approved):
glad to see us leveraging the custom filter further :)

StephDriver (approved):
tested and found that:

- both pages reported now show `[anonymised data]`
- this only applies to the section editor assigned to that article and when the setting is enabled... when not enabled the names are shown, and when setting is enabled but section editor is not assigned to that article (e.g. viewing list of all articles in review) they can see the author name for all other articles.

It is my understanding this is then intended result - that the anonymisation should only be for the assigned articles not across all articles at the review stage.

Issue #1928, opened 2020-11-13
Add option to add geospatial metadata to articles to improve discovery


nuest:
[This feature is motivated by the EarthArXiv preprint server, but I imagine it being useful for other journals as well. I'll stick to preprints in the following for simplicity.]

**Is your feature request related to a problem? Please describe.**

It can be hard to keep track of new preprints coming out, or to identify work that is relevant to a researcher's work. You can subscribe to feeds or trust social media, but these are also largely based on manual searching and keywords in the title or abstract, which are readily accessible pieces of information. I'm convinced that _geospatial metadata_ could play a great role to help researchers find interesting works. We are working on an OJS plug-in to achieve just that and you can find more [in this abstract](https://septentrio.uit.no/index.php/SCS/article/view/5590/5347).

I think this would be a great feature for Janeway too, especially for EarthArXiv or other instances where articles have a clear spatial component.

**Describe the solution you'd like**

_On submission_ the author is provided the option to enter geospatial metadata. A map is shown where the author can either draw a rectangle (bounding box) or an arbitrary geometry. This geometry is stored in the database.

The same could be done for a "temporal bounding box", so a period of time that the submission is relevant for.

_The reader_ on the website is given the options to

1. see all articles on a map, and by clicking on a geometry, the name and title are shown and the user can click on a link to open the article page
2. can filter search results by providing a time period and/or zoom into a specific area/draw an area of interest on a map

**Additional context**

I'm happy to contribute code here, though not in the near future, possibly autumn next year. The [OJS plug-in under under development](https://github.com/tnier01/geoOJS) might provide some helpful snippets on the UI/mapping side.

I'm sure this is not a simple feature, because to leverage it fully you need a spatial index on the database, which might not be available out of the box, etc.

nuest:
Just FYI: I'm still interested in this and working in a related project (https://projects.tib.eu/optimeta). I'm looking for a student to explore this topic in a thesis. So please keep the issue open :smile:

nuest:
Possibly interested servers/journals because their work has a clear :world_map: spatial component :earth_africa: :earth_asia: :earth_americas: 

- https://www.ecoevorxiv.com/
- https://eartharxiv.org/

nuest:
Short update: we have secured funding :tada: for this development (yes, we have never contributed to Janeway before, but we're optimists) as part of the project KOMET :comet: https://projects.tib.eu/komet

[...]

PR #4483, opened 2024-11-08
Support for the Research Organization Registry (ROR)

### LINKED ISSUE #3168: Add support for Research Organization Registry (ROR)
## Problem
To show the impact of our publishing activities, we often want to see how many affiliates of a given university have published in our journals. Authors are encouraged to enter the names of their universities when they publish through Janeway, but they do not always give the same names for their universities. As a result, if we tell the computer to give us all the authors who typed in a given name, we may be missing some out--it will be imprecise and messy.

This is the problem solved by the Research Organization Registry (ROR). They assign unique identifiers to organizations, like DOIs for digital objects, ORCIDs for people, and ISBNs for books.

To use ROR in Janeway, we need to build the data model for it in our database and provide interfaces that let individual users select an authoritative organization name, and that allow press managers to safely match existing institution names to ROR identifiers.

## Proposed user journeys
- [x] As a user registering for the first time, I should not be asked for institution, affiliation, and country right away, but instead have an option on my profile page to add an affiliation, which should lead me through the data entry and ROR selection steps (if applicable) on a dedicated page (change theme to admin, for consistency with UI in submission?)
image
image

## Draft data model

```mermaid
classDiagram
    Account "0..1" <-- "*" Affiliation
    FrozenAuthor "0..1" <-- "*" Affiliation
    PreprintAuthor "0..1" <-- "*" Affiliation
    Affiliation "*" --> "1" Organization
    Organization "*" --> "*" Location
    OrganizationName "*" --> "1" Organization
    Location "*" --> "0..1" Country
    class Account{
        property affiliation() str
        property institution() str
        property department() str
    }
    note for Account "CharField institution deprecated\nCharField department deprecated"
    class FrozenAuthor{
        property affiliation() str
        property institution() str
        property department() str
    }
    note for FrozenAuthor "CharField institution deprecated\nCharField department deprecated"
    class PreprintAuthor{
        property affiliation() str
    }
    note for PreprintAuthor "CharField affiliation deprecated"
    class Affiliation{
        CharField title
        CharField department
        BooleanField is_primary
        DateField start
        DateField end
        property current() bool
    }
    class Organization{
        URLField ror
        CharField ror_status
    }
    class OrganizationName{
        CharField value
        CharField language
    }
    class Location{
        CharField name
        CharField latitude
        CharField longitude
        IntegerField geonames_id
    }
```

## Additional context
- See https://github.com/BirkbeckCTP/hourglass/pull/125 for one example of how to integrate the form and API
- RORs are required in Crossref 5.3: https://www.crossref.org/documentation/schema-library/metadata-deposit-schema-5-3-1/
- Guide for publishers: https://ror.readme.io/docs/publishers
- REST API: https://ror.readme.io/docs/rest-api
- Existing affiliation data should be matched with ROR to back-populate ROR IDs where possible

joemull:
Closes #3168.

## Overview
This piece of work adds a ROR data model, including affiliations, organizations, organization names, and locations. It includes an import routine to fetch and process ROR's full database. I've included as much backwards compatibility as possible for importers that do not know about ROR. It also adds a user interface for editing one's own affiliations, and an admin interface for all of the relevant models.

## Command line interface

```sh
❯ python src/manage.py import_ror_data
DEBUG 2024-12-09 18:41:33,666 connectionpool P:131157 T:135176669522816 Starting new HTTPS connection (1): zenodo.org:443
DEBUG 2024-12-09 18:41:34,697 connectionpool P:131157 T:135176669522816 https://zenodo.org:443 "GET /api/communities/ror-data/records?sort=newest HTTP/1.1" 200 None
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 100/100 [00:00<00:00, 101.06it/s]
INFO 2024-12-09 18:41:38,658 import_ror_data P:131157 T:135176669522816 successful
```

## User interface

image
image
image
image
image
image
image


MartinPaulEve:
If you enter an organization name that isn't found (e.g., in test import, "Birkbeck") and then click on one of the pagination items at the bottom, you are redirected to an erroring page eg:

http://127.0.0.1:8000/JRNL/profile/organization/search/?q=birkbeck&page=3&paginate_by=25

This errors because the item isn't found, but the search query is inserted into the pagination link

src/submission/tests.py
```diff
@@ -66,28 +66,32 @@ def create_journal():
...
+    @classmethod
+    def create_authors(cls):
         author_1_data = {
+            'email': 'one@example.org',
             'is_active': True,
             'password': 'this_is_a_password',
             'salutation': 'Prof.',
             'first_name': 'Martin',
+            'middle_name': '',
             'last_name': 'Eve',
```
MartinPaulEve, line 78: Top test author

src/utils/migrations/0035_rorimport_rorimporterror.py
```diff
@@ -0,0 +1,37 @@
...
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='RORImport',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('started', models.DateTimeField(auto_now_add=True)),
+                ('stopped', models.DateTimeField(blank=True, null=True)),
+                ('status', models.CharField(choices=[('ongoing', 'Ongoing'), ('unnecessary', 'Unnecessary'), ('successful', 'Successful'), ('failed', 'Failed')], default='ongoing')),
```
MartinPaulEve, line 20: Should these choices reference RORImportStatus rather than duplicating?

MartinPaulEve (changes requested):
This mostly looks really good. Just a few small things.

- [ ] Pagination in ROR search currently redirects to page not found error
- [x] List of affiliations should show which is primary?
- [ ] RORImportStatus should not be duplicated?

src/core/logic.py
```diff
@@ -868,7 +868,7 @@ def check_for_bad_login_attempts(request):
     time = timezone.now() - timedelta(minutes=10)
 
     attempts = models.LoginAttempt.objects.filter(user_agent=user_agent, ip_address=ip_address, timestamp__gte=time)
-    print(time, attempts.count())
+    logger.info(time, attempts.count())
```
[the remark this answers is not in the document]
joemull: Yes probably, though I'm just removing the noisy print statement here, not changing this feature on this branch.

src/submission/tests.py
joemull: Without a doubt!

src/utils/migrations/0035_rorimport_rorimporterror.py
joemull: I believe this is the standard way Django records choices in migrations, even when the choices are expressed in a class. The reason is that if the migration referenced a variable in the models file, then if the choices list ever changed there, the migration might stop working.

Example:
https://github.com/openlibhums/janeway/blob/bc8a3279b15d2a88d6390c4136655fcd83056ca4/src/journal/migrations/0031_issue_old_issue_type.py#L35

joemull:
> * [ ]  Pagination in ROR search currently redirects to page not found error

This turned out to be a Django bug I believe. I think it's obscure because most people don't encounter it, since they submit their search query to change the list of items before selecting a page.

But I've gone ahead and fixed it for this view.

> * [ ]  List of affiliations should show which is primary?

Yes, good call. I added a label.

I also updated the save method to make the affiliation primary if it's the first one.

image

> * [ ]  RORImportStatus should not be duplicated?

That's standard, I'm pretty sure. See comment inline.

joemull:
Martin approved this off GitHub on 13 Dec. Moving it along to @ajrbyers and @mauromsl.

src/core/models.py
```diff
@@ -1906,3 +1923,558 @@ def log_hijack_ended(sender, hijacker, hijacked, request, **kwargs):
...
+        Intended for use in batch importers where ROR data is not available
+        in the data being imported.
+        Does not support ROR ids, ROR name types, or location data.
+        """
+        # Is there a single exact match in the
+        # canonical name data from ROR (e.g. labels)?
+        try:
+            organization = cls.objects.get(labels__value=value)
+            created = False
+        except cls.DoesNotExist or cls.MultipleObjectsReturned:
```
ajrbyers, line 2095: This wont catch `cls.MultipleObjectsReturned`. It should be a tuple `except (cls.DoesNotExist, cls.MultipleObjectsReturned):`

src/core/models.py
```diff
@@ -1906,3 +1923,558 @@ def log_hijack_ended(sender, hijacker, hijacked, request, **kwargs):
...
+            # Or maybe one in the past or alternate
+            # name data from ROR (e.g. aliases)?
+            try:
+                organization = cls.objects.get(aliases__value=value)
+                created = False
+            except cls.DoesNotExist or cls.MultipleObjectsReturned:
```
ajrbyers, line 2101: Same issue here. This should be a tuple.

src/core/models.py
```diff
@@ -1906,3 +1923,558 @@ def log_hijack_ended(sender, hijacker, hijacked, request, **kwargs):
...
+                # account / frozen author / preprint author?
+                try:
+                    organization = cls.objects.get(
+                        affiliation__account=account,
+                        affiliation__frozen_author=frozen_author,
+                        affiliation__preprint_author=preprint_author,
+                        ror__exact='',
+                    )
+                    created = False
+                except cls.DoesNotExist or cls.MultipleObjectsReturned:
```
ajrbyers, line 2113: Same issue here. This should be a tuple.

src/core/models.py
```diff
@@ -1906,3 +1923,558 @@ def log_hijack_ended(sender, hijacker, hijacked, request, **kwargs):
...
+        Backwards-compatible API for finding a matching organization name.
+        Intended for use in batch importers where ROR data is not available
+        in the data being imported.
+        Does not support ROR ids, ROR name types, or location data.
+        """
+        # Is there a single exact match in the
+        # canonical name data from ROR (e.g. labels)?
+        try:
+            organization = cls.objects.get(labels__value=value)
+            created = False
```
ajrbyers, line 2094: If an exception is raised `created` wont be set. Should it be set before the try block?

src/templates/admin/elements/layout/key_value_above.html
ajrbyers, line 1: Nice.

ajrbyers (changes requested):
This is generally very good. The CBVs are growing on me... slowly.

There are some comments inline that should be addressed. Here are some general comments in addition:

- Institution and Affiliation fields are removed without preservation.
- Mauro made a suggestion to handle setting inst/dept etc using the Manager, this should be added.
- When my affiliation was snapshotted it did not include the start/end date or the title.
- There is a conflict currently listed


Screenshot 2025-01-15 at 10 52 31

src/utils/management/commands/import_ror_data.py
```diff
@@ -0,0 +1,53 @@
...
+    help = "Fetches ROR data and generates Organization records."
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--test_full_import',
+            help='By default, the command only runs 100 records when DEBUG=True.'
+                 'Pass --test_full_import to import the entire dump in development.',
+            action='store_true',
+        )
+        return super().add_arguments(parser)
```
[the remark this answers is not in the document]
mauromsl: It is still good practice to call `super()` when overriding a method like this, in case the underlying implementation changes or a mixin is added in the inheritance chain

src/core/views.py
```diff
@@ -2775,3 +2780,153 @@ def post(self, request, *args, **kwargs):
...
+
+    def get_context_data(self, *args, **kwargs):
+        context = super().get_context_data(*args, **kwargs)
+        context['account'] = self.request.user
+        affiliation_id = self.kwargs.get('affiliation_id')
+        context['affiliation'] = core_models.Affiliation.objects.get(
+            pk=affiliation_id
+        )
+        context['organization'] = context['affiliation'].organization
+        return context
```
mauromsl, line 2932: Have we discussed introducing the use of Django's generic CBVs?
They don't seem to provide any value in terms of reusability, have added cognitive load over existing views and more importantly, they are inconsistent with the rest of the codebase

joemull: > Have we discussed introducing the use of Django's generic CBVs? They don't seem to provide any value in terms of reusability, have added cognitive load over existing views and more importantly, they are inconsistent with the rest of the codebase

Not sure I agree that "they don't seem to provide any value in terms of reusability". Let's schedule a time to talk about your concerns.

ajrbyers:
@joemull running the migrations on a clean OLH database gives me:

```python
Running migrations:
  Applying utils.0035_rorimport_rorimporterror... OK
  Applying core.0100_location_organization_affiliation... OK
  Applying core.0101_migrate_affiliation_institution...Traceback (most recent call last):
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/db/backends/utils.py", line 89, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.StringDataRightTruncation: value too long for type character varying(200)


The above exception was the direct cause of the following exception:

  File "/Users/ajrbyers/Code/janeway/src/manage.py", line 16, in <module>
    execute_from_command_line(sys.argv)
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/core/management/base.py", line 412, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/core/management/base.py", line 458, in execute
    output = self.handle(*args, **options)
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/core/management/base.py", line 106, in wrapper
    res = handle_func(*args, **kwargs)
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/core/management/commands/migrate.py", line 356, in handle
    post_migrate_state = executor.migrate(
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/db/migrations/executor.py", line 135, in migrate
    state = self._migrate_all_forwards(
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/db/migrations/executor.py", line 167, in _migrate_all_forwards
    state = self.apply_migration(
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/db/migrations/executor.py", line 252, in apply_migration
    state = migration.apply(state, schema_editor)
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/db/migrations/migration.py", line 132, in apply
    operation.database_forwards(
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/db/migrations/operations/special.py", line 193, in database_forwards
    self.code(from_state.apps, schema_editor)
  File "/Users/ajrbyers/Code/janeway/src/core/migrations/0101_migrate_affiliation_institution.py", line 71, in migrate_affiliation_institution
    create_affiliation(
  File "/Users/ajrbyers/Code/janeway/src/core/migrations/0101_migrate_affiliation_institution.py", line 47, in create_affiliation
    organization = create_organization(apps, old_institution, old_country)
  File "/Users/ajrbyers/Code/janeway/src/core/migrations/0101_migrate_affiliation_institution.py", line 30, in create_organization
    OrganizationName.objects.create(
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/db/models/manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/db/models/query.py", line 658, in create
    obj.save(force_insert=True, using=self.db)
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/db/models/base.py", line 814, in save
    self.save_base(
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/db/models/base.py", line 877, in save_base
    updated = self._save_table(
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/db/models/base.py", line 1020, in _save_table
    results = self._do_insert(
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/db/models/base.py", line 1061, in _do_insert
    return manager._insert(
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/db/models/manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/db/models/query.py", line 1805, in _insert
    return query.get_compiler(using=using).execute_sql(returning_fields)
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/db/models/sql/compiler.py", line 1822, in execute_sql
    cursor.execute(sql, params)
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/db/backends/utils.py", line 102, in execute
    return super().execute(sql, params)
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/db/backends/utils.py", line 67, in execute
    return self._execute_with_wrappers(
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/db/backends/utils.py", line 80, in _execute_with_wrappers
    return executor(sql, params, many, context)
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/db/backends/utils.py", line 89, in _execute
    return self.cursor.execute(sql, params)
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/db/utils.py", line 91, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/Users/ajrbyers/.venvs/janeway/lib/python3.9/site-packages/django/db/backends/utils.py", line 89, in _execute
    return self.cursor.execute(sql, params)
django.db.utils.DataError: value too long for type character varying(200)
```

Looks like `OrganizationName.value` isn't large enough for the old institution field:

https://github.com/openlibhums/janeway/blob/d9c1ea56fcb2a0e12c5a8798e0518320e36e3da7/src/core/models.py#L277-L282

ajrbyers (changes requested):
See comment on migration error.

joemull:
OK expanded it @ajrbyers

src/review/logic.py
```diff
@@ -824,24 +824,22 @@ def process_reviewer_csv(path, request, article, form):
         reader = csv.DictReader(csv_file)
         reviewers = []
         for row in reader:
-            try:
-                country = core_models.Country.objects.get(code=row.get('country'))
-            except core_models.Country.DoesNotExist:
-                country = None
-
```
ajrbyers, line 831: This is removed and not handled so this data is lost.

src/utils/models.py
```diff
@@ -3,10 +3,12 @@
 __license__ = "AGPL v3"
 __maintainer__ = "Birkbeck Centre for Technology and Publishing"
 
-import json as jason
```
ajrbyers, line 6: Change this back......................... JASON! 😄

ajrbyers (changes requested):
Thanks for this, Joe. Its looking good. I've found a couple of bugs that need to be squashed and added a couple of comments for you to look over.

src/utils/models.py
joemull: Who is Jason and what did he do to Janeway???? 😄

src/utils/testing/helpers.py
```diff
@@ -62,6 +62,7 @@ def create_user(username, roles=None, journal=None, **attrs):
             journal=journal
         )
 
+    user.save()
```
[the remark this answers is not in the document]
joemull: Thanks for catching that. It was an earlier attempt at backward compatibility before writing the ModelManager.

src/review/logic.py
joemull: Thanks for catching. Again an issue with the old backwards compatibility approach.

src/core/forms/forms.py
```diff
@@ -940,3 +940,33 @@ class AccountRoleForm(forms.ModelForm):
...
+        model = models.Affiliation
+        fields = '__all__'
+        widgets = {
+            'account': forms.HiddenInput,
+            'frozen_author': forms.HiddenInput,
+            'preprint_author': forms.HiddenInput,
+            'organization': forms.HiddenInput,
+            'start': HTMLDateInput,
+            'end': HTMLDateInput,
+        }
```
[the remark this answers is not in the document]
joemull: Ah, see what you're saying now. I'll keep working on it and remove the account-related fields from the form. I'll use a different form for the frozen author editing interface, but that one probably also won't include account.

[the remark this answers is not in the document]
joemull: Blocked by https://github.com/openlibhums/janeway/issues/4625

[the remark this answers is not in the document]
joemull: Unblocked and changes made.

src/core/forms/forms.py
```diff
@@ -954,20 +954,33 @@ class Meta:
...
         }
 
+    def __init__(self, *args, **kwargs):
+        self.organization = kwargs.pop('organization', None)
+        super().__init__(*args, **kwargs)
+
+    def save(self, commit=True):
+        affiliation = super().save(commit=False)
+        request = get_current_request()
+        affiliation.account = request.user
```
ajrbyers, line 978: This works but why not just pass the user object in like organization?

ajrbyers (changes requested):

ajrbyers (approved):

src/utils/models.py
```diff
@@ -348,3 +354,170 @@ def fetch(url, up_auth_file='', up_base_url='', ojs_auth_file=''):
...
+        """
+        timestamps = {}
+        for ror, ts in organizations.values_list("ror", "ror_record_timestamp"):
+            timestamps[ror] = ts
+        filtered_data = []
+        for record in tqdm.tqdm(ror_data, desc="Finding new or updated ROR records"):
+            ror = record.get("id", {})
+            last_modified = record.get("admin", {}).get("last_modified", {})
+            timestamp = last_modified.get("date", "")
+            if ror and timestamp and timestamp > timestamps.get(ror, ''):
```
mauromsl, line 511: Can we make sure the defaults for these `.get()` are also floats for explicitness? we can go with `timestamps.get(ror, float())` or `timestamps.get(ror, float('-inf'))` if the timestamp could somehow be 0.0

mauromsl (changes requested):
Thanks Joe, lots of work went into this one I can tell.

some important comments regarding backwards compatibility, there rest are just some polish and nitpicking,but very happy with the work at large here

src/submission/models.py
```diff
@@ -1930,6 +1930,31 @@ def pinned(self):
...
+        # create or update affiliation
+        if institution or department or country:
+            Affiliation.naive_get_or_create(
+                institution=institution,
+                department=department,
+                country=country,
+                frozen_author=frozen_author,
+            )
+
+        return frozen_author, created
```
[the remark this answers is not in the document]
joemull: @mauromsl I can expand this, but I believe it will also need to include get methods, because if you pass a field to `get_or_create` that does not exist on the model, the lookup will fail: `django.core.exceptions.FieldError: Cannot resolve keyword 'country' into field.` OK with you if I also override `get` for all three models in addition to overriding `create` and `update`?

src/cron/management/commands/send_publication_notifications.py
```diff
@@ -47,12 +50,12 @@ def handle(self, *args, **options):
                 setting_name='send_reader_notifications',
                 journal=journal,
             ).value:
-                print('Sending notification for {}'.format(journal.name))
+                logger.info('Sending notification for {}'.format(journal.name))
```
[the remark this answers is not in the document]
joemull: I believe I changed this so that they would not interrupt tests. Is there something wrong with logger.info here?

src/core/views.py
```diff
@@ -2800,3 +2805,221 @@ def post(self, request, *args, **kwargs):
...
+            organization=organization,
+        )
+        if form.is_valid():
+            affiliation = form.save()
+            messages.add_message(
+                request,
+                messages.SUCCESS,
+                _("Affiliation created: %s") % affiliation,
+            )
+            return redirect(reverse('core_edit_profile'))
```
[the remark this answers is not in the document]
joemull: Sure, I've added a clean method.

src/utils/models.py
joemull: This timestamp will always be a string, actually. It's just `YYYY-MM-DD` from the ROR JSON, no Python object being created. Is that alright?

joemull:
OK @mauromsl, I've addressed some of the smaller things but have questions about others.

joemull:
Two more things in addition to the above comments, @mauromsl:

- I've updated the exclusive fields check to use the constraint like I did on the credit branch, but also added an option for all fields to be blank, which is a logical conclusion of their individual `blank=True` settings. I'll update the credit branch as well.
- I've gone ahead and moved storage of ROR IDs to the short, non-URL form, even though you didn't mention it in your review, because of our conversation the other day about storing URLs in databases. This has a slight UX hit in that you can't search by the full ROR URL. But that is not a big use case, and we don't really want end-users dealing with raw ROR IDs anyway, so I can live with it.

src/core/models.py
```diff
@@ -230,6 +236,28 @@ def create_superuser(self, email, password, **kwargs):
...
         return AccountQuerySet(self.model)
 
+    def get_or_create(self, defaults=None, **kwargs):
+        """
+        Backwards-compatible override for affiliation-related kwargs
+        """
+        # check for deprecated fields related to affiliation
+        institution = kwargs.pop('institution', '')
+        department = kwargs.pop('department', '')
+        country = kwargs.pop('country', '')
```
mauromsl, line 246: we need to check and pop these keys from the `defaults` argument as well.

src/core/models.py
```diff
@@ -230,6 +236,28 @@ def create_superuser(self, email, password, **kwargs):
     def get_queryset(self):
         return AccountQuerySet(self.model)
 
+    def get_or_create(self, defaults=None, **kwargs):
+        """
+        Backwards-compatible override for affiliation-related kwargs
+        """
```
mauromsl, line 242: we should instead do this at the Queryset level, since django will delegate this call to `AccountQuerySet` and the queryset also exposes these ORM operations.

Also, don't forget about `get`, `create`, `filter` and `update_or_create`

The bulk operations can be ignored IMO

src/core/models.py
```diff
@@ -1919,3 +1968,641 @@ def log_hijack_ended(sender, hijacker, hijacked, request, **kwargs):
...
+        related_name='acronyms',
+        blank=True,
+        null=True,
+        help_text="This name is a ROR-provided acronym.",
+    )
+    language = models.CharField(
+        max_length=10,
+        blank=True,
+        choices=submission_models.LANGUAGE_CHOICES,
+    )
```
mauromsl, line 2024: Is there a rationale behind avoiding the use of `django-modeltranslations` here? Does it have to do with the need for a distinction between translations and `ror_display_for`?

src/core/models.py
```diff
@@ -1919,3 +1968,641 @@ def log_hijack_ended(sender, hijacker, hijacked, request, **kwargs):
...
+
+    def __str__(self):
+        return self.value
+
+
+def validate_ror_id(ror_id):
+    ror_regex = '^0[a-hj-km-np-tv-z|0-9]{6}[0-9]{2}$'
```
mauromsl, line 2031: It is good practice to [compile the regex](https://docs.python.org/3/library/re.html#re.compile) explicitly as a constant, both for performance and reusability. Also, remember to flag the string as a regex with `r'^0[a-hj-km-np-tv-z|0-9]{6}[0-9]{2}$'`

src/core/models.py
```diff
@@ -1919,3 +1968,641 @@ def log_hijack_ended(sender, hijacker, hijacked, request, **kwargs):
...
+            ).update(
+                is_primary=False,
+            )
+
+    @classmethod
+    def get_primary(
+        cls,
+        account=None,
+        frozen_author=None,
+        preprint_author=None,
```
mauromsl, line 2456: I don't think we ever want three arguments and instead we can use a single mixed-typed argument (e.g. 'author') which can be an instance of a Model that has the `affiliation_set` interface

src/core/models.py
```diff
@@ -1919,3 +1968,641 @@ def log_hijack_ended(sender, hijacker, hijacked, request, **kwargs):
...
+        try:
+            affiliation, _created = Affiliation.objects.get_or_create(
+                account=account,
+                frozen_author=frozen_author,
+                is_primary=True,
+            )
+            affiliation.department = value
+            affiliation.save()
+        except ValueError:
+            logger.warn('The department could not be set.')
```
mauromsl, line 2566: if we use a warning level, we need more context on this message (e.g.   `logger.warn('The department could not be set to "%s"' % value)`

src/core/models.py
```diff
@@ -1919,3 +1968,641 @@ def log_hijack_ended(sender, hijacker, hijacked, request, **kwargs):
...
+        # account / frozen author / preprint author has been saved already
+        try:
+            affiliation, _created = Affiliation.objects.get_or_create(
+                account=account,
+                frozen_author=frozen_author,
+                is_primary=True,
+            )
+            affiliation.department = value
+            affiliation.save()
+        except ValueError:
```
mauromsl, line 2565: Given how we are running the query, we also now need to handle `django.db.IntegrityError` for the `check_exclusive_fields_constraint` declaration on `Affiliation`

src/core/models.py
```diff
@@ -1919,3 +1968,641 @@ def log_hijack_ended(sender, hijacker, hijacked, request, **kwargs):
...
+        decimal_places=6,
+        blank=True,
+        null=True,
+    )
+    longitude = models.DecimalField(
+        max_digits=9,
+        decimal_places=6,
+        blank=True,
+        null=True,
+    )
```
mauromsl, line 2597: Are we making use of these fields? If not, I reckon it is worth holding off on exposing an API that we might need to later make backwards compatible if we end up using [PostGIS](https://postgis.net/), similar to the ongoing [KOMET](https://github.com/openlibhums/janeway/issues/1928) project

src/cron/management/commands/send_publication_notifications.py
```diff
@@ -47,12 +50,12 @@ def handle(self, *args, **options):
...
                 journal=journal,
             ).value:
-                print('Sending notification for {}'.format(journal.name))
+                self.stdout.write('Sending notification for {}'.format(journal.name))
                 readers = journal.users_with_role('reader')
                 bcc_list = [reader.email for reader in readers]
 
                 if bcc_list:
-                    print("Sending notifications to {}".format(
+                    logger.info("Sending notifications to {}".format(
```
mauromsl, line 58: we still have a few messages not being received by `stdout`

src/submission/models.py
```diff
@@ -1932,6 +1932,31 @@ def pinned(self):
...
+        if institution or department or country:
+            Affiliation.get_or_create_without_ror(
+                institution=institution,
+                department=department,
+                country=country,
+                frozen_author=frozen_author,
+            )
+
+        return frozen_author, created
+
```
mauromsl, line 1958: as with `Account`, we should be thorough here and make sure we cover the main ORM methods and do so at the `Queryset` level

src/submission/tests.py
```diff
@@ -51,8 +51,8 @@ def test_new_journals_has_submission_configuration(self):
         if not self.journal_one.submissionconfiguration:
             self.fail('Journal does not have a submissionconfiguration object.')
 
-    @staticmethod
-    def create_journal():
+    @classmethod
+    def create_journal(cls):
```
mauromsl, line 55: what does this change achieve? the `cls` argument is unused

mauromsl (changes requested):
Mega job here @joemull 

I've added a few comments regarding backwards compatibility and database integrity but very happy overall.

I'm testing out the ROR data import procedure locally to try and anticipate any possible hiccups in production but I don't want to hold you up further, good work!

src/core/models.py
joemull: Thanks for raising this. Yes, ROR data sometimes has multiple names per language. I've written it up here: https://github.com/openlibhums/janeway/issues/4657

Coincidentally, the importer was bugged because ROR gives 2-character language codes but our `LANGUAGE_CHOICES` uses 3-character codes. So I've fixed that by using a small ISO-639 utility. I've also documented which version of ISO-639 we hardcode, for future reference.

src/core/models.py
joemull: Yeah, good point. Removed these fields.

mauromsl (changes requested):
Awesome! a couple minor nitpicks and we are good to go. This is turning into a very nice piece of kit :)

mauromsl (approved):

PR #4488, opened 2024-11-14
Update docs copyright and author statements.

ajrbyers:


docs/source/conf.py
```diff
@@ -20,8 +20,8 @@
 # -- Project information -----------------------------------------------------
 
 project = 'Janeway'
-copyright = '2018, Andy Byers, Mauro Sanchez & Martin Paul Eve'
-author = 'Andy Byers, Mauro Sanchez & Martin Paul Eve'
+copyright = '2018-2024 Birkbeck, University of London'
+author = 'OLH Publishing Technologies Team'
```
joemull, line 24: Can I make a request that we use the exact organization name for the author? It is possible someone coming across this will not know what OLH is, and the name of our team is not an organization either, so it makes more sense to me to have 'Open Library of Humanities' as the author.

joemull (changes requested):
Copyright line is great, though I'd suggest changing the author line.

StephDriver (approved):

Issue #4215, opened 2024-05-31
Icon A11y improvements


StephDriver:
Icons are not being handled accessibly.  

Icons are non-text content as per [WCAG 1.1.1.](https://www.w3.org/TR/WCAG22/#non-text-content)
The key technique is [H86](https://www.w3.org/WAI/WCAG22/Techniques/html/H86.html)
 They must also be semantically identified, as per [ARIA24 Technique](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA24.html).  

Most of our icons come from Font Awesome, but there are a few others, such as the ORCiD logo, that need individual handling.

For Font Awesome:
- [x] consider implementing [Font Awesome's auto accessibility feature ](https://docs.fontawesome.com/web/dig-deeper/accessibility)
- [x] if not using auto accessibility then update manually (syntax further down that same page).  For each icon:
    - consider whether it is decorative or semantic
    - update icon syntax accordingly.
- [x] update ORCiD logos to be accessible.
- [x] share icons have already been updated in #3936

StephDriver:
from #4194 

- Sample 1, item 9. The icon after author's name is not identified, simply 'link' `<a itemprop="email" href="[redacted]" class="fa fa-envelope email-link"></a> `
- Sample 2, item 21. The 'eye' icon next to views has no description at all, found by voice over, but has no description to give. `<i class="fa fa-eye"></i>`
- Sample 2, item 22. likewise, the 'download' icon. `<i class="fa fa-download"></i>`
- Sample 2, item 59. envelope icon following first author is missing alt text, such that is is read simply as 'link' with no description at all. `<a itemprop="email" href="mailto:[redacted]" class="fa fa-envelope email-link"></a>`
- Sample 2, item 74. `<i class="fa fa-check-circle-o" style="color: green"></i>` no alt for the icon, but also not marked as decorative.
- Sample 5.1, item 14. the icons for calendar and book that adorn the article details have no description, but are also not marked as decorative, so they can be navigated to and then give silence.
- Sample 5.2, item 11. article date and volume icons are focusable without having any description
- Sample 5.3, item 20. date and volume icons are focusable but not described.

StephDriver:
@StephDriver  to find a list of all icons we actually use.

StephDriver:
Icons in use:

1. fa-align-justify
1. fa-angle-right
1. fa-arrow-circle-down
1. fa-arrow-circle-o-left
1. fa-arrow-circle-o-right
[...]

PR #4490, opened 2024-11-14
Front of House table A11y updates

### LINKED ISSUE #4055: Table A11y updates

StephDriver:
tables identified via a search for 'table' in the templates within the themes directory.

first commit includes a comment above each table stating what needed fixing, these comments are then removed for the final commit. 

For the OLH theme, there were icons that needed updating within the table (as per #4215) which convey information that is relevant to their position within the table, so I updated those too.

closes #4055

### DIFF
--- a/src/themes/OLH/templates/elements/section_block.html
+++ b/src/themes/OLH/templates/elements/section_block.html
@@ -1,50 +1,47 @@
 <h3>{% trans 'Sections' %}</h3>
+{# TableA11y superflous p tags, single row of headings, very ambiguous data -> single row and column headings and make icons screen-reader compatible #}
 <table>
     <thead>
         <tr>
-            <th>
-                <p>{% trans "Section or article type" %}</p>
-            </th>
-            <th>
-                <p>{% trans "Public Submissions" %}</p>
-            </th>
-            <th>
-                <p>{% trans "Peer Reviewed" %}</p>
-            </th>
-            <th>
-                <p>{% trans "Indexed" %}</p>
-            </th>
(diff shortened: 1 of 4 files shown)


joemull (approved):

StephDriver:
As per discussion this morning, having edited translated text, django-admin commands need to run.

StephDriver:
following discussions on themes, I have now checked for tables in templates included by Clean from outside the clean directory - there are none.

It is possible that there are some outside themes directory tables used by OLH and Material - but as per the prioritisation of the A11y work, it is Clean that is being updated, and OLH and material only in as much as they are easy changes alongside clean, so hunting out additional templates with tables for OLH and material is beyond scope of this issue.

StephDriver:
this work has been combined with icon work as the same lines were being edited, see 
- #4492

PR #4491, opened 2024-11-15
Unified style and accessibility upgrades for login, registration, and profile pages

### LINKED ISSUE #4380: Bring authentication and account pages into back office
## Stories

As a designer and developer, I would like to improve the interfaces around authentication and account management without the overhead of updating and maintaining three separate sets of templates that each use different CSS frameworks and methodologies. The current setup is especially expensive if you think about the fact that we usually use frontend theme CSS for presentational and content-focused things, but the account pages are interactive, so we need to create interactive components and pages that handle forms including patterns like selection, confirmation, deletion, etc. An example is the work I am doing for #3168.

As any Janeway user, I would like to navigate things like registering, resetting my password, and editing my profile without being confused about where I am, or about how to get back to where I was before. Basically, I shouldn't have to deal with repeated switching between theme layers, and at the end I should be redirected to where I was, or a clear button that lets me get back to the homepage (if I was on the public-facing site) or the manager or dashboard (if I was in the back office).

As a theme designer for Janeway journal sites and articles, I do not want to have to redesign interaction-heavy pages to get a consistent look and feel. It would be nice if the themes were lighter and easier to cover with fewer templates and CSS components, and I could delegate the interactive pages to a single back office theme.

## Proposed solution
- [x] Create or update back-office templates for each of these views:
      - user_login
      - ~user_login_orcid~ - not needed as no template used
      - ~user_logout~ - not needed as no template used
      - get_reset_token
      - reset_password
      - register
      - orcid_registration
      - activate_account
      - edit_profile
- [x] Deprecate the corresponding templates in the theme template directories
- [x] Write view tests to check that each of these views finds the correct back-office template
- [x] Manually test the new interfaces, perhaps with some user testing by people who have never used Janeway

## Other context
In addition to user testing after the change, I think it would be beneficial to do some rapid UX research beforehand on this one. Methods that might work well:

- [First click test](https://www.optimalworkshop.com/101guides/first-click-testing-101), to see if casual users know how to get between the front and back offices
- User journey mapping, to pin down the exact routes users take that bring them through the auth and account pages

joemull:
Closes #4380.

## Implementation notes
* Brought all pages where a user manages their own account into the back office, and deprecated the theme templates
* Adjusted the admin base as needed to handle unauthenticated users
* Implemented a contextual title element so that the relevant site name is appropriately provided in addition to the page title
* Wrote view tests to make sure new templates are hit
* Used "secondary" style buttons for accessibility (color contrast)
* Adjusted link colors for accessibility (color contrast)
* Improved form accessibility in a few places
* Made heading structure consistent, though I would have preferred not to use h2 for these elements (h1 is taken by the header and used for the site title, which is not ideal in my opinion)

## Limitations
- [ ] We have got to sort out the heading structure in `core/base.html` -- it results in two h1s in most cases, and one of them is often empty. My preference is that h1 is used for the actual page, not the site name, and that the site name is simply appended to the title element, so it is available in the browser tab.
  - Update: Created https://github.com/openlibhums/janeway/issues/4503
- [ ] The interests field is a consistent accessibility trap, until we can get away from jquery / tagit. We may be able to re-implement this bit of the page as a [Django formset](https://docs.djangoproject.com/en/5.1/topics/forms/formsets/)

Screenshot_20241115_164725
Screenshot_20241115_164748
Screenshot_20241115_164758
Screenshot_20241115_164806
Screenshot_20241115_164832
Screenshot_20241115_164857
Screenshot_20241115_164957
Screenshot_20241115_165013
Screenshot_20241115_165050
Screenshot_20241115_165123
Screenshot_20241115_165141
Screenshot_20241115_165210
Screenshot_20241115_165224
Screenshot_20241115_165430
Screenshot_20241115_165509
Screenshot_20241115_165826


StephDriver (changes requested):
Joe, I think something's been missed, this is how I manually tested it:

- log in (works so far, and was taken to new back office log in page)
- select 'Profile' from drop down menu on top right
- was taken to a front of house template, which was themed (testing with ?theme=clean allowed me to change the theme).

either I've misunderstood the scope of this issue, or something's gone wrong. I was expecting the profile page to be back office.

src/templates/admin/elements/forms/accessible_messages.html
```diff
@@ -0,0 +1,19 @@
+{% comment %}
+  Note that messages only display through this element if they are not acti
+{% endcomment %}
+{% for message in messages %}
```
ajrbyers, line 4: The comment above is incomplete. I don't think we should duplicate messages here. I know some of the account views use messages—convert those to non_field_errors so we don't have to duplicate message output.

ajrbyers (changes requested):
Thanks for this, its good to reduce the number of overall templates.

I've added some comments inline. Some more general comments:

- If you enter your username and password incorrectly when the page reloads the form is larger, this feels very strange and should be a consistent experience for the user.
- Can you review all of the comments throughout this PR and ensure they're appropriate. See for example `settings.css` I don't think the comment here is useful.
- The forms feel a bit lost in the middle of the screen, especially on wider screens.
- These forms introduce a layout system of their own, is there a reason we're not utilising the established layout for the back office here?

src/templates/admin/core/accounts/edit_profile.html
```diff
@@ -0,0 +1,140 @@
...
+        <h3>{% trans 'Profile Details' %}</h3>
+        <form method="POST" enctype="multipart/form-data">
+          {% csrf_token %}
+          {% include "admin/elements/accounts/user_form.html" %}
+          <div>
+            <button
+              type="submit"
+              name="edit_profile"
+              class="button secondary">
+              <span class="fa fa-save"></span>
```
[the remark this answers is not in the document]
joemull: Alright, I'll remove it.

src/templates/admin/elements/forms/accessible_messages.html
joemull: > The comment above is incomplete.

Thanks--I'll remove it.

joemull: Appreciate this thinking on not duplicating messages. Some of these views operate without Django forms, so it would not be possible to rely on non-field errors for everything. I've opted to keep the messages in the body of the page and disable the toastr messages just on these pages, by putting the toastr script in a block that can be overwritten with nothing.

joemull:
>     * If you enter your username and password incorrectly when the page reloads the form is larger, this feels very strange and should be a consistent experience for the user.
>     * The forms feel a bit lost in the middle of the screen, especially on wider screens.

Thanks for these comments and the chat we had, which helped clarify a few things for me. I've opted to left-align everything to be more consistent with other back-office pages, and I've returned to using the title-section element. This puts the semantic name of each of these pages in an h1, which is good (was h2 in the first version), but it means we need to resolve the h1 in the header before this will be accessible, as discussed.

>     * These forms introduce a layout system of their own, is there a reason we're not utilising the established layout for the back office here?

I've tried to use what I can, but there are some float-based components that were not feasible to re-use. I've opted to marginally expand the utilities so that we can develop a framework-agnostic set of layout tools with modern patterns like grid and flex, while we shift frameworks. Let's discuss this more because I have a few things coming down the pike, and I'd like to expand the utilities in the back office more formally.

StephDriver (changes requested):
- ✅  log in takes me to the new back office page.
- ✅ clicking 'forgotten your password' on new back office log in page takes me to back office
- :x: clicking 'register' on new back office log in page takes me to front of house
- :x: register from navigation bar takes me to front of house
- :x: once logged in, the profile link in account menu from nav bar takes me to front of house

as per screen share discussion, this could be a domain v path issue.

ajrbyers:
@StephDriver let’s find some time to dig around this next week.

StephDriver:
> I still can't replicate this issue @StephDriver. Are we sure your browser isn't caching the old pages?
> 
> @mauromsl or @ajrbyers, could you tell me what about Steph's dev setup might cause this issue? I can't think of a scenario in which this view
> 
> https://github.com/openlibhums/janeway/blob/042b33e65bec9c89512eaa3818d2843a4fc26f31/src/core/views.py#L382
> 
> would return a template under `/OLH/` or `/clean/` or something else, even if there were a domain / path difference. It should return a 404 if it really can't find the path, not fall back to a path explicitly excluded by the `admin` bit, right?

I don't think so... mostly because I tried it with multiple different browsers and had same results.  But I agree it needs further digging... so watch this space.

StephDriver:
Trying this again:
✅ log in takes me to the new back office page.
✅ failed / rejected log in remains in back office.
✅ clicking 'forgotten your password' on new back office log in page takes me to back office
✅  clicking 'register' on new back office log in page takes me to front of house
✅ register from navigation bar takes me to front of house
✅ once logged in, the profile link in account menu from nav bar takes me to front of house

StephDriver (approved):
see comment, all tests fine now.

StephDriver (approved):

src/templates/admin/core/accounts/edit_profile.html
```diff
@@ -0,0 +1,159 @@
...
+          <div class="title-area">
+            <h2>{% trans "Register for Article Notifications" %}</h2>
+          </div>
+          <form method="POST">
+            <p>{% blocktrans %}
+              Use the button below to register to receive notifications of new articles
+              published in this journal.
+            {% endblocktrans %}</p>
+            {% csrf_token %}
+            {% if reader %}
```
ajrbyers, line 66: Spotted during testing that this is never set so the unsubscribe link is never displayed. Presumably its supposed to be in the context like:

```python
...
"reader": models.Role.objects.get(slug='reader'),
...
```

ajrbyers (changes requested):
This is very impressive work @joemull, I really like this. I have spotted one issue though, check out the comments.

src/templates/admin/core/accounts/edit_profile.html
joemull: OK I've fixed that.

PR #4499, opened 2024-11-20
feat: add JATS default article type to sections and articles

MartinPaulEve:
This PR adds a field to sections and articles that denotes the JATS article-type.

At submission, the article's JATS article type is set to the same as the section to which it is being submitted. Editors can, then, later change the JATS article type if they so wish, under the "Edit Metadata" pane.

The JATS stub generator then uses this field to generate a JATS "article" element with attribute "article-type" set to the value of the field. If no value is set, then "research-article" is used as the default.

Tests cover this fallback condition and so continue to pass.


src/submission/models.py
```diff
@@ -35,6 +35,7 @@
 from django.utils.functional import cached_property
 from django.utils.html import mark_safe
 import swapper
+from rest_framework.fields import ChoiceField
```
ajrbyers, line 38: This import is not used.

ajrbyers (changes requested):
This is looking good. A few comments:

- Can you check that lengths added are PEP8 compliant? Some look like they might be too long
- I have a concern that if the article section is changed this could leave the article's type out of sync.

src/submission/models.py
MartinPaulEve: - [X] Removed unused import
- [X] Checked PEP8 (I think my new lines were fine, but this will also be handled by a formatter soonish?)
- [X] I have added documentation to the docs files and the model's help_text fields explaining how the synchronisation does/doesn't work

MartinPaulEve:
Please hold on this: there's some debate as to which standard we should be using for the controlled vocabulary of this field

src/submission/models.py
```diff
@@ -640,6 +643,17 @@ class Article(AbstractLastModifiedModel):
     language = models.CharField(max_length=200, blank=True, null=True, choices=LANGUAGE_CHOICES,
                                 help_text=_('The primary language of the article'))
     section = models.ForeignKey('Section', blank=True, null=True, on_delete=models.SET_NULL)
+    jats_article_type_override = DynamicChoiceField(max_length=255,
+                                                    dynamic_choices=get_jats_article_types(),
+                                                    choices=tuple(),
+                                                    blank=True, null=True,
+                                                    help_text="The type of article as per the JATS standard. The initial state of this field is set by the submission section's article type.",
```
mauromsl, line 650: This help text needs adjusting now, the initial state is no longer set by the article type.

"An override of the article type when encoding into JATS. If the article has a section, it will use it as its article type"

src/core/forms/forms.py
```diff
@@ -480,7 +480,7 @@ class Meta:
         fields = [
             'name', 'plural', 'number_of_reviewers',
             'is_filterable', 'sequence', 'section_editors',
-            'editors', 'public_submissions', 'indexing',
+            'editors', 'jats_article_type', 'public_submissions', 'indexing',
```
mauromsl, line 483: should this be `jats_article_type_override` now?

src/core/janeway_global_settings.py
```diff
@@ -679,3 +679,37 @@ def __len__(self):
...
+    ("product-review", ("Product Review")),
+    ("rapid-communication", ("Rapid Communication")),
+    ("reply", ("Reply")),
+    ("reprint", ("Reprint")),
+    ("research-article", ("Research Article")),
+    ("retraction", ("Retraction")),
+    ("review-article", ("Review Article")),
+    ("translation", ("Translation")),
+    ("other", "Other")
+)
```
mauromsl, line 715: Is it worth flagging these for translation as you had previously done?

mauromsl (changes requested):
Nice, a few minor comments and we are good to go.

src/submission/models.py
MartinPaulEve: Done

src/core/forms/forms.py
MartinPaulEve: No: this is for the section, which is not an override

src/core/janeway_global_settings.py
MartinPaulEve: Unsure whether we want to load translation in the global settings file? Also: these are fixed in English for the jats type... Unsure we should allow translation as the user needs to see what is entered in the field.

mauromsl: doh

mauromsl: sounds good

Issue #3351, opened 2023-01-17
Enable submitting authors and editors/moderators to request ORCID iDs from co-authors


alainna:
**Is your feature request related to a problem? Please describe.**
As a Moderators/Editors, and as a Submitting Author, I should be able to send an email from the item view page to ask that the co-author(s) connect their ORCID iD to Janeway via authentication. No one should not be required to manually input an ORCID iD. 

Bonus: Ideally, no registration/activation should be necessary. 

**Describe the solution you'd like**
1. If there is no ORCID iD, or if there’s no valid access token, and “requesting ORCID authentication” is enabled in the settings, then the submitting author or moderator/editor should sees “Ask to verify ORCID iD” (or something similar) in author list: 
![image](https://user-images.githubusercontent.com/15632391/212969633-da6e3495-66a0-4384-ae7e-17a79ab898cc.png)

2. Clicking “Ask to verify” opens pop-up window to send an email to the co-author. The iD request email should be a modifiable template. 
![image](https://user-images.githubusercontent.com/15632391/212969734-54472ac9-cf25-4512-9560-230910d7bcfd.png)

3. The co-author shouldn’t be required to log in – use similar workflow to the “Invite Comment from Reviewer”

joemull:
Hi @alainna, I've been following the ORCID feature idea trail for a while, and I'm a fan!

I wanted to get your thoughts on how this might relate to two features that we are planning to build for release 1.6:
* https://github.com/BirkbeckCTP/janeway/issues/145
* https://github.com/BirkbeckCTP/janeway/issues/2655

Would 145 and 2655 be a basis on which this idea would build? Would there be any conflicts? Any adjustments you'd make?

I like the idea that authors wouldn't have to log in to Janeway, just authenticate their ORCID in Janeway. That would allow them to bypass the janeway account activation / password reset loop. We don't have that in 145 and 2655 yet.

alainna:
@joemull flagging in @hardyoyo too-- 

> https://github.com/BirkbeckCTP/janeway/issues/145

This is great and a good point to confirm ORCID iDs! It sounds like something that this feature could build off of -- but I'll let Hardy confirm 

> https://github.com/BirkbeckCTP/janeway/issues/2655

[...]

PR #4512, opened 2024-11-26
feat: Added support for the CRediT Taxonomy for contributors

### LINKED ISSUE #3037: Add support for CRediT (Contributor Roles Taxonomy) in Janeway
As a submitting author, I would like to fully represent the contributions of all colleagues who contributed to the creation of our work which we are submitting to a Janeway journal or preprint server. 

Enable submitting authors or editors to assign each author a contributor role based on [CRediT](https://credit.niso.org/). 

Ideally these roles would also be asserted in metadata about the article/preprint where possible. 

It also should be a toggle-able feature that is optional for each journal/server.

It could also be nifty to combine as part of https://github.com/BirkbeckCTP/janeway/issues/3351 The co-author could authenticate their ORCID iD and affirm their CrediT role and affiliation. 🤔

MartinPaulEve:
This PR adds the CRediT taxonomy to Janeway. This is a prerequisite for the OA Switchboard work.

The PR:

* Adds a data model for CRediT and some logic and associated methods for display
* Adds CRediT display to article pages (and doesn't display if no CRediTs are specified) and the JATS stub
* Adds a setting for turning CRediT on and off per journal

Not in scope, still to do as part of Janeway 1.8:

- [ ] Add a CRediT interface to the submission process where users can add taxonomy elements and delete them from any author
- [ ] Add CRediT display and modification to the editor "Edit Metadata" facility

closes #3037


joemull (changes requested):
So glad to see this work being done! Thanks again @MartinPaulEve. 

I am looking forward to testing out the interface. I wasn't able to do so yet because I ran into a migration error:

`CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0083_creditrecord_preprint_author, 0085_alter_article_jats_article_type_override_and_more in submission).`

What we like to do nowadays with git conflicts or split migration graphs is rebase dev branches on master and change the dependency of the dev migration. This allows us to avoid creating duplicate merge migrations in multiple people's dev instances. Could you do that please before I go further with my review?

joemull (changes requested):
Thanks again for this Martin.

I wrote some changes inline that I'd like to see before merging. Some of them are important to take care of regardless of how the UI works, while some of them could be informed by the conversation we have around https://github.com/openlibhums/janeway/issues/4519 . More details inline.

Two global comments:

- Shouldn't there be an interface for adding and removing CREDIT roles in the repository submission system, if there is a field for repository author on the role object? Or did I miss that integration?
- Shouldn't the canonical URI to each CRediT role be captured to make it more machine-readable in your plugin and other places like our own JATS? Taking a quick look at the [implementation guidelines](https://jats4r.niso.org/credit-taxonomy/), it seems like the URLs would be useful for name authority. I think this is especially the case since the strings used contain questionable characters for downstream consumption, like the ampersand in ["Writing - Review & Editing"](https://credit.niso.org/contributor-roles/writing-review-editing/).

joemull:
FYI @MartinPaulEve: We now have something of a plan for the submission page (see https://github.com/openlibhums/janeway/issues/4519#issuecomment-2520960819).

So, we'd like to ask if you would remove the back-office UI elements from this PR. I think that's everything in `templates/admin` and `submission/views.py`.

src/core/models.py
```diff
@@ -565,6 +566,13 @@ def snapshot_self(self, article, force_update=True):
                 setattr(frozen_author, k, v)
             frozen_author.save()
 
+        # now update the CRediT records
+        credit_records = CreditRecord.objects.filter(article=article, author=self)
+
+        for credit_record in credit_records:
+            credit_record.frozen_author = frozen_author
+            credit_record.save()
+
```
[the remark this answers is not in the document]
MartinPaulEve: OK, so I wasn't sure what force_update meant in "if frozen_author and force_update", and assumed it was a unique case, but that we would always want to save the credit_records when we freeze an article. Is that not right?

src/templates/admin/submission/submit_authors.html
```diff
@@ -77,6 +78,30 @@ <h2>{% trans 'Current Authors' %}</h2>
                                 <td><i class="fa fa-sort"></i></td>
                                 <td>{{ order.author.full_name }}</td>
                                 <td>{{ order.author.email }}</td>
+                                <td>
+                                    {% for author, role in article.credit_roles.items %}
+                                        {% if author.pk == order.author.pk %}
+                                            {% for credit in role %}<form style="display:inline-block" method="POST" id="remove_credit-{{order.author.pk}}-{{credit.pk}}" name="remove_credit-{{order.author.pk}}-{{credit.pk}}">{% csrf_token %}<input type="hidden" name="role" id="role" value="{{credit.pk}}"/><input type="hidden" name="remove_credit" id="remove_credit" value="remove_credit"><a href="#" onclick="document.getElementById('remove_credit-{{order.author.pk}}-{{credit.pk}}').submit();">{{ credit|title }}</a></form>{% if not forloop.last %}, {% endif %}{% endfor %}
```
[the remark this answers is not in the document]
MartinPaulEve: 1. Sure, but below you say you want this change nixed. Before I do anything on this, please conform if you actually want this changed or deleted.
2. My interface is just an example really as I wasn't give a spec and you seem to be redesigning the entire interface underneath this work.

src/core/models.py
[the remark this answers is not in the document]
joemull: Yes, I agree it's a good idea to update credit records in this spot, it just needs to be inside the first `if` clause / indented four spaces.

WRT to `force_update`, it is defined on `snapshot_authors` if you search for it (granted, it should be defined on `snapshot_self`). It is "Whether or not to update existing records". So this `if` block is supposed to run when there are existing frozen authors and when they need to be updated, which is a lot of the time, like in the pre-publication stage, IIRC.

With the current indentation, you are changing the code below from `if / else` to [`for / else`](https://book.pythontips.com/en/latest/for_-_else.html#else-clause), which is not what we want, because the `else` will always run. We only want lines 577 etc. to run if the conditions on line 564 are not met.

src/submission/views.py
```diff
@@ -823,6 +839,22 @@ def edit_metadata(request, article_id):
...
+                CreditRecord.objects.get(pk=request.POST.get('role')).delete()
+
+            if request.POST and 'add_credit' in request.POST:
+                # extract author ID
+                author_id = request.POST.get('author_id', None)
+
+                if author_id:
+                    role = request.POST.get(f'credit-{author_id}', None)
+
+                    if role != "None":
```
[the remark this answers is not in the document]
joemull: Please do delete it. We had to change direction yesterday afternoon because of the user needs around this interface.

src/templates/admin/submission/edit/metadata.html
```diff
@@ -165,6 +166,30 @@ <h2 id="edit-metadata-authors">Authors</h2>
...
+                                                    {% endfor %}
+                                                </select>
+                                            </div>
+                                            <div class="large-4 columns" style="vertical-align:middle">
+                                                <input type="hidden" name="author_id" id="author_id" value="{{ f_author.pk }}">
+                                                <input type="submit" name="add_credit" id="add_credit" value="Add" class="small success button"/>
+                                            </div>
+                                        </div>
+                                </form>
+                            </td>
```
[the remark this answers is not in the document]
joemull: Yes, please do delete.

src/templates/admin/submission/submit_authors.html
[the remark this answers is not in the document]
joemull: Yes, please do delete.

joemull:
Adding a note here to record my conversation with @MartinPaulEve about URIs and name authority: Martin is going to implement the URIs that I mentioned in my comment above, perhaps as keys in the choices, perhaps as part of a separate model.

The URIs are:

```
Term	URI
Conceptualization	https://credit.niso.org/contributor-roles/conceptualization/
Data curation	https://credit.niso.org/contributor-roles/data-curation/
Formal analysis	https://credit.niso.org/contributor-roles/formal-analysis/
Funding acquisition	https://credit.niso.org/contributor-roles/funding-acquisition/
Investigation	https://credit.niso.org/contributor-roles/investigation/
Methodology	https://credit.niso.org/contributor-roles/methodology/
Project administration	https://credit.niso.org/contributor-roles/project-administration/
Resources	https://credit.niso.org/contributor-roles/resources/
Software	https://credit.niso.org/contributor-roles/software/
Supervision	https://credit.niso.org/contributor-roles/supervision/
Validation	https://credit.niso.org/contributor-roles/validation/
Visualization	https://credit.niso.org/contributor-roles/visualization/
Writing – original draft	https://credit.niso.org/contributor-roles/writing-original-draft/
Writing – review & editing	https://credit.niso.org/contributor-roles/writing-review-editing/
```

More details in the [ANSI/NISO spec document](https://www.niso.org/publications/z39104-2022-credit).

joemull:
@MartinPaulEve as mentioned off GH, I'm also linking our [original feature request for CRediT roles](https://github.com/openlibhums/janeway/issues/3037). The requester asked for these things in their wish-list:

- [X] 1 Enable submitting authors or editors to assign each author a contributor role based on [CRediT](https://credit.niso.org/).

- [ ] 2 Ideally these roles would also be asserted in metadata about the article/preprint where possible.

- [ ] 3 It also should be a toggle-able feature that is optional for each journal/server.

- [ ] 4 It could also be nifty to combine as part of https://github.com/openlibhums/janeway/issues/3351 The co-author could authenticate their ORCID iD and affirm their CrediT role and affiliation. 🤔

@StephDriver and @mauromsl and I had a brief chat about this and were hoping you could confirm for us whether these remaining things are in scope? 

My sense is that 2 is done for reader display on Janeway, but it's not yet in the JATS. Is that right? Were you planning to put it in the JATS or not?

For 3, I think it should be in scope, because CRediT roles might not work for all journals and disciplines, and editors will want the option to avoid collecting metadata that is not useful to them, since it would be inconsistent if some articles had this information and others didn't. I think it should be on by default, but controllable at the journal level. A boolean-type `core.models.Setting` should do the trick, and you can add one in `src/utils/install/journal_defaults.json`.

My sense is that 4 is not in scope because it has to do with a different issue and involves UI design. Just let me know what you think.

MartinPaulEve:
OK, so I've added CRediT to the JATS, including the work on canonical URLs (which Mauro suggested to store in the choices list, rather than in the database).

4 is definitely out of scope for now, but 3 is an interesting question.

My *suggestion* for this is that the new interface that you build, when re-doing submission, should simply hide the ability to add CRediT information if it's disabled, because all of the display stuff I've done will NOT show anything if CRediT information is not provided.

joemull (changes requested):
Hi @MartinPaulEve, thanks for this progress.

I still think we need to think more about how the `snapshot_self` function works with CRediT roles. What about this scenario: An author submits an article on which they are drafting author, and then, while it's in production, they submit another one to the same journal on which they are the editing author. How does Janeway not erroneously pick up the editing author role from article 2 when it's updating the frozen authors for article 1, say in typesetting or prepub? This is a harder problem to solve, because I think it might involve creating frozen authors when author metadata is first entered during submission, and not updating them when we snapshot the author at a later date, just allow editing the frozen author. That's perhaps beyond the scope of your work, but I want to raise it here so @mauromsl and @ajrbyers are aware of it, and we'll have to address it before we release 1.8.

> the new interface ... should simply hide the ability to add CRediT information if it's disabled

Yeah that's what I was thinking, but how does an editor or press manager disable it? Have I missed a setting you have introduced? Any thoughts on my suggestion above: "A boolean-type `core.models.Setting` should do the trick, and you can add one in `src/utils/install/journal_defaults.json`." The setting should be introduced in core Janeway before you try to make your OA switchboard plugin hook into it, because otherwise you won't know how to check for whether CRediT roles are expected or not, for a given journal's metadata. Ideally, this setting should control whether this metadata is displayed on the article page, *not* whether CRediT records are present. Editors will expect a single toggle on / off to control it everywhere.

> so I've added CRediT to the JATS, including the work on canonical URLs (which Mauro suggested to store in the choices list, rather than in the database)

Super, though I'm wondering why we have spaces and ampersands in the DB keys? Any reason why we shouldn't use the less error-prone short slug form of each role, e.g. `formal-analysis`? We can also change this after this PR is merged if you don't have time to keep making changes like this.

Again, really appreciate this contribution. As I have upcoming annual leave I am going to request changes and also request Andy's and Mauro's review so you have some other perspectives sooner rather than later.

ajrbyers:
> I still think we need to think more about how the snapshot_self function works with CRediT roles. What about this scenario: An author submits an article on which they are drafting author, and then, while it's in production, they submit another one to the same journal on which they are the editing author. How does Janeway not erroneously pick up the editing author role from article 2 when it's updating the frozen authors for article 1, say in typesetting or prepub? This is a harder problem to solve, because I think it might involve creating frozen authors when author metadata is first entered during submission, and not updating them when we snapshot the author at a later date, just allow editing the frozen author. That's perhaps beyond the scope of your work, but I want to raise it here so @mauromsl and @ajrbyers are aware of it, and we'll have to address it before we release 1.8.

I've checked this over and can't see how this scenario would play out as `snapshot_self` filters `CreditRecord` on the current article.

I'd agree with the other two suggestions if there is time to make them @MartinPaulEve.

MartinPaulEve:
Yeah, sorry Joe, but I'm afraid I don't understand the comment about article mix-up; it's keyed to an article and updates the FrozenAuthor at the time they are created.

I have added a toggle option under "general" for turning CRediT on and off.

The last one is a challenge. The field is just a CharField limited to the choices of CRediT as defined by the spec. If you want a separate "key" for each then you'll need to add a "value" to each that has the display -- but you might just as well use the name as the lookup/what is stored rather than add additional text to the database that we're not going to use. If you want to change this, I'll leave it to you if that's OK.

src/submission/models.py
```diff
@@ -76,6 +81,67 @@ def article_media_upload(instance, filename):
...
+        "Writing - Original Draft",
+        "https://credit.niso.org/contributor-roles/writing-original-draft/",
+    ),
+    "Writing - Review & Editing": (
+        "Writing - Review & Editing",
+        "https://credit.niso.org/contributor-roles/writing-review-editing/",
+    ),
+}
+
+CREDIT_ROLE_CHOICES = [(key, values[0]) for key, values in CREDIT_ROLES.items()]
```
joemull, line 143: As mentioned in my last review, this still seems unnecessarily complicated to me. Couldn't we have a simpler choices set like this?

```py
CREDIT_ROLE_CHOICES = [
    ("conceptualization", "Conceptualization"),
    ("data-curation", "Data Curation"),
    ("formal-analysis", "Formal Analysis"),
    ("funding-acquisition", "Funding Acquisition"),
    ...
    ("writing-review-editing": "Writing - Review & Editing")
]
```

Then to get the URI form for downstream metadata, a method:

```py
class CreditRecord(AbstractLastModifiedModel):
    def uri(self):
        return f"https://credit.niso.org/contributor-roles/{self.role}/"
```

And for the string representation, we can get the choice label:

```py
    def __str__(self):
        return self.get_role_display()
```

joemull (changes requested):
@ajrbyers asked me to review this one more time to double-check the author leaking issue, and take a look at the new setting.

I think I was mistaken before about the account leaking its roles, as pointed out above, so that's no longer an issue.

But I would still like to see changes regarding a few other things mentioned before. I've put additional comments inline to make them extra clear.

src/submission/models.py
mauromsl: The reason to not tie together the URL to a given pattern based on key, was that we anticipate additional roles to be added that do not have an entry in credit.niso.org

In fact, in Martin's use case, there were already a few entries that were custom IIRC

As for using the niso slugs for the credit roles, that's fine. I think the only prior objection against using the actual URL as the database keys/identifiers.

joemull:
I've now made the discussed updates to the branch:

* Made sure the snapshot function could be used to remove credit records
* Made credit records exclusive to author, frozen author, or preprint author (the repository integration is TODO), by means of an exclusive field validator
* Cascaded deletion to the CRediT record from the abovementioned account-like records, as well as from the article, since preprint and article should not be assigned on the same CRediT record when we implement CRediT for preprints. We should call the exclusive fields validator on that pair of fields as well.
* consolidated the credit role choices
* updated the credit admin
* tested the new additions
* fixed two bugs

joemull (approved):

src/utils/install/journal_defaults.json
```diff
@@ -5132,5 +5132,24 @@
...
+        },
+        "setting": {
+            "description": "Whether or not to use CRediT for this journal",
+            "is_translatable": false,
+            "name": "use_credit",
+            "pretty_name": "Enable CRediT",
+            "type": "boolean"
+        },
+        "value": {
+            "default": "on"
```
ajrbyers, line 5148: Should this be on by default? It causes a state change that editors may not be aware of.

src/submission/admin.py
```diff
@@ -110,6 +113,32 @@ def _article(self, obj):
...
+        'author__email',
+        'frozen_author__first_name',
+        'frozen_author__last_name',
+        'frozen_author__frozen_email',
+        'preprint_author__account__first_name',
+        'preprint_author__account__last_name',
+        'preprint_author__account__email',
+    )
+
+    raw_id_fields = ('author', 'frozen_author', 'preprint_author')
```
ajrbyers, line 133: `article` should be included in `raw_id_fields`

ajrbyers (changes requested):
This is looking good, I've added some comments inline to look at. 

- I noticed that credit roles do not show on the `submit_review` template. It would be good to include them here so that the author gets an accurate summary of the submisson
- The roles also do not show on the article archive where we display author data, can they be added?
- The article pages show the credit roles in different ways due to how they display author information. However on all themes the the roles block shows even for authors with no listed credit roles. See screenshots for each theme below.
- There is no interface for an editior/staff to edit a credit role - I can't remember if this is something we decided would be worked on alongside the editor/staff interface for editing an authors ROR?


Screenshot 2025-02-12 at 21 37 38
Screenshot 2025-02-12 at 21 37 30
Screenshot 2025-02-12 at 21 37 36

src/utils/install/journal_defaults.json
joemull: Yeah, good point, and there are many journals for whom CRediT does not provide the right taxonomy.

I am hesitant to turn it off by default and do so silently, because that may mean lots of people don't use it who could benefit from it and would want to use it if they knew about it.

Here's a possible solution in two parts:
1) turn it off by default
2) create an interactive upgrade guide in Janeway that shows up in a prominent banner above the top of the manager, for press managers and journal managers when they log into a new version that has important decisions for them to make. They can dismiss it without opening it to accept defaults, or they can go through the items to make decisions, after which it will also go away.

joemull: If #2 takes too long to do for 1.8, we could create it and include some important settings retroactively that may have been missed, like this one. What do you think?

ajrbyers: Sounds like a great idea for surfacing changes after upgrades but yes potentially too much for 1.8. Can you write this up and mark it for discussion at the next BRF?

ajrbyers (changes requested):
Thanks for sorting the admin widget and setting issue. Can you address the comments in: https://github.com/openlibhums/janeway/pull/4512#pullrequestreview-2601947277 ?

Alternatively if you've got too much on with this and #4483 can this be handed off to me?

joemull:
> Thanks for sorting the admin widget and setting issue. Can you address the comments in: [#4512 (review)](https://github.com/openlibhums/janeway/pull/4512#pullrequestreview-2601947277) ?

Apologies @ajrbyers, I missed those comments with the length of this PR conversation. Will get on them.

joemull:
Alright, I've fixed the display bug with the existing template changes to the article page:

Screenshot from 2025-02-18 13-08-12
Screenshot from 2025-02-18 13-08-27
Screenshot from 2025-02-18 13-08-44

And as discussed offline I've moved the other UI features to a separate issue: https://github.com/openlibhums/janeway/issues/4622

ajrbyers (approved):

src/submission/models.py
```diff
@@ -2141,6 +2252,53 @@ def is_correspondence_author(self):
...
+        return f"https://credit.niso.org/contributor-roles/{self.role}/"
+
+    @staticmethod
+    def all_roles(self):
+        return CREDIT_ROLE_CHOICES
+
+    def save(self, *args, **kwargs):
+        exclusive_fields = ['author', 'frozen_author', 'preprint_author']
+        model_utils.validate_exclusive_fields(self, fields=exclusive_fields)
+        super().save(*args, **kwargs)
```
mauromsl, line 2299: When I suggested checking for mutual exclusion, I meant as a database constraint, as models and querysets have other ways to write to db. e.g:

```py
class Meta:
        constraints = [
            models.CheckConstraint(
                check=(
                    models.Q(author__isnull=True, frozen_author__isnull=True, preprint_author__isnull=False) |
                    models.Q(author__isnull=True, frozen_author__isnull=False, preprint_author__isnull=True) |
                    models.Q(author__isnull=False, frozen_author__isnull=True, preprint_author__isnull=True)
                ),
                name='single_author_relation_constraint'
            )
        ]
```

For extra fancy points, you could generalize this approach for your list of `exclusive_fields`

src/templates/common/encoding/article_jats_1_2.xml
```diff
@@ -45,6 +45,17 @@
...
+                        {% if credit_record.frozen_author == author %}
+                            <role
+                                vocab="credit"
+                                vocab-identifier="https://credit.niso.org/"
+                                vocab-term="{{ credit_record.get_role_display }}"
+                                vocab-term-identifier="{{ credit_record.uri }}">
+                                {{ credit_record.get_role_display }}
+                            </role>
+                        {% endif %}
+                    {% endfor %}
```
mauromsl, line 58: This seems cumbersome, as we will repeatedly iterate all credit records and check they match the right author. can't we instead call `author.credits` and iterate that subset?.

mauromsl (approved):
Thanks for picking this up Joe, I've added a few comments. The most important one for me is the data integrity around ensuring we set a database constraint to enforce mutual exclusion on the credit objects and its author foreign keys.

mauromsl (changes requested):
wrong radio button

joemull:
> Thanks for picking this up Joe, I've added a few comments. The most important one for me is the data integrity around ensuring we set a database constraint to enforce mutual exclusion on the credit objects and its author foreign keys.

Thanks Mauro, all good points. I have refactored and tested the methods in addition to creating that DB constraint. Let me know if there is still an issue with the methods.

joemull:
> wrong radio button

Which button?

mauromsl (approved):
Nice, thanks Joe!

mauromsl:
waiting on https://github.com/openlibhums/janeway/pull/4483

joemull:
#4483 is merged so I will rebase this now.

joemull:
Nevermind--waiting on https://github.com/openlibhums/janeway/pull/4663

PR #4515, opened 2024-11-28
Fixed an issue where users could be link to an empty Review Interest

### LINKED ISSUE #4504: Review Interest interface doesn't stop users from registering an empty string  as an interest
**Describe the bug**
On one of our production installations, we found a server error that was raised from multiple entries existing for the empty string.

The first problem was that an empty string can end up in the list. The second, is that there are no constraints in the database to guard against duplicate Interests from being inserted.

**Janeway version**
v1.7.2

mauromsl:
closes #4504


ajrbyers (approved):

StephDriver (approved):
tests fine, but does leave existing empty string and/or duplicate interests in the data, probably should tidy those up.

PR #4524, opened 2024-12-04
fix annotated files if typesetting plugin is not installed

Beethro:
Clicking on Document Management causes an error since the recent addition of annotated_files to the file_type template tag if the typesetting plugin is not installed.
Suggested fix.


joemull (approved):
Thank you for this @Beethro.

I replicated the error with a fresh installation without the plugin and confirmed the fix is working.

PR #4530, opened 2024-12-06
fix: change jats_type_override field to be Foundation styled

MartinPaulEve:
Adds Foundation styling to the JATS article type override field


mauromsl (approved):

src/templates/admin/submission/edit/metadata.html
```diff
@@ -94,7 +94,7 @@ <h2 id="edit-metadata">Edit Metadata</h2>
                         <div class="large-6 columns">
                             <div class="large-12 columns">
                                 <label>{{ info_form.jats_article_type_override.label }}</label>
-                                {{ info_form.jats_article_type_override }}
+                                {{ info_form.jats_article_type_override|foundation }}
```
mauromsl, line 97: The foundation filter will render the label, I don't think we need the explicit render anymore, as it might render twice

mauromsl (changes requested):

mauromsl (approved):

PR #4543, opened 2024-12-12
Merge of typesetting plugin to core

mauromsl:


ajrbyers:
There remains, I think, one todo for this PR:

- [ ] handle cases where Production is still in use in: src/journal/views.py

PR #4546, opened 2024-12-16
Center-align account management pages

### LINKED ISSUE #4545: Center-align login and account management pages
## Problem
When we originally created the new back-office account pages, they were center-aligned. The `title-section` block that gives the name of the current page was not used; instead a heading was put inside the white box, center-aligned with the rest of the component.

But for accessibility (avoiding an empty h1) and consistency with the rest of the back office UI (everything else is left-aligned), we temporarily moved this to left-aligned.

image

## Proposed solution
With further discussion we agreed this can be avoided by center-aligning the content of the `title-section` block. The h1 spans the full width of the page, which does not make for a very organized layout but is better than an empty h1, on the one hand, and a left-aligned login box, on the other.

joemull:
Closes #4545

### DIFF
--- a/src/static/admin/css/admin.css
+++ b/src/static/admin/css/admin.css
@@ -1,3 +1,7 @@
+:root {
+  --small-inline-margin: 0.625rem; /* as defined in app.css */
+}
+
 .small td {
     font-size: 0.8rem;
 }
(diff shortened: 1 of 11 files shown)


ajrbyers (approved):

StephDriver (approved):
tested, Login, Profile, Register all are now centre aligned.

PR #4575, opened 2025-01-09
Adds API endpoints for Janeway core and Plugin versions.

### LINKED ISSUE #4350: As a system administrator, I want to poll my installation versions using the REST API
## User story

When managing multiple installations of Janeway, I often want an up-to-date snapshot of which version each one is on.

## Proposed solution

Add a Janeway version endpoint to the REST API.

ajrbyers:
Closes #4350

### DIFF
--- a/src/api/urls.py
+++ b/src/api/urls.py
@@ -15,6 +15,8 @@ router.register(r'preprints', views.PreprintViewSet, 'preprint')
 router.register(r'licences', views.LicenceViewSet, 'licence')
 router.register(r'keywords', views.KeywordsViewSet, 'keywords')
 router.register(r'accounts', views.AccountViewSet, 'accounts')
+router.register(r'versions', views.VersionViewSet, 'versions')
+router.register(r'plugins', views.PluginViewSet, 'plugins')
 
 # Wire up our API using automatic URL routing.
 # Additionally, we include login URLs for the browsable API.
--- a/src/api/serializers.py
+++ b/src/api/serializers.py
@@ -4,6 +4,7 @@ from core import models as core_models
 from journal import models as journal_models
 from submission import models as submission_models
 from repository import models as repository_models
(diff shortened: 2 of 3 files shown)


src/api/urls.py
```diff
@@ -15,6 +15,8 @@
 router.register(r'licences', views.LicenceViewSet, 'licence')
 router.register(r'keywords', views.KeywordsViewSet, 'keywords')
 router.register(r'accounts', views.AccountViewSet, 'accounts')
+router.register(r'versions', views.VersionViewSet, 'versions')
+router.register(r'plugins', views.PluginViewSet, 'plugins')
```
ajrbyers, line 19: I left these as single quote for consistency.

joemull (changes requested):
When I proposed this feature I should have added a bit more detail on the use case. It is often important to get the *exact* version an installation is on, when doing support, so you know how to help them. 

The way you've done it, by relying on the Janeway Version rather than the Git tag for this information, we are not always getting the exact version. There could be RCs, or there could be a version tagged but not propagated to the Version object, like the current 1.7.3 release (the Version object shows as 1.7.2). Is there a safe way to get and display the Git tag?

Also, I had actually imagined this would be public, not behind an Editor permissions decorator, to make it easy to write a script. (Security implications?) As it is, I don't know how you'd write a script for this. Do we offer token authentication? It is not in our documentation, so can you document it? https://janeway.readthedocs.io/en/v1.7.2/dev/api.html

ajrbyers:
No worries. I think there is a git module we can use and we can inject the tag into the JSON. I still think this endpoint is useful. I think it should be behind authentication. I’ll add some info to docs.

ajrbyers:
- [ ] Expose the Janeway version from src/janeway/__init__.py
- [ ] Make this view public, and display more information if the view is authenticated

PR #4599, opened 2025-01-29
Fix a bug on the article page with a misleading downloads metric

### LINKED ISSUE #4299: Hide Downloads from Metrics when there are no Downloadable galleys?
Question arising from working on #3911 which involved tidying up the empty Download lists when there were no galleys avaliable for download. 

When an article has no downloadable Galleys, should the download metric be hidden from view?

Reasoning - having a metric with a zero just looks bad. And depending on screenwidth, it may not be onscreen at the same time as the 'there are no downloads' statement that now appears when there are none available.

Consideration - when an article used to be downloadable, but now isn't, i.e. has a non-zero metric for downloads, then it probably should continue to show that metric.

Proposed solution: have logic `if galleys` and `if article.metrics.downloads` i.e. downloads avaliable and non-zero for past downloads.

joemull:
Fixes #4299.

### DIFF
--- a/src/themes/OLH/templates/journal/article.html
+++ b/src/themes/OLH/templates/journal/article.html
@@ -233,26 +233,29 @@
                                 {% if article.is_published or proofing %}
                                     {% if not request.journal.disable_metrics_display %}
                                         <div class="top">
-                                            <div class="row">
-                                                <div class="large-{% if article.citation_count and not journal_settings.article.suppress_citations_metric %}4{% else %}6{% endif %} columns">
+                                            <div class="flex">
+                                                <div class="columns">
                                                     <p class="number">
                                                         {{ article.metrics.views }}
                                                         <span>{% trans "Views" %}</span>
                                                     </p>
                                                 </div>
-                                                <div class="large-{% if article.citation_count and not journal_settings.article.suppress_citations_metric %}4{% else %}6{% endif %} columns">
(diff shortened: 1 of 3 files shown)


ajrbyers (approved):

StephDriver (commented):
@joemull 
What was the expected behaviour for when there are views, but no downloads?
because this is what I'm seeing:
image
 but I thought from the issue, that the downloads part of that would be hidden, so it would just show the views.

ajrbyers:
> @joemull What was the expected behaviour for when there are views, but no downloads? because this is what I'm seeing: image but I thought from the issue, that the downloads part of that would be hidden, so it would just show the views.

It should only hide the Downloads if there are no downloable galleys.

StephDriver:
> It should only hide the Downloads if there are no downloable galleys.

the reasoning given in the original issue is:

> Reasoning - having a metric with a zero just looks bad. 

So this does seem to apply to cases of nothing yet having been downloaded, so where there are views, but no downloads.  However that is a very niche case, as any article with downloads  avaliable and that has been viewed, is unlikely to remain on zero for long outside of a development environment!  Ok, instead of borrowing trouble, lets go with this and we can always come back later for the other if it does become a user issue.

StephDriver (approved):
(comments above, but go ahead!)

PR #4603, opened 2025-01-29
Don't look for an attribute on None object

gamboz:
Not particularly important, but I thought I'd report :slightly_smiling_face: 

In the decorator `article_stage_accepted_or_later_or_staff_required`, there is an `if` clause that might try to access an attribute of a None object.

To replicate, one should log-in to a journal and manually enter the URL of a non existing article in order to trigger the view "article_file_download" (`views.serve_article_file`).

E.g. https://olh.openlibhums.org/article/id/20000/file/1234/  (ATM article 20000 does not exist).

HTH


joemull:
@ajrbyers @mauromsl can we add this to the 1.8 issues please?

ajrbyers (approved):

joemull (approved):

PR #4608, opened 2025-01-31
Fix bug that kept filter from working on some search views

### LINKED ISSUE #4607: Rummage Portal field has wrong field type
## Problem
When I want to filter supporters by country, the filter is not applied.

This is because the widget provides multiple choices, but the form field is a single choice field:

https://github.com/openlibhums/janeway/blob/3cad5113161b7b0119706266c04d3bf08db6ebd4/src/core/forms/forms.py#L698-L702

As a result the data is misaligned when the form is validated.

## Proposed solution

Use `MultipleChoiceField`.

joemull:
Fixes #4607.

Note that this is hot-fixed on OLH as well, so that the Supporters Manager country field works.

The bug does not directly affect most Janeway users because only the Supporters plugin uses this field, to my knowledge.

### DIFF
--- a/src/core/forms/forms.py
+++ b/src/core/forms/forms.py
@@ -698,9 +698,9 @@ class CBVFacetForm(forms.Form):
                     count = values_list.count(value)
                     label_with_count = f'{label} ({count})'
                     choices.append((value, label_with_count))
-
                 choices = sorted(choices, key=lambda x: x[1])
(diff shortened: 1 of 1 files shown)


ajrbyers (approved):

mauromsl (approved):
Looks good Joe, can you rebase and merge?

PR #4611, opened 2025-02-05
Merges the typesetting plugin into Janeway core as a Django application

ajrbyers:
The aim of this PR is not to overhaul anything but simply to make the Typesetting plugin a core Janeway application.

A brief summary:

- Moves the Typesetting plugin into core and maintains its git history
- Maintains the `plugins/typesetting/` URL path but adds `typesetting/` too
- Renames the workflow element from `Typesetting Plugin` to `typesetting` for consistency
- Updates the default workflow to include Typesetting, removes production and proofing
- Plugin loading will ignore the Typesetting plugin
- Merges events and settings into core modules


ajrbyers:
I've unassigned @joemull and converted this to draft. There are some diffs showing up that shouldn't so will need to remove them.

PR #4614, opened 2025-02-11
Refactor get_ip_address to avoid 500 errors

yakky:
This refactoring aims to prevent 500 errors when the `HTTP_X_FORWARDED_FOR` and `REMOTE_ADDR` headers are manipulated by potentially malicious users and do not contain a valid IP address.

For example, the following request could cause such an issue:
`wget "http://<domain>/article/pubid/<pubid>/" -O /dev/null --header="X-Forwarded-For: some-random-text,1.2.3.4"`

## Models changes
The `LoginAttempt.ip_address` field needs to be updated to allow `None` values. This is necessary because if the `get_ip_address` method returns `None`, we need to be able to save `None` when there are failed login attempts with invalid or manipulated headers.

The same applies to the `Contact.client_ip` field, which is set in `journal.views.contact` using `new_contact.client_ip = shared.get_ip_address(request)`. Without this change, an error will occur if the IP address is None.

Although it's highly unlikely that ipware will ever return `None`, actually Its `get_client_ip` method can return `None` values.

## Code changes
We also need to handle TypeError in the `get_iso_country_code` method, as `geoip2.database.Reader().country` will fail if the IP address is `None`. It seems appropriate to modify this method to return `None`, as its output is used by `iso_to_country_object`, which already returns `None` if the corresponding Country object is not found among the registered Countries.

### DIFF
--- a/src/core/models.py
+++ b/src/core/models.py
@@ -1595,7 +1595,7 @@ class Contact(models.Model):
     sender = models.EmailField(max_length=200, verbose_name=_('Your contact email address'))
     subject = models.CharField(max_length=300, verbose_name=_('Subject'))
     body = JanewayBleachField(verbose_name=_('Your message'))
-    client_ip = models.GenericIPAddressField()
+    client_ip = models.GenericIPAddressField(blank=True, null=True)
     date_sent = models.DateField(auto_now_add=True)
 
     content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE, related_name='contact_c_t',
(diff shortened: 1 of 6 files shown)


yakky:
Original code by @lorenzogram

src/core/models.py
```diff
@@ -1793,7 +1793,7 @@ def __str__(self):
 
 
 class LoginAttempt(models.Model):
-    ip_address = models.GenericIPAddressField()
+    ip_address = models.GenericIPAddressField(blank=True, null=True)
```
mauromsl, line 1796: Is it even worth registering a login attempt if we do not have an IP address?
 
As this model is only used to count the number of failed attempts from a given IP, we might use the python logger as INFO for correlation but ignore the write to db

mauromsl (changes requested):
Thank you @yakky (and @LorenzoGrama) , good catch!

I just made a comment regarding `LoginAttempt.ip_address` being now nullable, which I think we do not need. thoughts?

src/core/models.py
mauromsl: yes please, and thanks

ajrbyers (approved):

mauromsl (approved):

Issue #4221, opened 2024-05-31
Contextual link fixes arising from a11y audit


StephDriver:
[edited 16 Dec to make clear the scope]

## Action List:

- fix items found by manual screen reader testing below, (some may have been fixed by icon work).
- ...and from that get idea of which kind of links are non-compliant (e.g. links within logic loops).  
- hunt for similar across front end (prioritising Clean theme) and fix/standardise

edit 10 June, moved the following out of scope:
- (DOI links updated as per format from Crossref below to be included in #4236 )
- (internal cross references within articles are out of scope for this issue, and will be addressed in https://github.com/openlibhums/janeway/issues/4245, #4084 and elsewhere. ) 
 
This is about **non-article** links only.

## Link Requirements

- Each link on a page should have unique text (for visual only links, this means the text for screen reader). 
    - particularly important for links created inside logic loops.
- the purpose of the link should be clear (not 'click here') This may require aria as an alternative to visual context (e.g. for 'read more').
- the user should be able to tell whether the link is going
   - another part of the page
   - another page on the site
   - external
- if the link involves an icon, look at the work done on #4215 
- In general DO NOT use the word 'link' in a link's text, as a screen reader will read out the word 'link' before any link, an there is a visual difference too, making this superfluous.

### WCAG

[WCAG 2.4.4](https://www.w3.org/TR/WCAG22/#link-purpose-in-context)
> The purpose of each link can be determined from the link text alone or from the link text together with its programmatically determined link context, except where the purpose of the link would be ambiguous to users in general.

### Crossref DOI links
see [Cross-ref](https://www.crossref.org/blog/accessibility-for-crossref-doi-links-call-for-comments-on-proposed-new-guidelines/) for info on DOI links

TLDR:
non-contextual link but with Aria markup so that it is read as if contextual by screen reader

![Image](https://github.com/BirkbeckCTP/janeway/assets/5330770/50b70fe3-312f-475c-98d6-7f249a10d1ec)

### Manual Screen Reader Testing Results

The screen reader testing (#4194) found the following links did not have appropriate context:

- DOI must show the full link, but should have aria markup so that it doesn't get read out as all the digits, see below.  addressed by #4236 
- Sample 2, item 69. License was done better in Sample 1, where there was a brief description of what the license meant.
[...]

PR #4619, opened 2025-02-12
Removes share links from the article page and replaces them with a share link and copy button

### LINKED ISSUE #4390: Change X/Twitter sharelink to Bluesky
It was [announced](https://www.openlibhums.org/news/750/) today that OLH is leaving Twitter / X.

Our sharelinks, as per issue #4353 were recently updated to replace the twitter icon with X, but in light of this announcement, we should consider replacing this with Bluesky.

ajrbyers:
- Drops all share links on all themes
- Adds a field with either the DOI or canonical URL
- Adds copy to clipboard common JS script
- Closes #4390 

Screenshot 2025-02-12 at 16 53 21
Screenshot 2025-02-12 at 16 53 26
Screenshot 2025-02-12 at 16 53 30

### DIFF
--- a/src/themes/OLH/templates/elements/journal/share.html
+++ b/src/themes/OLH/templates/elements/journal/share.html
@@ -1,0 +1,25 @@
+{% with article.get_doi_url|default:article.url as share_url %}
+  <div class="section">
+    <h3>{% trans "Share" %}</h3>
+    <div class="input-group">
+      <input
+              id="share-link"
+              class="input-group-field"
+              type="text"
+              aria-label="{% trans "Article's sharing URL" %}"
+              value="{{ share_url }}"
+              readonly
+      >
+      <div class="input-group-button">
+        <button
+                class="button"
+                type="button"
+                onclick="copyToClipboard()"
+                aria-label="{% trans 'Copy Share Link' %}"
+        >
+          <span class="fa fa-copy"></span>&nbsp;{% trans "Copy" %}
+        </button>
+      </div>
+    </div>
+  </div>
+{% endwith %}
--- a/src/themes/OLH/templates/elements/journal/share.html
+++ b/src/themes/OLH/templates/elements/journal/share.html
@@ -17,7 +17,7 @@
                 onclick="copyToClipboard()"
                 aria-label="{% trans 'Copy Share Link' %}"
         >
-          <span class="fa fa-copy"></span>&nbsp;{% trans "Copy" %}
+          <i aria-hidden="true" class="fa fa-copy"></i>&nbsp;{% trans "Copy" %}
         </button>
       </div>
     </div>
--- a/src/themes/OLH/templates/elements/journal/share.html
+++ b/src/themes/OLH/templates/elements/journal/share.html
@@ -2,11 +2,11 @@
   <div class="section">
     <h3>{% trans "Share" %}</h3>
     <div class="input-group">
+      <label for="share-link" class="sr-only">{% trans "Article URL" %}</label>
       <input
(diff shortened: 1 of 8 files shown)


src/themes/OLH/templates/elements/journal/share.html
```diff
@@ -0,0 +1,25 @@
...
+              readonly
+      >
+      <div class="input-group-button">
+        <button
+                class="button"
+                type="button"
+                onclick="copyToClipboard()"
+                aria-label="{% trans 'Copy Share Link' %}"
+        >
+          <span class="fa fa-copy"></span>&nbsp;{% trans "Copy" %}
```
StephDriver, line 20: Decorative icon - this is our standardised syntax.

<i aria-hidden="true" class="fa fa-copy"></i>

src/themes/clean/templates/elements/journal/share.html
```diff
@@ -0,0 +1,23 @@
...
+            readonly
+    >
+    <div class="input-group-append">
+      <button
+              class="btn btn-primary"
+              type="button"
+              onclick="copyToClipboard()"
+              aria-label="{% trans 'Copy Share Link' %}"
+      >
+        <span class="fa fa-copy"></span>&nbsp;{% trans "Copy" %}
```
StephDriver, line 19: update syntax for icon

src/themes/material/templates/elements/journal/share.html
```diff
@@ -0,0 +1,19 @@
...
+            aria-label="{% trans "Article's sharing URL" %}"
+            value="{{ share_url }}"
+            readonly
+    >
+    <button class="btn btn-small waves-effect waves-light teal darken-1 share-button"
+            type="button"
+            onclick="copyToClipboard()"
+            aria-label="{% trans 'Copy Share Link' %}"
+    >
+      <i class="fa fa-copy"></i>&nbsp;Copy
```
StephDriver, line 16: update syntax for icon

StephDriver (changes requested):
A few A11y tweaks needed

1. icon syntax (see comments above).
2. Share link text is a naked DOI, which means this text is read aloud by voiceover, each letter and number separately and only afterwards tells you it is the article share link.  i.e. the aria-label is read out after the text itself, not instead of it. 

see also #4221 for info on accessible DOI links (I know this isn't a link, but that's because its a special case to provide it as text to copy - I think it should still be read aloud in an understandable format).

ajrbyers:
> A few A11y tweaks needed
> 
>     1. icon syntax (see comments above).
> 
>     2. Share link text is a naked DOI, which means this text is read aloud by voiceover, each letter and number separately and only afterwards tells you it is the article share link.  i.e. the aria-label is read out after the text itself, not instead of it.
> 
> 
> see also #4221 for info on accessible DOI links (I know this isn't a link, but that's because its a special case to provide it as text to copy - I think it should still be read aloud in an understandable format).

1. Is now resolved.

ajrbyers:
> A few A11y tweaks needed
> 
>     1. icon syntax (see comments above).
> 
>     2. Share link text is a naked DOI, which means this text is read aloud by voiceover, each letter and number separately and only afterwards tells you it is the article share link.  i.e. the aria-label is read out after the text itself, not instead of it.
> 
> 
> see also #4221 for info on accessible DOI links (I know this isn't a link, but that's because its a special case to provide it as text to copy - I think it should still be read aloud in an understandable format).

2. is now also updated.

StephDriver (approved):
thanks - all three themes tested with voice over and sounding good.

PR #4626, opened 2025-02-17
Fix migration bugs in submission and core

### LINKED ISSUE #4625: Missing migration dependency
**Describe the bug**
`django.db.migrations.exceptions.NodeNotFoundError: Migration core.0100_contact_ip_address_null dependencies reference nonexistent parent node ('cms', '0099_alter_accountrole_options')`

**Janeway version**
be301351209e87c3deb871002ca9608b4604a818

joemull:
Fixes #4625.


ajrbyers (approved):

src/core/migrations/0102_merge_20250217_0954.py
```diff
@@ -0,0 +1,14 @@
...
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0100_contact_ip_address_null'),
+        ('core', '0101_delete_blank_keywords'),
+    ]
```
mauromsl, line 11: These migrations shouldn't conflict, I believe the solution is to address their dependencies

src/submission/migrations/0085_merge_20250217_0955.py
```diff
@@ -0,0 +1,14 @@
+# Generated by Django 4.2.18 on 2025-02-17 09:55
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('submission', '0082_article_abstract_es_article_title_es_section_name_es_and_more'),
+        ('submission', '0084_remove_article_jats_article_type_and_more'),
```
mauromsl, line 10: similar to the above

mauromsl (changes requested):

src/core/migrations/0102_merge_20250217_0954.py
joemull: Sure, happy to do that. My understanding has been that if a migration is already merged, don't change its dependencies. Sounds like that's not always the case?

src/core/migrations/0100_contact_ip_address_null.py
```diff
@@ -4,7 +4,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('cms', '0099_alter_accountrole_options'),
+        ('core', '0101_delete_blank_keywords'),
```
mauromsl, line 7: 100 should not depend on 101

mauromsl (changes requested):

src/core/migrations/0102_merge_20250217_0954.py
mauromsl: This conflict arises from changing the dependencies in this PR so it is worth setting them right. Not so sure how the other happened...This conflict arises from changing the dependencies in this PR so it is worth setting them right. Not so sure how the other happened...

PR #4656, opened 2025-03-14
Include securitytags wherever pii is used

bibliotechy:
Multiple review templates use the `se_can_see_pii` template tag, but did not load the securitytags in the template, resulting in errors. This PR fixes those issues.

### DIFF
--- a/src/templates/admin/review/decision.html
+++ b/src/templates/admin/review/decision.html
@@ -1,5 +1,6 @@
 {% extends "admin/core/base.html" %}
 {% load i18n %}
+{% load securitytags %}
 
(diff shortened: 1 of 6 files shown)


ajrbyers:
Security tags needs to be added as a default tag in settings.py rather than loaded like this.

bibliotechy:
@ajrbyers Will do!

ajrbyers:
> @ajrbyers Will do!

See: https://github.com/openlibhums/janeway/blob/35874511402fb248438921f592afbfb3b224e28a/src/core/janeway_global_settings.py#L173

if you override this in your settings file make sure this is added 👍

PR #4665, opened 2025-03-19
Add typesetting plugin stage to migrations

### LINKED ISSUE #4664: Migration needed for workflow stages introduced by typesetting
**Describe the bug**
Django is suggesting a migration for the the workflow stages because of the recent merge of typestting into core.

**Janeway version**
052cf7d71467756831a43b78f64d39c12b367fa9

joemull:
Fixes #4664


ajrbyers:
Should this be a DynamicChoiceField rather than adding migrations for it?

joemull:
Happy to redo it that way if you like?

mauromsl (approved):

mauromsl:
> Should this be a DynamicChoiceField rather than adding migrations for it?

It is already a DynamicChoiceField, but it is not using `dynamic_choices` for built-in stages, since those are always present

ajrbyers (approved):

PR #4671, opened 2025-03-20
Fix navigation between all users and journal users

### LINKED ISSUE #4577: Reversing press URLs on journal sites in path mode
## Problem description

When forming Janeway URLs in path mode, the press level of a site is often identified by the *absence* of a journal code, rather than the presence of a key or argument or path segment.

As a result, there is currently no way to dynamically build a named URL at the press level when working within the context of a journal page. This makes it difficult to create journal-level interfaces that refer to press interfaces and guide users between them.

Example:

https://github.com/openlibhums/janeway/blob/69823297b203c4edb3a4ab6b127ba81e32c65fe2/src/templates/admin/core/manager/users/list.html#L28-L33

You can do `request.press.site_url` or `journal.press.site_url` in most cases, but these are just context variables pointing to the press landing page. They do not take arguments and so cannot be used to reverse named urls.

The template tag `site_url` also does not work because, on path mode, it prepends the journal code. 

https://github.com/openlibhums/janeway/blob/69823297b203c4edb3a4ab6b127ba81e32c65fe2/src/core/templatetags/fqdn.py#L33-L44

This means the formed URL will always refer back to the journal. If you do `{% site_url "press_all_users" %}` that will just put out `example.org/JOURNAL/press/user/all`.

## Proposed solution

An additional template tag `press_url` that allows you to reverse any press url, regardless of current page / request context. It could be used like this:

```
If a broader search is needed,
          try <a href="{% press_url 'press_all_users' %}">All Users
          at the press level</a>.
```

## Alternative considered

An additional argument on the `site_url` template tag for a boolean called `press`. If `True`, the tag will use press as the `site_type` rather than the `site_type` of the request.

### LINKED ISSUE #4667: fix bug that prevents movement between press and journal levels

joemull:
Fixes #4667 - bug report.
Fixes #4577 - underlying problem.

Adds an additional menu for links *back* from the press to the journal level:

image

"Test A" in this screenshot is the name of the press.


joemull:
@Astronaut-resting Is it alright if I assign you for first review of this? It fixes the link from the journal level to the press level, and also adds a box at the press level with links back to the journal users pages.

Astronaut-resting (approved):
looks good!

joemull:
Ah, I had a thought about this one: we may need to ensure the list of journal links is sorted A-Z.

src/templates/admin/core/manager/users/list.html
```diff
@@ -27,12 +28,28 @@
...
+                  {{ journal.name }}
+                </a>
+              </li>
+            {% endfor %}
+          </ul>
         </aside>
       {% endif %}
+      {% comment %} The above URLs are hardcoded because there is no other
+      way to form a press URL dynamically on a journal page at the moment,
+      or to form a journal URL dynamically on a press page.{% endcomment %}
```
mauromsl, line 52: we do have one: see fqdn.external_journal_url.

src/press/tests/test_views.py
```diff
@@ -31,3 +38,50 @@ def test_press_nav_account_links_do_not_have_return(self):
...
+        self.client.force_login(self.press_manager)
+        response = self.client.get(
+            '/user/all/',
+            SERVER_NAME=self.press.domain,
+        )
+        content = response.content.decode()
+        self.assertIn(
+            f'{self.press.domain}/{self.journal_one.code}/user/all/',
+            content
+        )
```
mauromsl, line 87: While I appreciate the existance of tests, these types of tests we are introducing here look more like acceptance tests than unit tests. They are relatively slow and they are just testing that you've put the correct URL on the template, so I don't think we should include them as long as we use the correct template tag to generate urls and that the tag itself is properly unittested.

mauromsl (changes requested):

mauromsl (approved):

joemull:
Thanks for your updates Mauro. Before you review this Andy, I will fix a typo and also sort the journals A-Z.

src/press/models.py
```diff
@@ -185,6 +185,37 @@ def journals(**filters):
             return journal_models.Journal.objects.filter(**filters)
         return journal_models.Journal.objects.all()
 
+    @property
+    def journals_by_sequence(self):
+        from journal import models as journal_models
+        return journal_models.Journal.objects.all().order_by('sequence')
```
ajrbyers, line 191: Is this used somewhere else?

ajrbyers (changes requested):
Looks good - though added a comment about an potentially unused method.

src/press/models.py
joemull: No. I will remove it.

ajrbyers (approved):

PR #4682, opened 2025-03-26
Updates the offset_date tag to take the users preferred timezone into account

### LINKED ISSUE #4666: New Publication Date Buttons Set Time as GMT
Image

**Describe the bug**
The blue Publication Date buttons introduced in Janeway 1.7 (which appear when you select 'Set Publication Date' on the Pre-Publication Checklist page for an article are behaving differently than when a user selects the publication date using the calendar picker - and neither are taking into account the Preferred Timezone setting from the user's profile.

**Janeway version**
1.7.4

**To Reproduce**
In the Set Publication Date modal, 
- Select the blue 'Now' button (or the 1 day, 2 day, etc. buttons). The time and date will be set based on GMT
- Clear out all time and date information
- Select the Calendar widget and select today's date or (if your browser has it - some don't) select the Today button. Time and date are set based on the user's machine timezone settings.

**Expected behavior**
The GMT timezone may cause unexpected behavior when the Publish button is selected, and general confusion when the user sees an expected time.

Having each option consistently use the timezone of the user's machine seems better.

However it seems like the real timezone being used should prefer the user's 'Preferred Timezone' profile setting and, if there isn't one (which there isn't for the majority of users), fall back to the user's machine's timezone.

Note that the functionality calendar widget varies depending on the browser. Chrome includes the 'Today' option, which stamps the current time and date. Firefox and Safari lack this feature.

**Related Issue**
https://github.com/openlibhums/janeway/issues/2220

ajrbyers:
- This  closes #4666 however there are browser issues that we cannot fix (such as Chrome's built-in datetime-local widget providing its own Now button that uses the browsers timezone.

### DIFF
--- a/src/core/tests/test_templatetags.py
+++ b/src/core/tests/test_templatetags.py
@@ -1,4 +1,4 @@
-from datetime import timedelta
+from datetime import datetime, timedelta
 
 import pytz
 from django.utils import timezone
@@ -35,7 +35,14 @@ class TestFqdn(TestCase):
         self.assertEqual(url, f'http://{self.journal_two.press.domain}/{self.journal_two.code}/user/all/')
 
 
-@freeze_time("2025-03-26 12:00:00")
+ny_tz = pytz.timezone("America/New_York")
+frozen_dt = timezone.make_aware(
+    datetime(2025, 3, 26, 12, 0, 0),
+    timezone=ny_tz,
+)
+
+
+@freeze_time(frozen_dt)
 class TestOffsetDateTag(TestCase):
     @classmethod
     def setUpTestData(cls):
@@ -63,7 +70,7 @@ class TestOffsetDateTag(TestCase):
             days=2,
             input_type="datetime-local",
         )
-        expected = "2025-03-28T12:00"
+        expected = (timezone.now() + timedelta(days=2)).strftime("%Y-%m-%dT%H:%M")
         self.assertEqual(result, expected)
 
     def test_offset_date_with_preferred_timezone_datetime(self):
@@ -71,13 +78,15 @@ class TestOffsetDateTag(TestCase):
         Test offset_date tag with a valid preferred timezone and datetime-local input.
         """
         context = {"request": self.request_tz}
-        # 12:00 UTC is 12:00 GMT in March (not in daylight saving yet)
+        london = pytz.timezone("Europe/London")
+        expected_due = timezone.now().astimezone(london) + timedelta(days=1)
+        expected = expected_due.strftime("%Y-%m-%dT%H:%M")
+
         result = dates.offset_date(
             context,
             days=1,
             input_type="datetime-local",
         )
-        expected = "2025-03-27T12:00"
         self.assertEqual(result, expected)
 
     def test_offset_date_with_invalid_timezone_datetime(self):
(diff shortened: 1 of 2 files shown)


src/core/tests/test_templatetags.py
```diff
@@ -28,3 +33,76 @@ def test_stateless_site_url_for_journal_path(self):
...
+        )
+        expected = "2025-03-28T12:00"
+        self.assertEqual(result, expected)
+
+    def test_offset_date_with_preferred_timezone_datetime(self):
+        """
+        Test offset_date tag with a valid preferred timezone and datetime-local input.
+        """
+        context = {"request": self.request_tz}
+        # 12:00 UTC is 12:00 GMT in March (not in daylight saving yet)
```
mauromsl, line 74: Does this mean the test will start failing in the summer? :D

src/core/tests/test_templatetags.py
```diff
@@ -28,3 +33,76 @@ def test_stateless_site_url_for_journal_path(self):
...
+        result = dates.offset_date(
+            context,
+            days=1,
+            input_type="datetime-local",
+        )
+        expected = "2025-03-27T12:00"
```
mauromsl, line 80: given that we are freezing time with a timezone-unaware string, I believe we could have unintended consequences for the translation between the host timezone and the timezone selected for the test user `"Europe/London"`. Namely daylight savings - If I'm reading all this correctly the result will be one hour off the expected for 6 months every year

mauromsl (changes requested):
I think I picked up a problem on the test, otherwise looks good

ajrbyers:
> I think I picked up a problem on the test, otherwise looks good

Ah good catch XD

src/core/tests/test_templatetags.py
mauromsl: Ah, since we are freezing the date, that part becomes a non-issue, but I think we still have an issue if the host system date is not in UTC when running the test

StephDriver (approved):
Looks fine as is, but I am getting some 'interesting' behaviour when putting through non integer values for days (e.g. 0.5) that while they'd pass these tests, are not as I expect them as the human user in my own timezone.  I wonder whether for future proofing a comment in the code that this is not intended for anything other than whole day offsets might be good, so if we then were to change extend the functionality we'd know to rework it?  

It works anyway, so approving, and leaving that tweak up to you.

ajrbyers:
> Looks fine as is, but I am getting some 'interesting' behaviour when putting through non integer values for days (e.g. 0.5) that while they'd pass these tests, are not as I expect them as the human user in my own timezone. I wonder whether for future proofing a comment in the code that this is not intended for anything other than whole day offsets might be good, so if we then were to change extend the functionality we'd know to rework it?
> 
> It works anyway, so approving, and leaving that tweak up to you.

Playing around with this and can see good results for non-integer numbers:

```
In [13]: timezone.now().strftime("%Y-%m-%dT%H:%M")
Out[13]: '2025-04-14T10:03'

DEBUG 2025-04-14 11:03:28,412 selector_events P:7410 T:140199447126016 Using selector: EpollSelector
In [14]: dates.offset_date({}, days=1.5, input_type='datetime-local')
Out[14]: '2025-04-15T22:03'
```

UTC0 timezone in this example, as no request.user is set.

PR #4689, opened 2025-04-01
front of house human readable dates

### LINKED ISSUE #4210: Make Dates more readable
Dates are all over our pages - articles, news items, citations, data... etc.

Dates are not consistently interpreted by screen readers, there are variations between how a US and UK citizen might interpret them (day and month ordering) and they cause readability issues - unlike a character within a word, a digit within a number can't often be determined from context.

When dates are written as numbers separated by punctuation, e.g. `10-11-2024` or `2024-11-10` this is immediately visibly identifiable as a date, even if there might be US/UK confusion over whether it was the 10th of November or the 11th of October.  When read aloud, his recognition that it is a date doesn't happen until one is part was through the number as the above examples are read as numbers e.g. "ten eleven two-thousand and twenty-four". Even with all punctuation being read "ten hyphen eleven hyphen two-thousand and twenty-four" the date is part read before any recognition is possible.

In short, dates rely upon typographical symbols to provide important context, making them effectively non-text content [WCAG 1.1](https://www.w3.org/TR/WCAG22/#text-alternatives)

Consider:
Either dates are written in full as suggested by the UK Gov approach (below) or there needs to be some markup at least for screenreaders. My preferred solution is to write them in full, because that makes them unambiguous across the board.

#### Additional info
- [Accessibility Oz on date formatting](https://www.accessibilityoz.com/2021/06/the-trouble-with-blind-dates-screen-readers-and-date-formatting/)
- [UK Gov style for writing dates](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#dates)
- [WCAG 1.1 techniques](https://www.w3.org/WAI/WCAG22/quickref/?showtechniques=311%2C312%2C111#non-text-content)

StephDriver:
closes #4210

### DIFF
--- a/src/core/templatetags/dates.py
+++ b/src/core/templatetags/dates.py
@@ -2,6 +2,7 @@ from datetime import datetime, timedelta
 
 import pytz
 from django import template
+from django.conf import settings
 from django.template.exceptions import TemplateSyntaxError
 from django.utils import timezone, formats
 from django.utils.translation import gettext as _
@@ -54,16 +55,16 @@ def offset_date(context, days=0, input_type="date"):
 @register.filter(name='date_human', is_safe=True)
 def date_human(value):
     """Convert a date to a human readable Day Month(text) Year format e.g. 3 January 2025"""
-    if value == "":
-        logger.warning("date_human filter received no value - expected a datetime object")
-        return ""
-    elif isinstance(value, datetime):
+    if isinstance(value, datetime):
         return formats.date_format(
             value,
             format="j F Y",
             use_l10n=True,
         )
     else:
-        raise TemplateSyntaxError(
-            "The value filtered by `date_human` must be a `datetime.datetime`"
-        )
+        error_message = "The value filtered by `date_human` must be a `datetime.datetime`"
+        if settings.DEBUG:
+            raise TemplateSyntaxError(error_message)
+        else:
+            logger.error(error_message)
+            return ""
--- a/src/core/templatetags/dates.py
+++ b/src/core/templatetags/dates.py
@@ -1,13 +1,14 @@
-from datetime import timedelta
+from datetime import datetime, timedelta
 
 import pytz
 from django import template
+from django.template.exceptions import TemplateSyntaxError
 from django.utils import timezone, formats
 from django.utils.translation import gettext as _
-from datetime import datetime
-
+import logging
 
 register = template.Library()
+logger = logging.getLogger(__name__)
 
 @register.simple_tag(takes_context=True)
 def offset_date(context, days=0, input_type="date"):
@@ -53,10 +54,16 @@ def offset_date(context, days=0, input_type="date"):
 @register.filter(name='date_human', is_safe=True)
 def date_human(value):
     """Convert a date to a human readable Day Month(text) Year format e.g. 3 January 2025"""
-    if isinstance(value, datetime):
+    if value == "":
+        logger.warning("date_human filter received no value - expected a datetime object")
+        return ""
+    elif isinstance(value, datetime):
         return formats.date_format(
             value,
             format="j F Y",
             use_l10n=True,
         )
-    return ""
+    else:
+        raise TemplateSyntaxError(
+            "The value filtered by `date_human` must be a `datetime.datetime`"
+        )
--- a/src/core/templatetags/dates.py
+++ b/src/core/templatetags/dates.py
@@ -3,6 +3,7 @@ from datetime import timedelta
 import pytz
 from django import template
 from django.utils import timezone
+from django.utils.translation import gettext as _
 from datetime import datetime
 
 
@@ -53,5 +54,6 @@ def offset_date(context, days=0, input_type="date"):
 def date_human(value):
     """Convert a date to a human readable Day Month(text) Year format e.g. 3rd January 2025"""
     if isinstance(value, datetime):
-        return value.strftime('%d %B %Y')  # e.g., "14 March 2024"
+        # Translators: this creates a human readable date e.g., "14 March 2024"
+        return _('%(day)s %(month)s %(year)s.') % {'day': value.strftime('%d'), 'month': value.strftime('%B'), 'year':value.strftime('%Y'), }
     return value
--- a/src/core/templatetags/dates.py
+++ b/src/core/templatetags/dates.py
@@ -3,7 +3,7 @@ from datetime import timedelta
 import pytz
 from django import template
 from django.utils import timezone
-
+from datetime import datetime
 
 
 register = template.Library()
@@ -48,3 +48,10 @@ def offset_date(context, days=0, input_type="date"):
         return due.strftime("%Y-%m-%d")
     elif input_type == "datetime-local":
         return due.strftime("%Y-%m-%dT%H:%M")
+
+@register.filter(name='date_human', is_safe=True)
+def date_human(value):
+    """Convert a date to a human readable Day Month(text) Year format e.g. 3rd January 2025"""
+    if isinstance(value, datetime):
+        return value.strftime('%d %B %Y')  # e.g., "14 March 2024"
+    return value
--- a/src/core/templatetags/dates.py
+++ b/src/core/templatetags/dates.py
@@ -59,4 +59,4 @@ def date_human(value):
             format="j F Y",
             use_l10n=True,
         )
-    return value
+    return ""
--- a/src/core/templatetags/dates.py
+++ b/src/core/templatetags/dates.py
@@ -2,7 +2,7 @@ from datetime import timedelta
 
 import pytz
 from django import template
-from django.utils import timezone
+from django.utils import timezone, formats
 from django.utils.translation import gettext as _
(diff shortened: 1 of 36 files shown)


StephDriver:
## Updated syntax for accessible dates with icons

```html
{% load dates %}

...

<i aria-hidden="true" class="fa fa-calendar"></i>
<time datetime="{{ article.date_published|date:"Y-m-d" }}">
    <span role="time">{{ article.date_published|date_human }}</span>
</time>
```

src/core/templatetags/dates.py
```diff
@@ -28,3 +29,11 @@ def offset_date(days=0, input_type="date"):
...
     elif input_type == "datetime-local":
         return due.strftime("%Y-%m-%dT%H:%M")
+
+@register.filter(name='date_human', is_safe=True)
+def date_human(value):
+    """Convert a date to a human readable Day Month(text) Year format e.g. 3rd January 2025"""
+    if isinstance(value, datetime):
+        # Translators: this creates a human readable date e.g., "14 March 2024"
+        return _('%(day)s %(month)s %(year)s.') % {'day': value.strftime('%d'), 'month': value.strftime('%B'), 'year':value.strftime('%Y'), }
+    return value
```
ajrbyers, line 39: I think we should use `from django.utils import formats` rather than this format. We can break this down to:

```python
    if isinstance(value, datetime):
        return formats.date_format(
            value,
            format="j F Y",
            use_l10n=True,
        )
    return value
```

And get the same result, but the month will be auto translated.

ajrbyers (changes requested):

src/core/templatetags/dates.py
StephDriver: done

src/core/templatetags/dates.py
```diff
@@ -28,3 +29,14 @@ def offset_date(days=0, input_type="date"):
...
+@register.filter(name='date_human', is_safe=True)
+def date_human(value):
+    """Convert a date to a human readable Day Month(text) Year format e.g. 3 January 2025"""
+    if isinstance(value, datetime):
+        return formats.date_format(
+            value,
+            format="j F Y",
+            use_l10n=True,
+        )
+    return value
```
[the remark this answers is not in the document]
StephDriver: tests added - both a great learning curve for me, and it got me thinking more about what should happen if a non-date value was put into the template tag, and I've updated line 42 accordingly.

As I see it, if something that isn't a date is passed in, then that will be due to an error, and it is better that nothing is displayed than that something incorrect is displayed.  

This tag takes a datetime and return a string.  But if left with `return value` as the behaviour for non-datetime objects, then a date-like string could be passed through and return as that same string, but not be included in the translation.  e.g. `date_human(datetime(2024,4,20))` and `date_human("20 April 2024") would both return "20 April 2024" for journals set to English. The latter would not translate, but would not be as easy to detect as an error in browser, especially when developing in English.  A missing date is more likely to be noticed and queried.

ajrbyers (approved):

src/core/templatetags/dates.py
```diff
@@ -28,3 +29,14 @@ def offset_date(days=0, input_type="date"):
...
+@register.filter(name='date_human', is_safe=True)
+def date_human(value):
+    """Convert a date to a human readable Day Month(text) Year format e.g. 3 January 2025"""
+    if isinstance(value, datetime):
+        return formats.date_format(
+            value,
+            format="j F Y",
+            use_l10n=True,
+        )
+    return ""
```
joemull, line 42: As mentioned last week, I think this could go one step further in being helpful to template developers by doing:

```py
else:
    raise TemplateSyntaxError("The value passed to `date_human` must be a `datetime.datetime`")
```

Curious if @ajrbyers also likes this approach?

https://docs.djangoproject.com/en/5.1/topics/templates/#django.template.TemplateSyntaxError

src/core/tests/test_templatetags.py
```diff
@@ -28,3 +31,94 @@ def test_stateless_site_url_for_journal_path(self):
...
+        """Test date_human with all supported languages from settings"""
+        for key, test_date in self.test_dates.items(): 
+            for lang_code, language in settings.LANGUAGES:
+                with self.subTest(date=key, code=lang_code):
+                    activate(lang_code)
+                    result = dates.date_human(test_date)
+                    expected = self.expected_formats[key].get(lang_code)
+                    self.assertEqual(result, expected, 
+                        f"Failed for {lang_code}:{language}.  Expected '{expected}', actual '{result}'."
+                    )
```
joemull, line 102: I could be wrong but I think we'll want to override the settings here to hard-code the languages in `janeway_global_settings.py`, since someone might put different languages in their Django settings, causing this test to fail. (The syntax used elsewhere is a decorator `@override_settings`.)

StephDriver: I've been thinking a lot about this and I think I need to understand more about how someone might have different languages to those in the global settings.  Could you explain further about this scenario?

In the meantime, prompted by this I've also made the tests more robust by checking for mismatches between `settings.LANGUAGES` and the test data so that any language that cannot be tested, or is being tested but isn't supported is reported back - i.e. we will be prompted to update this test when the supported languages change and while doing so it won't be satisfied until we have full test data for any new languages. It also will warn us if we are testing for languages that don't exist as options.

src/core/tests/test_templatetags.py
```diff
@@ -28,3 +31,94 @@ def test_stateless_site_url_for_journal_path(self):
...
+    
+    def test_date_human_browser_languages(self):
+        """Test data_human uses application language regardless of browser language setting"""
+        supported_languages = [lang[0] for lang in settings.LANGUAGES]
```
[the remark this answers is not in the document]
StephDriver: again - it's now more robust, but I'm not yet convinced overriding and hardcoding is the way to go.

src/core/templatetags/dates.py
StephDriver: I have implemented this, but tweaked it a bit.  The `if not value` was failing on tests for zero as it also produces a false.  So I've got it specifically testing for the empty string, and then inputs such as `0` and `None` end up with the `TemplateSyntaxError` it's only empty strings that will get a 'free pass' with the page being rendered missing a date.

I noted that this also means if there is a syntax error this can get a free pass on the page working, i.e. no Error thrown, but the date also doesn't show, for example:

`{{ article.date_acepted|date_human }}` where there is a spelling error in the variable just disappears off screen being handled like an empty string. 

I'm unsure whether this is good or not - on the one hand it means a small syntax error is effectively squashed, the page just doesn't show a date, but otherwise the page is fine - which feels like a better user experieence than say every article page going down to a TemplateSyntaxError if a spelling mistake gets into the template. But on the other hand, we get no warning that there is a problem.  I feel like perhaps we should have a debug warning for the empty string, something that doesn't stop the page rendering, but does clue in developers that there is a problem.

thoughts?

StephDriver:
I've made quite a few tweaks to this since the first reviews, which implement some but not all of what was suggested.  So here's my rationale:

### hardcoding languages into test rather than using settings.LANGUAGES
I decided against this for two reasons:
1. I dislike repeating data when I could look it up, one place of truth and all that.
2. I liked the reasoning of making sure the test was applicable when languages change, but by hardcoding the languages here, they wouldn't then update with changes, so I felt it more robust approach would be to have the test check all languages listed had tests data, and report a fail if the didn't. This way developers are notified they need to update the test.  If I'd hardcoded the languages in, the test would potentially become out of synch with the actual languages being used.  

### TemplateSyntaxError
I felt it was an improvement to throw an error when non valid dates are presented, but I do not want to risk this error being encountered in production for such a small part of a page. I do not want to present **incorrect** dates, but I feel hiding the invalid date is a good approach as the rest of the page will still load, and anyone wanting that particular part is aware it is not there and could query that, while for others they may not even notice and be able to go about their business as usual.   But for developers, it is important they are stopped in their tracks and required to deal with it if their new code includes invalid dates. This is particularly important as something as easy to miss as a spelling error in a variable name could result in a non `datetime` object being passed into the tag.  This means the error behaviour should be different when `DEBUG=True` and also means that the tests themselves needed updating to use `override_settings` and check for behaviour for both `True` and `False`.

### Empty Strings
Previously I had a special case for the empty string so that 'nothing happened' and a empty string was returned, and the introduction of the TemplateSyntaxError was for other invalid input only.  On further thought I felt this needed the same error handling as other invalid data.  This is because there are two 'easy' errors which could be handled as empty strings:
1. a spelling error in a variable name to one that does not exist is interpreted through django templates as passing an empty string, e.g. `{{ article.date_acepted|date_human }}` versus `{{ article.date_accepted|date_human }}`
2. writing in what seems to be a date, if without quotation marks, could be interpreted as a variable that does not exist as in 1. e.g. `{{ today|date_human }}`
As when `DEBUG=False` all human date errors are hidden from the page, this will only show up to developers anyway, and could catch a few mistakes.

ajrbyers (approved):

joemull (approved):
Looks great!

Issue #145, opened 2017-09-10
Allow Coauthor Confirmation


ajrbyers:
If enabled, coauthors are asked to confirm the submission is ok, expires after set period.

hlflanagan:
+1 to getting this feature in!

ajrbyers:
> +1 to getting this feature in!

This will be folded into 1.3.7 so not long!

PR #4697, opened 2025-04-07
Author interface for ROR, CRediT, and ORCiD in submission workflow

### LINKED ISSUE #1485: Wiki update ORCID Information
**Update information about the API connection**
Currently, the information provided for the ORCID integration is not updated, since the API v1.2 is no longer available.

**Updated information**
ORCID_API_URL = 'https://pub.orcid.org/v2.1/' or 'https://pub.orcid.org/v2.0/'

More info: https://members.orcid.org/api/news/public-api-now-only-supports-api-20

### LINKED ISSUE #3111: Allow adding authors by ORCID (alone)
Improvement in adding authors to a submission.

It should be possible to add authors via ORCID ID alone.  The necessary information should be pulled from the orchid registry, and if not available the user should be asked to provide the missing information.

In reference to #2357 a list of ORCIDs should work as well, not as csv but copy&pasted into the orchid fields.

Gathering information entries from ORCID and via ORCID ID is the only way to uniquely describe an author with a single entry.  
Other ways still need to be valid, since not everybody uses / wants to use an ORCID. 

**Additional context**
This expands on #2292 since not only the field needs to be evaluated but also the information available when accessing the registry entry.

### LINKED ISSUE #4519: Update the author submission screen
Currently the author submission screen is spread across two columns which doesn't give us a lot of room to display information especially now that we are adding ROR and credit taxonomy information. This change would also be good for consistency as all other submission interfaces are single column.

- [ ] Sketch a new layout for the authors submission interface

joemull:
Closes #4519.
Closes #3111.
Fixes #1485.

This follows the proposed design of #4519 as closely as possible. The ORCiD-based flow works really well, CRediT roles are easy to add, it's easy to change the correspondence author, and authors are easy to re-order.

However, there's a bit more work to do around editing co-author affiliations. We may need to move straight on to https://github.com/openlibhums/janeway/issues/145, which was out of scope for this PR but is a natural way to solve the problem of allowing maximum control over affiliation data, while not introducing security risks where bad actors could edit any user's affiliations.

image
image
image


ajrbyers:
@joemull can you check you've pushed everything? I think there are some missing imports on `submission.logic`.

joemull:
I don't get any import errors @ajrbyers--was it on server start or on loading a page?

ajrbyers:
> I don't get any import errors @ajrbyers--was it on server start or on loading a page?

Check `submission.logic` there are a bunch of function calls for things that are not imported:

`orcid.get_orcid_record_details(cleaned_orcid)`
`generate_dummy_email(orcid_details)`
`utils_shared.generate_password()`

joemull:
Ah, right you are. I put the author search functionality in that file at the last minute. To guard against simple things like this, I should probably write quick view tests for each branch of the view's POST logic, eh?

joemull:
We did a demo of this in the design meeting last week. Here are notes from that.

* Maybe the help text for inviting a co-author to update their affiliations should explain that an account exists, and the co-author should try to log in and will be guided through the process. Something like: “This account may be inactive, and if so the user will be guided through the account activation process.”
* For adding credit roles, use the same hide/display interaction as for correspondence authors, rather than displaying them all in a horizontal spread of buttons.
* Avoid CSS reflow of Edit and Delete buttons on affiliation when a second author is added.
* Consider changing the “Add affiliation manually” to “Add my affiliation” or similar to avoid the user expecting this button to appear as well for co-authors.
* On other buttons, add the object of the action too, like “Remove author”, “Edit affiliation”, etc.

joemull:
My other todos when this comes back from review:

* Add translation tags?
* Check accessibility of the “Change” dropdown - aria roles and states etc.
* Add tests for each branch of the view logic
* Author names should be h3s

src/submission/forms.py
```diff
@@ -261,63 +262,32 @@ class Meta(ArticleInfo.Meta):
 
 
 class AuthorForm(forms.ModelForm):
```
ajrbyers, line 264: This class needs tidying up. Remove commented out fields.

src/submission/tests/test_views.py
```diff
@@ -0,0 +1,239 @@
...
+            cls.journal_one,
+            '',
+        )
+        logic.add_user_as_author(cls.kathleen, cls.article)
+
+    @override_settings(URL_CONFIG='domain')
+    def test_submit_authors_get(self):
+        self.client.force_login(self.kathleen)
+        response = self.client.get(
+            f'/submit/{self.article.pk}/authors/',
```
ajrbyers, line 59: Why aren't we reversing the URL here? If the URL pattern changes the test will break.

ajrbyers (changes requested):
Some general thoughts:

- When attempting to update affiliations from ORCID using (https://orcid.org/0000-0003-2126-266X) I was informed that there were no affiliations. My public profile shows I have them. Are we using something else to find the user on ORCID? The email address this user has does not match one I have on ORCID.

- When posting a CREDIT for an author can we return the user to that record so they don't need to scroll to find them?

- Should we include a link for the submitting author to share with their co-author in this text "If this author's name, email, ORCiD, or affiliations need to be updated, please invite them to edit their profile on the Open Library of Humanities website."?

- Lastly, could the CREDIT record items be the remove buttons and be on a single line? This would reduce the height of each record which was an issue we discussed.

Some code comments in line.

src/submission/logic.py
```diff
@@ -275,15 +282,147 @@ def order_fields(request, fields):
...
+                new_author = candidates.first()
+                break
+
+    # Otherwise, check if there is at least one public email address in the
+    # ORCiD record, or make a fake email, and then make an account.
+    if not new_author and orcid_details:
+        if emails:
+            email = emails[0]
+        else:
+            email = generate_dummy_email(orcid_details)
```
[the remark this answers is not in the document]
joemull: Yeah, it's a bit wonky. If they have a valid orcid, I was anticipating they'd log in using orcid. That's the idea behind this message:

https://github.com/openlibhums/janeway/blob/5713e3de798279c6edbd324568df5b2cacdf1106/src/templates/admin/submission/submit_authors.html#L86-L92

src/submission/tests/test_views.py
joemull: The first few view tests I wrote for Janeway were based on the Django docs, which do not reverse URLs but write them out.

https://docs.djangoproject.com/en/4.2/topics/testing/tools/#overview-and-a-quick-example

Later I noticed this was different from the way they've been written in Janeway before. At this point I've added 50+ tests with literal URLs. Shall I reverse them?

joemull:
@ajrbyers the change to the credit role display you requested is done now.

Also forgot to mention yesterday that I could not reproduce the error you found with updating your affiliations from ORCID. It works by querying the public ORCID API using a cleaned version of the ORCID of the request user. Anything you were doing that would trip that up?

Obviously in the editor UI I'll have to make it possible for an editor to bulk update someone else's affils, and to do so for a frozen author.

ajrbyers (changes requested):
Thanks for this - looks good. Just one final thing from me: when you edit your own details although the `next` querystring is set there is no way to get back from the edit profile page.

ajrbyers (approved):

src/submission/logic.py
```diff
@@ -275,15 +282,149 @@ def order_fields(request, fields):
...
+    messages.add_message(
+        request,
+        messages.SUCCESS,
+        sentences[change_order],
+    )
+    return author
+
+
+def add_author_from_search(search_term, request, article):
+    query = Q(email=search_term)
```
mauromsl, line 334: for my own sanity, can we split this function? There are just too many branches. At a first step we want to check if the search term is either an Orcid, an email or something else. Then we delegate to `add_author_via_email` and `add_author_via_orcid`

Also try to keep all updates to a record in the same area of the codebase.

src/submission/tests/test_workflow.py
```diff
@@ -541,41 +536,7 @@ def test_url_based_orcid_cleaned(self):
...
-            }
-        )
-        self.assertTrue(
-            form.is_valid(),
-        )
-        self.assertEqual(
-            form.cleaned_data.get('orcid'),
-            '0000-0003-2126-266X',
-        )
+            clean_orcid_id('Mauro-sfak-orci-dtst')
```
mauromsl, line 539: :man_dancing:

src/utils/orcid.py
```diff
@@ -78,13 +78,27 @@ def get_orcid_record(orcid):
...
+
+def get_affiliations(summary):
+    affils = []
+    affils.extend(
+        [affil for affil in summary["employments"]["employment-summary"]]
+    )
+    if not affils:
+        affils.extend(
+            [affil for affil in summary["educations"]["education-summary"]]
+        )
```
mauromsl, line 98: It makes sense for get_affiliations to return these, but I think we are adding them automatically to the user when they are added via orcid, do we want that?

src/submission/logic.py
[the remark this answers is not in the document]
mauromsl: We are not flagging the account as active, so they won't be able to login with orcid.
And the email address is fake, so they won't be able to login either. I reckon for these authors (no email address) we should NOT create an account, only a `FrozenAuthor` record.

mauromsl (changes requested):
.

src/utils/orcid.py
joemull: Yeah, I was replicating the logic we used to use for a single affiliation. Do you think it would be unexpected for all your education affils to be pulled through if you are a student? Too much data? Get just the last one?

src/submission/logic.py
```diff
@@ -275,15 +282,149 @@ def order_fields(request, fields):
...
+        orcid_details = {}
+        emails = []
+
+    # First check if there is a Janeway account with an email address
+    # that matches an email address in a public ORCiD record.
+    if not new_author and emails:
+        for email in emails:
+            candidates = core_models.Account.objects.filter(email=email)
+            if candidates.exists():
+                candidates.update(orcid=cleaned_orcid)
```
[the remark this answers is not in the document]
joemull: Yeah, it's a good question.

I think the best option is to check for an account that already has that ORCID in JW. If it exists, use it. This function should be doing that anyway. It would avoid a situation where multiple ORCIDs are affiliated with the same account.

If there is no match, then I still do think we should check the ORCID record's email addresses against email addresses in Janeway, and update the first account found with the ORCID.

The reason is: this is what we do during ORCID login--I was replicating the behavior we have already used for a few years there:

https://github.com/openlibhums/janeway/blob/e73827dc9db693756a3fe448f7a00d9963b2f821/src/core/views.py#L180-L185

If we don't link existing accounts, it increases the likelihood of duplicate accounts, which is a big headache for usability. The alternative I see is to use frozen authors only for co-authors during this step, as you suggest below, but that opens a few cans of worms. See my response below on that.

As for the security risk you mention, yes, we are trusting ORCID. [They say](https://info.orcid.org/ufaqs/why-cant-i-read-all-email-addresses/): "Unverified emails do not display in the public view of the record and only email addresses that have been verified will appear in API results."

If we don't want to trust them, we need to change the login with ORCID flow as well, because the same risk applies there. Do you want to do that?

joemull:
Thanks @mauromsl, see my comments above.

GitHub didn't give me a "Reply" text box on your last comment for some reason, so I'll reply here: 

> We are not flagging the account as active, so they won't be able to login with orcid.
> And the email address is fake, so they won't be able to login either. I reckon for these authors (no email address) we should NOT create an account, only a `FrozenAuthor` record.

The login with ORCID flow doesn't check for an inactive account. Are you saying this will error out with an inactive account? 

https://github.com/openlibhums/janeway/blob/e73827dc9db693756a3fe448f7a00d9963b2f821/src/core/views.py#L165-L168

As for whether to create frozen authors here, I originally was headed in that direction, because I fundamentally don't like the feature of Janeway where you can create an account for someone else as a non-trusted party. But it raises all kinds of questions about how you'd re-link records to authors, and the view during author submission would have to handle either author or frozen author records. Let's discuss offline.

joemull:
Apropos: https://github.com/openlibhums/janeway/issues/4751

src/utils/orcid.py
mauromsl: @ajrbyers thoughts?

mauromsl:
We have decided to change the submission system to use FrozenAuthor records for co-authors, since it solves some problems that have arisen in this PR, some pre-existing:
- FrozenAuthors can be safely edited by submitting authors during submission.
- Account issuance shouldn't be at the will of the submitting author.
- ORCID account integration for a third party account will not be triggered by the submitting author
- Submitting authors can safely edit their co-authors affiliations for a submission without altering requiring acccess to edit their account profiles.

Expected implications:

- Article-level snapshotting needs to account for the pre-existance of FrozenAuthors.
- Where a co-author is added and matches an existing account in Janeway (via email or ORCID) we should link the newly created FrozenAuthor record to the account but avoid changing the Account model itself.
- Need to reconsider how we store ArticleAuthorOrdering since it relies on Accounts prior to snapshotting

ajrbyers:
image

joemull:
OK @ajrbyers this is ready for your review again. I tested it manually and made a few small fixes to the UI, including to the broken "Back to all authors" button.

src/templates/admin/submission/edit/author.html
```diff
@@ -49,6 +49,7 @@
         <h2>{% trans "Name, bio, and identifiers" %}</h2>
         {% if not form %}
           <form method="GET">
+            <input type="hidden" name="next" value="{{ request.GET.next }}">
```
mauromsl: I see we guard against a value of `None` elsewhere in the code base. Should we do the same here to avoid `None` being serialized as text in the HTML? (i.e. `{{ request.GET.next|default:""}}`

PR #4708, opened 2025-04-08
Submission fields can now be unchcked. Review form checkboxes now display labels.

### LINKED ISSUE #4581: Checkbox name field doesn't show on custom fields
**Describe the bug**
When creating a custom field in submission or review and using the Checkbox field type, the Name field (title/header) doesn't show when rendering the field: 
image

This is no bueno as users may not know what they are being asked to tick, especially if this field is required and no help text has been added.

For example, see:
https://www.openlibhums.org/doctest/review/requests/22041/?access_code=825560cc-6c61-4c08-ae05-e8c197049159

**Janeway version**
1.7 

**Additional context**
Tested in both form previews and actual forms.

ajrbyers:
Closes #4581

### DIFF
--- a/src/submission/tests.py
+++ b/src/submission/tests.py
@@ -831,7 +831,7 @@ class SubmissionTests(TestCase):
             article=self.article,
             field=self.boolean_field,
         )
-        self.assertEqual(field_answer.answer, 'True')
+        self.assertEqual(field_answer.answer, 'on')
 
     def test_text_field_sets(self):
         post_data = {
--- a/src/submission/tests.py
+++ b/src/submission/tests.py
@@ -52,21 +52,6 @@ class SubmissionTests(TestCase):
         if not self.journal_one.submissionconfiguration:
             self.fail('Journal does not have a submissionconfiguration object.')
 
-    @staticmethod
-    def create_journal():
-        """
-        Creates a dummy journal for testing
-        :return: a journal
-        """
-        update_xsl_files()
-        update_settings()
-        journal_one = journal_models.Journal(code="TST", domain="testserver")
-        journal_one.title = "Test Journal: A journal of tests"
-        journal_one.save()
-        update_issue_types(journal_one)
-
-        return journal_one
-
     @classmethod
     def create_authors(cls):
         author_1_data = {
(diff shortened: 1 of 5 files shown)


src/templates/admin/elements/generate_form.html
```diff
@@ -4,8 +4,8 @@
 <div class="form-group">
     <div class="{% if field.field.widget.attrs.div_class %}{{ field.field.widget.attrs.div_class }}{% else %}large-12 columns{% endif %}">
         {% if field.field.widget.attrs.is_checkbox %}
-        {% if field.errors %}<span class="red">{{ field.errors|join:", " }}</span><br/>{% endif %}
-        {{ field }} &nbsp;{% if field.help_text %}{{ field.help_text|safe }}{% endif %}
+          {% if field.errors %}<span class="red">{{ field.errors|join:", " }}</span><br/>{% endif %}
+          {{ field }} {{ field.label }}&nbsp;{% if field.help_text %}<small>{{ field.help_text|safe }}</small>{% endif %}
```
joemull, line 8: In the user interface for custom review fields, this puts out HTML like this:

```html
<div class="form-group">
          <input type="checkbox" name="3" is_checkbox="" id="id_3"> Test element&nbsp;
    </div>
</div>
```

That's not [how HTML labels should be marked up](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/label). There needs to be a `label` element and it should be linked to the `input` id with the `for` attribute. This will make the label clickable in place of the checkbox to change the selection, and it will make the whole thing more accessible in other ways as well.

src/submission/tests.py
```diff
@@ -764,6 +772,100 @@ def test_author_doesnt_see_non_public_section(self):
...
+            self.fail(f"Form is invalid: {form.errors}")
+
+        field_answer = models.FieldAnswer.objects.get(
+            article=self.article,
+            field=self.text_field,
+        )
+        self.assertEqual(
+            field_answer.answer,
+            'Sometimes first contact is last contact.',
+        )
```
joemull, line 868: These look good

joemull (changes requested):
There is some oddness with the review form HTML but otherwise looks good.

joemull (approved):

StephDriver (changes requested):
tested:
- ✅ it is now possible to check/uncheck fields in article `edit metadata` without the change reverting on save.
- :x: for review, the name is now showing as a label, but it comes after the checkbox and yet a colon is being added after the name.

image

image

Changes:
- please either remove the colon, or move the label before the checkbox.

StephDriver (approved):

PR #4714, opened 2025-04-10
Fix a bug with constraint names that caused a database error during upgrade

### LINKED ISSUE #4713: Constraint identifier too long for MySQL/MariaDB
**Describe the bug**
Applying migrations fails on
0104_location_organization_affiliation.py
due to a constraint name
"controlled_affiliation_account_frozen_author_preprint_author_exclusive_fields_constraint"
which exceeds the max allowed character limit (64 chars) on MariaDB / MySql.
(PostgreSQL may truncate silently)

**Janeway version**
master, commit 2f8190245860b3c844fbda8230bd81531d1d69e1

**To Reproduce**
Try to migrate from a database which hasn't had the above migration applied.

joemull:
Fixes #4713.


ajrbyers:
There is a non zero chance that this results in constraint name duplication. Django will truncate long constraint names using this method:

https://github.com/django/django/blob/21f8be76d43aa1ee5ae41c1e0a428cfea1f231c1/django/db/backends/utils.py#L283-L312

I test this out with this example:

```python

class Meta:
    unique_together = ('article', 'review_round', 'reviewer', 'editor', 'decision', 'date_declined', 'date_accepted', 'review_type', 'competing_interests')
```

It generates the following constraint in psql:

```
"review_reviewassignment_article_id_review_round__a7cd5089_uniq" UNIQUE CONSTRAINT, btree (article_id, review_round_id, reviewer_id, editor_id, decision, date_declined, date_accepted, review_type, competing_interests)
```

ajrbyers (changes requested):
See comment.

joemull:
OK how about this.

PR #4715, opened 2025-04-10
display author CRediT in back-office workflow

### LINKED ISSUE #4703: Add display of CRediT information throughout the workflow UI where it is a relevant piece of metadata
CRediT roles should be displayed for the frozen authors of the article through the workflows as appropriate.

"appropriate" - consider the layout and whether this would add clutter.  Start by adding it and see what you find.

StephDriver:
closes #4703

### DIFF
--- a/src/templates/admin/elements/metadata.html
+++ b/src/templates/admin/elements/metadata.html
@@ -171,6 +171,9 @@
     <table class="scroll small" id="unassigned">
         <tr>
             <th>Name</th>
+            {% if journal_settings.general.use_credit %}
+                <th>CRediT</th>
+            {% endif %}
             <th>Email</th>
             <th>Display Email Link?</th>
             <th>Primary?</th>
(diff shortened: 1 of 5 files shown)


joemull (changes requested):
So I think this does need some more changes:

* The template that needs to change for editors' *viewing* metadata is not `submission/edit/metadata.html` as you've picked up here, but `admin/elements/metadata.html`.
* The typesetter has a different view of this data that needs to be changed here: https://github.com/openlibhums/janeway/blob/427b28ece5f1eb1536f170c745a5bf3d70739775/src/typesetting/templates/typesetting/elements/typesetter/metadata.html#L97

joemull:
It might also be worth checking that the reviewer gets this information in their template as well.

StephDriver:
> It might also be worth checking that the reviewer gets this information in their template as well.

As far as I can see there are three metadata templates.
- `admin/submission/edit/metadata.html` is used for editing metadata in all workflows including typesetting
- `admin/elements/metadata.html` is the modal display of metadata the 'view metadata' button, which then has an 'edit metadata' link which goes to the above editing template
- `typesetting/elements/typesetter/metadata.html`is the display that appears within the cover page of a typesetting assignment.

I have now updated all three.

ajrbyers (approved):

PR #4721, opened 2025-04-15
Add Submissions and Publications to User History

RoseReatherford:
This adds both publication and submission history of a user to the history section. A link to the archive page of the article is also provided if the article isn't in the "UNSUBMITTED" stage. 

Preview: 
Screenshot 2025-04-11 at 1 09 48 PM

### DIFF
--- a/src/templates/admin/core/manager/users/history.html
+++ b/src/templates/admin/core/manager/users/history.html
@@ -17,6 +17,68 @@
         <div class="row expanded">
             <div class="large-6 columns">
                 <div class="title-area">
+                    <h2>Submissions</h2>
+                </div>
+                <div class="content">
+                    <table class="scroll small">
+                        <tr>
+                            <th>ID</th>
+                            <th>Article</th>
+                            <th>Stage</th>
+                            <th>Date Started</th>
+                            <th>Last Update</th>
+                        </tr>
+                        {% for submission in submissions %}
+                            <tr>
+                                <td>{{ submission.pk }}</td>
+                                {% if submission.stage == stages.STAGE_UNSUBMITTED %}
+                                    <td>{{ submission.safe_title }}</td>
+                                {% else %}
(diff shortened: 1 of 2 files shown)


src/templates/admin/core/manager/users/history.html
```diff
@@ -14,6 +14,68 @@
...
+                                {% else %}
+                                    <td><a href="{% url 'manage_archive_article' submission.pk %}">{{ submission.safe_title }}</a></td>
+                                {% endif %}
+                                <td>{{ submission.stage }}</td>
+                                <td>{{ submission.date_started }}</td>
+                                <td>{{ submission.date_updated }}</td>
+                            </tr>
+                        {% empty %}
+                            <tr>
+                                <td colspan="4">No submissions</td>
```
StephDriver, line 45: there are 5 columns, not 4. This ends up like this when there are no submissions:
image

StephDriver:
Changes Needs:

- `colspan=5` as in previous comment
- BUG: When read with VoiceOver, the Publications table is not read aloud. (to reproduce, start at top of page, on read all, the Submissions heading and then table are read out, then the publications heading, then the editorial heading and table and so on.  The publications table is skipped).
- Other tables here do not have Article ID columns, so this is odd having them just in these two tables. Please justify this decision, or remove the ID columns, or add ID columns to the other tables.
image

ajrbyers:
> Please justify this decision, or remove the ID columns, or add ID columns to the other tables.

I think adding ID cols to the other tables is the best approach.

StephDriver (changes requested):

RoseReatherford:
Completed review comments. Furthermore, changed the phrasing for no results and added links to all article titles for further consistency.

src/templates/admin/core/manager/users/history.html
RoseReatherford: Done!

StephDriver (approved):
This looks fantastic! Thanks for updating all the tiny pieces that make it so consistent now!

PR #4726, opened 2025-04-17
JATS and Crossref XML outputs now include CREDIT and ROR data

### LINKED ISSUE #4623: Downstream metadata integration of ROR
## User stories

User: Publishing librarian or research impact librarian

- [ ] I can see ROR data in Crossref deposits from Janeway
- [ ] I can see ROR data in Janeway's OAI-PMH feeds, including JATS XML

ajrbyers:
Closes #4623


src/journal/views.py
```diff
@@ -1319,6 +1320,35 @@ def publish_article_check(request, article_id):
...
+    }
+    xml_output = render_to_string(
+        'common/encoding/article_jats_1_2.xml',
+        context,
+    )
+    valid, error_message = xml_validation.validate_jats_with_remote_dtd(xml_output)
+
+    if not valid:
+        error_comment = f"<!-- JATS Validation Error: {error_message} -->\n"
+        xml_output = xml_output + error_comment
```
mauromsl, line 1348: Adding a comment at the top of the XML document will make it not XML valid and some browsers might display an error. if we need to add a comment, it should be after the initial XML declaration.

mauromsl (changes requested):
looks good, just a small comment on the display of the faulty JATS in the debug view

mauromsl (approved):

src/templates/common/identifiers/crossref_article.xml
```diff
@@ -1,88 +1,138 @@
...
+
+                        {% if affiliations %}
+                            <affiliations>
+                                {% for affil in affiliations %}
+                                    {% with org=affil.organization %}
+                                        <institution>
+                                            <institution_name>{{ org.name.value }}</institution_name>
+                                            {% if org.ror_id %}
+                                                <institution_id
+                                                        institution_id_type="ROR">
```
joemull: This should be:
```xml
type="ROR">
```
https://www.crossref.org/documentation/schema-library/markup-guide-metadata-segments/affiliations/

src/templates/common/identifiers/crossref_article.xml
```diff
@@ -1,88 +1,138 @@
...
+                                                    https://ror.org/{{ org.ror_id }}</institution_id>
```
joemull: Rather than manually join the domain to the id, you can can use `org.uri` here as you did in the JATS schema.

src/templates/common/identifiers/crossref_article.xml
```diff
@@ -1,88 +1,138 @@
...
+                                            {% endif %}
+                                        </institution>
+                                    {% endwith %}
+                                {% endfor %}
+                            </affiliations>
                         {% endif %}
+
                         {% if author.orcid %}
-                        <ORCID>https://orcid.org/{{ author.orcid }}</ORCID>
+                            <ORCID>https://orcid.org/{{ author.orcid }}</ORCID>
```
joemull: This will produce bad data where the ORCID field has the domain in addition to the id. I had the same problem in the other ROR branch being reviewed right now. I created this method for it, maybe useful here as well:

https://github.com/openlibhums/janeway/blob/e319a3d937f0e741e91559c67c3a15ac85cc75ef/src/submission/models.py#L2278-L2287

src/templates/common/identifiers/crossref_article.xml
```diff
@@ -1,88 +1,138 @@
...
+                        {% endif %}
+
+                        {% if article.object.journal|setting:'use_credit' and author.credits.exists %}
+                            {% for credit in author.credits %}
+                                <contrib_role>
+                                    <role_name>{{ credit.role|title }}</role_name>
```
joemull: I think it would be better to use `credit.get_role_display` because this method might uppercase something that isn't typically uppercased in the CRediT standard.

src/templates/common/identifiers/crossref_conference_paper.xml
```diff
@@ -1,57 +1,92 @@
...
+                            <contrib_role>
+                                <role_name>{{ credit.role|title }}</role_name>
+                                <role_identifier>{{ credit.uri }}/</role_identifier>
+                            </contrib_role>
+                            {% endif %}
+                        {% endfor %}
                     </person_name>
+                    {% endwith %}
+                {% endfor %}
+            </contributors>
```
joemull: If the contributors section is the same for articles and conferences, could we create a sub-template and include it in both? That would make it easier to maintain.

src/submission/models.py
```diff
@@ -1969,6 +2007,32 @@ def pinned(self):
...
+        if not self.abstract:
+            return ""
+
+        # Clean the abstract bleach
+        cleaned = clean(
+            self.abstract,
+            tags=utils_const.get_jats_allowed_tags(),
+            attributes=utils_const.get_jats_allowed_attrs(),
+            strip=True,
+        ).strip()
```
joemull: I am not sure about this method. I believe that in effect, it will strip all tags out of the abstract, because abstracts are authored in HTML, with tags like `i`, `em`, `b`, and `strong`, but the list of jats allowed tags you have written checks for `italic`, `bold`, etc.

Is having HTML in the abstracts a problem for validation?

src/submission/models.py
```diff
@@ -1969,6 +2007,32 @@ def pinned(self):
...
+
+        return mark_safe(f"<p>{cleaned}</p>")
```
joemull: What about multi-paragraph abstracts?

src/utils/xml_validation.py
```diff
@@ -0,0 +1,29 @@
...
+            load_dtd=True,
+            no_network=False,
+            resolve_entities=True,
+        )
+        parser.resolvers.add(NetworkResolver())
+
+        doc = etree.fromstring(xml_string.encode("utf-8"), parser)
+        return True, "Validation successful"
+    except etree.XMLSyntaxError as e:
+        return False, str(e)
```
joemull, line 29: Fancy!

joemull (changes requested):
Glad to have this work moving this direction.

I noted a few places where the crossref templates can be shored up and posed some questions about the jats bleacher.

src/submission/models.py
ajrbyers: Yeah, JATS abstracts don't support HTML. I'm writing an XSLT that will support the subset of available JATS tags and try to handle the rest.

src/templates/common/identifiers/crossref_article.xml
ajrbyers: Thanks - will wait for the other branch and then rebase to get this method in.

src/templates/common/encoding/article_jats_1_2.xml
```diff
@@ -33,52 +36,62 @@
...
-                            </role>
-                        {% endfor %}
-                    {% endif %}
-                    <xref ref-type="aff" rid="aff-{{ forloop.counter }}"/>
-                </contrib>
-            {% endfor %}
+                {% for item in article.frozen_authors_for_jats_contribs %}
+                    {% with author=item.author affiliations=item.affiliations %}
+                        <contrib contrib-type="author"{% if author.is_correspondence_author %} corresp="yes"{% endif %}>
+                            {% if author.orcid %}
```
mauromsl: call me superstitious, but shall we check `author.orcid_uri` here to avoid having different results from `author.orcid` and `author.orcid_uri`? Seems like if the orcid was malformed, `author.orcid_uri` will return a falsey value here

mauromsl (approved):
Minor nitpick and good to go

src/templates/common/encoding/article_jats_1_2.xml
ajrbyers: Yeah, a good shout.

PR #4753, opened 2025-04-25
Review form elements no longer have the width field

### LINKED ISSUE #4582: Field width configuration not functional
**Describe the bug**
Whilst 'width' can be configured for custom review forms, it does not actually change the width and all fields display as full width. This presumably will also be the case for custom submission fields. This has been tested with various field types; text-types (as they are large by default and that might have explained it, but alas), checkboxes (small, but have the Name-field bug (#4581), but alas) and select / upload (small and no known bugs, but alas). The problem does however persist, both for 1/3 and 1/2 width.

For examples, see:

https://www.openlibhums.org/doctest/review/requests/22158/?access_code=a9d0606c-f164-48f4-bc6e-970318fa7e1e
https://www.openlibhums.org/doctest/review/requests/22159/?access_code=39793246-5fb2-45cb-b059-d9d0a889f641
https://www.openlibhums.org/doctest/review/requests/22041/?access_code=825560cc-6c61-4c08-ae05-e8c197049159

**Janeway version**
1.7

**Additional context**
I suggest removing the width configuration option rather than fixing it. I do not recall ever seeing it used (and if it is, it being dysfunctional does not seem to bother anyone). Especially as it is a largely aesthetical configuration option that is non-essential. (If anything, it probably increases the chance of a field being missed by accident, if they are 1/3 or 1/2 width.)

ajrbyers:
Closes #4582

- Removes the width element from Form Elements.


mauromsl (approved):

StephDriver (approved):
checked entry and display of review forms, all working.

PR #4754, opened 2025-04-25
The view reviews template now splits reviews by round and only displays an appropriate download link

### LINKED ISSUE #4594: Double review file download link - one broken
**Describe the bug**
On review #21223 on Glossa, the author can see two file download links, only one is functional and the other returns an error. This was confusing for the author as they thought they were missing a file and thus contacted the editor who contacted support. Especially as there are two reviews (as is often the case), it confused the author into thinking there should have been two accessible file. (One review didn't have a file associated, just webform.)

Image

**Janeway version**
1.7

**Additional context**
See [FD 9668](https://janeway.freshdesk.com/a/tickets/9668)

### LINKED ISSUE #4609: Separate reviews by round when displaying to author
I was uploading a revised manuscript for an author today, I did this because it’s version 3 of the article and the author just sent it to me rather than following the directions to upload their revision.
 
I noticed that the reviews of the paper are collated together, and not separated by rounds, and that can be quite confusing for authors to discern which review references their latest version, see the screenshot below. Is there a way to separate the reviews, just using headers, so the authors can determine themselves which review is to round 1 and which review is to round 2 of their paper? it’s not in numerical order either so I had to go back to the live preprint page to see which review is for which round:

Image

**Additional context**
Submitted through FD 9736 - UCL

ajrbyers:
Closes #4594
Closes #4609 

Two birds, one template.

### DIFF
--- a/src/templates/admin/elements/review/view_reviews.html
+++ b/src/templates/admin/elements/review/view_reviews.html
@@ -1,10 +1,18 @@
+{% regroup reviews by review_round.round_number as reviews_by_round %}
+
 <p>Please click 'View Review' below to view the peer review report.</p>
 
+{% for round in reviews_by_round %}
+<p><strong>Round {{ round.grouper }}</strong></p>
+
 <ul class="accordion" data-accordion data-allow-all-closed="true" data-multi-expand="true">
-    {% for review in reviews %}
+    {% for review in round.list %}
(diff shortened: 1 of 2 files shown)


mauromsl (changes requested):
just a comment around how complex the if statements for displaying the review file have become

PR #4759, opened 2025-05-02
A11y improvements for endnote backlinks

### LINKED ISSUE #4446: Carat for endnotes too small a target
split from 
- #4245 
> ... all [endnote] links being 'carat' ^ which is also too small a target for accessible selection.

Examples from #4194

- **Sample 2 (Material Theme) , item 49.** '] used with the carat as the only link text within notes are problematic. All of them end with these, so there is nothing to differentiate them, and repeated 'carat, link' is unhelpful when traversing the document by links.

- **Sample 3 (Clean Theme), item 19.** the end not link back to the original reading position is not contextual (just read as 'carat') such that all end notes have links that are read out identically.... imagine navigating by link 'carat, link, carat, link, carat, link...'

Not observed in OLH theme during this testing, but the template should be reviewed alongside the changes to Material and Clean to make sure.

StephDriver:
closes #4446

### DIFF
--- a/src/transform/xsl/default.xsl
+++ b/src/transform/xsl/default.xsl
@@ -495,7 +495,7 @@
               <xsl:apply-templates/>
             <xsl:for-each select="//xref[@rid=$fn-id]">
               <xsl:variable name="i"><xsl:value-of select="string(position())"></xsl:value-of></xsl:variable>
-              <a class="footnotemarker"  href="#{$fn-id}-nm{$i}"> ⮭</a>
+              <a class="footnotemarker"  href="#{$fn-id}-nm{$i}" aria-label="reference to {$fn-id} in text" role="doc-backlink"> ---^ </a>
             </xsl:for-each>
           </li>
     </xsl:template>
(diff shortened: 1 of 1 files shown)


StephDriver:
Notes:
1. aria-label is in the xsl, not sure how to translate within that, so not marked for translations
2. cannot workout where line 5070 outputs (of if it outputs) updated similar to 498 for consisitency.
3. 498 is the one I've tested manually with Voiceover, this is now annouced as a 'backlink' with the aria label. The ID is not a number but often has suffix 'n' e.g. `n10` but this is intuitive when read aloud.

ajrbyers:
Thanks for this @StephDriver, two things:

1. The text should say footnote rather than reference
2. Can we add this return feature to references? Currently you can click a ref link to be taken to the footer but cannot be returned

StephDriver:
> 2. Can we add this return feature to references? Currently you can click a ref link to be taken to the footer but cannot be returned

In short, not in the same way, endnotes are expected to come from a single source, a one to one relationship with a location in the text. But references may be referred to multiple times and so are a many to one relationship with the reference list.

ajrbyers (approved):

mauromsl (approved):

PR #4777, opened 2025-05-09
a11y copy citation and share link clipboard updates

StephDriver:
closes 4236


StephDriver:
A11y Syntax Update:

```html
<span aria-hidden="true"> doi: </span>
<a href="https://doi.org/{{ article.identifier.identifier }}" 
    aria-label="{% trans 'DOI for ' %} {{ article.title|safe }}"
    >
    https://doi.org/{{ article.identifier.identifier }}
</a>
```

StephDriver:
I feel I may need to make some changes to unit tests as part of this PR, but I'm not really sure what, so would appreciate guidance from reviewers if they feel that should be done.

StephDriver:
from Design meeting:
- go for consistency with theme
- if within that consistency, can have a less loud button, that's fine, but if not, consistency wins.
- placement is ok.

StephDriver:
Friday WIP
- rebased on master
- updated the Clean copy buttons went through several variations and settled on this after discussion within team.
image


Next 
- [ ] update material and OLH too.

ajrbyers (changes requested):
Please review the changes to the copy DOI button for sharing. They are now padded away from the field and are no longer the same hight as the field, in the case of OLH.

StephDriver:
Ok..... after looking at this further, it gets very messy very quickly if the doi uses `aria-describedby` as the element with the article / preprint title is often nowhere near the doi itself (e.g. on the article page), and as the doi_display is also nested inside other templates such as how-to-cite it often ends up nested away from the element we want to refer back to in described-by. 

So I've reverted back to the original plan of using aria-label.  I still think the ideal would be describedby and this is something to consider when we create components for Janeway 2.0.

StephDriver:
fixed.

ajrbyers (approved):

StephDriver:
rebased.

mauromsl (approved):

Issue #4304, opened 2024-06-27
Clean theme Iframes overflowing article container


StephDriver:
This has been split off from #4261 as it has similar symptoms but not from the same cause.

> Not just References - other content overflows too.  This was noted while testing another fix on my dev install, so this article does not exist on Clean, it was the one flagged up for a different problem in #3911 then downloaded for use as a test subject, but it illustrates the problem.
> 
> 
> 
> <img width="922" alt="image" src="https://github.com/BirkbeckCTP/janeway/assets/5330770/0675a606-038f-4902-b554-988f8333b287">
> 
>  

 _Originally posted by @StephDriver in [#4261](https://github.com/BirkbeckCTP/janeway/issues/4261#issuecomment-2191816366)_

- [ ] content inside the article should be constrained by the article container, and not overflow into the sidebar.

StephDriver:
This was discussed at backlog refinement today, and was split from #4261.

StephDriver:
branch has been merged into master

PR #4781, opened 2025-05-15
A11y sprint css

### LINKED ISSUE #4264: Clean theme table of contents not scrolling
**Describe the bug**
In Clean on wider windows, the table of contents for an article is displayed to the right of the the text.  This remains static as the text scrolls.  When the window is shorter than the height of the table of contents, it still remains static until one scrolls to the bottom of the page, this can mean that for example, the References, often the last item is not displayed on screen until one reaches the footer.

**To Reproduce**
1. Go to [example article](https://politicalphilosophyjournal.org/article/id/15481/) with a wide, but short window.
2. Scroll down the page

**Expected behavior**
Either - when the table of contents cannot fit on the height of the screen it needs to be displayed another way, or it should be scrollable independent of the main contents.

**Screenshots**
image

The table of contents is static through the whole page scroll right up to this point with the last 5 items hidden, then it finally scrolls for the footer:
image

## Front-end Issues

 - Mac Chrome Version 125.0.6422.142
 - Same behaviour observed on Safari 17.5

Not expected, as this is about the wider window location for table of contents.

**Additional context**

Discovered during #4260 because this bug is made worse when viewing the page at sizes required in WCAG 1.4.12, since the change in size can lengthen the table of contents, and thus trigger this bug on windows that otherwise would have had enough space to display the full table.

StephDriver:
working across the list of CSS issues in the A11y sprint

closes #4264

### DIFF
--- a/src/themes/material/assets/mat.css
+++ b/src/themes/material/assets/mat.css
@@ -760,3 +760,8 @@ main {
     padding: 0 10px;
     min-width: 40px;
 }
+
+#toc-card {
+  max-height: 100vh;
+  overflow-y: auto;
+}
--- a/src/themes/clean/assets/css/clean.css
+++ b/src/themes/clean/assets/css/clean.css
@@ -478,3 +478,8 @@ header svg {
     list-style-type: none;
     padding-bottom: 8px;
 }
+
+.sticky-top {
+    max-height: 100vh;
+    overflow-y: auto
+}


StephDriver:
Note, I've had some trouble setting up an article with an iFrame to test the work for #4304  - I can see it working in general but have not been able to be 'complete' in testing it.

src/themes/material/assets/mat.css
```diff
@@ -759,4 +759,21 @@ main {
...
+}
+
+#toc-card {
+  max-height: 100vh;
+  overflow-y: auto;
+}
+
+#article iframe {
+    max-width: 100%;
+    width: 100%;
```
ajrbyers: I think we should remove `width: 100%` and let the iframe set its own width. `max-width: 100%` is enough to prevent overflow.

ajrbyers (changes requested):
All tested, looks good. I added a comment about iframe widths for you to look at. If I've got that wrong feel free to ignore and pass on to next reviewer.

src/themes/material/assets/mat.css
StephDriver: Good spot - I was trying various combinations of CSS and tweaking them, and think I left that redundant line in there by mistake.

ajrbyers (approved):

PR #4782, opened 2025-05-16
a11y: #4211 list syntax

### LINKED ISSUE #4211: Markup decorative typography as decorative
We have used bullets and vertical lines in the middle of text to differentiate between items, for example:

image
image

These are read out to by a screen reader as "vertical line" or "bullet" respectively.

1. this is confusing.  While the presentative elements are being spoken, they are not be communicated in a meaningful way that aids comprehension or navigation.
2. if items need to be differentiated, they should be formatted as a list.

[WCAG 1.3.1](https://www.w3.org/TR/WCAG22/#info-and-relationships)
> Information, structure, and relationships conveyed through presentation can be programmatically determined or are available in text.

StephDriver:
closes #4211

### DIFF
--- a/src/themes/material/assets/mat.css
+++ b/src/themes/material/assets/mat.css
@@ -458,7 +458,7 @@ h5 {
     color: #616161;
 }
 
-.footer-copyright p {
+.footer-copyright p, li {
     color: #616161;
 }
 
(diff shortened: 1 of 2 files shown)


StephDriver:
## Accessible Inline List Syntax

### Clean
```html
<ul class="list-inline">
    <li class="list-inline-item">item</li>
</ul>
```
example:
image


### Material
```html
<ul class="list-inline">
    <li>item</li>
</ul>
```
example:
image


### OLH

StephDriver:
typo in branch name, from 4219 should have been 4211, which meant it was referencing the wrong issue, close as part of renaming the branch to fix this. Comments transferred over to new PR #4783

PR #4789, opened 2025-05-22
article internal cross reference navigation improvements

### LINKED ISSUE #3242: Reference section: "back to text" button next to the citation when a reference is clicked
References don't have a jump mark back to the text if you click on the href, notes do have a circumflex for that.

Inject a "back to text" button next to the citation when it is clicked, using some javascript.

Readers should be able to jump back to the last position of the page after hitting a reference-href.

### LINKED ISSUE #4337: Reference highlighting when jumped to from citations on web version of article
At the moment when citations are clicked in the body text of an article, the webpage jumps to the bottom where the references hang out. At the moment this positions the selected reference at the top of the web browser window, however for references towards the bottom of the webpage they cannot reach the top of the browser as the webpage can't scroll any further down.

Potentially highlighting selected references somehow when they are jumped to (or invoking some kind of visual cue) so that the user knows which reference they should be looking at if it cannot be positioned as the topmost item in the browser window.

StephDriver:
closes #4337 
closes #3242


StephDriver:
Plan:

On page load:
- dynamic generate links to article (i.e. reverse links) for each reference list item. There may be more than one.

On clicking link to a reference in the article:
- go to reference item
- draw user attention to that item (highlighting for visuals + accessible alternatives)
- there should be a backlink for returning to the source location without having to remember which that was (i.e. not a free choice among multiple possible backlinks).

On returning to article via a backlink
- go to the source link in the article
- draw user attention (as above) 
- where either the backlink was inserted, or highlighted from among other links, this change must be reset.

Considerations:
- potential many to one relationship, with the same reference cited multiple times.
- having reached the reference list, another reference might catch the user's attention, so avoid confusion of clickin the 'back link' of a different citation expecting to return to the user's own prevoius position and ended up somewhere related to that item instead.
- a11y: make sure changes work for keyboard navigation and screen reader use, particularly having navigated does 'read on' work for the change of location?
- a11y: consider the impact of 'movement' in regards to both automatic scrolling and visual highlighting.

StephDriver:
New Behaviour:

1. backlinks are generated for all references on page load
2. backlinks have tooltip with the closest proceeding heading for the place linked from.
3. backlink `aria-label` is text from original link and heading as above (to make unique)
4. backlinks are not distracting.. Using similar syntax to update for endnote backlinks, but with handling of one to many with such numbered. The tooltip on mouse hover provides more information (see 2).
5. On activating an onpage link in an article (any, not just the backlinks) the scroll is now handled in the javascript so that the destination anchor is highlighted to 'draw attention' and screen reader/keyboard navigation also sets to that.
6. if the anchor is a block or heading, the attention is drawn to that
7. otherwise the attention is drawn to the containing block. 
8. this works with the table of contents too - drawing attention to the heading.
9. The scroll destination is offset from the top of the window - this means it is not hidden behind the OLH text sticky and lands in a more natural reading position.
10. `--draw-attention` colour is in the css as a variable, for easy customisation.  In the OLH theme, this is set to the `--primary-light-color`.

StephDriver:
Known Issues:

I have updated and debugged.  I have two remaining minor issues which I cannot resolve, but I think this may be mergeable as is, or if not, I'd value some insight in how to resolve them:

1. On Material, the draw attention highlighting works fine, except with the References list.  Every other type of in page link seems to work.  I cannot locate the cause of this.
2. On Clean, the scrolling is different, more jumpy less smooth.  Again I cannot work out why.

src/static/common/js/reversable-links.js
```diff
@@ -0,0 +1,196 @@
+let attentionTimeouts = new Map(); // Track timeouts per element
+
+const headerElements = ['h1, h2, h3, h4, h5, h6'];
+const blockElements = ['p', 'li', 'ul', 'ol',' div', 'section', 'article', 'aside', 'nav', 'header', 'footer', 'main'];
```
joemull, line 4: These lists are not formed the same way, with the first one being essentially a string. I *think* they are both being parsed because the CSS parsing is pretty flexible, but may be worth making them standard? Probably not related to the first bug you mention, but maybe?

src/static/common/js/reversable-links.js
```diff
@@ -0,0 +1,196 @@
...
+        // Only apply offset if element would scroll to top
+        let finalPosition = elementTop;
+        if (Math.max(0, elementTop) === elementTop) { 
+            finalPosition -= offset;
+        }
+        
+        window.scrollTo({
+            top: finalPosition,
+            behavior: 'smooth'
+        });
```
joemull, line 33: I don't know if this is worth it, but if it's easy, you could try using [scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) instead of window.scrollTo in case it allows you to get the smooth scroll behavior on clean, and avoid having to manually calculate the scroll position? My linter is also telling me pageYOffset is deprecated, for what it's worth. Only if it's easy, since it's fine not to scroll smoothly!

joemull (changes requested):
Works really nicely, and the tab index change is really nice for keyboard use.

Only two small things I had thoughts about, mentioned inline.

src/static/common/js/reversable-links.js
StephDriver: good spot!  I'll standardise and see what impact that has.

StephDriver:
I've done some further digging into Material and the references not being highlighted while everything else is:

- the `draw-attention` class is not added to reference items, so this is a js problem, not a css one.
- The click event to add it doesn't fire.

I conclude that there is some other js that is changing the click behaviour the links to the bibliography on material and that this is run after the js included in this PR.  But it only affects those links, as all other on the page seem to work fine.

I haven't tracked it down.

As our focus for A11y is the Clean theme, so spending more time fixing this is lower priority than moving on, I see two options:
1. put PR through as is.
2. remove draw attention functionality from Material theme.


thoughts?

src/static/common/js/reversable-links.js
StephDriver: I updated the `pageYOffset`.

I tried`scrollIntoView`, but it didn't make it smoother, and it lost the offset that was to ensure the destination wasn't hidden behind the sticky bar at the top, so I left the methods as they were.

joemull:
Thanks for checking into that--I think we should put this through as is!

joemull (approved):

ajrbyers (changes requested):
On OLH the draw attention color seems to match the default OLH link colour so you end up with this:

image

Perhaps attention color could be more subtle? A lighter color maybe?

StephDriver:
> Perhaps attention color could be more subtle? A lighter color maybe?

@ajrbyers - I did consider that - but the content is not expected to be read during the highlight, it only highlights to 'draw attention' and then fades away.  I thought it looked good picking up the theme colour - but we could just have the same yellow as on the other themes if you'd prefer.

PR #4796, opened 2025-06-09
dyslexia and text size features added to clean and material

### LINKED ISSUE #755: Social Media/Dyslexia Features across themes
We should make these features available on all themes as they are very useful.

StephDriver:
closes #755

### DIFF
--- a/src/static/common/js/text_readability.js
+++ b/src/static/common/js/text_readability.js
@@ -4,3 +4,8 @@ function resizeText(multiplier) {
   }
   document.body.style.fontSize = parseFloat(document.body.style.fontSize) + (multiplier * 0.2) + "em";
 }
+
+$("#dyslexia-mode").click(function(e) {
+    e.preventDefault();
+    return $('#article').toggleClass('dyslexia-friendly');
+});
--- a/src/themes/OLH/assets/js/app.js
+++ b/src/themes/OLH/assets/js/app.js
@@ -20,10 +20,6 @@ $(".search-toggle").click(function() {
   $(".global-search input").focus();
 });
 
-$("#dyslexia-mode").click(function(e) {
-    e.preventDefault();
-    return $('#article').toggleClass('dyslexia-friendly');
-});
 
 
 function kanbanInit() {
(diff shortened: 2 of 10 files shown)


StephDriver:
From design meeting,
- buttons by abstract, not bar.
- consistency with other buttons in theme
- check how this works on Mobile, especially in regards to longer 'abstract' headings.  Make sure don't overlap, expect them to be 'over the top' of the heading.
- make sure buttons that are on/off are clearly on or off (e.g. dyslexia mode).

StephDriver:
## Controls
I have had a rethink on the buttons.  There are different constraints in material and clean as to how they handle mobile and desktop such that two different solutions were needed.

For Material, on desktop I have a card at the top of the sidepane which is sticky and stacks above the sticky table of contents when that is reached.  On mobile, there is a separate card above the abstract, which remains sticky (as the table of contents is not shown on mobile, there is no need to stack there). 

For clean, as the desktop/mobile change is handled by reflow, placing the text controls in teh sidebar automatically placed them at the end of hte article in mobile, which is far too late!  So Instead I have placed them under the article banner image, but visually laid out on the far right hand side, as if at the top of the sidebar. 

## Javascript functions
I have completely rewritten the javascript so that it can handle the more complex nesting structure of Material.  I tried several simpler approaches, but none were sufficient. This affected all three buttons - as none of them were reaching all the required elements.  

The changes I've made make the resizing more repeatable (size down is now almost reversed by a size up without ending up losing the proportional relationship between elements, e.g. sizes of headings and paragraphs).  The difficulty here is calculating floats, which are then applied as font sizes which truncates them, so over repeats these drift apart.  I have placed limits on how small text or large text can become, which helps limit this drift.  And removes the problem of loss of proportional relationship once the minimum is reached by some elements while others could continue reducing until all are the same.  While my solution is not perfect, it seems unlikely many users will go up and down with enough repeats for the drift to be noticeable to the eye.  A perfect numerical solution would require a base size for all values to be stored which seems disproportionate.

ajrbyers (changes requested):
For material can we pop them in the box below rather than have them in a box of their own? It feels odd.

StephDriver:
rebased

StephDriver:
rebased

ajrbyers (approved):

PR #4798, opened 2025-06-09
Editors and others can view and manage new author metadata

### LINKED ISSUE #1676: Author Manager
Instead of displaying authors as part of Edit Metadata we should add a new difinitive page for managing accounts, authors and their frozen author records.

Page should:
- Display who is linked by ArticleAuthorOrder and allow you to edit them
- Display FAs and allow you to edit them
- Allow one way syncing from AAO model to FAs.

### LINKED ISSUE #4622: As an editor I want to edit ROR-based affiliations
## User story

As an editor I want to edit ROR-based affiliations in the Edit Metadata screen for frozen authors.

## Proposed solution
- [ ] Create a set of views based on the self-edit views that an editor, or section editor, or press manager can use to edit affiliations of frozen author records.
- [ ] Prevent editors, section editors, and press managers from being able to edit affiliations on just any account.

## Alternatives considered
- Allow editors, section editors, and press managers to edit account affiliations directly.

### LINKED ISSUE #4711: Unable to Add Existing Authors – Creates Blank Author Records
Sure! Here's the full bug report formatted as requested:

---

**Describe the bug**  
When editing the metadata of an article, there is no option to select or add existing authors. Attempting to add a new author causes the entire form to reset, resulting in a blank submission. This leads to the creation of an empty author record and an empty author list in the article metadata.

**To Reproduce**  
1. Log in to the Janeway admin dashboard.
2. Navigate to an existing article and click **Edit Metadata**.
3. Try to add an author.
4. The form reloads and creates a blank author entry.
5. Author list appears empty.

**Expected behavior**  
The system should allow selecting existing authors . The author should be saved and displayed correctly in the metadata section.

**Screenshots**  

https://prnt.sc/ZCg53svXXLc4

---

joemull:
Things to check in review:
- PII visibility
- navigation, back buttons, clarity of interfaces
- lingering pieces from old UIs that might be confusing
- opportunities to make it DRYer

* Closes #4622
* Closes openlibhums/bau#265
* Closes #4711
* Closes #1676


src/submission/views.py
```diff
@@ -230,7 +231,7 @@ def submit_info(request, article_id):
                     reverse(
                         'submit_authors',
                         kwargs={'article_id': article_id},
-                    ),
```
ajrbyers, line 233: Why was this removed?

src/submission/views.py
```diff
@@ -914,74 +875,79 @@ def edit_metadata(request, article_id):
...
+    """
+    article = get_object_or_404(
+        models.Article,
+        pk=article_id,
+        journal=request.journal,
+    )
+    new_author_form = forms.EditFrozenAuthor()
+    last_changed_author = None
+
+    if request.POST and 'add_author' in request.POST:
```
ajrbyers: Check for `request.POST` once to avoid clutter in this block.

src/submission/views.py
```diff
@@ -1323,11 +1256,11 @@ def get_context_data(self, *args, **kwargs):
...
         author = get_object_or_404(
             models.FrozenAuthor,
-            Q(pk=int(self.kwargs.get('author_id'))) &
-            Q(article=article) &
-            (Q(author=self.request.user) |
-            Q(article__owner=self.request.user))
+            pk=int(self.kwargs.get('author_id')),
+            article=article,
         )
+        if not author.can_edit(self.request.user):
```
ajrbyers, line 1270: This is used a few times across views. Any chance a reusable security decorator would be suitable?

src/submission/models.py
```diff
@@ -2104,6 +2107,38 @@ def owner(self):
         else:
             return None
 
+    def can_edit(self, user):
```
ajrbyers, line 2111: I think you should take another look over this method. The default position should always be False (this is a security principle: deny by default). return False at the end of the method outside of the if blocks.

Do an early check for self.article, if its not found the default return will be False.

```python
if self.article:
    if (check a thing is true):
        return True

return False
```

The rest of the logic can then probably be simplified into a single if statement.

src/templates/admin/elements/author_summary_table.html
```diff
@@ -0,0 +1,113 @@
...
+                {% endif %}
+              {% endif %}
+              {% if author.is_correspondence_author %}
+                <div class="padding-block-0-25">
+                  <span class="fa fa-comments"></span>
+                  <span>{% trans "Correspondence" %}</span>
+                </div>
+              {% endif %}
+            {% endif %}
+            {% if can_see_pii and any_editor and author.author %}
```
ajrbyers: I found viewing this in the submit summary page a bit odd. Maybe exclude it from that particular page?

src/templates/admin/elements/orcid_display.html
```diff
@@ -0,0 +1,9 @@
+{% load static %}
+
+<a href="{{ author.orcid_uri }}" target="_blank">
```
ajrbyers: This should check if orcid_uri is empty or you end up with a link to the current page. Its not likely to happen on live but here is an example where the ORCID is not valid.

image

src/templates/admin/elements/author_summary_table.html
```diff
@@ -0,0 +1,113 @@
...
+                Only display emails to non-editors
+                if they are OK to display.
+                It's fine to display all emails to submitting authors.
+              {% endcomment %}
+              {{ author.email|se_can_see_pii:article }}
+              {% if any_editor %}
+                {% if author.display_email %}
+                  <div class="padding-block-0-25">
+                    <span class="fa fa-eye"></span>
+                    <span>{% trans "OK to display" %}</span>
```
ajrbyers: We had a brief discussion about this wording. Consensus was this makes it sound like the editor needs to take an action to display this, can we update the wording so it more closely matches what this field does?

ajrbyers: @mauromsl suggests: "Displayed on article" / "Not displayed on article" or "hidden" / "public" as it affects more than just the display.

src/submission/logic.py
```diff
@@ -473,3 +473,120 @@ def add_author_from_search(search_term, request, article):
...
+                % {
+                    "author_name": author.full_name(),
+                    "email": author.email
+                },
+        )
+    return author
+
+
+def save_correspondence_author(request, article):
+    account = get_object_or_404(
```
ajrbyers, line 521: I'm not sure a 404 is best here, redirect with message?

src/utils/forms.py
```diff
@@ -24,6 +24,9 @@
 
 
 ENTITIES_MAP = (("&amp;", "&"), ("&gt;", ">"), ("&lt;", "<"))
+ORCID_REGEX = re.compile(
```
ajrbyers: This exists in utils.orcid:

https://github.com/openlibhums/janeway/blob/eaabc7a09baf3488f97ae1c1d6f347b6b4f60571/src/utils/orcid.py#L165-L172

src/submission/logic.py
```diff
@@ -473,3 +473,120 @@ def add_author_from_search(search_term, request, article):
...
+def add_new_author_from_form(request, article):
+    new_author_form = EditFrozenAuthor(request.POST)
+    author = None
+
+    frozen_email = request.POST.get("frozen_email")
+    if frozen_email:
+        try:
+            author = models.FrozenAuthor.objects.get(
+                Q(article=article) &
+                (Q(frozen_email=frozen_email) |
```
mauromsl: These fields are not part of a `unique_together` constraint, however we are using `.get` . do we need to handle a potential `MultipleObjectsReturned` exception?

src/submission/logic.py
mauromsl: I think semantically a 400/404 is correct, but it is true that given the nature of the request requiring a full page reload it will feel odd.

Can we turn this toggle into an ajax request, like we do elsewhere for these type of toggles (with the aim to replace it with htmx interactions soon)?

I think it'd make for a much greater user experience in the interim, while making the code way easier to follow across the edit MVT stack

src/submission/logic.py
```diff
@@ -473,3 +473,120 @@ def add_author_from_search(search_term, request, article):
...
+            frozen_author=author,
+        )
+    else:
+        return None
+
+
+def add_credit_role(request, article):
+    author = get_object_or_404(
+        models.FrozenAuthor,
+        pk=int(request.POST.get('author_pk')),
```
mauromsl: if we are not using a Django form, we should take care of sanitizing this input before we hand it over to the ORM and raise validation errors as required (rather than raise unexpected typ errors)

src/submission/logic.py
```diff
@@ -473,3 +473,120 @@ def add_author_from_search(search_term, request, article):
...
+                    "role": record.get_role_display(),
+                },
+        )
+    return author
+
+
+def remove_credit_role(request, article):
+    record = get_object_or_404(
+        models.CreditRecord,
+        pk=int(request.POST.get('credit_pk')),
```
mauromsl: as above, sanitize before the ORM call /type cast

src/submission/models.py
```diff
@@ -1226,6 +1226,9 @@ def non_correspondence_authors(self):
         else:
             return self.author_accounts
 
+    def is_submitted(self):
+        return self.stage != STAGE_UNSUBMITTED
```
mauromsl: Should this return `True` for articles that are submission complete? how about imported articles? I'd avoid establishing this API until these are resolved, and since this is esentially a single statement, I doubt the usefulness of its use is worth more than the potential misuse it could introduce

src/submission/models.py
mauromsl: agree

src/submission/models.py
```diff
@@ -2323,6 +2372,7 @@ def get_or_snapshot_if_email_found(cls, email, article):
         except cls.DoesNotExist:
             try:
                 account = core_models.Account.objects.get(email=email)
+                account.add_account_role('author', article.journal)
                 author = account.snapshot_as_author(article)
```
mauromsl: seeing the following line, I wonder if we could delegate this to `Account.snapshot_as_author` method instead, so we can avoid requiring this step elsewhere

src/submission/models.py
```diff
@@ -2344,6 +2399,7 @@ def get_or_snapshot_if_orcid_found(cls, orcid, article):
                 account = core_models.Account.objects.get(
                     orcid__contains=orcid,
                 )
+                account.add_account_role('author', article.journal)
```
mauromsl: as above

src/submission/views.py
```diff
@@ -599,13 +565,15 @@ def delete_frozen_author(request, article_id, author_id):
...
-                next_url,
-                kwargs={'article_id': article_id}
+        if next_url:
+            return redirect(next_url)
+        else:
+            return redirect(
+                reverse(
+                    'submit_authors',
+                    kwargs={'article_id': article_id}
+                )
```
mauromsl, line 626: Just a question since I've seen this a couple times: why does this change need unwinding / what issue is it trying to address?

mauromsl (changes requested):
Added some comments inline, they are mostly minor. I've recommended moving one of the use cases to ajax ( correspondence author toggles), since seeing the amount of refactoring and multibranch code required to have a second use case has been quite the event. When we consider the amount user experience pain it introduces, combined with the code quality, makes me think it is worth pursuing, as it won't take much effort

src/submission/views.py
joemull: Oh just consistency with the other ways we usually call redirect--I noticed it when doing a text search.

src/submission/views.py
joemull: Yes, I can replace some of these with a decorator.

src/templates/admin/elements/author_summary_table.html
joemull: I think you must have had the editor role when testing the author interface? If you only had the author role, you would not see it.

ajrbyers: > I think you must have had the editor role when testing the author interface? If you only had the author role, you would not see it.

Yeah I was aware of that during test, its still _very_ odd to see.

src/submission/logic.py
joemull: I think it will be cleaner to do a filter and then get the first one. The UI will announce that the author already exists--so if there is something wrong with the existing author record, the user can remove (delete) it and add a new one.

src/submission/models.py
joemull: OK, how about I change it to `is_unsubmitted`? I'd like a method since it's easier to use in templates, and I want to avoid hard-coding the stage name as a string all over the place.

src/submission/views.py
joemull: In this case, the logic was wrong--it should not have tried to redirect users **with** the next URL, because the intended user action was unsuccessful. The user needs to be sent back to the starting point for the intended action, so they can take a different action.

So the best thing to do is to redirect them **to** the `next_url` with the error message. For authors, this place would be the **Author Info** page in the submission workflow. For editors, this would be the **Edit author metadata** page.

We keep 'submit_authors' as a backup in case the referring page did not include a next URL parameter. For editors, this will be wrong, but it is better to have a backup than none.

src/submission/models.py
joemull: I'll see what I can do.

As for returning False by default, I modeled it on this one, which returns True at the end. Maybe worth reconsidering the logic of that one too?

https://github.com/openlibhums/janeway/blob/e319a3d937f0e741e91559c67c3a15ac85cc75ef/src/submission/models.py#L1518-L1541

ajrbyers: Yes it probably should be changed. It's ancient code from the fabled times of Mercurial.

src/submission/models.py
mauromsl: good shout, `self.stage == STAGE_UNSUBMITTED` is more precise for our use case, so I'd be happy with `is_unsubmitted`

src/templates/admin/elements/author_summary_table.html
joemull: I'll add a check for `not is_unsubmitted`.

src/submission/logic.py
joemull: This isn't really a toggle as you suggest, and it does imply quite some effort. To take the correspondence author change for example, it requires reloading the full author list, because the current correspondence author needs to be no longer marked as such, and the new one needs to be marked as correspondence author. Same for author order--needs the full list reloaded.

But I've gone ahead and created a partial view for the current author list that handles three actions: changing the corr author, reordering authors, and adding and removing credit roles. This makes the UX waaaaaay smoother. It was not simple--it required a new template tag pattern to keep next urls working as expected, as well as more custom JavaScript than you might expect to make sure the system status was clear to the user. I'm also not sure it was worth the time to write this if HTMX is around the corner.

But it's working pretty well now, and we have slightly simpler branching in the views.

src/submission/models.py
```diff
@@ -2104,6 +2108,36 @@ def owner(self):
...
+            elif not user.is_anonymous and user.is_editor(
+                request=None,
+                journal=self.article.journal,
+            ):
+                return True
+            elif self.owner == user and self.article.is_unsubmitted():
+                # FrozenAuthor.owner is a property that considers both
+                # FrozenAuthor.author and FrozenAuthor.article.owner.
+                return True
+        else:
```
ajrbyers: Drop the else here so that False is always returned if there is no other return.

ajrbyers (changes requested):
One change, looks great. The UX feels much better, thanks for putting in the effort there.

ajrbyers (approved):

src/core/logic.py
```diff
@@ -68,6 +68,34 @@ def reverse_with_next(url_name, next_url, args=None, kwargs=None):
...
+        reversed_url = reverse(url_name)
+
+    if not full_page_path:
+        return reversed_url
+
+    final_url = utils_logic.add_query_parameters_to_url(
+        reversed_url,
+        {'full_page_path': full_page_path},
+    )
+    return final_url
```
mauromsl: are we going to have one of these for every possible query parameter?

I reckon something like the following would suffice

```py
def reverse_with_query(name, args=None, kwargs=None, query_params=None):
home')})
    reversed = reverse(name, args=args, kwargs=kwargs)

    if query_params:
        reversed = add_query_parameters_to_url(reversed, query_params)
    return reversed
```

Also, rather than silently swallow `kwargs` when they are provided alongside `args` ,let `reverse` determine behaviour and outcome

src/core/templatetags/next_url.py
```diff
@@ -24,6 +24,33 @@ def url_with_return(context, url_name, *args, **kwargs):
...
+    """
+    A template tag to use when reversing a URL for a partial view
+    before accessing it with JavaScript. This is what it makes possible:
+      1. passes full_page_path to the partial view as a GET parameter
+      2. the partial view puts it into the context dictionary
+      3. the partial view's template calls url_with_return as normal
+      4. url_with_return checks for it and uses it rather than its own request.get_full_path
+    Without this pattern, url_with_return would encode a next parameter
+    that points to the partial view, not the full page, resulting in the user
+    loading up a bare HTML snippet with no nav or CSS on completing a task.
```
mauromsl: There are more comments than code for a function that takes its main argument from context rather than from signature. I think you might have been typing out your train of thought rather than providing useful details. It might also be more self explanatory with my proposed change

```py
return reverse_with_query(
    url_name,
    args=args,
    kwargs=kwargs,
    query_params={"full_page_path": request.get_full_path()},
)

File: src/core/views.py [hunk 20]
```diff
@@ -724,7 +724,7 @@ def affiliation_update_from_orcid(request, how_many='primary'):
             _("No affiliations were found on your public ORCID record "
               "for ID %(orcid_id)s. "
               "Please update your affiliations on orcid.org and try again.")
-                % {'orcid_id': request.user.orcid },
+                % {'orcid_id': clean_orcid_id },
```
mauromsl: should this be `cleaned_orcid` ?

src/security/decorators.py
```diff
@@ -740,6 +739,11 @@ def wrapper(request, *args, **kwargs):
 
 
 def article_edit_user_required(func):
+    raise DeprecationWarning('Use user_can_edit_article instead.')
+    return user_can_edit_author(func)
```
mauromsl: shouldn't this be `user_can_edit_article` ?

src/submission/views.py
```diff
@@ -387,7 +345,63 @@ def submit_authors(request, article_id):
 
 
 @login_required
-@article_edit_user_required
+@require_POST
+@user_can_edit_article
+def edit_current_authors(request, article_id):
+    """
+    Async partial view accessed via JavaScript.
```
mauromsl: I would refrain from using the word "Async" here, nothing about this view is asynchronous. It could be confusing when we start adding actual async views to Janeway.

src/submission/logic.py
```diff
@@ -473,3 +478,129 @@ def add_author_from_search(search_term, request, article):
...
+    except TypeError:
+        raise ValidationError('Invalid author ID submitted')
+    author = get_object_or_404(
+        models.FrozenAuthor,
+        pk=author_id,
+        article=article,
+    )
+    credit_form = CreditRecordForm(request.POST)
+    if credit_form.is_valid() and author:
+        record = credit_form.save()
```
mauromsl, line 571: There is no server-side validation to protect against adding the same credit role twice to the same author. a 500 is raised here, might be worth handling this in the `CreditRecordForm` as part of its validation

```
IntegrityError at /edit/article/2891/current-authors/

duplicate key value violates unique constraint "submission_creditrecord_frozen_author_id_role_abd0ca00_uniq"
DETAIL:  Key (frozen_author_id, role)=(7866, data-curation) already exists.
```

joemull: Right, thanks for catching. Because I remove choices that correspond to existing roles, I just need to remember to pass in the frozen_author to the form on initialization, as the form will then catch and error out any value that is not one of the provided choices.

src/core/logic.py
joemull: Sure, I'm happy to go ahead and abstract this a bit more.

src/core/logic.py
```diff
@@ -68,30 +62,24 @@ def reverse_with_next(url_name, next_url, args=None, kwargs=None):
...
-    if args:
-        reversed_url = reverse(url_name, args=args)
-    elif kwargs:
-        reversed_url = reverse(url_name, kwargs=kwargs)
-    else:
-        reversed_url = reverse(url_name)
+    reversed_url = reverse(url_name, args=args, kwargs=kwargs)
 
-    if not full_page_path:
+    if not any(query_params.values()):
```
mauromsl: The default value for the `query_params` is `None`, so I think this could technically raise our dreaded `AttributError: Nonetype object has no attribute "values"` exception.

Any reason to flip the logic and call `.values()` explicitely over what was proposed during review?

mauromsl: Another concern I have with this approach, is that "falsey" values would be silently discarded such as `0` or `[]`

joemull: When designing the next URL patterns, I found that it was easy for view or template authors to make a mistake with data types and corrupt the value of the next URL with falsey non-string values, especially with layers of percent-encoding and decoding happening, and downstream logic in views can't detect this easily. It just gets the stringified version of whatever is in request.GET.get('next'), and it doesn't make sense to also check whether that value is a sensible string. The effect is that the user can be provided with a garbled link rather than a sensible backup. That is why, with both these functions, I have tried to avoid passing falsey values.

But on balance, I think we have to lean on good use of the function, rather than safeguard against bad use, since it could lead to unwanted side effects, as you suggest. It opens the door to regressions that go unnoticed, but I will have to live with that until we have better front-end testing architecture.

PR #4802, opened 2025-06-10
front of house link a11y

### LINKED ISSUE #4221: Contextual link fixes arising from a11y audit

StephDriver:
Close #4221

### DIFF
--- a/src/themes/OLH/templates/journal/article.html
+++ b/src/themes/OLH/templates/journal/article.html
@@ -2,6 +2,7 @@
 {% load static %}
 {% load hooks %}
 {% load i18n %}
+{% load dates %}
 
 {% block page_title %}{% if article.preprint %}{% trans "Preprint" %} {% endif %}{% trans "Article" %}{% endblock %}
 
@@ -303,7 +304,7 @@
                                             {% if article.is_published or proofing %}
                                                 <p id="article_date_published">
                                                     {% trans "Published on" %} <br>
-                                                    {{ article.date_published|date:"Y-m-d" }}
+                                                    {{ article.date_published|date_human }}
                                                 </p>
                                             {% endif %}
                                         </div>
@@ -439,10 +440,10 @@
                                  <li>{% trans "Original ISSN" %}: {{ article.ISSN_override }}</li>
                                  {% endif %}
                                  {% if journal_settings.article.display_date_submitted and article.date_submitted %}
-                                 <li>{% trans "Submitted on" %}: {{ article.date_submitted|date:"Y-m-d" }}</li>
+                                 <li>{% trans "Submitted on" %}: {{ article.date_submitted|date_human }}</li>
                                 {% endif %}
                                 {% if journal_settings.article.display_date_accepted and  article.date_accepted %}
-                                <li>{% trans "Accepted on" %}: {{ article.date_accepted|date:"Y-m-d" }}</li>
+                                <li>{% trans "Accepted on" %}: {{ article.date_accepted|date_human }}</li>
                                 {% endif %}
                              </ul>
                          </div>
--- a/src/themes/OLH/templates/journal/article.html
+++ b/src/themes/OLH/templates/journal/article.html
@@ -324,7 +324,7 @@
                                         <div class="large-6 columns" data-equalizer-watch>
                                         {% endif %}
                                             <p>{% trans "License" %}</p>
-                                            <a class="has-tip scroll-link" rel="license" data-allow-html="true" id="license" title="{{ article.license.text|safe }}" data-tooltip="2cc6nw-tooltip" aria-haspopup="true" data-fade-out-duration="1000" title="" aria-describedby="pyltep-tooltip" data-yeti-box="pyltep-tooltip" data-toggle="pyltep-tooltip" data-resize="pyltep-tooltip" data-events="resize" href="{{ article.license.url }}"><p>{{ article.license.name }}</p></a>
+                                            <a class="has-tip scroll-link" rel="license" data-allow-html="true" id="license" title="{{ article.license.text|safe }}" data-tooltip="2cc6nw-tooltip" aria-haspopup="true" data-fade-out-duration="1000" aria-describedby="pyltep-tooltip" data-yeti-box="pyltep-tooltip" data-toggle="pyltep-tooltip" data-resize="pyltep-tooltip" data-events="resize" href="{{ article.license.url }}"><p>{{ article.license.name }}</p></a>
                                         </div>
                                     </div>
                                 </div>
--- a/src/themes/OLH/templates/repository/preprint.html
+++ b/src/themes/OLH/templates/repository/preprint.html
@@ -3,6 +3,7 @@
 {% load hooks %}
 {% load i18n %}
 {% load foundation %}
+{% load dates %}
 
 {% block title %}{{ preprint.title }}{% endblock %}
 
@@ -124,8 +125,8 @@
                     {% endif %}
                     <h2>{% trans "Metadata" %}</h2>
                     <ul>
-                        <li>{% trans "Published" %}: {{ preprint.date_published|date:"Y-m-d" }}</li>
-                        <li>{% trans "Last Updated" %}: {{ preprint.current_version.date_time|date:"Y-m-d" }}</li>
+                        <li>{% trans "Published" %}: {{ preprint.date_published|date_human }}</li>
+                        <li>{% trans "Last Updated" %}: {{ preprint.current_version.date_time|date_human }}</li>
                         {% if preprint.license %}<li>
                           {% trans "License" %}: <a href="{{ preprint.license.url }}" data-toggle="tooltip"
                                                      title="{{ preprint.license.text }}">{{ preprint.license.name }}</a>
@@ -144,7 +145,7 @@
                             <li><a href="{% url 'repository_file_download' preprint.id version.file.id %}" >
                                 <i aria-hidden="true" class="fa fa-download">&nbsp;</i>
                                 <span class="sr-only">{% trans 'Download' %}</span>
-                                {% trans 'Version' %} {{ version.version }} - {{ version.date_time|date:"Y-m-d" }}
+                                {% trans 'Version' %} {{ version.version }} - {{ version.date_time|date_human }}
                             </a></li>
                             {% empty %}
                             <li>{% trans "This is the only version of the preprint." %}</li>
--- a/src/themes/clean/templates/core/news/index.html
+++ b/src/themes/clean/templates/core/news/index.html
@@ -1,5 +1,6 @@
 {% extends "core/base.html" %}
 {% load truncate %}
+{% load dates %}
 
 {% block title %}{{ journal_settings.news.news_title }}{% endblock title %}
 
@@ -12,7 +13,7 @@
         {% for item in news_items %}
             <div class="col-md-12">
                 <h4>{{ item.title|safe }}</h4>
-                <h6>{{ item.byline }} on {{ item.posted|date:"Y-m-d" }}</h6>
+                <h6>{{ item.byline }} on {{ item.posted|date_human }}</h6>
                 <p>{{ item.body|truncatewords_html:50|safe }}</p>
                 <a href="{% url 'core_news_item' item.pk %}">{% trans 'Read More' %}</a>
                 <hr/>
--- a/src/core/templatetags/dates.py
+++ b/src/core/templatetags/dates.py
@@ -2,6 +2,7 @@ from datetime import datetime, timedelta
 
 import pytz
 from django import template
+from django.conf import settings
 from django.template.exceptions import TemplateSyntaxError
 from django.utils import timezone, formats
 from django.utils.translation import gettext as _
@@ -54,16 +55,16 @@ def offset_date(context, days=0, input_type="date"):
 @register.filter(name='date_human', is_safe=True)
 def date_human(value):
     """Convert a date to a human readable Day Month(text) Year format e.g. 3 January 2025"""
-    if value == "":
-        logger.warning("date_human filter received no value - expected a datetime object")
-        return ""
-    elif isinstance(value, datetime):
+    if isinstance(value, datetime):
         return formats.date_format(
             value,
             format="j F Y",
             use_l10n=True,
         )
     else:
-        raise TemplateSyntaxError(
-            "The value filtered by `date_human` must be a `datetime.datetime`"
-        )
+        error_message = "The value filtered by `date_human` must be a `datetime.datetime`"
+        if settings.DEBUG:
+            raise TemplateSyntaxError(error_message)
+        else:
+            logger.error(error_message)
+            return ""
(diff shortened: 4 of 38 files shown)


StephDriver:
Work in Progress:
- [ ] create filter for external link, as a base for other link templates.
- [ ] find links (642 across 163 files in Themes)
- [ ] work through all {# LinkA11y #} tagged links in the html, and add new link filter templates as needed.
- [ ] work through js links and update syntax individually to match the new link type filter templates.
- [ ] consider links in common templates that are used in Front of House.

StephDriver:
Syntax Notes:

``` django
{% load render_string links %}
```

```django
 {% render_string '<a ...... </a>' as link_html %}
 {{ link_html|external_link:"contextual" }}
```

```django
 {% render_string '<a aria-labelledby=.......... </a>' as link_html %}
 {{ link_html|external_link }}
```

StephDriver:
## Link Syntax

After much discussion, we have decided not to use templates, and instead to review and update each link in situ.

### Contextual v ARIA
If an external or internal link is contextual, then no aria markup is required.  Otherwise, add an aria-labelledby, aria-describedby or aria-label.

### External Links

```
<a  class="external-link" target="_blank" >

    inner html content

    <i aria-hidden="true" class="fa fa-external-link" style="vertical-align: super; font-size: 0.8em;"
        title="{% trans 'External link, opens in new tab.' %}"> </i>
    <span class="sr-only">{% trans "External link, opens in new tab." %}</span>

    any closing inner html tags

</a>
```

### Read more Links

### Internal Links

### Actions:
If an action rather than navigation, then markup as `{# ButtonA11y #}` for individual consideration at the end.

StephDriver:
## Buttons

for navigation dropdowns, use `dropdown-button` css to get original style.

### Close alert messages example
```django
        {% for message in messages %}
            <div class="alert alert-{{ message.tags }} alert-dismissable">
                <button class="close" data-dismiss="alert" aria-describedby="message-{{ message.id }}">
                    <span aria-hidden="true">&times;</span>
                    <span class="sr-only">{% trans 'Close message' %}</span>
                </button>
                <span id="message-{{ message.id }}">{{ message|linebreaksbr }}</span>          
            </div>
        {% endfor %}
```

src/themes/OLH/templates/journal/article.html
```diff
@@ -302,19 +347,24 @@ <h2>
                             {% include "elements/journal/share.html" %}
                             <div class="section">
                                 {% if article.is_published or proofing %}
-                                    <h2>{% trans "Downloads" %}</h2>
+                                    <h2>{% trans "Files" %}</h2>
```
mauromsl, line 337: This change doesn't seem to be related to the accessibility changes? I think downloads was a good fit + it has a good translation + its translation carries more meaning than its translation under many contexts. I think its best to stick to the purpose of the issue and avoid spurious changes.

src/themes/OLH/templates/repository/preprint.html
```diff
@@ -128,13 +128,28 @@ <h2>{% trans "Metadata" %}</h2>
...
+                                title="{{ preprint.license.text }}"
+                                target="_blank"
+                                >
+                                {{ preprint.license.name }}
+                                {% include "elements/icons/link_external.html" %} 
+                            </a>
                         {% endif %}
                         </li>
-                        <li>Subjects: {% include "common/repository/subject_display.html" %}</li>
+                        <li>{% trans "Subjects: " %} {% include "common/repository/subject_display.html" %}</li>
```
mauromsl: Avoid trailing whitespace inside translation target.

src/themes/OLH/templates/repository/preprint.html
```diff
@@ -128,13 +128,28 @@ <h2>{% trans "Metadata" %}</h2>
...
                         {% if preprint.keywords.all %}
-                          <li>Keywords: {% for keyword in preprint.keywords.all %}{{ keyword.word }}{% if not forloop.last %}, {% endif %}{% endfor %}</li>
+                            <li>{% trans "Keywords: " %}
```
mauromsl: Avoid trailing whitespace inside translation target.

src/themes/clean/templates/core/login.html
```diff
@@ -44,8 +44,8 @@ <h2 class="text-center">{% trans "Log in with your account" %}</h2>
                                         class="btn btn-primary btn-block">{% trans "Log in" %}</button>
                             </p>
                         </form>
-                        <p class="text-center"><a
-                                href="{% url 'core_get_reset_token' %}">{% trans "Forgotten your password?" %}</a>
+                        <p class="text-center">
+                            <a href="{% url 'core_get_reset_token' %}">{% trans "Forgotten your password?" %}</a>
```
mauromsl: Try to avoid formatting changes that are not related to the issues we are addressing, as they can make the merge process more complicated than it should. If we want to reformat the HTML, we should look at doing it within its own PR and with a deterministic process (e.g a linter)

src/themes/clean/templates/core/news/index.html
```diff
@@ -12,10 +12,10 @@ <h1>{{ journal_settings.news.news_title }}</h1>
         </div>
         {% for item in news_items %}
             <div class="col-md-12">
-                <h2>{{ item.title|safe }}</h2>
+                <h2 id="news-item-{{ item.pk }}">{{ item.title|safe }}</h2>
                 <p class="byline">{{ item.byline }} on {{ item.posted|date_human }}</p>
                 <p>{{ item.body|truncatewords_html:50|safe }}</p>
-                <a href="{% url 'core_news_item' item.pk %}">{% trans 'Read More' %}</a>
+                <a href="{% url 'core_news_item' item.pk %}" aria-describedby="news-item-{{ item.pk }}">{% trans 'Read more' %}</a>
```
mauromsl: If changing the key to this translation target we need to:
 - Update the key in the translation files
 - Update the translations so they match the default case

src/themes/clean/templates/elements/journal/issue_list.html
```diff
@@ -1,13 +1,13 @@
 <ul>
     {% for issue in issues %}
         <li>
-            <a href="{% url 'journal_issue' issue.id %}" 
-                aria-label='{{ issue.display_title_a11y }} ({{ issue.articles.all.count }} {% trans "items" %})' 
+            <a href="{% url 'journal_issue' issue.id %}"
+                aria-label='{{ issue.display_title_a11y }} ({{ issue.articles.all.count }} {% trans "items" %})'
```
mauromsl: Try to avoid formatting changes that are not related to the issues we are addressing.

src/themes/clean/templates/elements/journal/issue_list_by_decade.html
```diff
@@ -6,7 +6,9 @@
 <ul class="list-inline">
     {% for decade, issues in issues_by_decade.items %}
         <li class="list-inline-item">
-            <a href="#decade-{{ decade }}">{{ decade }}</a>
+            <a href="#decade-{{ decade }}">
+                {{ decade }}
+            </a>
```
mauromsl: Try to avoid formatting changes that are not related to the issues we are addressing

src/themes/clean/templates/elements/journal/issue_sidebar.html
```diff
@@ -6,8 +6,8 @@
             <h2>{% trans 'Downloads' %}</h2>
             <p>
                 <a href="{% url 'journal_issue_download_galley' issue.pk issue.galley.pk %}">
-                    <i aria-hidden="true" class="fa fa-download"></i>&nbsp
-                    {% trans 'Download Issue' %} (PDF)
+                    {% trans 'Issue' %} (PDF)
+                    {% include "elements/icons/link_download.html" %}
```
mauromsl, line 10: Isn't "Download Issue" a better title here? Even if redundant with the icon, it gives a more clear prompt to translators of what the text means.

src/themes/clean/templates/elements/journal/issue_sidebar.html
```diff
@@ -20,7 +20,7 @@ <h2>
                 {% for issue in issues %}
                 {% if issue.issue_type.code == 'issue' %}
                     <li><a href="{% url 'journal_issue' issue.id %}" aria-label='{{ issue.display_title_a11y }}'>
-                        {{ issue.display_title }} 
+                        {{ issue.display_title }}
```
mauromsl: Try to avoid formatting changes that are not related to the issues we are addressing

src/themes/clean/templates/journal/full-text-search.html
```diff
@@ -38,17 +38,23 @@ <h1>{% trans 'Article Search Results' %}</h1>
                                 {% include 'elements/forms/errors.html'  with form=form%}
                                 {% bootstrap_form form %}
                             {% else %}
-                                {% trans "You are currently browsing by keyword" %}. <a href="{% url 'search' %}">{% trans 'Search for an article' %}.</a>
+                                {% trans "You are currently browsing by keyword." %}
```
mauromsl: If changing the key to this translation target we need to:
 - Update the key in the translation files
 - Update the translations so they match the new default punctuation

mauromsl (changes requested):
Thanks @StephDriver what a monumental piece of work, well done!

I've added a few comments regarding unrelated changes / changes that have an impact on translations.

I would avoid, if possible, changes to translation keys that are not directly related to accessibility (even if those changes are desireable). I think we should address them separately as they could potentially cause:
 - Difficulty to resolve merge conflicts.
 - Changes to pre-existing translation work.
 - Overhead in code review
 - Personal preference / inconsistent changes (e.g. Change the casing of some links like "Read More" to "Read more" but not "Download Issue" to "Download issue")
 - Changes to translation keys that become more ambiguous for our translators (e.g. "Download Issue" to "Issue"), even while being redundant with the new icons.
 
 Also, best to avoid formatting changes to HTML code that does not relate at all to the issues. It over complicates the merge process and could lead to more inconsistencies / bad merge resolutions. Ideally we want to resolve these formatting inconsistencies on their own

src/themes/OLH/templates/journal/article.html
mauromsl: Just realised it is to avoid the View/Download interaction below, so please ingore this one

src/themes/clean/templates/elements/journal/issue_sidebar.html
StephDriver: the download icon template is semantic, so it already has the word download in it.   So if I didn't change line 10 this would read "Download issue (download.)"

src/themes/clean/templates/core/login.html
StephDriver: Good catch - I think this one is from the way I worked - by tagging links to identify them all, then removing each tag as I reviewed that link. This file shouldn't have ended up being touched at all because it is deprecated so I must have made a mistake when removing it's tags.  I'll change it back and rebase so that it doesn't end up as part of this branch at all.

src/themes/clean/templates/core/news/index.html
StephDriver: On looking into this, every single PO file had it as "Read More" so I'm going to standardise the English to the same!

StephDriver:
There was a team discussion about the styling of the icons.  Ideally we would have wanted them to both recede, and be underlined as part of the link - but this doesn't seem possible.  The icon will only underline when the css targets it directly, or specifically told to inherit the text-decoration. But as the icon has a slightly lower opacity in order to subtly recede, the underlining ends up a different colour to the rest of the link.  Styling an enclosing element has not fixed this, as the icon will default to `text-decoration: none`. The only way around I could think of was to use a bottom border instead of text-decoration, but this then would look quite different from all other links, and so all would need to be restyled to have their underlining suppressed, and such is likely to conflict with custom css.

TLDR. the path of least waste of time was to have links as normal, and the icon will be next to it, receded but not be underlined but is clickable as part of that link so that the behaviour remains intuitive.

mauromsl (approved):

PR #4804, opened 2025-06-11
Add arXiv identifier choice

gamboz:
Add more identifier types:
- arxiv id, useful to "link" papers to https://arxiv.org preprints
- preprint id, useful to "link" papers to internal/non-Janeway preprint repositories 
- proton id from [IOP](https://iopscience.iop.org/)'s production/publication system "Proton"


src/identifiers/models.py
```diff
@@ -29,6 +29,8 @@
     ('doi', 'DOI'),
     ('uri', 'URI Path'),
     ('pubid', 'Publisher ID'),
+    ('arxiv', 'ArXiv ID'),  # https://arxiv.org/
+    ('preprintid', 'Preprint ID'),
```
gamboz: The `'preprintid'` choice is something that is useful when migrating papers from other systems, however please consider it a "weak desiderata" :slightly_smiling_face:

ajrbyers (changes requested):
Can you find the last time this field was altered in a migration and add these choices into the migration so that they're in the migration graph.

gamboz:
I've moved this to "draft" because I'll also add a part to send the arXiv id to crossref

E.g.
```xml
<program xmlns="http://www.crossref.org/relations.xsd">
     <related_item>
       <intra_work_relation relationship-type="hasPreprint" identifier-type="arxiv">2604.12345</intra_work_relation>
     </related_item>
  </program>
 ```

PR #4808, opened 2025-06-12
fix: #4807 material fig table downloads

### LINKED ISSUE #4807: material - missing figure and table downloads
found while updating link a11y that the figure and table downloads were not always appearing on material theme article page. 

cause - the selection of divs for the download functions were not matching the xsl that defined those divs.

e.g.

``` javascript
var figs = $("div[id^=F]");
```

when figures had ids in format `fig1`

similar for tables.

StephDriver:
Close #4807

### DIFF
--- a/src/themes/material/assets/material.js
+++ b/src/themes/material/assets/material.js
@@ -1,12 +1,12 @@
 function figure_downloads() {
-	var figs = $("div[id^=F]");
+	var figs = $("div.fig-inline-img-set");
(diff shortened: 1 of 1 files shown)


StephDriver:
I have changed the selectors to match the classes that are defined in default.xsl, rather than match by initial letter.

mauromsl (approved):

ajrbyers (approved):

PR #4815, opened 2025-06-18
Adds conditional accept and desk rejection email templates

### LINKED ISSUE #1827: Allow for Additional Templates
Right now, there’s really only one template to invite a review and only one template for acceptances. Is there possible for more? For example, when inviting reviewers to evaluation a revision, the language often changes a bit. Also, sometimes submission are accepted conditionally when the revisions are minor so not really a R&R but not really an accept … likewise, a desk reject is different from a reject based on peer review. 

Allow Editors to make additional templates and link them to the workflow so they can select which kind of template to use.

ajrbyers:
This PR addresses the requirements of https://github.com/openlibhums/janeway/issues/1827 but not in the way that particular issue proposes. It has been open for nearly 5 years and the complexity of the proposal makes it impractical. We have most often been asked for:

- An option for conditional accept ✅ 
- An option for a separate desk rejection template ✅ 

This PR adds:
- A new template for the revision workflow that covers the conditional acceptance path, this template will work in the draft decisions workflow
- A new template for desk rejection, this template is used when an editor rejects an article that is still unassigned, otherwise the standard review rejection template is used

I don't think there is any need to alter the revision template as it stands. If editors want conditional text based on major vs minor revisions this can be achieved already with conditionals in the template.

Closes #1827

### DIFF
--- a/src/utils/install/journal_defaults.json
+++ b/src/utils/install/journal_defaults.json
@@ -5533,7 +5533,7 @@
             "type": "rich-text"
         },
         "value": {
-            "default": "<p>Dear {{ article.correspondence_author.full_name|se_can_see_pii:article }},</p><p>We are pleased to inform you that your article has been conditionally accepted for publication, subject to the successful completion of the required revisions.</p><p>Details of the changes required are below:</p><p>{{ revision.editor_note }}</p><p>You can view and respond to the revision requests here: {{ do_revisions_url }}</p><p>Please complete your revisions by {{ revision.date_due }}.</p><p>We look forward to receiving your updated manuscript.</p><p>Regards,<br>{{ request.user.signature|safe }} </p>\n"
+            "default": "<p>Dear {{ article.correspondence_author.full_name|se_can_see_pii:article }},</p><p>We are pleased to inform you that your article has been conditionally accepted for publication, subject to the successful completion of the required revisions.</p><p>Details of the changes required are below:</p><p>{{ revision.editor_note|safe }}</p><p>You can view and respond to the revision requests here: {{ do_revisions_url }}</p><p>Please complete your revisions by {{ revision.date_due }}.</p><p>We look forward to receiving your updated manuscript.</p><p>Regards,<br>{{ request.user.signature|safe }} </p>\n"
         },
         "editable_by": [
             "editor",
@@ -5571,7 +5571,7 @@
             "type": "char"
         },
         "value": {
-            "default": "Article Conditional Acceptance"
+            "default": "Conditional Acceptance of Your Article"
         },
         "editable_by": [
             "editor",


src/review/models.py
```diff
@@ -63,6 +63,16 @@ def review_decision():
...
+def draft_decision_choices():
+    return (
+        (ED.ACCEPT.value, 'Accept Without Revisions'),
+        (ED.MINOR_REVISIONS.value, 'Minor Revisions Required'),
+        (ED.MAJOR_REVISIONS.value, 'Major Revisions Required'),
+        (ED.CONDITIONAL_ACCEPT.value, 'Conditional Accept'),
+        (ED.DECLINE.value, 'Reject'),
+    )
+
+
```
mauromsl: I don't get why we need to split these,are we trying to avoid presenting Conditional accept as an option on non-draft reviews? how would the transition from a draft decision of `'Conditional Accept'` to a review decision happen? It looks to me like it would insert `''Conditional Accept'` as the review decision on the review request, which would go through (there is no DB level check) but could present problems editing review requests via admin / Django forms (since it is not part of the review request choice list).

src/review/views.py
```diff
@@ -2475,7 +2477,7 @@ def draft_decision_text(request, article_id):
             author_review_url=author_review_url,
         )
 
-    elif decision in {ED.MINOR_REVISIONS.value, ED.MAJOR_REVISIONS.value}:
+    elif decision in {ED.MINOR_REVISIONS.value, ED.MAJOR_REVISIONS.value, ED.CONDITIONAL_ACCEPT.value}:
```
mauromsl: This is just a personal opinion: we are overloading a bit the meaning of `get_revision_request_content`, especially because we have to couple the two functions in order to load a different email template for it. I get that the context is the same for both **currently** but I think I'd prefer conditional accept to have its own content drawing logic (e.g. `logic.get_conditional_accept_request_content()`, even if it means repeating ourselves a bit

src/review/models.py
mauromsl: P.S.: This is were the fault would take place I think

https://github.com/openlibhums/janeway/blob/d8bee5af419a2db847864afb469428f10a0b59d3/src/review/views.py#L2484

ajrbyers: I discussed this with @mauromsl last night and we decided that this was fine but that `review_decision` is no longer in use with this change so I will remove it.

ajrbyers:
Thanks for the review there @mauromsl. 

I've made the changes we discussed and couldn't help go beyond a little and implement the standard email form for this view.

src/utils/install/journal_defaults.json
```diff
@@ -5520,5 +5520,81 @@
...
+        },
+        "setting": {
+            "description": "Email sent to authors when sending a conditional accept revision request.",
+            "is_translatable": true,
+            "name": "conditional_accept",
+            "pretty_name": "Conditional Accept",
+            "type": "rich-text"
+        },
+        "value": {
+            "default": "<p>Dear {{ article.correspondence_author.full_name|se_can_see_pii:article }},</p><p>We are pleased to inform you that your article has been conditionally accepted for publication, subject to the successful completion of the required revisions.</p><p>Details of the changes required are below:</p><p>{{ revision.editor_note }}</p><p>You can view and respond to the revision requests here: {{ do_revisions_url }}</p><p>Please complete your revisions by {{ revision.date_due }}.</p><p>We look forward to receiving your updated manuscript.</p><p>Regards,<br>{{ request.user.signature|safe }} </p>\n"
```
joemull: The editor note needs to be marked safe:

image

src/utils/install/journal_defaults.json
```diff
@@ -5520,5 +5520,81 @@
...
+        },
+        "setting": {
+            "description": "Subject for Conditional Accept Email",
+            "is_translatable": true,
+            "name": "subject_conditional_accept",
+            "pretty_name": "Subject Conditional Accept",
+            "type": "char"
+        },
+        "value": {
+            "default": "Article Conditional Acceptance"
```
joemull: I think "Conditional Acceptance of Your Article" would be slightly clearer as a subject line.

ajrbyers:
Thanks @joemull, updated.

joemull (approved):

PR #4818, opened 2025-06-23
Readers can see ROR affiliations on article and public profile pages

joemull:
image
image
image


src/templates/common/svg/orcid-mono.svg
ajrbyers, line 1: We should follow ORCID's brand guidelines https://info.orcid.org/documentation/integration-guide/orcid-id-display-guidelines/#h-unauthenticated-orcid-ids-0

src/templates/common/svg/ror-mono.svg
ajrbyers, line 1: As with ORCID, ROR have their own guidelines for their logo

ajrbyers (changes requested):
See comment, one thing though that jumps out is the ORCID logo looks a bit large for the surrounding text.

PR #4821, opened 2025-06-24
Workflow logs now record the user who moved the article between workflow elements

### LINKED ISSUE #2181: Log moving articles between stages with audit log
It would be helpful to know who moved an article from one stage to another

ajrbyers:
Closes #2181


StephDriver (approved):
would be nice to have this information on the logs part of logs documents and more - but that may be for another time.

src/core/models.py
```diff
@@ -1843,6 +1843,11 @@ class WorkflowLog(models.Model):
         on_delete=models.CASCADE,
     )
     timestamp = models.DateTimeField(default=timezone.now)
+    user = models.ForeignKey(
+        'core.Account',
+        on_delete=models.SET_NULL,
+        null=True,
```
mauromsl, line 1849: Do we not need blank=True to avoid shenanigans when editing via admin?

mauromsl (approved):

PR #4822, opened 2025-06-25
Make Clean-theme navigation dropdowns and links more accessible

### LINKED ISSUE #4227: Clean Theme Navigation Fixes
From screenreader testing (https://github.com/BirkbeckCTP/janeway/issues/4194), looking at the narrow screen navigation the following issues were identified:

- [ ] Sample 3, item 5. menu items with dropdowns don't distinguish themselves from those which are links to other pages.
- [ ] moved to #4837 as user content - Sample 3, item 6. articles item (read out as "articles, link") opens a dropdown the first item of which is articles (read out as "articles, link). This is highly confusing!
- [ ] moved to #4837 as user content - Sample 3, item 7. likewise the second item under 'submissions' is also called 'submissions'
- [ ] moved to #4837 as user content - Sample 3, item 8. and again, 'About' has the first subitem as 'about'
- [ ] moved to  #4838  - Sample 3, item 9. Login and Register are a new list of two items, but part of the menu navigation landmark... confusing.

**Dev notes**
This a11y issue stems from the fact that top level nav items can have a path/link associated with them, while also having subnav. In order to provide the dropdown menu on click events, while also allowing the link to be navigable, the submenu renders a duplicate of the top level nav item.

Instead we have agreed the following actions:

 - Make `has_sub_nav` and `link` mutually exclusive
 - To add a data migration that will update any nav items that meet both conditions and wipe the `link attribute`. A sub nav item will be created using the same name and link as its parent.

joemull:
Fixes #4227.


joemull:
I see what you mean, but I don't think we should try to write a migration that edits the nav item labels. We would inevitably get it wrong.

The way I see it, the main purpose of this fix is to make it (more) possible for users to write accessible navigation. At the moment, the UI does not prevent them from creating a nav item that serves the dual purpose of being a dropdown trigger AND a link. Once users have a UI that encourages them to keep these separate, they will be prompted to design their menus more flexibly.

As for the login and register buttons, should we make that a separate issue?

PR #4825, opened 2025-06-26
Fixes a bug that stopped ORCID login from succeeding when OIDC is also enabled

ajrbyers:
If OIDC is enabled and we don't call `authenticate()` we need to be explicit with which backend to use when calling `login()`

### DIFF
--- a/src/core/views.py
+++ b/src/core/views.py
@@ -442,7 +442,11 @@ def register(request, orcid_token=None):
                 if new_user.email == initial.get("email"):
                     new_user.is_active = True
                     new_user.save()
-                    login(request, new_user)
+                    login(
+                        request,
+                        new_user,
+                        backend='django.contrib.auth.backends.ModelBackend',
+                    )
                     return redirect(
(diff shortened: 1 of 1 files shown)


src/core/views.py
```diff
@@ -244,7 +244,11 @@ def user_login_orcid(request):
                 if candidates.exists():
                     # Store ORCID for future authentication requests
                     candidates.update(orcid=orcid_id)
-                    login(request, candidates.first())
+                    login(
+                        request,
+                        candidates.first(),
+                        backend='django.contrib.auth.backends.ModelBackend',
+                    )
```
mauromsl, line 254: This is slightly different from the flow we do right above: https://github.com/openlibhums/janeway/blob/f014424815b9ea06990e983d15e1879c9cf5a687/src/core/views.py#L233

I think I prefer this approach to mutation. Can we change the above to use the same `kwarg` in `login()`, for consistency'sake?

mauromsl (changes requested):
Minor nitpick

PR #4833, opened 2025-07-01
Complete sitemaps

### LINKED ISSUE #2329: Improvements to basic sitemap
I use the sitemap when I'm looking for a custom page (like https://bok.idpro.org/site/article-refresh-process/) as if I was a regular user. Right now, the XML isn't really human friendly.

I'd like to see a more visually human-friendly sitemap. As an example, https://www.emerald.com/insight/sitemap/publications

### LINKED ISSUE #2799: Sitemaps: lastmod should be better

### LINKED ISSUE #4231: Sitemaps should expose all links for each site
**User stories**
As a user of the sitemaps, I can browse all publicly available pages from the sitemap of a Janeway site (journal/press/preprints) as if I was using the standard navigation from the front-end theme.

**context**

[WCAG 2.4.5 - Multiple Ways](https://www.w3.org/TR/WCAG22/#multiple-ways)
> More than one way is available to locate a Web page within a set of Web pages except where the Web Page is the result of, or a step in, a process.

Our themes have a navigation menu, but that is just one way. Some journals have a sitemap, but not all, and as described in https://github.com/BirkbeckCTP/hourglass/issues/247 sitemaps must include the static pages as well as the articles.  Journals have a search bar, but this only searches articles, and not the static pages or news items (e.g. [typing in 'submission' doesn't bring up the submission page ](https://www.glossa-journal.org/search/?article_search=submission&title=on&abstract=on&sort=relevance )).

Consider:
- [x] improved sitemaps
- [x] improved search
- [x] breadcrumbs

joemull:
## Issues addressed
* Closes #4231
* Closes #2799 
* Closes #2329 

## Example
image

## Improvements
* Maintained existing hierarchy
* Added site content at press and journal levels
* Added human-readable names for links
* Accessibility improvements to HTML markup
* Reworked XSLT to accommodate new namespace `janeway`, which was necessary for including human-readable names without breaking the sitemaps schema
* Updated tests

## To do
* Fix display of translatable charfields at press level. I'm getting raw links at the press level, because `NavigationItem.link_name` is not coming through:
image

### DIFF
--- a/docs/md/contributing-guide.md
+++ b/docs/md/contributing-guide.md
@@ -63,3 +63,9 @@ Also think about how new pages fit into existing documentation, and add cross-re
 We are centred in the UK, so we adopt British spelling in all user-facing writing, taking the first variant listed in the [Oxford English Dictionary](https://www.oed.com/dictionary/centre_n1) when in doubt.
 
 However, for much of our code, we use American spelling for consistency with a third party, like CSS `color` or Creative Commons `license`.
+
+## A-Z list
+
+We record spelling or usage of words here if they are not specified on the Google style guide’s A-Z list and not listed in the OED.
+
+* sitemap, sitemaps
--- a/docs/md/sitemaps-and-robots-files.md
+++ b/docs/md/sitemaps-and-robots-files.md
@@ -1,0 +1,55 @@
+## Sitemaps and robots files
+
+Janeway can auto-generate sitemaps and robots files for press, journal, and repository websites.
+
+## Sitemaps
+
+Janeway sitemaps follow the XML-based standard specified by sitemaps.org.
+
+They are also navigable by end users as an alternative means of navigation. An XSL style sheet is provided that implements an accessible HTML tree with named links and labelled page regions. Each website should have a link to its sitemap in the footer, so that users can discover it.
+
+You can generate sitemaps for your Janeway sites by running the following management command:
+
+```sh
+python3 src/manage.py generate_sitemaps
+```
+
+Running this command will generate sitemaps in `src/files/sitemaps/` with the following hierarchy:
+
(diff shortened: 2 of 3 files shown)


src/journal/models.py
```diff
@@ -690,6 +701,30 @@ def rejected_and_archived_articles(self):
             '-date_declined'
         )
 
+    @property
+    def navigation_items_for_sitemap(self):
+        NavigationItem = apps.get_model('cms.NavigationItem')
```
ajrbyers, line 706: Are you using get_model here to avoid a circular dependency?

src/journal/models.py
```diff
@@ -690,6 +701,30 @@ def rejected_and_archived_articles(self):
...
+            is_external=False,
+        ).order_by('sequence')
+
+    @property
+    def active_news_items(self):
+        """
+        Get the active news items belonging to the press,
+        excluding any journal news.
+        """
+        NewsItem = apps.get_model('comms.NewsItem')
```
ajrbyers, line 721: Same here, why get_model?

ajrbyers (changes requested):
Just a couple of questions about the use of apps.get_model.

src/journal/models.py
joemull: Yeah that's why.

src/journal/models.py
joemull: Yes

StephDriver (changes requested):
Other than the bug reported above, this looked fine - I checked in on one of the ones that had run before the bug halted the generation.

PR #4840, opened 2025-07-07
Provide tip to people configuring Current Issue element

### LINKED ISSUE #4242: Lists of articles a11y improvements
Journals may have long lists of articles under each issue, and when searching a journal, a long list of articles may be returned.  In both instances, this is difficult to navigate. There is nothing to indicate how many items are in the list, nor how far down the list you are when scrolling (other than the window's scroll bars).

This is a particular issue when visiting on a smartphone as the list can take a large number of screens to scroll through. for example [OLH theme home page example](https://www.iastatedigitalpress.com/mmb/)

Similarly, if navigating by keyboard, that's a lot of tabbing.

Consider: 
- pagination.
- search results indicating how many articles have been found.
- An A..Z set of skip to links to jump down the list 
- or a way to further filter results without performing the search again, e.g. listing them by issue?

joemull:
Fixes #4242.

### DIFF
--- a/src/templates/admin/core/manager/settings/index_home.html
+++ b/src/templates/admin/core/manager/settings/index_home.html
@@ -68,6 +68,23 @@
                 </div>
             </div>
         </div>
+        <div class="large-6 columns">
+            <div class="box">
+            <div class="title-area">
+                    <h2>Tips</h2>
+                </div>
+                <div class="content">
+                  <ul>
+                    <li>
+                      <p>If the current issue has more than fifteen or twenty articles,
+                      it is not recommended to use the <strong>Current Issue</strong>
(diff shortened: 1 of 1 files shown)


StephDriver (approved):
👍 the *Mechanical Turk* solution to having too long a list 😆

mauromsl (changes requested):
@joemull is this supposed to be the change? https://github.com/openlibhums/janeway/pull/4840/commits/79db2c5bd4095db4ec23409976351190958cfc6e

I can see this PR bringing in a lot of other unrelated changes.

There are also a couple of conflicts but I think they relate to the erroneous commits. Probably worth a rebase

PR #4841, opened 2025-07-07
Clear up usage of Account.salutation

### LINKED ISSUE #4137: Remove redundant name prefix field
**Describe the bug**
The name prefix field was added to the core account model recently, but I think salutation should have been used instead.

I don't believe we need both of these:

https://github.com/BirkbeckCTP/janeway/blob/e48da784d523b1e4fde5c5b0ef8d8ef03f818fff/src/core/models.py#L234

https://github.com/BirkbeckCTP/janeway/blob/e48da784d523b1e4fde5c5b0ef8d8ef03f818fff/src/core/models.py#L240-L241

**Janeway version**
1.5.2

joemull:
Closes #4137.

This PR does a few things to clarify and simplify how salutation is used on the Account model:

* Deprecates the name_prefix field mistakenly introduced in Janeway 1.5
* Ensures templates use salutation in salutations
* Removes the mistaken mapping of Account.salutation to FrozenAuthor.name_prefix
* Documents current state and recommended use of person name fields for developers

Note that only a few accounts are affected by the deprecation of `name_prefix`, since only the press-level Hourglass register and profile pages gave users access to the field. We should be able to fix these accounts when we update the Hourglass theme to ignore the name_prefix field.


docs/md/person-name-fields.md
```diff
@@ -0,0 +1,26 @@
...
+
+Janeway has ten or so fields related to names of people. They are spread across the `Account` and `FrozenAuthor` models.
+
+These fields are generally subjected to plain text validation to ensure no markup is injected, since they are included in some strings that are marked safe and parsed as HTML.
+
+## Account name fields
+
+We have three fields for first, middle, and last name. Separate fields are useful needed for producing citations, consistent alphabetization, and downstream metadata requirements from ORCID, Crossref, JATS.
+
+However, there are well-documented problems with separate name fields and making them required fields. See [this W3C article about different conventions](https://www.w3.org/International/questions/qa-personal-names), and this [list of falsehoods programmers believe about names](https://shinesolutions.com/2018/01/08/falsehoods-programmers-believe-about-names-with-examples/). If possible, future Janeway development on name fields should make them more flexible.
```
ajrbyers: I don't think this should be in the docs, please move this to an issue on changing how we handle name fields.

ajrbyers (changes requested):
I've added one comment about docs. Other than that it looks good.

ajrbyers (approved):

src/core/models.py
```diff
@@ -349,6 +348,12 @@ class Account(AbstractBaseUser, PermissionsMixin):
 
     uuid = models.UUIDField(default=uuid.uuid4, editable=False)
 
+    name_prefix = models.CharField(
+        max_length=10,
+        blank=True,
+        help_text="The 'name_prefix' field is deprecated. Use 'salutation'.",
+    )
+
```
mauromsl, line 356: We can't deprecate this field without considering the existing records that make use of the field. It will make it unuseable and potentially change names for anyone who have migrated from Bepress. I suggest we do not deprecate this field up until we have a clear path on how we want to handle names moving forward.

mauromsl: I've mixed up name_prefix and suffix, disregard

mauromsl (approved):

PR #4843, opened 2025-07-08
pagination arrows a11y update

### LINKED ISSUE #4247: Add context for page numbers for screenreaders
Page numbers need context for screen readers.  The following examples were identified in #4194 

- Sample 4, item 41. pages: 1-13 is read as "pages one thirteen" because the - is considered invisible punctuation as it could in say 'copy-editing` - this should read as "Pages one to thirteen" or similar. It should be described to be read that way.
- Sample 5.1, item 16. the page numbers at the end of the article details, '1-9' are read as 'one nine' because the screenreader doesn't know that punctuation is important in this place - suggest marking up both to describe these as pages, and to indicate it is a range, e.g. 'pages one to nine' - but need to be careful with implementing this as I believe the UK is 'one to nine' but the US preference may be for 'one through nine'.

### LINKED ISSUE #4842: make left and right arrows accessible
When working on the Icon accessibility #4215 the use of html characters for left and right arrows weren't spotted.

So:
update where`&laquo;` and `&raquo;` are used across the front of house to be accessible.

StephDriver:
closes #4842 
closes #4247

### DIFF
--- a/src/themes/clean/templates/press/core/news/index.html
+++ b/src/themes/clean/templates/press/core/news/index.html
@@ -28,8 +28,7 @@
                 {% if news_items.has_previous %}
                     <li class="list-inline-item">
                         <a href="?page={{ news_items.previous_page_number }}" class="btn btn-outline-primary">
-                            &laquo;
-                            <span class="sr-only">{% trans 'previous page' %}</span>
+                            {% include "elements/icons/page_previous.html" %}
                         </a>
                     </li>
                 {% endif %}
@@ -51,8 +50,7 @@
                 {% if news_items.has_next %}
                     <li class="list-inline-item">
                         <a href="?page={{ news_items.next_page_number }}" class="btn btn-outline-primary">
-                            &raquo;
-                            <span class="sr-only">{% trans 'next page' %}</span>
+                            {% include "elements/icons/page_next.html" %}
                         </a>
                     </li>
                 {% endif %}
--- a/src/themes/clean/templates/press/core/news/index.html
+++ b/src/themes/clean/templates/press/core/news/index.html
@@ -22,26 +22,41 @@
             <h2>{% trans 'This journal currently has no news items to display' %}.</h2>
         {% endfor %}
     </div>
-    <div class="pagination-block">{{ page }}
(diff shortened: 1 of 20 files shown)


src/themes/clean/templates/press/core/news/index.html
```diff
@@ -22,20 +22,41 @@ <h6>Posted by {{ item.posted_by.full_name }} on {{ item.posted|date_human }}</h6
...
-                   class="btn {% if news_items.number == page %}btn-primary{% else %}btn-outline-primary{% endif %}">{{ page }}</a>
-                &nbsp;
-            {% endfor %}
-            {% if news_items.has_next %}
-                <a href="?page={{ news_items.next_page_number }}" class="btn btn-outline-primary">&raquo;</a>
-            {% endif %}
-        </ul>
-    </div>
+    <section aria-label="pagination">
+        <div class="pagination-block">{{ page }}
```
joemull: I think the `{{ page }}` here is a bug. You caught it in the other paste of this code and removed it. Maybe combine these two instances with a sub-template to make them easier to maintain?

joemull (changes requested):
Looks good to me. I found one tiny bug. Also consider combining some of the instances of verbose blocks into single templates which are included in multiple places, to make it easier to maintain going forward.

StephDriver:
> Looks good to me. I found one tiny bug. Also consider combining some of the instances of verbose blocks into single templates which are included in multiple places, to make it easier to maintain going forward.

I've had a look into making pagination templates (beyond those which currently exist) but there is a lot of difference between the logic used in each place and each theme needs different classes, all together, this woudl end up with each new template being used once.

However I take the point on standardising - and what can be standardised is the part that the user interacts with, e.g. the previous page and next page buttons could be pulled out as templates.  This would also make it easier to update when we get rid of the `&laquo;` etc to replace with an icon set.  And it would ensure all page change links are marked up accessibly.

StephDriver:
page icons:

```django
{% include "elements/icons/page_previous.html" %}
```

```django
{% include "elements/icons/page_next.html" %}
```

ajrbyers (approved):
Great work here.

PR #4845, opened 2025-07-08
Upgrades to contact messages

### LINKED ISSUE #1720: Press shouldn't use contact view from journal
The press contact view is using `journal.views.contact`.
This makes sense, since the logic is the same regardless of the site type, but it is not obvious from the view that it is being used by the press which has led to #1719 

Move the shared logic to core, since it will be used by all types of sites . Then we can have wrapper views on journal/press with the appropriate decorators.

### LINKED ISSUE #4138: Let referring link specify which contact to select in contact form
## User journey
As an end user reading a press or journal website, if I am told to contact a particular staff person or editor, and I am linked to the `/contact/` page to do so, I may be a bit confused if the pre-selected person is different than the person I was told to contact.

## Proposed solution
Check for a query parameter like `recipient=email@example.org` and pass that argument into the form here: 

https://github.com/BirkbeckCTP/janeway/blob/e48da784d523b1e4fde5c5b0ef8d8ef03f818fff/src/journal/views.py#L1886-L1896

## Potential synergy

https://github.com/BirkbeckCTP/janeway/issues/1720

joemull:
Closes #4138.
Closes #1720.

~Please also review and merge docs PR: https://github.com/openlibhums/memory-alpha/pull/155.~

## Notes

* Closes the linked issues
* Connects the contact person and contact message models to account, so that we do not have names and emails outside of the Account object
* Renames the contact person and contact message models to make them easier to read and work with
* Removes stored client IPs to comply with GDPR
* Adds the Django auto-documentation views, which are helpful when searching for template tags and filters, and which show you view and url pairs
* Documents a few things
* Tests for the contact manager and the press and journal contact pages

## Screenshots for contact system

Screenshot from 2025-07-21 17-32-46
Screenshot from 2025-07-21 17-33-10
Screenshot from 2025-07-21 17-33-20
Screenshot from 2025-07-21 17-33-46
Screenshot from 2025-07-21 17-34-03
Screenshot from 2025-07-21 17-34-14

## Screenshots for admin docs feature

I enabled this in the process of doing this work. It is enormously helpful to view a comprehensive list of things in Janeway sometimes.

Screenshot from 2025-07-21 17-35-02
Screenshot from 2025-07-21 17-35-18
Screenshot from 2025-07-21 17-40-34


src/core/migrations/0111_rename_contacts_contactperson_and_more.py
ajrbyers, line 1: This migration failed with live data.

django.db.utils.IntegrityError: duplicate key value violates unique constraint "core_account_email_key"
DETAIL:  Key (email)=(X.Xxxxx@xxx.ac.uk) already exists. 

I can give you the link to this user account, but it looks like an import.

src/templates/admin/core/manager/contacts/related_info.html
ajrbyers, line 1: Wide content does not display well.

image

src/templates/admin/core/manager/contacts/index.html
```diff
@@ -57,7 +60,7 @@ <h2>Current Contacts</h2>
                             {% empty %}
                             <tr>
                                 <td colspan="5">
-                                    This journal has no contacts. <a href="{% url 'core_new_journal_contact' %}">Add a new contact.</a>
```
ajrbyers, line 60: Was there a specific reason for this change?

src/templates/admin/core/manager/contacts/contact_person_form.html
ajrbyers, line 1: Saving this page presents no confirmation that an action has been taken.

ajrbyers (changes requested):
I've added a few comments inline for you to go over, however, I have a fundamental question about this implementation. We have a contact form to obscure email addresses from being scraped. If we require a user to create a URL with the users email address in it we may be defeating this purpose. I think the account UUID field is the least painful option here. The Contacts page could present contact link that a user could simply copy and paste so they don't have to form it themselves?

Open to thoughts on this.

src/templates/admin/core/manager/contacts/index.html
joemull: It felt redundant since there was an **Add** button in the typical place, and having an extra link presented some complexities that I thought we could avoid: we'd have to have a link for both journal and press levels, and we'd have to make sure the pattern of next URLs was correct.

joemull:
> I've added a few comments inline for you to go over, however, I have a fundamental question about this implementation. We have a contact form to obscure email addresses from being scraped. If we require a user to create a URL with the users email address in it we may be defeating this purpose. I think the account UUID field is the least painful option here. The Contacts page could present contact link that a user could simply copy and paste so they don't have to form it themselves?
> 
> Open to thoughts on this.

All good suggestions, thanks! Updated with some changes:

* Used uuid as suggested
* Added link copy button as suggested
* Went ahead and standardised other buttons and confirm delete behavior for consistency
* Fixed other display issues
* Updated migration to fix bug, and tested with OLH data

joemull:
Note: Going to rebase this as there are conflicts keeping the CI job from running.

ajrbyers (approved):

src/core/migrations/0110_rename_contact_contactmessage_and_more.py
```diff
@@ -0,0 +1,213 @@
...
+            account = matching_accounts.first()
+        else:
+            account = Account.objects.create(
+                email=contact_person.email,
+                username=contact_person.email.lower(),
+                first_name=" ".join(contact_person.name.split()[:-1]),
+                last_name=contact_person.name.split()[-1],
+            )
+        contact_person.account = account
+        contact_person.save()
```
mauromsl: There are cases where the username might not match the email. Ultimately, we probably want to remove the username field and use email as the username to avoid the duplicity.

Meanwhile, I think in this migration you could end up in a situation where matching by username fails, however the create below also fails because of a duplicate email address, throwing an exception and aborting the migration. We should check both fields before we run a create

Additionally, when checking an address against a username, the correct way to sanitize the input would be to "normalize" the address, rather than run `email.lower()` see it [here](https://github.com/django/django/blob/main/django/contrib/auth/base_user.py#L19-L31)

src/core/models.py
```diff
@@ -1970,32 +1975,69 @@ class Contacts(models.Model):
...
+        return self.account.full_name() if self.account else ""
+
+    @property
+    def display_email(self):
+        return self.account.pretty_wrapping_email if self.account else ""
+
+
+class ContactMessage(models.Model):
+    account = models.ForeignKey(
+        Account,
```
mauromsl: This is more of a question than a comment. Should this model have a FK to `Contact` rather than `Account`. I suppose the main question is, would we want to track/audit the messages as "User messaged this person" or as "User messaged this contact, who is linked to this person". This is only relevant when a contact is shifted from one person to another (e.g. an editor in chief or press manager is replaced) but I thought I'd ask

src/core/views.py
```diff
@@ -1870,117 +1878,184 @@ def article_image_edit(request, article_pk):
...
+    :return: HttpResponse object
+    """
+    next_url = request.GET.get("next", "")
+    contact_person = get_object_or_404(
+        models.ContactPerson,
+        pk=contact_person_id,
+        content_type=request.model_content_type,
+        object_id=request.site_type.pk,
+    )
+    contact_people = request.site_type.contact_people
```
mauromsl, line 1984: I don't see where this is used

src/core/views.py
```diff
@@ -1870,117 +1878,184 @@ def article_image_edit(request, article_pk):
...
+    """
+
+    next_url = request.GET.get("next", "")
+    contact_person = get_object_or_404(
+        models.ContactPerson,
+        pk=contact_person_id,
+        content_type=request.model_content_type,
+        object_id=request.site_type.pk,
+    )
+    contact_people = request.site_type.contact_people
```
mauromsl, line 2029: I don't think we use this variable

src/core/logic.py
```diff
@@ -1221,3 +1221,58 @@ def create_organization_name(request):
...
+def get_contact_form(request, recipient_uuid):
+    subject = request.GET.get("subject", "")
+    contact_people = request.site_type.contact_people
+    account = None
+    if recipient_uuid:
+        try:
+            account = models.Account.objects.get(
+                contactperson__in=contact_people,
+                uuid=recipient_uuid,
+            )
```
mauromsl: If I know the UUID of another account, I think this potentially allows an attacker to get a contact from from a different site. Instead, and since `request` is an argument to this function, we can flip the query to retrieve a `Contact` person and match against the `request.site_type`

mauromsl (changes requested):
a few comments inline

src/core/logic.py
joemull: Hmm, thinking this through, I don't think that would add any additional security. The `site_type` is already filtering the query, because it is used in forming the list of `contact_people`, which the account has to match one of. See `AbstractSiteModel.contact_people`. But is there something I'm not seeing?

src/core/models.py
joemull: As contacts are labelled with a person's name, I think the messages should be linked to that account, because that's the best way to provide data privacy. It's also not possible to change the account that a `ContactPerson` is linked to, and we should probably keep it that way for privacy reasons as well. If people have a use case for a shared contact, I think it makes sense for them to use a shared email address, for the time being.

src/core/views.py
joemull: It is passed into the template context so the user can see which contact people currently exist in the left-hand sidebar when adding, removing or editing a contact person.

src/core/views.py
joemull: It is used in the context by a sub-template, as above. I've added `with contact_people=contact_people` in the template inclusions to show the trail more clearly.

src/core/migrations/0110_rename_contact_contactmessage_and_more.py
joemull: OK, I will try using `_iexact` to check for matches on both `username` and `email` when querying, and then I will use `normalize_email` to form the string for `username` when creating a new account.

FYI, we already lowercase the entire string for `username` in our main `create_user` method:

https://github.com/openlibhums/janeway/blob/311b19619833eb51c18e876b014be0a05b0d6396/src/core/models.py#L412-L420

This is why I took the lowercasing approach, so that it would match our data.

Are you suggesting we need to change this?

joemull (commented):
OK I've written some replies and changed a few things--commits incoming once I test it on OLH data again.

src/core/models.py
mauromsl: Ok, makes sense, but I think we still have a problem in that, if an account is deleted, (say GDPR reasons or the user leaves the platform), the new send_contact_message will likely raise our dearest exception `AttributeError: Nonetype object has no attribute "email"`.

I'm not sure how we should defend against that, since it is fairly niche, but at least we should probably:

 - Document how to change contacts: Now contacts have to be deleted and recreated rather than just changing the email address / related account
 - Let the user know that deleting the contact will not delete their messages + maybe tell them how they can find those messages?
 - When an active `ContactPerson` has had its underlying account deleted/set to null have some sort of warning or prompt?

mauromsl (changes requested):
Thanks for the clarifications Joe, I've added one comment with some final suggestions inline

src/core/models.py
joemull: Good suggestions, but "tell them how they can find those messages" raises an interesting design challenge. Right now, contact messages are stored in the DB, but *not* as logs (most emails are stored via logs). However, there is no interface outside of the Django admin to pull them up.

I think we have two main options. Which approach do we prefer?

a. Create an interface for viewing contact messages in the back office. It could consist of a link from the dashboard + a list page that lets you sort and search + an item page to view the message + a link to message the user back (and have *that* logged).
b. Change how contact messages are stored, so that they go into logs like other emails, and then also make it possible to view logs that are not attached to any article. Right now I don't know of a way they anyone access a log item that does not have an associated article.

I slightly prefer b, because it is more consistent with where users expect messages to be recorded. However, it also is counter-intuitive, because users only know about logs as a feature of article details, via "Logs, docs, and more".

Thoughts? Other possibilities? Also tagging @StephDriver and @ajrbyers for comment.

StephDriver: 1. I think UI and where the messages are actually stored are two separate things.  
2. I like `a` as an interface, I think that makes sense for users - you'd have a place to see messages. 
3. I think all these messages should go into the logs, and potentially in future be viewable there too - in terms of notification centre dreams, that would make sense too, everything going through the switchboard. 

So all in all, I think a combination of the two options is needed.  The user-interface should be based around a, but the underlying construct should be through the logs, as in b. 

BUT, this is so close to the notification system work it really feels like a rush to try to do this solo, and for 1.9.

ajrbyers: I prefer option b. Contact messages are emails like any other system communication, so they should be stored in logs for consistency.

The counter-intuitiveness issue is actually an opportunity - we can use this to begin introducing non-article linked logs. Start simple: add a way to view logs that aren't attached to articles, filterable by type. Contact messages become the first use case, and we build the minimal interface needed to access them, and a link to that page can apply a filter to show contact messages for that particular journal.

src/core/forms/forms.py
```diff
@@ -106,6 +104,37 @@ class Meta:
...
+        ]
+        self.fields["subject"].initial = subject
+
+        if contact_person:
+            self.fields["contact_person"].initial = contact_person.pk
+
+
+class JournalContactForm(ContactPersonForm):
+    def __init__(self, *args, **kwargs):
+        return DeprecationWarning("Use ContactPersonForm instead.")
```
mauromsl: There can't be a return statement inside a `__init__` call.

I would just log a warning or raise if in debug mode

src/core/janeway_global_settings.py
```diff
@@ -52,6 +52,7 @@
 
 INSTALLED_APPS = [
     "modeltranslation",
+    "django.contrib.admindocs",
```
mauromsl, line 55: Leave it in this time, but I reckon this should've been on its own PR.

src/core/logic.py
```diff
@@ -1257,3 +1258,74 @@ def create_organization_name(request):
...
+            subject=subject,
+            contact_people=contact_people,
+            contact_person=contact_person,
+        )
+    return contact_form, contact_people
+
+
+def send_contact_message(contact_form, request):
+    # Todo: do we need this if the field is bleached rich text?
+    body = contact_form.cleaned_data["body"].replace("\n", "<br>")
```
mauromsl: or use [`linebreaksbr`](https://github.com/django/django/blob/10d1d2c55ed8ec1e8421e00674b3b6b8448318eb/django/template/defaultfilters.py#L485)

src/core/forms/forms.py
```diff
@@ -106,6 +104,37 @@ class Meta:
         )
 
 
+class ContactMessageForm(CaptchaForm):
+    contact_person = forms.TypedChoiceField(
+        label=_("Who would you like to contact?"),
+    )
+    sender = forms.EmailField(
+        max_length=200,
```
mauromsl: we probably want a constant to keep `Logentry.actor_email` and this in sync.

src/core/migrations/0110_rename_contacts_contactperson_and_more.py
```diff
@@ -0,0 +1,267 @@
...
+    """
+    email = email or ""
+    try:
+        email_name, domain_part = email.strip().rsplit("@", 1)
+    except ValueError:
+        pass
+    else:
+        email = email_name + "@" + domain_part.lower()
+    return email
+
```
mauromsl: While model methods cannot be accessed from migrations, manager methods can (so long as they get serialized) https://docs.djangoproject.com/en/4.2/topics/migrations/#model-managers

Should be a bit more robust than copying the function over

src/core/migrations/0111_rename_contacts_contactperson_and_more.py
```diff
@@ -0,0 +1,267 @@
...
+        if matching_accounts.exists():
+            account = matching_accounts.first()
+        else:
+            normalized_email = normalize_email(email)
+            account = Account.objects.create(
+                email=normalized_email,
+                username=normalized_email,
+                first_name=" ".join(contact_person.name.split()[:-1]),
+                last_name=contact_person.name.split()[-1],
+            )
```
mauromsl: There can still be some shenanigans here:

- For the username, we should compare against `contact_person.lower()` since that is what we do during user creation.

- When username doesn't match (e.g username <-> email de-synch) it gets a bit tricky to match email address. There can be a case where both sides of the OR query fail, so we need to handle a potential exception during `Account.objects.create()` for an `IntegrityError`. I'd also suggest comparing against the nornmalized email address in the SQL query to minimise query misses and avoid exceptions being raised during `migrate`.

Something like:
```py
    for contact_person in ContactPerson.objects.all():
        email = contact_person.email
        matching_accounts = Account.objects.filter(
            models.Q(username__iexact=email.lower()) | models.Q(email__iexact=email),
        )
            account = matching_accounts.first()
        else:
            normalized_email = normalize_email(email)
            try:
              account = Account.objects.create(
                  email=normalized_email,
                  username=normalized_email,
                  first_name=" ".join(contact_person.name.split()[:-1]),
                  last_name=contact_person.name.split()[-1],
              )
            except IntegrityError:
                # missmatched formatting on account emails, skip
                pass 
[...]
```

src/core/migrations/0110_rename_contacts_contactperson_and_more.py
```diff
@@ -0,0 +1,267 @@
...
+        else:
+            normalized_email = normalize_email(email)
+            account = Account.objects.create(
+                email=normalized_email,
+                username=normalized_email,
+                first_name=" ".join(contact_person.name.split()[:-1]),
+                last_name=contact_person.name.split()[-1],
+            )
+        contact_person.account = account
+        contact_person.save()
```
mauromsl: should we clear the email address after finding the account, to avoid de-normalising the data?

```py
if account:
    contanct_person.account = account
    contact_person.email = None
    contact_person.save()

File: src/core/migrations/0111_rename_contacts_contactperson_and_more.py [hunk 14]
```diff
@@ -0,0 +1,267 @@
...
+        contact_person.account = account
+        contact_person.save()
+
+
+def infer_contact_message_target(apps, account):
+    Journal = apps.get_model("journal", "Journal")
+    Press = apps.get_model("press", "Press")
```
mauromsl: What a royal pain this.

If I'm not mistaken, the list of contacts is populated from `core.Contacts.objects.all()`

Shouldn't we check there first, as our source of most truthy value?

src/core/migrations/0111_rename_contacts_contactperson_and_more.py
```diff
@@ -0,0 +1,267 @@
...
+            models.Q(username__iexact=email) | models.Q(email__iexact=email),
+        )
+        if matching_accounts.exists():
+            account = matching_accounts.first()
+        else:
+            normalized_email = normalize_email(email)
+            account = Account.objects.create(
+                email=normalized_email,
+                username=normalized_email,
+            )
```
mauromsl: as above, consider querying against `email.lower()` and handle potential `IntegrityError` on `Account.objects.create()`

src/journal/logic.py
```diff
@@ -397,6 +397,10 @@ def set_article_image(request, article):
 
 
 def send_contact_message(new_contact, request):
+    raise DeprecationWarning(
+        "`journal.logic.send_contact_message` is deprecated. "
+        "Use `core.logic.send_contact_message` instead."
+    )
```
mauromsl: we want to log, not raise, as it can cause a 500 depending on how logging is configured + the version of Python

src/utils/models.py
```diff
@@ -189,6 +206,20 @@ def bulk_add_simple_entry(
...
+    def viewable_as_contact_message_by(self, account):
+        """
+        Determines whether a user can view the log entry,
+        if it is categorized as a "Contact Message".
+        Denies access for other log entries of other types,
+        so it should not be used in those cases.
+        """
+        if self.types == "Contact Message":
+            if Addressee.objects.filter(
+                email__iexact=account.email,
```
mauromsl: Normalise and compare to avoid MailCleaner-type attacks

mauromsl (changes requested):
a few comments inline around the complexity of email comparisons and related data migrations.

src/core/forms/forms.py
joemull: I can't work out what you mean. In sync in what way?

src/core/migrations/0111_rename_contacts_contactperson_and_more.py
joemull: > * For the username, we should compare against `contact_person.lower()` [I think you mean `email.lower()`] since that is what we do during user creation.

> There can be a case where both sides of the OR query fail

How? Here's my thinking: The [`__iexact` ending is a case-insensitive match](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#iexact). This obviates the need for `lower()` on the right side, and will catch all possible case variations of an existing account username or email. If `models.Q(username__iexact=email) | models.Q(email__iexact=email)` returns no matches, we can be confident there is no account with a username or email in any case variation that has the same characters as the contact person email. What am I not seeing?

> except IntegrityError:
                # missmatched formatting on account emails, skip
                pass 

To double-check, is this for a case where the contact person email is not a valid email address, so the creation fails because of checks on `models.EmailField`?

src/core/migrations/0110_rename_contacts_contactperson_and_more.py
joemull: ~Alright, I've turned on `use_in_migrations` but it is still giving me an error that it does not exist. Any ideas?~

It needed a migration--it is working now.

src/core/migrations/0111_rename_contacts_contactperson_and_more.py
mauromsl: I'm thinking cases I've seen where

- email has been inserted without normalization (due to a db operation or an import routine ignoring application-level checks)
- email and account are different to each other (e.g. account created with one email address, like a dummy email, and then later have their email address added).

for data migration code specifically, we always want to be overzealous and protect against edge-cases that are possible at the database level. 

In fact, I just thought of another case where we'll need to protect against `MultipleObjectsReturned`), so it is worth handling it as well.

src/core/forms/forms.py
mauromsl: Sorry, I meant the `max_length`, by using a constant, so that if one changes, the other does too. (since we copy from one to the other)

mauromsl (changes requested):

mauromsl (changes requested):
replied to comments, we are nearly there

src/core/migrations/0111_rename_contacts_contactperson_and_more.py
joemull: I asked additional clarifying questions offline. Here's a record of that:

The `iexact` will likely catch everything, but for readability and unforeseen cases, we should prepare emails and usernames in exactly the same way as we do on the Account, and we should add the IntegrityError for good measure.

joemull:
OK @mauromsl, I've addressed your comments. I've also tested it with a larger database and optimized the db access in the migration so it's quicker. There was also a bug with log entry dates because `auto_now_add` can't be manually overridden, so I made another change to that model.

src/core/migrations/0111_rename_contacts_contactperson_and_more.py
```diff
@@ -112,21 +138,23 @@ def move_contact_message_to_log_entry(apps, schema_editor):
...
-    # Link up the addressees with their log entries
-    addressees_to_create = []
-    for i, addressee in enumerate(addressees_to_link_up):
-        addressee.log_entry = log_entries[i]
-        addressees_to_create.append(addressee)
+    if addressees_to_link_up:
+        # Link up the addressees with their log entries
+        addressees_to_create = []
+        for i, addressee in enumerate(addressees_to_link_up):
+            addressee.log_entry = log_entries[i]
```
mauromsl: Something I've learned recently: Bulk creation **will not** return IDs on mysql databases, even when `ignore_conflicts` is set to `False` (which is the default behaviour, because it doesn't support `RETURNING`

from the [docs](https://docs.djangoproject.com/en/6.0/ref/models/querysets/#bulk-create):
```
If the model’s primary key is an [AutoField](https://docs.djangoproject.com/en/6.0/ref/models/fields/#django.db.models.AutoField) or has a [db_default](https://docs.djangoproject.com/en/6.0/ref/models/fields/#django.db.models.Field.db_default) value, and ignore_conflicts is False, the primary key attribute can only be retrieved on certain databases (currently PostgreSQL, MariaDB, and SQLite 3.35+). On other databases, it will not be set.
```

So in order to both support MySQL AND try and be performant, we want to put each batch on its own transaction and do a single commit per entry with `.save()` or `.create()` and committing the transaction. The problem with running separate transactions within the same migration is that, in the event of failure, committed changes will no longer be reverted, so we also need to make sure we do not create duplicate log entries

src/core/migrations/0111_rename_contacts_contactperson_and_more.py
```diff
@@ -1,90 +1,116 @@
...
+
+    return account
+
+
+def connect_contact_person_to_account(apps, schema_editor):
+    ContactPerson = apps.get_model("core", "ContactPerson")
+    Journal = apps.get_model("journal", "Journal")
+    Role = apps.get_model("core", "Role")
+    AccountRole = apps.get_model("core", "AccountRole")
+    for contact_person in ContactPerson.objects.all():
```
mauromsl: Is it worth filtering here by `NULL` accounts?
`for contact_person in ContactPerson.objects.filter(account__isnull=False)`

mauromsl (changes requested):
Couple changes suggested inline for the data migration

src/core/migrations/0111_rename_contacts_contactperson_and_more.py
joemull: I don't anticipate any contact people would have accounts when this is run, but it can't hurt.

src/core/migrations/0111_rename_contacts_contactperson_and_more.py
joemull: OK, I will give this a try. By the way, I believe this might be the bug behind #5248, because I had to use a similar strategy to link the different pieces of the ROR data model together without it taking an hour to run.

joemull:
Need to resolve #5259 before I can finish testing this. Setting status as blocked.

src/core/migrations/0111_rename_contacts_contactperson_and_more.py
joemull: Done and tested on MySQL.

joemull:
> Need to resolve #5259 before I can finish testing this. Setting status as blocked.

- Rebased in order to get the MySQL bug fix.
- Tested manually on MySQL. 
- Made a small adjustment for logged-in users filling out contact form to match user expectation.

mauromsl (approved):

src/core/logic.py
```diff
@@ -1370,3 +1371,78 @@ def resolve_alt_text_target(request):
...
+        "types": "Contact Message",
+        "target": request.site_type,
+        "actor_email": contact_form.cleaned_data["sender"],
+    }
+
+    # Avoid sending request.user through the logging system,
+    # because the contact page is public-facing and logged-in users
+    # will not expect it to record their account email as the LogEntry.actor
+    # rather than the email they put in the From field.
+    request.user = None
```
mauromsl, line 1426: This is a dangerous proposition, as `request.user` is global for the entire request lifecycle. I'd recommend either handling this downstream or changing the code to take an explicit user that is set to `None` for this instance

mauromsl:
@joemull I had to revert the merge of this PR, as I was a little to eager to merge and only realised there is a problem after merging.

Since we can't reopen the PR, would you mind opening a new one and linking them through

https://github.com/openlibhums/janeway/pull/4845#discussion_r3461181260

Apologies for the confusion

mauromsl:
now closed by #5367

PR #4848, opened 2025-07-09
Auto-record Janeway version bumps using signals

### LINKED ISSUE #4826: Automate version upgrades
Before we started using semantic and deterministic versioning within Janeway, we adopted the pattern of writing up a Django migration for every version of Janeway. This process is cumbersome and prone to error (See missing migration for Janeway v1.7.3)

An automated version logging system for Janeway, that keeps an auditable record of version changes, without requiring developer time on every release

**Additional context**
Django provides a number of signals during migration that can be used to run our custom logging. Alternatively we can override and extend the builtin "migrate" command

mauromsl:
Removes the need for adding migration bumps manually.

Closes #4826

### DIFF
--- a/src/core/apps.py
+++ b/src/core/apps.py
@@ -12,6 +12,7 @@ class CoreConfig(AppConfig):
     name = 'core'
 
     def ready(self):
+        from core import upgrade # register upgrade signals
         from core.model_utils import SearchLookup
         models.CharField.register_lookup(SearchLookup)
         models.TextField.register_lookup(SearchLookup)
--- a/src/core/upgrade.py
+++ b/src/core/upgrade.py
@@ -1,0 +1,24 @@
+"""
+Routines for handling Janeway version upgrades.
+"""
+__copyright__ = "Copyright 2025 Open Library of Humanties"
+__author__ = "Open Library of Humanities"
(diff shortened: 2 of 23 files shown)


ajrbyers (approved):

joemull:
Just to confirm the behavior, this may result in gaps in the Version objects that someone has in their database, right? If they upgrade from 1.4.4 to 1.7.5, they won't have anything for version 1.5.x, 1.7.x, etc.

mauromsl:
> Just to confirm the behavior, this may result in gaps in the Version objects that someone has in their database, right? If they upgrade from 1.4.4 to 1.8.0, they won't have anything for version 1.5.x, 1.6.x, 1.7.x, etc.

that is correct, and also the expected behaviour IMO. Currently, installing Janeway from scratch populates the version table with every version in exsitance

ajrbyers:
> Just to confirm the behavior, this may result in gaps in the Version objects that someone has in their database, right? If they upgrade from 1.4.4 to 1.8.0, they won't have anything for version 1.5.x, 1.6.x, 1.7.x, etc.

It would make the audit trail cleaner this way, right? It would show the install went straight from 1.4.4 to 1.8.0.

PR #4856, opened 2025-07-15
Fix the migrations for workflow log user and salutation

### LINKED ISSUE #4855: Mismatched field parameters on migration for WorkflowLog.user
## Problem

The core 0108 migration needs `blank=True` to match the model:

https://github.com/openlibhums/janeway/blob/55462f1ed04ee324b04b5daef7d0f60a345593f8/src/core/migrations/0108_workflowlog_user.py#L18

## Version

55462f1ed04ee324b04b5daef7d0f60a345593f8

joemull:
Fixes #4855.


src/core/migrations/0108_salutation_name_20250707_1420.py
```diff
@@ -322,7 +322,7 @@ def use_salutation_name_in_email_salutations(apps, schema_editor):
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('core', '0107_alter_controlledaffiliation_options_and_more'),
+        ('core', '0108_workflowlog_user'),
```
mauromsl, line 325: Since this is not on master yet, is it worth renaming the migration to `0109_*`?

joemull: Sure thing

PR #4895, opened 2025-08-05
bypass for RUSAGE to prevent errors on MacOS

StephDriver:
when using dev-settings on mac, we get the following error:
`AttributeError: module 'resource' has no attribute 'RUSAGE_THREAD'`

previously addressed by manually removing`    "utils.middleware.TimeMonitoring",` from these settings each time it was installed on mac. 

This try/catch addresses this directly.

### DIFF
--- a/src/utils/middleware.py
+++ b/src/utils/middleware.py
@@ -75,5 +75,8 @@ class TimeMonitoring(BaseMiddleware):
 
     @staticmethod
     def _get_usage():
-        utime, stime, *_ = resource.getrusage(resource.RUSAGE_THREAD)
+        try:
+            utime, stime, *_ = resource.getrusage(resource.RUSAGE_THREAD)
+        except AttributeError:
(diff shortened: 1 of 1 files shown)


src/utils/middleware.py
```diff
@@ -75,5 +75,8 @@ def _diff_usages(cls, start, end=None):
 
     @staticmethod
     def _get_usage():
-        utime, stime, *_ = resource.getrusage(resource.RUSAGE_THREAD)
+        try:
+            utime, stime, *_ = resource.getrusage(resource.RUSAGE_THREAD)
+        except AttributeError:
+            utime, stime, *_ = resource.getrusage(resource.RUSAGE_SELF)
```
ajrbyers: Add a broad exception here so that this middleware (which is DEBUG only anyway) never kills the process.

ajrbyers (changes requested):
See comment.

Issue #4899, opened 2025-08-07
Add a settings for rendering all XML articles with the builtin XSL stylesheet


mauromsl:
**Is your feature request related to a problem? Please describe.**
During development, it can be difficult to work around the XSLT system to develop features onto the built in XSL stylesheet.

**Describe the solution you'd like**
A setting that allows developers to override the XSLT file system and use the builtin XSL for all articles

PR #4900, opened 2025-08-07
feat: Allow devs to force builtin XSL render #4899

mauromsl:
Adds a new setting (`FORCE_BUILTIN_XSL`) that allows developers to ingore the XSLT file system and always render XML to HTML galleys off the builtin XSL provided by Janeway.


StephDriver (approved):
thank you!

PR #4901, opened 2025-08-08
fix: #4008 enlarge table link removed from enlarged table itself

### LINKED ISSUE #4008: When viewing an enlarged table I cannot see any way to dismiss or close it
**User Story**
As a reader I cannot see any way to dismiss the enlarged table modal in the material theme:

Image

I can press Escape to dismiss it or click outside it, but I really should see some way to close it as well.

StephDriver:
closes #4008

### DIFF
--- a/src/themes/OLH/templates/journal/article.html
+++ b/src/themes/OLH/templates/journal/article.html
@@ -500,10 +500,12 @@
     <script src='https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.2/MathJax.js?config=TeX-MML-AM_CHTML'></script>
     {% hook 'article_js_block' %}
     <script>
-    $('.table-expansion').each(function(){
-        var child = $(this).children(":first")
-        child.append('<br /><small><a data-open="table-' + $(this).attr('id') + '">View Larger Table</a></small>');
-    });
+        $('.table-expansion').each(function(){
+            if (!$(this).closest('.reveal').length) {
+                var child = $(this).children(":first")
+                child.append('<br /><small><a data-open="table-' + $(this).attr('id') + '">View Larger Table</a></small>');
+            }
+        });
     </script>
     <script src="{% static "common/lightbox/js/lightbox.js" %}"></script>
     <script src="{% static "common/js/clipboard.js" %}"></script>
--- a/src/themes/clean/templates/journal/article.html
+++ b/src/themes/clean/templates/journal/article.html
@@ -357,11 +357,17 @@
     <script src='https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.2/MathJax.js?config=TeX-MML-AM_CHTML'></script>
     <script src="{% static 'material/toc.js' %}"></script>
     {% hook 'article_js_block' %}
-     <script>
-    $('.table-expansion').each(function(){
-        var child = $(this).children(":first")
-        child.append('<br /><button type="button" class="btn btn-link" data-toggle="modal" data-target="#table-'+ $(this).attr('id') +'">{% trans 'View Larger Table' %}</button>')
-    });
+    <script>
+        $('.table-expansion').each(function(){
+            // Only add the link if this table is not inside a modal
+            if (!$(this).closest('.modal').length) {
+                var child = $(this).children(":first")
+                // Check if the link already exists to prevent duplicates
+                if (!child.find('button[data-target^="#table-"]').length) {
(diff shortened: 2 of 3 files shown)


StephDriver:
Work Done:
- [ ] prevent 'enlarge table' link appearing on the enlarged table as well as in the article
- [ ] close buttons to top right of modal (consistency)
- [ ] cell lines contrast to 1:3.5
- [ ] check for keyboard navigation (and fix)
- [ ] check for screen reader accessibility (and fix)

StephDriver:
## Screen Reader Testing
with Voiceover on mac
- buttons to open larger table need described by to link to identify which table.
- On material only, when opening with voiceover, the popup doesn't open at the top of the page but shows a midscroll (and starts reading in the middle of the table), but when opening by mouse or keyboard without voiceover, it opens with at the heading at the top.
- on OLH the open button claims to be a 'menu popup' and it's not!

ajrbyers (approved):

src/transform/xsl/default.xsl
```diff
@@ -1119,7 +1119,7 @@
     </xsl:template>
 
     <xsl:template match="table-wrap/label" mode="captionLabel">
-        <span class="table-label">
+        <span class="table-label" id="tab{count(preceding::table-wrap)+1}-label">
```
mauromsl, line 1122: just being a bit of nit-picking here: should we not want to count labels, rather than table-wraps here, since this template matches `<label>` nodes?

Otherwise the number of hits of the template on `table-wrap/label` could get out of sync with the IDs here. In the worst case, we would simply over-count (i.e IDs won't be sequential) whenever a table-wrap is encountered that does not have a `<label>`.

I don't see any changes in this pr relying on the `id="tab{id}-label"` pattern, so this might be irrelevant

mauromsl (changes requested):
Good work @StephDriver, just a minor nitpick

src/transform/xsl/default.xsl
StephDriver: EDIT - I'm kicking myself - there already is a unique table id used as a prefix for other table elements - so I'e updated to use that instead of introducing  counter.

---

whether they are sequential or not doesn't matter, as long as they are unique - this is to create a unique reference to that information.

the changes that rely on it aren't referring to `id=tab{id}-label` but `aria-describedby="{id}-label` because `tab{id}` in xml becomes `{id}` in the django and this is being used to identify which table is being closed by the close button of the modal.

It is a good point about not all table-wraps having labels - I'll have a further think.

StephDriver: Ok, after further investigation:

As far as I can tell, every table will have a `table-wrap` in a one to one relationship, but whether the `table-wrap` includes a label is less clear - it could have none, one or many, and all three scenarios need handling in order to have a unique identifying html element that can be used as the H2 of the modal, which is then also used to provide the modal-close button with an `aria-describedby`.

My aim is that:
- one and only one - use the table-label
- more than one - use the first table-label
- none - insert a hidden element in the original table (which becomes the H2 of the modal) which simple states 'Table {num}` with sequential numbering.  It seems very unlikely this will happen, but it provides for the edge case.

While investigating this, I also found some problematic screen reader behaviour on material where having opened the modal, one lands on the new sticky close button and can't get the content to read (olh and clean are fine)  I'll going to have a go at fixing it (if quick) and as they weren't accessible before either, if I can't fix it quickly I'll mark that up as a new issue.

StephDriver: ok, the table modal heading was being created in the javascript, so I've made changes there to ensure it can handle the three situations above.  I also found a bug I'd introduced where the table js was being applied to all modals and not just the tables (i.e. to the how to cite modals too). I've not sorted out the material screen reader issue, but as it wasn't accessible before, I think that can now be a standalone issue apart from this, especially as there is other work blocked by this PR.

src/transform/xsl/default.xsl
```diff
@@ -1119,7 +1119,7 @@
     </xsl:template>
 
     <xsl:template match="table-wrap/label" mode="captionLabel">
-        <span class="table-label">
+        <span class="table-label" id="{concat(parent::table-wrap/@id, '-label')}">
```
mauromsl: As discussed we need to make sure we defend against `table-wrap` not having an `@id` to match, since the resulting span ids would just read `"-label"` and would be identical. We could use the previous implementation you had, as long as we count preceding `table-wrap` with `label` combinations

`id="tab-{count(preceding::table-wrap[label])+1}-label"`

mauromsl: or, to be more explicit: preceding `label`s whose parent is a `table-wrap`
`count(preceding::label[parent::table-wrap])`

mauromsl (approved):

PR #4905, opened 2025-08-11
a11y clean - update loops to non-visible list syntax and style

### LINKED ISSUE #4890: Clean Loop list A11y
There are places where lists of items are only shown visually, and not included as lists in the HTML (e.g. separated by `<br/>` or commas).  These should be marked up as lists.  

Lists are important syntax where loops have been used.  Visually a user may know 'this is a handful of items' but for a screen reader user, using list syntax tells them how many items, and where they are in the list of items, as well as allowing them to skip quickly past the list where an unknown length of comma separated items would only provide that visually.

So:
- search for all places in the Clean theme where loops are used and ensure  it is marked up as a list. 

Note: to keep the visual consistent, there are already styles from the other list A11y work for inline lists, and those separated by specific characters (e.g. a bar) (see `clean.css`).

Consider: whether templates for common list types could be useful.

StephDriver:
closes #4890

### DIFF
--- a/src/templates/common/elements/journal/article_authors_br.html
+++ b/src/templates/common/elements/journal/article_authors_br.html
@@ -1,3 +1,4 @@
+<!--This template is deprecated as of 1.9.-->
 <ul class="list-no-bullet">
   {% for author in article.frozen_authors.all %}
     <li>
--- a/src/templates/common/elements/journal/article_authors_br.html
+++ b/src/templates/common/elements/journal/article_authors_br.html
@@ -1,8 +1,7 @@
-<p>
+<ul class="list-no-bullet">
   {% for author in article.frozen_authors.all %}
-    {% if forloop.first == False %}
-      <br>
-    {% endif %}
-    {% include "common/elements/journal/article_author_display.html" %}
+    <li>
+      {% include "common/elements/journal/article_author_display.html" %}
+    </li>
   {% endfor %}
-</p>
+</ul>
--- a/src/themes/clean/assets/css/clean.css
+++ b/src/themes/clean/assets/css/clean.css
@@ -690,12 +690,6 @@ span + .btn-background {
     text-decoration: underline;
 }
 
-.no-bullet-list{
-    list-style-type: none;
-    padding-left: 0;
-    margin-left: 0;
-}
-
 /*
  * WCAG 2.4.7 Focus Visible - Two-Color Focus Indicator (Technique C40)
  * Requires at least 1:9 contrast ratio between the two focus colours. 
--- a/src/themes/clean/templates/journal/article.html
+++ b/src/themes/clean/templates/journal/article.html
@@ -76,16 +76,16 @@
                 {% endif %}
                 {% if article.is_published or proofing %}
                     <div class="d-lg-none d-xl-none d-md-none">
-                            <h2>{% trans 'Files' %}</h2>
-                            <p>
-                            {% if galleys %}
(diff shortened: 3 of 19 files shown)


src/themes/OLH/assets/scss/app.scss
```diff
@@ -1483,3 +1483,48 @@ span + .btn-background {
...
 }
+
+input + div .btn-background {
+    height: 2.77778rem;
+}
+
+span + .btn-background {
+    height: auto;
+    min-height: 1.8rem;
+}
```
joemull: Below this point, it *should* be safe to move this to `common.css`

src/themes/clean/assets/css/clean.css
```diff
@@ -649,10 +649,40 @@ span + .btn-background {
...
+
+.bar-sep li:not(:last-child)::after {
+    content: " | ";
+}
+
+.list-no-bullet{
+    list-style: none;
+    margin: 0;
+    padding: 0;
 }
```
joemull: Move this to `common.css`?

src/templates/common/elements/affiliations_br.html
```diff
@@ -1,8 +1,7 @@
...
-    {% include "common/elements/affiliation_display.html" %}
-    {% if not forloop.last %}
-      <br>
-    {% endif %}
+    <li>
+      {% include "common/elements/affiliation_display.html" %}
+    </li>
   {% endfor %}
-</div>
+</ul>
```
joemull: This template was deleted in #4960 on the 1.8 release branch, so it may be worth restoring these changes to avoid a conflict.

src/templates/common/elements/journal/article_authors_br.html
```diff
@@ -1,8 +1,7 @@
...
-    {% if forloop.first == False %}
-      <br>
-    {% endif %}
-    {% include "common/elements/journal/article_author_display.html" %}
+    <li>
+      {% include "common/elements/journal/article_author_display.html" %}
+    </li>
   {% endfor %}
-</p>
+</ul>
```
joemull, line 8: I should have deleted this in #4960 but it looks like i accidentally left it in. I *did* remove all includes to it, so it should be safe to mark as deprecated.

src/themes/clean/templates/journal/article.html
```diff
@@ -345,11 +345,11 @@ <h2>{% trans 'CRediT Roles' %}</h2>
...
-                              {% for record in credit_records %}
-                                {{ record }}{% if not forloop.last %}, {% endif %}
-                              {% endfor %}
-                            </span>
+                            {% for record in credit_records %}
+                                <ul class="list-inline comma-sep">
+                                    <li>{{ record }}</li>
+                                </ul>
+                            {% endfor %}
                           </li>
```
joemull, line 356: The changes to lines 348-352 may be worth restoring to avoid conflicts with #4960, where this whole CRediT section has been removed.

src/themes/material/assets/mat.css
```diff
@@ -830,17 +830,40 @@ a {
...
+}
+
+.bar-sep li:not(:last-child)::after {
+    content: " | ";
+}
+
+.list-no-bullet{
+    list-style: none;
+    margin: 0;
+    padding: 0;
```
joemull: Safe to consolidate in `common.css`?

joemull (approved):

Issue #1979, opened 2021-01-19
Reviewer file upload page - reminder text to prompt reviewer to check any files are anonymised


rhb123:
A brief reminder on the Reviewer file upload page could be useful to prompt the reviewer to check that their files and comments are fully anonymised before uploading. Although these should always be checked by the editor as well, if the reviewer does this correctly it means that the editor can simply check this and approve it to confirm it has been done, which may save some unnecessary edits. Many thanks.

ajrbyers:
@rhb123 can you supply some example text?

rhb123:
No problem at all - how's this:

Please check that all comments and files have been fully anonymised before uploading your review.

ajrbyers:
Please do.

PR #4910, opened 2025-08-13
Update review_form.html

S-Haime:
Closing #1979 

Been a _long_ time since I've done this so I may have borked it. Tried to include a conditional that this only shows on blind and double-blind, rather than open peer review, but can always take that out.


mauromsl (approved):
Look at you @S-Haime !

PR #4918, opened 2025-08-26
ext-link in refs outputting twice

### LINKED ISSUE #4335: <ext-link> tagged within <ref> results in duplicated hyperlinks
**Describe the bug**
URLs tagged with `<ext-link>` within `<ref>` (most frequently DOIs) are being duplicated in the rendered HTML version of the article. Specifically, it seems that Janeway is rendering one instance of the `<ext-link>` where it was originally tagged, and then duplicating the link at the very end of the `<ref>`. I deduced this because a reference ending in two URLs listed both of the URLs as tagged, in the correct order and with a period and space between them, but then both URLs are immediately repeated (without any spaces) after the second URL - see example 1 in the Screenshots section below. I also found a handful of examples where plain text appears _after_ the `<ext-link>` within the tagged `<ref>`; in each of those cases the link is rendered where it was tagged, followed by the plain text, as expected, and _then_ the duplicate link appeared at the end - See example 2 in Screenshots below.

On Discord, Andy speculated that "I think we're outputing the text inside the `<ext-link>` tags and extraging the `xlink:href` and displaying that. The text inside the `<ext-link>` should be used for the inside of the `<a>` it shouldn't be duplicated." Hypothetically this could be worked around by wrapping the URL in plain `<ext-link>` tags without any attributes (specifically `xlink:href`) but I'm not sure if that would continue to work properly after the fix is implemented. Andy also suggested _only_ using the attributes and leaving the `<ext-link>` tag itself empty. However, I have a suspicion that the link that appears in the correct location is the text wrapped in `<ext-link>`, and that the link duplicated at the end is the extracted `href`, so I'm not sure an empty `<ext-link>` would be an appropriate workaround if we want the text to appear where it's meant to appear.

**Janeway version**
1.5.1, and (at Andy's suggestion) switching back to 1.4.3 fixed the problem.

**To Reproduce**
1. Go to https://journals.publishing.umich.edu/sdi/article/id/5296/
2. Click on "References"
3. Note that every URL in the References section is duplicated

**Expected behavior**
The text inside the `<ext-link>` should be visibly displayed as hyperlinked text, while the `href` should be used as the destination of that hyperlink (and _not_ visibly displayed)

**Screenshots**
Example 1:
```<ref id="R97"><mixed-citation publication-type="other"><person-group person-group-type="author"><string-name><surname>Wu</surname>, <given-names>S.</given-names></string-name>, <string-name><surname>Li</surname>, <given-names>J.</given-names></string-name>, <string-name><surname>Zhang</surname>, <given-names>P.</given-names></string-name>, &amp; <string-name><surname>Zhang</surname>, <given-names>Y.</given-names></string-name></person-group> (<year>2021</year>). <article-title>Natural language processing meets quantum physics: A survey and categorization</article-title>. <source>Proceedings of the 2021 Conference on Empirical Methods in Natural Language Processing</source>, <month>November</month> <day>7</day>&#x2013;<day>11</day>, <fpage>3172</fpage>&#x2013;<lpage>3182</lpage>. Retrieved from <ext-link xmlns:xlink="http://www.w3.org/1999/xlink" ext-link-type="uri" xlink:type="simple" xlink:href="https://aclanthology.org/2021.emnlp-main.254.pdf">https://aclanthology.org/2021.emnlp-main.254.pdf</ext-link>. <ext-link xmlns:xlink="http://www.w3.org/1999/xlink" ext-link-type="doi" xlink:type="simple" xlink:href="https://doi.org/10.18653/v1/2021.emnlp-main.254">https://doi.org/10.18653/v1/2021.emnlp-main.254</ext-link></mixed-citation></ref>```
image

Example 2:
image

## Front-end Issues

**Additional context**

### PR DESCRIPTION
**Pull request by StephDriver**
closes #4335

### DIFF
--- a/src/transform/xsl/default.xsl
+++ b/src/transform/xsl/default.xsl
@@ -1860,7 +1860,7 @@
   <xsl:template match="mixed-citation">
       <!-- Render each mixed-citation as-is https://jats.nlm.nih.gov/archiving/tag-library/1.1/element/mixed-citation.html -->
       <!-- Only exceptions are that we want titles <source> in italics and hyperlinked uris elements-->
-      <xsl:apply-templates select="source | node()" mode="nscitation"/>
+      <xsl:apply-templates select="source | node()[not(self::ext-link)]" mode="nscitation"/>
       <xsl:apply-templates select="ext-link"/>
   </xsl:template>
 

### CODE REVIEW

File: src/transform/xsl/default.xsl [hunk 1]
```diff
@@ -1860,7 +1860,7 @@
   <xsl:template match="mixed-citation">
       <!-- Render each mixed-citation as-is https://jats.nlm.nih.gov/archiving/tag-library/1.1/element/mixed-citation.html -->
       <!-- Only exceptions are that we want titles <source> in italics and hyperlinked uris elements-->
-      <xsl:apply-templates select="source | node()" mode="nscitation"/>
+      <xsl:apply-templates select="source | node()[not(self::ext-link)]" mode="nscitation"/>
       <xsl:apply-templates select="ext-link"/>
```
mauromsl, line 1864: I think the correct solution here is to remove the redundancy on line 1864. The `nscitation` template for `ext-link` adds additional behaviour I think we want to preserve

mauromsl (changes requested):
I'm taking this over to fix a few things that have cropped up with links in the XSLT

ajrbyers:
Superseded by https://github.com/openlibhums/janeway/pull/5365

PR #4930, opened 2025-08-29
Fixed an issue were custom form elements of type "upload" were not being saved properly

### LINKED ISSUE #4862: Review file disappears silently
## User story

On two different occasions, a single reviewer has managed to upload PDF files in to the review form that Janeway cannot save. The way they are producing these files must be problematic, but that is not the real problem.

When they upload the files to Janeway, the form submits and no error appears. But when the editor reviews the review assignment, there is no file attached.

An example faulty PDF can be found attached to Freshdesk ticket 10482 (contains PII): https://janeway.freshdesk.com/a/tickets/10482 

Note that the PDF can be fixed by running it through ocrymypdf.

## Proposed solution

1. Introduce a check in Janeway that can detect the problem with this file, either by checking its specs or trying to do the same thing with it that is failing when it is subsequently accessed.
2. Display this problem to the user rather than submitting the form, so that they can take steps to produce a better file.

## Janeway version
1.7.4

mauromsl:
closes #4862 

We had initially discussed removing this feature, however, as I was working on that, I realised removing the feature in a bacwards-compatible manner was actually harder than finish this feature off, so I did that. Since we know other platforms do support this feature, I figured we might want to preserve it to also ease their migration process.


src/review/views.py
```diff
@@ -3341,3 +3345,13 @@ def reviewer_shared_review_download(request, article_id, review_id):
             )
 
     raise Http404("You do not have permission to download this file.")
+
+
+def review_attachment_download(request, assignment_id, file_uuid):
```
ajrbyers, line 3351: There is no access control here.

ajrbyers (commented):
On the review summary page this displays without a label:

Image

Same on the editor version:

Image

Also when editing the answer there is no label or link to download:

image

When editing the answer by uploading a new file it fails with a form error:

image

src/review/views.py
mauromsl: > There is no access control here.

The access control is the UUID, just like we do for review assignments with secret codes, since reviewers can complete assignments while not being authenticated through django auth. Otherwise we'll need multiple views / very complex decorator logic for each access type: The submitting reviewer (when authenticated), submitting reviewer (using a secret code), fellow peer reviewers when the peer review is shared, the author during revisions, the editor and public users for open review.

ajrbyers (approved):

PR #4936, opened 2025-09-01
button a11y

### LINKED ISSUE #4854: Button accessibility updates
When updating links, I found some links that should be buttons and vice-versa.  Opening up a new issue to review buttons afterwards (both those identified during the link work, and those already listed as buttons, as some probably need to be links instead).

this work needs to wait until #4221 is complete.

StephDriver:
closes #4854

### DIFF
--- a/src/core/views.py
+++ b/src/core/views.py
@@ -2791,6 +2791,12 @@ class GenericFacetedListView(generic.ListView):
     # None or integer
     action_queryset_chunk_size = None
 
+    def get(self, request, *args, **kwargs):
+        if "clear_all" in request.GET:
+            return redirect(request.path)
+        else:
+            return super().get(request, *args, **kwargs)
+
     def get_paginate_by(self, queryset):
         paginate_by = self.request.GET.get("paginate_by", self.paginate_by)
         if paginate_by == "all":
--- a/src/themes/OLH/templates/elements/journal/article_list_filters.html
+++ b/src/themes/OLH/templates/elements/journal/article_list_filters.html
@@ -7,7 +7,7 @@
         <form method="GET" id={{ facet_form.id }}>
             {{ facet_form|foundation }}
             <button action="" class="button" type="submit">{% trans 'Apply' %}</button>
-            <button type="button" class="button warning" onclick="window.location.href='{{ request.path }}'">{% trans 'Clear all' %}</button>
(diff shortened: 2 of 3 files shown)


StephDriver:
- updated OLH narrow screen menu toggles to buttons with aria.
- end of Monday - working on clean pagination. updated to buttons, but need to get the CSS working as previously provided by materialize targetting `a`.
- next: OLH `citation_modals.html` - use buttons to open modals.  (arguably navigation, but as you don't actually move away from the page, but also can't interact with other parts of the page, a button is clearer).

StephDriver:
> Another smallter thing--is there a reason you're using the inline event handler attribute onclick rather than writing JS in a script block? We don't have a consistent style yet obviously, but it would be good to move toward one eventually, so want to understand your reasoning.

Because eons ago as a newbie on Janeway who knew JS - I was told to write JS as a last resort.  I took that as a Janeway approach - but from your comment, I'm now wondering if it was a 'don't rely on using JS and learn the other technologies.

StephDriver:
> is there anywhere you have written about the a11y distinction you are drawing between a and button?

Yes on the [Links A11y work](https://github.com/openlibhums/janeway/pull/4802#issuecomment-2959466688), where I first identified the need for a separate button issue.  It's also in or soon will be in the codex.

joemull (changes requested):
OK, I concede we should use a button here. Thanks for drawing that distinction--I was not super aware of its practical implications for roles and keystroke behavior before, but I read a bit and I think I get it now, beyond just the abstract distinction of action vs. navigation.

I still don't love the use of the `onclick` attribute, because this is generally pretty hacky and mixes behavior and markup in a way we only exceptionally do in JW so far (notwithstanding that we will do so with HTMX sometime soon).

I think if we were to do this properly (and we will eventually have to change *LOTS* of forms on the back office if we impose the `button` / `a` distinction there), we would let the button submit the form (i.e., do not override default button behavior), and handle it in the `get` method on the class-based view `GenericFacetedListView`. Something like:

```html
<button name="clear_all">Clear All</button>
```

```py
class GenericFacetedListView(generic.ListView):
...
    def get(self, request, *args, **kwargs):
        if `clear_all` in request.GET:
            return redirect(request.path)
        else:
            return super().get(request, *args, **kwargs)
```

This would effectively clear all the other query parameters on the GET request, which would clear the filter selections for the user.

joemull (approved):

PR #4953, opened 2025-09-08
setup a11y info page

### LINKED ISSUE #4694: Front of House Accessibility Information Page
We believe readers should have easy access to accessibility information and signposting for accessibility related queries.  

We propose creating a standard static page for journals and presses that provides accessibility information to be included as one of the site options.  This idea grew from a team discussion about #4510 which was about providing accessibility information on downloadable files.

### Functionality
It should:
- be navigable from all Front of House pages.
- include information we have written about the accessibility of the journal theme in use, including downloadable files.  It must be clear whether information is from Janeway, or customised.
- signpost users to report a11y bugs or raise questions.  
- allow journal managers to include their own accessibility information, especially those who use custom styling or have extended the themes.
- at press level, there should be an option which pulls together all the custom journal a11y statements while including the Janeway information just once, (and only those themes in use on the press).

### Example User Stories:

**As a reader who has encountered an accessibility barrier**, I need to be able to determine if it is 'my end' or 'theirs' in order to know what to do about it.  If the issue is with the site, then I would like to be able to report it, but if it is difficult to do so, I may not bother.

**As someone working with students, some of whom have access needs**, I wish to know whether a resource may be accessed by my students.

**As an author writing about disability or disabled people**, I need to know the journal I am submitting to will not publish my article in a format excludes people with that disability from reading the article.

**As journal staff** we wish to maximise our reach and readership.

**As journal staff** our parent organisation has specific requirements for accessibility statements with which we must comply.

**As journal staff** with limited resources, we would like to provide accessibility information but to not have the time, resources or expertise to do so, and likewise as a small team, we are using Janeway without custom modifications.

**As developers on Janeway** we want to know about access barriers so that we can work to reduce them.  In the spirit of `nothing about us without us` we particularly wish to know about barriers from the people who have been affected by them.

StephDriver:
closes #4694

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,5 +7,4 @@ src/files/issues
 src/files/press
 src/files/temp
 src/files/xsl
-src/plugins/*
 
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -6,6 +6,7 @@ RUN apt-get install -y pylint
 RUN apt-get install -y gettext
 RUN pip3 install -r requirements.txt --src /tmp/src
 RUN pip3 install -r dev-requirements.txt --src /tmp/src
+RUN find "/vol/janeway/src/plugins/" -print -iname "*requirements.txt" -exec pip3 install -r {} --src /tmp/src \;
 RUN pip3 install mysqlclient
 RUN if [ -n "$(ls -A ./lib)" ]; then pip3 install -e lib/*; fi
 RUN cp src/core/janeway_global_settings.py src/core/settings.py
--- a/requirements.txt
+++ b/requirements.txt
@@ -54,7 +54,7 @@ Pillow==10.2.0
 python-crontab==2.2.3
 python-dateutil==2.8.1
 python-docx==0.8.11
-python-magic==0.4.13
+python-magic==0.4.27
 pytz==2024.1
 requests==2.32.4
 six==1.16.0


src/press/models.py
```diff
@@ -184,6 +184,17 @@ class Press(AbstractSiteModel):
...
+        blank=True,
+        null=True,
+        verbose_name="Press Accessibility Information",
+        help_text="This is information about the accessiblity of user-content.  It will appear on the accessibility page under the press name.",
+    )
+    allow_journal_ally_info = models.BooleanField(
+        default=False,
+        verbose_name="Allow Journal Accessibility Information",
+        help_text="If enabled, journals can have their own accessibility information as well.",
+    )
```
joemull: String-based fields should not have `null=True`, generally speaking, because it makes their data type less predictable in Python code, and it can result in `None` showing up in templates unless you remember to override it with the `|default:""` template filter. (I know, a lot of Janeway string fields have it! But we should avoid it for new fields.)

src/templates/admin/elements/forms/group_journal.html
```diff
@@ -93,4 +93,14 @@ <h2>Other</h2>
     {% include "admin/elements/forms/field.html" with field=edit_form.use_credit %}
 </div>
 
+{% if request.press.allow_journal_ally_info %}
+    <div class="title-area">
+        <h2>Accessibility</h2>
+    </div>
+    <div class="content">
+        <p>{% trans "Display journal specific accessibility information." %}.</p>
```
joemull: Note double full-stop

src/journal/models.py
```diff
@@ -197,6 +197,12 @@ class Journal(AbstractSiteModel):
     contact_info = JanewayBleachField(
         null=True, blank=True, verbose_name="Contact Information"
     )
+    journal_a11y_info = JanewayBleachField(
+        null=True,
+        blank=True,
+        verbose_name="Journal Accessibility Information",
+        help_text="This is information about the accessibility of user-content. It will appear on the accessibility page under the journal name.",
+    )
```
joemull: We try to avoid adding new fields to the journal model these days, and we prefer using a `Setting` for things like this. Would this work as a new `Setting`, which can be created by adding to `journal_defaults.json`?

src/press/models.py
joemull: By the way, a field on the press object is best for press-level things like this, so these are good as fields here (even though I say in another comment that we should use a setting at the journal level). We don't have a comparable press-level setting object.

src/templates/admin/elements/forms/group_journal.html
StephDriver: 😩 good spot!

PR #4956, opened 2025-09-09
a11y: lightbox focus return to trigger element on close

### LINKED ISSUE #4239: Figure modal a11y improvements
from #4194 

- Sample 4, item 21. link to enlarge the figure is just 'figure 1' this needs to be more descriptive.
- Sample 4, item 22. having enlarged the figure, the modal is not navigable - no way to dismiss it.
- Sample 4, item 32. similar issues with Table 1 as with Figure 1 above. Cannot navigate to the close button, and end up outside the modal, but without it being dismissed.

StephDriver:
closes #4239

### DIFF
--- a/src/static/common/lightbox/js/lightbox.js
+++ b/src/static/common/lightbox/js/lightbox.js
@@ -31,6 +31,7 @@
   function Lightbox(options) {
     this.album = [];
     this.currentImageIndex = void 0;
+    this.triggeringElement = null; // Store reference to element that opened lightbox
     this.init();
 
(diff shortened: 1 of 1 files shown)


StephDriver:
As this is a modification of Lightbox - if approved, I'd like to submit the change as a PR to them too, and that's why I have included more comments than we'd normally do.

StephDriver:
I have just tested this with teh upcoming vocus visible work, and the close button doesn't show focus, so putting it back into draft.

PR #4960, opened 2025-09-10
Fixed a display bug where the article title was not displayed due to numerous authors

### LINKED ISSUE #4952: OLH theme display bug when many authors have long affiliations
## Problem

When an article has many authors who have long affiliations, the text is too tall for the space provided by the OLH theme:

Image

https://doi.org/10.48694/inggrid.4073

## Proposed solution

Consider moving the author metadata out of the hero image box to another place on the page that is not constrained by height.

joemull:
Fixes #4952.

I've decided on a minimal intervention here, which is to make the author list scrollable when it is taller than 6 rem.

Moving the authors out of the box as I suggested earlier would entail much more work on this theme and does not fit our strategy for upgrading the themes.

I've opted against hiding the email address behind an icon, because I want users to be able to read it, or select and copy it, without having to hover. Even if we did hide the email address, what if an article has 20 authors? We need a solution for displaying this content that does not depend on knowing the maximum length of the author list.

Another possibility, if people don't like the scrolling solution: when there are more than 3 authors, display a compact, comma-separated list of their names (truncated at about 70 words), with a button below that can expand this out into a line-broken list of full author details, potentially in a popover. This would just take more styling work, again, than we now want to spend on the current themes.

### DIFF
--- a/src/static/common/css/common.css
+++ b/src/static/common/css/common.css
@@ -327,3 +327,11 @@ a.twitter-x{
   font-weight: bold;
   line-height: 1.25;
 }
+
+/* Hotfix for creating a scrollbar when
+ * there are numerous authors, due to limited
+ * display space on OLH and clean themes */
+:not(.show-for-small-only) .article-authors-br {
+  overflow-y: scroll;
+  max-height: 6rem;
+}
--- a/src/templates/common/elements/journal/article_authors_br.html
+++ b/src/templates/common/elements/journal/article_authors_br.html
@@ -1,8 +1,10 @@
-<p>
-  {% for author in article.frozen_authors.all %}
-    {% if forloop.first == False %}
-      <br>
-    {% endif %}
-    {% include "common/elements/journal/article_author_display.html" %}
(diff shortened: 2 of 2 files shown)


mauromsl (changes requested):
I think displaying the email addresses is making the problem worse than it needs to be. some emails are incredibly verbose and I think using an icon/buttin is fine, plus the email can still be copied. I think the usability factor is how easy it is to read the person name and affiliation, rather than being able to read the email address. It also makes things less busy when using the sidebar
image
_(note how this author has 3 URLS underlying, making it difficult to know what each does, unless you hover and inspect the URL below)_
vs

image

We could also use a similar icon to link out to public profile, similar to how we do for ORCID.

As for the scroll bar on the OLH theme, I personally hate it but I'm unable to provide a better solution, other than moving the author metadata to the sidebar as suggested OR using an expansible list.

joemull:
Thanks @mauromsl --please read the updated description above for an explanation of updates.

PR #4964, opened 2025-09-12
a11y updates to full-text-search pages

### LINKED ISSUE #4839: Full-text search results are not fully navigable
### Describe the inaccessible feature
On all themes, when `ENABLE_FULL_TEXT_SEARCH` is enabled in DJango-level settings, a template is used (`journal/full-text-search.html`) that is not accessible.

* No page heading (h1) is provided
* No pagination for results list
* No indicator of how many results were returned

### Which users are impacted?
 2. Readers (i.e. those accessing the published site but not logging in to Janeway)

### Janeway version
1.7.2

### To Reproduce
Steps to reproduce:
1. Turn on ENABLE_FULL_TEXT_SEARCH and generate indexes
2. Run an article search
3. See page

### Screenshots
Screenshot of inaccessible article search results

## Preferred Solution
Add headings, pagination, and a result count, to mirror the search page you get when ENABLE_FULL_TEXT_SEARCH is turned off

## Additional context
Created as an offshoot from https://github.com/openlibhums/janeway/issues/4242

StephDriver:
closes #4839

### DIFF
--- a/src/journal/views.py
+++ b/src/journal/views.py
@@ -2235,7 +2235,7 @@ def full_text_search(request):
 
     paginate_by = request.GET.get("paginate_by", 25)
     if paginate_by == "all":
-        paginate_by = len(articles) if articles else 25
+        paginate_by = articles.count() if articles else 25
 
     paginator = Paginator(articles, paginate_by)
     page_number = request.GET.get("page")


src/journal/views.py
```diff
@@ -2229,12 +2232,30 @@ def full_text_search(request):
...
+        page_obj = paginator.get_page(1)
+
     template = "journal/full-text-search.html"
     context = {
-        "articles": articles,
+        "articles": page_obj,
+        "page_obj": page_obj,
+        "is_paginated": page_obj.has_other_pages(),
+        "paginate_by": paginate_by,
         "article_search": search_term,
```
joemull: Should this be "search_term" to match the templates?

src/themes/OLH/templates/journal/full-text-search.html
```diff
@@ -10,29 +10,45 @@
...
-            <hr aria-hidden="true" />
-            {% for article in articles %}
-                {% include "elements/journal/box_article.html" with article=article %}
-            {% empty %}
-                <p>{% trans 'No articles to display' %}.</p>
-            {% endfor %}
-
+        <div class="column large-12">
+            {% if search_term %}
+                <h1>Searching for: <i>{{ search_term }}</i></h1>
```
joemull, line 19: Should this variable be changed to match the context from the view, currently called `article_search`? Or, if changing the context variable to `search_term`, this can stay.

src/journal/views.py
```diff
@@ -2229,12 +2232,30 @@ def full_text_search(request):
...
         "keyword": keyword,
         "form": form,
+        "facet_form": form,
+        "order_by_choices": form.fields["sort"].choices,
+        "order_by": sort,
```
joemull: I don't believe `order_by_choices` and `order_by` will have any effect. They are not used in the templates for these views, are they?

src/journal/views.py
```diff
@@ -2229,12 +2232,30 @@ def full_text_search(request):
...
     context = {
-        "articles": articles,
+        "articles": page_obj,
+        "page_obj": page_obj,
+        "is_paginated": page_obj.has_other_pages(),
+        "paginate_by": paginate_by,
         "article_search": search_term,
         "keyword": keyword,
         "form": form,
+        "facet_form": form,
```
joemull: Because the variable `facet_form` is associated with the `GenericFacetedListView`, which is not used here, can we avoid this variable? The only time it is used in the template, you can just use `form` instead.

src/themes/OLH/templates/journal/full-text-search.html
```diff
@@ -10,29 +10,45 @@
...
         </div>
+        <section class="invisible-landmark" aria-label="Results">
+            <div class="columns large-8">
+                {% for article in articles %}
+                    {% include "elements/journal/box_article.html" with article=article %}
+                {% empty %}
+                    <p>{% trans 'No articles to display' %}.</p>
+                {% endfor %}
+                {% if articles %}
+                    {% include "elements/pagination.html" with form_id=facet_form.id %}
```
joemull: This can just be `form_id=form.id`.

src/journal/views.py
```diff
@@ -2220,6 +2220,9 @@ def full_text_search(request):
     search_term, keyword, sort, form, redir = logic.handle_search_controls(
         request,
     )
+
+    form.id = "search_form"
```
joemull: Can we set the id on the `SearchForm` class with `self.id = ...` in `__init__()`?

src/themes/clean/templates/journal/full-text-search.html
```diff
@@ -21,18 +21,31 @@ <h1>{% trans 'Article Search Results' %}</h1>
...
-
-	    </div>
+            <section aria-label="Results">
+                {% for article in articles %}
+                    {% include "elements/article_listing.html" with article=article %}
+                {% empty %}
+                    <p>{% trans "No articles to display" %}.</p>
+                {% endfor %}
+                {% if articles %}
+                    {% include "elements/pagination.html" with form_id=facet_form.id %}
```
joemull: Same as OLH: `facet_form` can be `form`

src/themes/material/templates/journal/full-text-search.html
```diff
@@ -11,29 +11,45 @@
...
-	    </div>
+        <section aria-label="Results">
+            <div class="col m8">
+		        {% for article in articles %}
+		            {% include "elements/article_listing.html" with article=article %}
+		        {% empty %}
+                    <p>{% trans "No articles to display." %}</p>
+		        {% endfor %}
+		        {% if articles %}
+		            {% include "elements/pagination.html" with form_id=facet_form.id %}
```
joemull: `form`

src/journal/views.py
```diff
@@ -2229,10 +2230,25 @@ def full_text_search(request):
             site=request.site_object,
         )
 
+    paginate_by = request.GET.get("paginate_by", 25)
+    if paginate_by == "all":
+        paginate_by = len(articles) if articles else 25
```
mauromsl: In terms of implementation, `qs.count()` is always preferred for querysets over `len(qs)`

I wonder, however, if we really want to support "all" as a feature, since we do not have a dynamic loader for this, we could put a lot of strain on the server. I don't think the user will be happy waiting for potentially thousands of results to be returned and displayed all at once

mauromsl (changes requested):
nice, just one comment regarding the value "all" for the pagination  queryparameter

src/journal/views.py
StephDriver: We use `if paginate_by == "all"` in several places, in commits spanning several developers over several years. 

While you raise a good point, I think addressing it here is out-of-scope, and that should be its own issue, considered through triage etc, as to whether to change our practice and update across the whole codebase.  Using it here is consistent with its use elsewhere.

mauromsl: fair enough, we still want `articles.count()` over `len(articles)` to avoid the implicit query and then we are good to merge

PR #4965, opened 2025-09-15
constrain default journal language to journal languages

### LINKED ISSUE #4313: `default_journal_language`  has no effect
## Problem

The `default_journal_language` setting (`core.models.Setting`) is not used to determine what content to put the language in, so far as I can tell. This is done by Django's `translation.get_language` based on the user's browser settings.

This causes user confusion because they think Janeway isn't working when in fact users are seeing different things than they are anyway.

## Solution
Deprecate the setting and document that for end users.

StephDriver:
closes #4313

### DIFF
--- a/src/utils/language.py
+++ b/src/utils/language.py
@@ -47,12 +47,20 @@ def get_constrained_language():
         return _original_get_language()
 
     # Handle Single Language only journals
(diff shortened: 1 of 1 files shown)


StephDriver:
Discussed at pair programming. Will be changing approach to something on the classes rather than the monkey-patch.

PR #4966, opened 2025-09-16
Make sure login page notice displays well

### LINKED ISSUE #4962: backoffice login container is too narrow when there are custom login messages
**Describe the bug**
When adding a message to the login screen, it can be difficult to read with the new layout of the backoffice login screen

**Janeway version**
v1.8.0-RC-7

**Expected behavior**
The wrapping div should be wider to accommodate for custom text more comfortably, even if the form inputs and buttons themselves remain at their current width

**Screenshots**

Image

joemull:
Fixes #4962.

Screenshot from 2025-09-16 15-57-57
Screenshot from 2025-09-16 15-58-19

### DIFF
--- a/src/templates/admin/core/accounts/login.html
+++ b/src/templates/admin/core/accounts/login.html
@@ -16,7 +16,7 @@
 
 {% block aside %}
   {% if journal_settings.general.display_login_page_notice %}
-    <div class="padding-block-1">
+    <div class="login-notice padding-block-1">
       <div class="card padding-block-2 padding-inline-2">
         {{ journal_settings.general.login_page_notice|safe }}
       </div>


src/templates/admin/core/accounts/login.html
```diff
@@ -14,6 +14,16 @@
   </div>
 {% endblock %}
 
+{% block aside %}
+  {% if journal_settings.general.display_login_page_notice %}
+    <div class="padding-block-1">
+      <div class="card padding-block-2 padding-inline-2">
+        {{ journal_settings.general.login_page_notice|safe }}
+      </div>
+    </div>
```
mauromsl: It might be worth giving this `<div>` a CSS target so we can adjust more easily when needed. I think it'd be safe to give it an `id` attribute but a `css` class for selection would be ok too

joemull: I think if CSS needs to target this via a class, then we can add a class at that time, right? What's the advantage of doing it before we know the use case?

ajrbyers: > I think if CSS needs to target this via a class, then we can add a class at that time, right? What's the advantage of doing it before we know the use case?

I can see Mauro's thinking. If we need to alter it at a customer request we can do so via custom styling plugin if the `<div>` can be targeted..

mauromsl: I've added one so I can pin an RC for tomorrow

joemull: Does the custom styling plugin load on back office templates? More to the point, should it? I don't think we want the complexity of allowing people to customize back office templates in that way, because it will make maintenance and changes painstaking and an exercise in letting people down.

ajrbyers: > Does the custom styling plugin load on back office templates? More to the point, should it? I don't think we want the complexity of allowing people to customize back office templates in that way, because it will make maintenance and changes painstaking and an exercise in letting people down.

Keep forgetting these aren't theme templates anymore. There are other methods for adding custom CSS to the back office if needed. This gives us that option if it's required. It also allows external folks that option. Id probably  push back on a request to add something custom but having the option if we need it is good.

PR #4977, opened 2025-09-22
Material keyboard a11y

### LINKED ISSUE #4226: Material Theme Side Navigation Screen-reader Fixes
From screenreader testing (https://github.com/BirkbeckCTP/janeway/issues/4194), looking at the narrow screen navigation the following issues were identified:

- [x] Sample 2, item 2. home, link, list start - but nothing visible on the page and is a duplicate of the link above.
- [x] Sample 2, item 3. Visually, at phone screen width there is a hamburger menu, but to the screen reader, no such thing exists, but instead the whole uncollapsed menu is read out despite being hidden. These menu items are not identified as a menu or as navigation.
- [x] Sample 2, item 4. items in the menu (articles, submissions, journal policies, about) are read followed by "arrow underscore drop underscore down" and identified as links. `<i class="material-icons right">arrow_drop_down</i>`
- [x] Sample 2, item 8. "Menu, link, end, navigation, end, banner". This is the hamburger menu, and when activated the menu does pop up, but you can't navigate the list, as you already have - what pops up visually is the list that I've just described above, but it has already been navigated, when advancing one reaches the article title instead. Further, there is no way now to dismiss this menu without refreshing the whole page, activating the menu link again does nothing.

### LINKED ISSUE #4549: Material Articles list pagination not keyboard navigable
Found while testing for icon a11y.  The pagination is not keyboard focusable at the bottom of the /articles/ page listing all articles.

The links to pages show up as links in voiceover rotor, but they are skipped when tabbing, and I've tried tabbing through the whole page without managing to land on them.  The content is read out with Voiceover too, so this seems to be keyboard navigation issue only.

template is `elements/pagination.html`.

Expected behaviour:
I should be able to tab into page links after tabbing out of the final keyboard focusable item within an article card. 

image

### LINKED ISSUE #4917: Material Keyboard Navigation Duplicated
2.4.7 Focus Visible

Head point: c96d43d19

Navigation point: all pages

When navigating by keyboard on Chrome desktop view, with the theme ``material``, you must navigate through the nav bar (Home, Articles, Issues, etc.) twice before reaching the main page components below. When navigating through the components a second time, no focus is indicated; perhaps the narrow screen navigation is being included. This is true for navigating all pages in this theme.

StephDriver:
closes #4226 
closes #4917 
closes #4549

### DIFF
--- a/src/themes/material/assets/mat.css
+++ b/src/themes/material/assets/mat.css
@@ -1180,3 +1180,7 @@ p + .btn-background {
       display: inline-block;
       white-space: normal;
   }
+
+  .nav-wrapper a:focus {
+    display: block;
+  }


joemull (approved):
This looks ok to me--I am mostly impressed that you dove into Materialize's JS!

One thing I found when testing is that there is some layout shift when the `keyboard-focusable` class is added to the ancestor, when manually tabbing:

[Screencast from 2025-11-13 14-39-11.webm](https://github.com/user-attachments/assets/5522ba93-c8fb-427b-bf0a-6ad240fe9e69)

I don't think this is a huge thing, but in case it's an easy fix, I just thought I'd mention it. It is also perplexing to me because `outline` should not cause a layout shift at all, which makes me think there is another class or ruleset being added. It is hard to test, though, because the state is controlled by Materialize JS, rather than the browser `:focus-visible` state, which is what I'd expect, and which I'd be able to emulate in dev tools.

src/themes/material/assets/material.js
```diff
@@ -35,3 +36,70 @@ $('a[href^="#"]:not(a[href$="!"])').click(function() {
...
+                var activeElement = document.activeElement;
+                var isFocusInSidenav = sidenavElement.contains(activeElement) || 
+                                     sidenavElement === activeElement;
+                
+                // If focus is not in sidenav and sidenav is open, close it
+                if (!isFocusInSidenav && sidenavInstance.isOpen) {
+                    sidenavInstance.close();
+                    if (originalTrigger && originalTrigger.offsetParent !== null) {
+                        originalTrigger.focus();
+                    }
```
mauromsl, line 87: I'm a bit out of my depth here, so I thought I'd ask. Why is it important to close the sidenav when focus changes out of it?

StephDriver: because it pops over the page and covers it. If you don't close it on change of focus, then when navigating by keyboard, one tabs through the main page underneath the sidenav, unable to see what you're focusing.  Hits the edgecase where you navigate into the nav, and then out again without activating any of the nav items.  Which won't happen as much now we have the skip to main that allows keyboard users to skip past the nav, but I've certainly looked in at the optoins in a menu before deciding I didn't want any of them and wanted to go back ot the main page.

PR #4982, opened 2025-09-23
setup a11y info pages on all three themes

### LINKED ISSUE #4694: Front of House Accessibility Information Page

StephDriver:
closes #4694

### DIFF
--- a/src/utils/install/journal_defaults.json
+++ b/src/utils/install/journal_defaults.json
@@ -5612,6 +5612,7 @@
             "default": ""
         },
         "editable_by": [
+            "editor",
             "journal-manager"
         ]
     }


src/core/models.py
```diff
@@ -1101,6 +1101,13 @@ class Setting(models.Model):
 
     is_translatable = models.BooleanField(default=False)
 
+    is_press_dependent = models.CharField(
+        max_length=100,
+        null=True,
+        blank=True,
+        help_text="If set, this setting is only available when the specified press setting is enabled.",
+    )
```
joemull: Two suggestions for this field definition:

1. Remove the `null=True` parameter since it's a string-based field. String-based fields should have an empty string as the blank value, which is the Django default, not `None`. This makes the data type more predictable in Python code and obviates the need for the `|default:""` filter in templates.
2. The name suggests it's a boolean, because it starts with `is_`, so can we rename it to something like `depends_on_press_field`?

src/core/views.py
```diff
@@ -1091,6 +1091,25 @@ def settings_index(request):
...
 
+    if request.journal:
+        press = request.journal.press
+        filtered_settings = []
+        for setting in settings:
+            if (
+                hasattr(setting, "is_press_dependent")
+                and setting.is_press_dependent
+                and setting.is_press_dependent.strip()
+            ):
```
joemull: Because the field `is_press_dependent` exists on every object instance (even if the JSON doesn't have it), you can structure this as `for setting in settings.filter(is_press_dependent__isnull=False):` or similar depending on how you change the field name and data type (above comment).

src/press/models.py
```diff
@@ -184,6 +184,17 @@ class Press(AbstractSiteModel):
     disable_journals = models.BooleanField(
         default=False, help_text="If enabled, the journals page will no longer render."
     )
+    ally_info = JanewayBleachField(
+        blank=True,
+        null=True,
```
joemull: Can we nix `null=True` for the same reason as above?

src/core/views.py
```diff
@@ -1091,6 +1091,25 @@ def settings_index(request):
...
+                # This setting is press-dependent, check if the press setting allows it
+                press_setting_value = getattr(press, setting.is_press_dependent, False)
+                if press_setting_value:
+                    filtered_settings.append(setting)
+            else:
+                # Non-press-dependent settings are always included
+                filtered_settings.append(setting)
+
+        settings = filtered_settings
```
joemull: I'm not so sure the settings should be invisible at the journal level. We already often have the problem that people don't understand why they can't see something that other people tell them should be there. This would re-create that problem in a new place unless someone knows that these settings are interrelated at the press and journal level. Staff might not always know how to work this feature, or communicate it to the journal editors.

What if instead of disappearing, the settings were read-only and could not be edited?

src/utils/install.py
```diff
@@ -50,6 +50,11 @@ def update_settings(
                 "is_translatable": item["setting"].get("is_translatable"),
             }
 
+            if "is_press_dependent" in item["setting"]:
+                setting_defaults["is_press_dependent"] = item["setting"][
+                    "is_press_dependent"
+                ]
```
joemull: The press-dependency field should be accessed with `.get()`, esp. if it only exists on some of the entries in `journal_defaults.json`, or there will be a `KeyError` when running this on new installations.

src/utils/install/journal_defaults.json
```diff
@@ -5596,5 +5596,24 @@
...
+            "name": "a11y_public_info",
+            "pretty_name": "Accessibility Information for the Public",
+            "type": "rich-text"
+        },
+        "value": {
+            "default": ""
+        },
+        "editable_by": [
+            "journal-manager"
+        ]
```
joemull, line 5617: Just to record what we discussed off GitHub, the `is_press_dependent` field will need to be extended to all items in the JSON file, for consistency and maintainability.

src/core/views.py
StephDriver: I did wonder about this for some time while working on the branch. My feeling was that in this specific instance, no or at least not yet. because I think the accessibility bit needs to be rolled out more slowly. Likely first to Press managers, so they can think about their part, and whether they want journal staff to see this or not. And only then to make it visible.

What I want to avoid is journal staff seeing the setting, and it's also the first time they've seen a setting that is not active.  Particularly as by making this an 'a11y' group setting, it appears by default at the start of the all settings list.

Longer term, yes, I do think they should be visible but inactive and signposted as 'needs X setting on press'.  But I don't want to introduce both inactive settings and accessibility page at the same time.

src/utils/install/journal_defaults.json
StephDriver: I looked at doing this - but the file is not consistent as is - the setting attributes aren't always in the same order  - so making it consistent feels like a stand alone task.

src/core/views.py
joemull: > in this specific instance, no or at least not yet. because I think the accessibility bit needs to be rolled out more slowly

That's fine if you want to tailor the changes to the current use case, but in that case, this should not be written as a generalised feature of all settings. It should instead be just a set of checks that is put in place when these particular setting pairs are dealt with in views. Janeway already has some examples of one setting that's logically dependent on the other, and we just check the one before doing anything with the other. Couldn't you do that?

In order to hide the journal-level settings from the All Settings list, you would need to just add a filter in the queryset that manually checks the appropriate press field and excludes your journal-level settings by name.

If we merged this as is, I'd worry that we are creating a general pattern that would make other settings that adopt this pattern less usable, by hiding things from the user without any way for them to know why they are not seeing what they expect. [Visibility is really foundational to usability](https://www.nngroup.com/articles/visibility-system-status/), especially when we know that Janeway users often have the problem of not seeing what other people see, and being confused by the difference.

src/utils/install/journal_defaults.json
joemull: OK, maybe another argument not to implement this as a general feature of all settings, but just a quirk of the relationship of these particular settings. (See my other comment for more.)

joemull:
BTW, apologies this sat for so long on my plate! I did not realize it was waiting on me.

StephDriver:
Notes:
- removed depends on press setting
- removed migration for that setting

Next:
- replace with other logic to create same effect for a11y info only.
- rebase (to get latest clean footer)

src/utils/install/journal_defaults.json
```diff
@@ -5596,5 +5596,23 @@
...
+            "is_translatable": true,
+            "name": "a11y_public_info",
+            "pretty_name": "Accessibility Information for the Public",
+            "type": "rich-text"
+        },
+        "value": {
+            "default": ""
+        },
+        "editable_by": [
+            "journal-manager"
```
joemull, line 5616: This needs to be editable by the "editor" role as well--not many people use the "journal-manager" role.

Or was there a reason "editor" was omitted?

src/templates/common/elements/a11y/janeway.html
ajrbyers, line 1: This looks oddly out of place and I cant see it referenced anywhere.

StephDriver: This is the Janeway section of the accessibility page -this is where our a11y statement will go, along with the toggle for turning on accessibility mode, details of why you might want to use accessibility mode etc.

ajrbyers (approved):

PR #4986, opened 2025-09-26
 olh screen reader improvements

### LINKED ISSUE #4220: OLH theme: box links used in lists of articles are not accessible
When navigating, for example an [OLH homepage](https://www.iastatedigitalpress.com/mmb/), by headings with a screenreader, the article headings are not described as links, but can be followed as if they are links to that article. When navigating by link, the link appears before the heading and is only described by the article number.

What seems to be happening is that the link itself is a box behind the other content, such that any content within that area of the screen then can be followed as if it is a link, but is not identified as a link.

Two examples of this were identified from screen reader testing #4194 
- Sample 5.1, item 11. when navigating by headings, the article headings are not described as links, but can be followed as if they are links to that article. When navigating by link, the link appears before the heading and is only described by the article number.
- Sample 5.1, item 22. popular article links are also just the article number, same behaviour as above with other articles in regards to the article title not being a link, but acting as if it were a link.

StephDriver:
closes #4220


src/themes/OLH/assets/scss/app.scss
```diff
@@ -1483,3 +1513,74 @@ span + .btn-background {
...
+    li a {
+        margin: calc(-1 * var(--focus-ring-size));
+        padding: var(--focus-ring-size);
+        box-sizing: border-box;
+    }
+
+    p a,
+    span a {
+      display: inline-block;
+    }
```
joemull, line 1621: I don't believe this is needed, since `outline` does not add any margin or padding or size to the element. I commented it out and tested it, and I did not get a layout shift when tabbing through.

It is also causing a bit of a visible gap between the focus ring and the element. Do you think we can remove it safely, or is there another reason for it I'm not seeing?

StephDriver: This was how I got the links that were within other text to remain stationary rather than move around when receiving/losing focus. The main culprit was where text had flowed such that the link had been split across two lines, for some links, this reflow ended up changing when focused. Using this style prevented that.  This was a style choice born of experimenting until something worked rather than theory.

StephDriver: https://github.com/user-attachments/assets/56a6f97d-6328-413a-9646-1a2a5b685afc

joemull: There is something wrong with the .mov file above that is keeping me from watching it, but that sounds fine--i just wanted to make sure this was needed.

src/core/templatetags/counter.py
```diff
@@ -0,0 +1,24 @@
...
+@register.simple_tag
+def get_next_id(prefix="id"):
+    """
+    Returns an incrementing id value unique within a single request/thread.
+    Primiarily for use within templates to create unique element ID
+    for use with aria references.
+    """
+    counter = _get_counter()
+    counter.value += 1
+    return f"{prefix}-{counter.value}"
```
joemull: This is one of those cases where you and I have had the same problem, and we've each solved it in different ways!

About 18 months go I wrote a template tag that serves the same purpose:

https://github.com/openlibhums/janeway/blob/c6f8f5df4194fc96ea00cb9ecf9b2a13096ca748/src/core/templatetags/uuid.py#L1-L9

You can see it used for aria labels in a few places:

https://github.com/openlibhums/janeway/blob/c6f8f5df4194fc96ea00cb9ecf9b2a13096ca748/src/templates/admin/submission/submit_correspondence_author.html#L3-L5
https://github.com/openlibhums/janeway/blob/c6f8f5df4194fc96ea00cb9ecf9b2a13096ca748/src/templates/admin/submission/submit_correspondence_author.html#L24
https://github.com/openlibhums/janeway/blob/c6f8f5df4194fc96ea00cb9ecf9b2a13096ca748/src/templates/admin/submission/submit_correspondence_author.html#L37

It looks like I did not document it, because this is before we had fixed or found a new home for the dev docs, which is probably why it was not easily discoverable.

So, let's think about which one to keep and document.

I think each approach has pros and cons:
* I like that your approach allows a prefix to be prepended to the ID. This adds to readability and avoids having to concatenate strings in the template, like I've had to do.
* I like that your approach results in shorter id values, so it's more legible to humans when inspecting HTML.
* I like that my approach avoids the complexity of depending on threading.
* I like that my approach uses UUIDs, which are a common way of producing unique IDs for interlinking HTML, in my experience. The difference is that most people shorten the UUID to 7 characters, I believe, or they use a different hash that is shorter (something like 10-15 chars). I tried to suggest shortening the uuid4 when I first wrote it but it was not clear to the reviewer, so we went with the full length.

What if:
- We keep my approach, but modify it in two ways.
    - We add an optional prefix.
    - We shorten the ids produced by my approach.
    - We document it (I can add a markdown file on this PR if that sounds OK with you).

Alternatively:
- We merge this as is, and refactor the two approaches together into one approach later. This would be more expedient, and I'm OK with it.

src/themes/OLH/assets/scss/app.scss
```diff
@@ -1548,3 +1578,139 @@ span + .btn-background {
...
+.medium-4 .olh-banner,
+.large-4 .olh-banner {
+  max-height: 216px;
+}
+.medium-6 .olh-banner,
+.medium-8 .olh-banner,
+.large-6 .olh-banner,
+.large-8 .olh-banner {
+  max-height: 288px;
+}
```
joemull: The CSS for the OLH banner has been duplicated, probably in rebasing. Lines 1582-1645 (your commit) are identical to lines 1517-1580 (my commit) and so can be removed.

StephDriver: thanks - I keep getting that kind of error on rebasing css, I thought I'd spotted it this time and removed - but must have done it more than once!

src/core/templatetags/counter.py
StephDriver: Yes - looks like that's exactly what happened. I did initially use a UUID but I was concerned about just how many could be needed on the same page if this becomes standard for element IDs (for aria) and that led me to look into how resource intensive it was, which led me to threading as being cheaper.  But this is the edge of my knowledge, so I could well have misunderstood that. 

I was concerned that the UUID does make for less-readable code, and hadn't though about shortening them.  That does sound like a good compromise, as long as they remain unique. I can't immediately see how that would work - as if an id is known to be unique at 14chars, then taking say the first 4 characters wouldn't assure uniqueness.

To my mind generating UUID was a bit of an overkill when the id only needed to be unique for the short time that a page was loaded before moving away, and by a single user.

PR #4988, opened 2025-09-29
Journals can have a publishing status in addition to being hidden from the press

### LINKED ISSUE #3970: Designate some journals as archived
As a press manager, I'd like to designate some journals as archived, so that I can list them separately from the actively publishing journals on the [journal list page](https://www.openlibhums.org/journals/).

This was requested by Caroline during the Hourglass theme launch.

I think we can handle this and future journal status functionality with a new "status" field on journals that has a dropdown / enumeration for the different states a journal can be in, such as:

* hidden
* coming_soon
* active
* archived

We could then change `journal.hide_from_press` to an object property that uses status.

Going forward, we can use properties like `public_journals` on the press to get the list of journals we want.

A Django object manager, but that might be overkill and confusing for something that is more properly an enumeration.

### LINKED ISSUE #4418: Published articles count may include hidden journals
## Problem
The `published_articles` method on `Press` may include articles in journals which are marked hidden from the press.

## Solution
Exclude such articles from the count.

### LINKED ISSUE #4701: OAI-PMH ListSets endpoint reveals and links to hidden journals and their content
**Describe the bug**
When setting up a test or placeholder journal (e.g for a migration) the OAI-PMG feed will ignore the "Hide from press" setting and display all journals and issues when using the `ListSets` verb 

**Janeway version**
v1.7.4

**To Reproduce**
1. Setup a test journal
2. Toggle On the "hide from press" checkbox. 
3. visit `/api/oai/?verb=ListSets` at the press level
4. See hidden journal being listed

**Expected behavior**
Hidden content should be hidden on press level OAI feeds

**Additional context**
We can consider having more granular control for hidden content (e.g "hide from press site", "hide from REST", "hide from OAI"...)

### LINKED ISSUE #4987: Random Featured Journals causes TypeError
At Press, if homepage elements of `journals` or `journals_and_html` are included and configured to display `random featured journals` then the homepage does not load due to a TypeError.

Found during development, on master at 204116d09

```
  File "/venv/Janeway/lib/python3.12/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/venv/Janeway/lib/python3.12/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "aneway/src/press/views.py", line 70, in index
    element_context = function(request, homepage_elements)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "janeway/src/core/homepage_elements/journals_and_html/hooks.py", line 34, in yield_homepage_element_context
    featured_journals = get_random_journals()
                        ^^^^^^^^^^^^^^^^^^^^^
  File "janeway/src/core/homepage_elements/journals_and_html/hooks.py", line 23, in get_random_journals
    return random.sample(set(journals), sample_size)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/homebrew/Cellar/python@3.12/3.12.10/Frameworks/Python.framework/Versions/3.12/lib/python3.12/random.py", line 413, in sample
    raise TypeError("Population must be a sequence.  "
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Exception Type: TypeError at /
Exception Value: Population must be a sequence.  For dicts or sets, use sorted(d).

```

joemull:
Closes #3970.
Fixes #4701.
Fixes #4418.
Fixes #4987.

For an explanation of decisions see the [documentation](https://github.com/openlibhums/janeway/blob/7117805f737517c4d6a0447477e8614016a271b5/docs/md/journal-visibility-and-publishing-status.md).


src/api/oai/views.py
```diff
@@ -2,6 +2,8 @@
 A django implementation of the OAI-PMH interface
 """
 
+import warnings
```
ajrbyers: Is this imported to fix something else? It doesn't look like its used in the diff.

src/comms/views.py
```diff
@@ -208,8 +209,20 @@ def news_list(request, tag=None, presswide=False):
     """
 
     news_objects = models.NewsItem.active_objects.all()
-
-    if not presswide or request.model_content_type.model != "press":
+    all_tags = models.Tag.objects.all()
+
+    if presswide or request.model_content_type.model == "press":
+        Journal = apps.get_model("journal.Journal")
```
ajrbyers: Is there a circular import if you import journal.models in this module?

src/comms/views.py
```diff
@@ -222,6 +235,14 @@ def news_list(request, tag=None, presswide=False):
         )
         tag = get_object_or_404(models.Tag, text=unquoted_tag)
 
+    all_tags = (
```
ajrbyers, line 239: Why was this moved?

src/core/models.py
```diff
@@ -928,8 +929,13 @@ def published_articles(self):
         )
         request = utils_logic.get_current_request()
         if request and request.journal:
-            articles.filter(journal=request.journal)
-
+            articles = articles.filter(journal=request.journal)
+        else:
+            articles = articles.filter(journal__hide_from_press=False)
+        Journal = apps.get_model("journal.Journal")
```
ajrbyers, line 957: Same questions as before, why are we using apps.get_model here?

src/identifiers/logic.py
```diff
@@ -50,6 +51,10 @@ def register_batch_of_crossref_dois(articles, **kwargs):
 
     use_crossref, test_mode, missing_settings = check_crossref_settings(journal)
 
+    Journal = apps.get_model("journal.Journal")
```
ajrbyers: Same questions as before.

src/journal/models.py
```diff
@@ -267,6 +267,19 @@ class Journal(AbstractSiteModel):
     # Boolean to determine if this journal should be hidden from the press
     hide_from_press = models.BooleanField(default=False)
 
+    class PublishingStatus(models.TextChoices):
+        ACTIVE = "active", _("Active")
```
ajrbyers, line 358: Should these definitions be constants like https://github.com/openlibhums/janeway/blob/master/src/review/const.py ?

src/themes/clean/templates/press/press_journal_set.html
```diff
@@ -0,0 +1,56 @@
+{% load static %}
+{% load settings i18n %}
+
+<div class="col-lg-8 col-md-10 journal-container">
+  {% for journal in journals %}
+    {% setting_var journal 'disable_journal_submission' as submission_disabled %}
```
ajrbyers: This is already in the context `{{ journal_settings.general.disable_journal_submission }}`. No need to fetch it again.

src/utils/importers/ojs.py
ajrbyers, line 1: This doesn't seem to have anything to do with this PR.

src/utils/importers/shared.py
ajrbyers, line 1: This doesn't seem to have anything to do with this PR.

src/utils/importer.py
ajrbyers, line 1: This doesn't seem to have anything to do with this PR.

src/utils/logic.py
```diff
@@ -3,6 +3,7 @@
 import hmac
 from urllib.parse import SplitResult, urlencode, urlparse, unquote
 
+from django.apps import apps
```
ajrbyers: This is imported and not used.

ajrbyers (changes requested):
Generally this is great work. I've only done a code review here as I'd like @mauromsl to look over it and then have it come back to me.

There is one major issue I see with this PR. Could you share the reason you’re using apps.get_model throughout this PR? I can see one or two places where it makes sense and in others where it seems a direct import would be better and more consistent.

I've made some comments and suggestions throughout, if you have any queries let me know.

src/api/oai/views.py
joemull: Just cruft from making a change and then deleting it

src/comms/views.py
joemull: Was being overly defensive due to journal models' tendency to create circular imports. Will change this and other instances where I can.

src/comms/views.py
joemull: So that it could be filtered by the journal. I realized that it wasn't being filtered, so it could potentially include junk tags from journals that are hidden from the press.

src/journal/models.py
joemull: Maybe--what's the advantage? Would it support translation?

I like defining them close to the field on the model, for readability, and it also removes the need to import anything other than the model. Django docs:

https://docs.djangoproject.com/en/5.2/ref/models/fields/#enumeration-types

I've also used this pattern in a few other places already:

https://github.com/openlibhums/janeway/blob/7117805f737517c4d6a0447477e8614016a271b5/src/utils/models.py#L416-L420

https://github.com/openlibhums/janeway/blob/7117805f737517c4d6a0447477e8614016a271b5/src/core/models.py#L2791-L2795

src/themes/clean/templates/press/press_journal_set.html
joemull: Thanks--I was just moving this into a new file, but I will make the change.

src/utils/importers/ojs.py
joemull: I was going through the `utils/importers` files with a fine-toothed comb to look for instances of journal-related querysets, to see if I needed to make changes related to `hide_from_press` or the test publishing status. As I was doing so, I remembered that we agreed these were deprecated during the ROR work. I'm not misremembering, am I? So that I didn't have to re-check them in a future piece of work on something else, I marked them deprecated.

joemull:
> I've made some comments and suggestions throughout, if you have any queries let me know.

Thanks Andy--I've made some changes and added some replies above.

src/utils/importers/ojs.py
ajrbyers: Its fine - just wanted to check they hadn't been pulled in to this PR by accident.

src/core/models.py
mauromsl: This is actually best practice to avoid circular imports, I like it and we should use it more often (but only where code is unit tested, as the lack of a top level import statement will no longer be enough to catch a bad import exception)

src/journal/models.py
mauromsl: Either works. Some constants make sense to exist inside a model because they are "owned" by a model. Others are shared across models and other APIs and makes sense for them to exist in an external namespace, easy to reach by all without leading to circular dependencies  (think review decisions)

In case of doubt, an external namespace will be more resilient to future changes.

src/press/models.py
```diff
@@ -441,20 +428,117 @@ def navigation_items_for_sitemap(self):
...
+        """
+        Get all journals that are visible to the press
+        and marked as 'Coming soon' in the publishing status field.
+        """
+        Journal = apps.get_model("journal.Journal")
+        return self.apply_journal_ordering(
+            self.public_journals.filter(
+                status=Journal.PublishingStatus.COMING_SOON,
+            )
+        )
```
mauromsl: These sound more like either individual custom managers or a single custom manager with multiple methods. The rule of thumb is to ask oneself, do my model instances need to invoke this property or is this a special way of querying model-instances? If the answer is the latter, it is a good indication that you are writing a Manager method rather than a Model property.

src/themes/clean/templates/press/press_journal_set.html
```diff
@@ -0,0 +1,56 @@
...
+                 src="{% if journal.default_cover_image %}{{ journal.default_cover_image.url }}{% endif %}"
+                 alt="{% if journal.name != " " and journal.name != "" %}{{ journal.name }}{% else %}Unnamed Journal{% endif %}">
+        </div>
+        <div class="col-lg-9 col-md-8 journal-description-box">
+            <span class="d-none">{{ journal.name|lower }}</span>
+            {% if level == "deep" %}
+              <h3 class="journal-name">{{ journal.name }}</h3>
+            {% else %}
+              <h2 class="journal-name">{{ journal.name }}</h2>
+            {% endif %}
```
mauromsl: This doesn't feel very intuitive to me and works for this very specific set of templates, but I don't think would be very reusable (I would like to avoid someone adding a `level == "very-deep" entry here). Can we use `header == "h3"` or similar, if that is the intention?

mauromsl (changes requested):
Thanks, really appreciate the clear documentation and tests.

Added a couple small comments in line but generally looks good

src/themes/clean/templates/press/press_journal_set.html
joemull: I like `header = "h3"` too, but we've got 20 or so templates that use the "level" + "deep" convention now, so I was being consistent with those. They were introduced by Steph during the breadcrumbs updates. How about you introduce a separate issue to change them?

src/press/models.py
joemull: Yeah, I agree a manager works well. I've gone ahead and switched everything over to one.

However, I am now running into a bug where the `order_by` method is not recognizing that `journal_name` has been added as an annotation. This didn't happen when it was a press method, and I don't think I've introduced anything that would have changed the behavior.

I've pushed a WIP--do you know what's going on?

My only guess is it's something with the order of applying ordering in the manager methods, maybe similar to whatever was behind this use of `extra()` rather than `order_by()`:

https://github.com/openlibhums/janeway/blob/1978afbc64fee52c803071115925c0ea6b60e0b3/src/core/model_utils.py#L282-L290

src/journal/models.py
joemull: Alright, since this one is owned by the field and isn't used elsewhere yet, I'd like to keep it here.

src/themes/clean/templates/press/press_journal_set.html
ajrbyers: I'll second this - its consistent with the the template base as it is but lets open an issue to examine it and potentially change it.

mauromsl (approved):

src/press/models.py
joemull: @mauromsl would appreciate a second glance at this comment.

src/journal/models.py
```diff
@@ -97,6 +98,97 @@ def issue_large_image_path(instance, filename):
...
+        """
+        localized_column = build_localized_fieldname(
+            "value",
+            settings.LANGUAGE_CODE,  # Assumed to be 'en' in default config
+        )
+        name = core_models.SettingValue.objects.filter(
+            journal=models.OuterRef("pk"),
+            setting__name="journal_name",
+        )
+        journals.annotate(
```
ajrbyers: `annotate` returns a new queryset rather than changing `journals` in place, and the result isn't captured here so `order`_by below runs on  the un-annotated `journals` and can't find `journal_name`. The result of the annotate needs to be assigned to `journals`.

ajrbyers (changes requested):

src/press/views.py
```diff
@@ -139,10 +142,12 @@ def conferences(request):
     """
     template = "press/press_journals.html"
 
-    journal_objects = journal_models.Journal.objects.filter(
-        hide_from_press=False,
-        is_conference=True,
-    ).order_by("sequence")
+    journal_objects = request.press.apply_journal_ordering(
```
ajrbyers: Does `apply_journal_ordering` exist for the `press? object` I couldn't find it.

src/comms/views.py
```diff
@@ -208,8 +209,19 @@ def news_list(request, tag=None, presswide=False):
...
-    if not presswide or request.model_content_type.model != "press":
+    if presswide or request.model_content_type.model == "press":
+        press_visible_journal_pks = [
+            journal.pk
+            for journal in journal_models.Journal.objects.filter(
+                hide_from_press=False,
+            )
+        ]
+        news_objects = news_objects.filter(
+            object_id__in=press_visible_journal_pks,
```
ajrbyers, line 224: If this is filtering by journal it should have a content_type as a press pk and journal pk could collide.

ajrbyers (commented):
Overall, great, I had two comments and then please fix conflicts.

🥇

joemull:
I've noticed two more problems while fixing the latest bugs:

1. The conferences page is not working, in several ways, because it shares a template with the journals page. The header says "Journals" and the query set is not picked up by the template.
2. There is a field called `is_archived` on the journal which needs to be rolled into the publishing status enumeration.

joemull:
This is now ready for review again.

src/comms/views.py
```diff
@@ -208,8 +210,21 @@ def news_list(request, tag=None, presswide=False):
     """
 
     news_objects = models.NewsItem.active_objects.all()
+    all_tags = models.Tag.objects.all()
```
ajrbyers: Not used before being overwritten below.

src/comms/views.py
```diff
@@ -208,8 +210,21 @@ def news_list(request, tag=None, presswide=False):
...
 
-    if not presswide or request.model_content_type.model != "press":
+    if presswide or request.model_content_type.model == "press":
```
ajrbyers: Is the intent with this change that the press _always_ shows journal level news?

ajrbyers (changes requested):
See comments

src/comms/views.py
joemull: No, seems there is a logic bug. Will fix it by changing "or" to "and".

ajrbyers (approved):

joemull:
I will rebase to fix the conflicts now.

PR #4989, opened 2025-09-30
Introduces changes for clarity against clean for sanity checks.

ajrbyers:
This doesn't close any issues on its own and will be closed without being merged. This issue has been opened soley for the purpose of checking it does not eliminate the changes that have been made for a11y so far.

This theme includes three palettes:

- Evergreen
- Ocean
- Cardinal

These can be changed by switching the import in clean.css.

## Plan

A rough plan has formed around this PR. If we find that we like it we have two options:

1. Release it as a new theme (clarity) and gradually move people off of clean and deprecate it
2. Merge it as a new version of clean and move everyone on clean over to it now

Personally I think option 1 that gives us time to shift people over. Increases some maintenance in the short term. But we could make clarity a sub theme and eliminate templates that are identical to clean.

Screenshot from 2025-09-30 13-20-50
Screenshot from 2025-09-30 13-10-04
Screenshot from 2025-09-30 13-21-11

### DIFF
--- a/src/themes/clean/build_assets.py
+++ b/src/themes/clean/build_assets.py
@@ -42,7 +42,7 @@ def process_journals():
                     "static",
                     "clean",
                     "css",
-                    f"journal{journal.id}_override.css"
+                    f"journal{journal.id}_override.css",
                 )
 
                 copy_file(file, override_css_file)


StephDriver:
making further a11y tweaks in #5007

PR #4992, opened 2025-10-01
Fixes a bug where the decline template would show for desk accept

ajrbyers:
The other changes are caused by ruff.


ajrbyers:
Closed in favour of https://github.com/openlibhums/janeway/pull/4994

PR #4998, opened 2025-10-02
Fixed an issue for a backwards compatible handler of frozen authors that affected back content plugin

### LINKED ISSUE #4991: backwards_compat_authors raises error
**Describe the bug**
Currently when `backwards_compat_authors` is called it tries to delete frozen authors for an article with 

```
instance.frozen_authors.filter(author__in=pk_set).delete()
```

frozen_authors is a function, perhaps it should be a method?

```    
def frozen_authors(self):
        return FrozenAuthor.objects.filter(article=self)
```

**Janeway version**
1.8

Note:

After testing a change I noticed a recursion issue that will need to be dealt with.

mauromsl:
closes #4991

### DIFF
--- a/src/submission/models.py
+++ b/src/submission/models.py
@@ -3374,7 +3374,7 @@ def backwards_compat_authors(
         for account in accounts:
             account.snapshot_as_author(instance)
     if action in ["post_remove", "post_clear"]:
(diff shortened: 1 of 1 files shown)


ajrbyers:
Has this been tested? I tried this particular fix locally and ended up with a recursion issue.

PR #5000, opened 2025-10-02
clarity a11y updates

StephDriver:
changes from reviewing clarity.

PR #5004, opened 2025-10-03
fix copy citation button not working on clean

StephDriver:
In [b-4890-loop-a11y-syntax](https://github.com/openlibhums/janeway/tree/b-4890-loop-a11y-syntax) while updating list syntax for loops across the theme, I updated the citation modals too. But on testing Clarity I found this breaks the copy citation function in two ways. First it places lists inside a span, which is invalid HTML and so they end up with an empty span and the lists afterwards. But the copy button is aiming at the span element, so effectively, it copies nothing. Changing out the span for a div doesn't work either, as having a list inside when copy/pasting ends up with bullets being added in when pasting into word. 

So the best solution seems to be to revert the original change.  I have also made the same change to Clarity in #5000.

This PR reverts 55fa11f52 a11y: #4890 update non-visible list syntax on clean

### DIFF
--- a/src/themes/clean/templates/elements/journal/citation_modals.html
+++ b/src/themes/clean/templates/elements/journal/citation_modals.html
@@ -10,14 +10,12 @@
                 </div>
                 <div class="modal-body">
                     <p>
-                        <span id="harvard-cite">
-                            <ul class="list-inline comma-ampersand-sep">
-                                {% for author in article.frozenauthor_set.all %}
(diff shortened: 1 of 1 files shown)


ajrbyers (changes requested):
Can I suggest having a hidden plain text version of this that the copy button can target? That way the syntax can remain for screenreaders and the copy/paste will work without inserting bullets?

StephDriver:
@ajrbyers I think we discussed this and I convinced you - if not could you add here what you want me to do?

PR #5013, opened 2025-10-14
Improve display of large images

### LINKED ISSUE #2641: Document press-level image settings
The press-level image settings could use some help text and an RTD page. This would build on #2632

### LINKED ISSUE #4222: Pixelation of large images
## User story

From *Field* editors: “I was wondering if there was a way of increasing the resolution of the 'large images' used as banner in articles. I can see that the system compresses images to 750px wide which means they are pixelated when rendered at 1200px. Given that we are a visual discipline, it would be fantastic if we could allow for a slightly higher resolution.”

## Investigation results

Janeway does not consistently handle the cropping of large images. When you upload a large image to most forms, it is sent through `resize_and_crop` with a max width of 750 and a max height of 324.

An example from *Field*:

- [Article](https://www.field-journal.org/article/id/145/)

- [Image](https://www.field-journal.org/manager/article/images/edit/145/)

But default images are left alone when they are saved through the Images page in the Journal Manager. This is because this view is built with `edit_settings_group`, which does not call `resize_and_crop`. An example from *Political Philosophy*:

- [Article](https://politicalphilosophyjournal.org/article/id/15252/)

- [Image](https://politicalphilosophyjournal.org/media/cover_images/f9c2a4fe-8f97-4292-86b8-b98549087ce2.png)

The other important thing to note is that all three themes allow the display of these images at more than 750px wide, inevitably leading to pixelation, whatever theme you use, if you upload images through any form other than the default settings page.

## Proposed solution

Given that lots of large images are not cropped already, and that the images that are cropped are pixelated in any theme, I think we should rethink the `resize_and_crop` routine, perhaps by enlarging the dimensions to meet the max width allowed by the themes. If we do this, we should also resize the default large image the same way, so that we are not storing and serving unecessary pixels.

## Other solutions

Happy to discuss alternatives if there is a good reason to keep the cropping to 750 x 324 that I am unaware of.

### LINKED ISSUE #4844: Developer docs for image sizing and cropping
## Problem description
It is tricky to anticipate how images are key images treated in Janeway's media storage and how they are displayed for each of the themes.

## Proposed solution
We need a brief summary of how Janeway crops uploaded images and in what circumstances, as well as how the CSS of each theme treats key images, such as the default large image, the article image, the article thumbnail, the issue cover, and the press image override.

On this basis we can then communicate key information to authors of end-user documentation. Info like this:
* Which images will be cropped immediately on uploading? How will they be cropped?
* Which images will be cropped on display? Which contexts (i.e., issue page vs. article page)? What are the recommended dimensions for such images in pixels?

joemull:
* Closes #4844.
* Closes #4222.
* Closes #2641.

This is a substantial rewrite of the templates and CSS around large images in the Janeway reader interface, to make things more legible and consistent, and to allow us to document behavior accurately. It also fixes the problems with image resolution that result from a discrepancy between original size and rendered size, by increasing the default resize-and-crop dimensions applied to large images being uploaded. The resize behavior was kept in the end because it will make a substantial performance difference for readers if we do not load gigantic images every time an article is accessed.

### DIFF
--- a/src/submission/models.py
+++ b/src/submission/models.py
@@ -2585,8 +2585,8 @@ class Article(AbstractLastModifiedModel):
             return self.issue.large_image.url
         elif self.journal.default_large_image:
             return self.journal.default_large_image.url
-        elif self.press.default_carousel_image:
-            return self.press.default_carousel_image.url
+        elif self.journal.press.default_carousel_image:
+            return self.journal.press.default_carousel_image.url
         else:
             return static(settings.HERO_IMAGE_FALLBACK)
 
--- a/src/core/tests/test_app.py
+++ b/src/core/tests/test_app.py
@@ -470,7 +470,7 @@ class CoreTests(TestCase):
             SERVER_NAME="testserver",
         )
         self.assertEqual(response.status_code, 200)
-        self.assertNotContains(response, "Reviews")
+        self.assertNotContains(response, "<h2>Reviews</h2>")
 
     @override_settings(URL_CONFIG="domain")
     def test_peer_reviews_for_author_consumption_overrides_hide_review_data(self):
--- a/src/templates/admin/core/article.html
+++ b/src/templates/admin/core/article.html
@@ -18,13 +18,16 @@
             {% hook 'edit_article' %}
         {% endif %}
         <div class="content">
-            {% include "elements/metadata.html" %}
         </div>
     </div>
 
-    {% if not journal_settings.general.hide_review_data_pre_release or article.peer_reviews_for_author_consumption %}
+    {% if not journal_settings.general.hide_review_data_pre_release or article.peer_reviews_for_author_consumption.exists %}
     <div class="box">
         <div class="title-area">
+            {% comment %}
+              If editing this heading, see
+              src.core.test_app.CoreTests.test_hide_review_details_on
+            {% endcomment %}
             <h2>Reviews</h2>
         </div>
         <div class="box">


src/core/logic.py
```diff
@@ -148,10 +148,23 @@ def send_confirmation_link(request, new_user):
...
     """
     Resize and crop an image to fit the specified size.
+    :param img_path: filepath to saved image
+    :param size: two-item list with [width, height] in pixels
+    :param crop_type: "top", "middle", or "bottom"
+    :param field_name: human-readable field name for help messages
+    :param original_filename: the original filename for help messages
     """
+    if not size:
+        size = [1500, 648]
```
mauromsl: Since we are changing this, can we put it under settings.py, so folks with custom themes can tweak the default? we can also use a unmutable type like a tuple to declare this at the signature level for better readability.

e.g:
```
# janeway_global_settings.py
DEFAULT_CROP_SIZE = (1500, 648)

#src/core/logic.py
def resize_and_crop(
    img_path,
    size=DEFAULT_CROP_SIZE,
    crop_type="top",
    field_name="",
    original_filename="",
):
```

docs/md/images-in-reader-interface.md
```diff
@@ -0,0 +1,128 @@
...
+
+The Material theme is not supported at the press level.
+
+#### Featured articles and popular articles
+
+Featured and popular article homepage elements have the same behavior as article pages.
+
+### Is the large image cropped?
+
+The large image is cropped to a maximum of 1500px wide by 648px tall, in preparation for display at up to 1100px wide (Clean), 1200px wide (OLH), or 1477px wide (Material). The top of the image is kept. (On an earlier version of Janeway, the middle portion was kept.)
```
mauromsl: What's the rationale behind the change from middle to top as the default?

mauromsl (changes requested):
Good work, just as small comment inline regarding the default crop size.

docs/md/images-in-reader-interface.md
joemull: I've helped people navigate this feature before, and the middle crop behavior is confusing to many people. People who are not very tech savvy have a hard time predicting what will come out, because they do not know how to crop their image to the exact dimensions called for, and resizing happens before the cropping.

I think that if we keep the top, then at least these users can do a better job of predicting what part of the image will be kept, so even a less precise crop job will have a good chance of targeting the desired portion of the image.

mauromsl: Interesting, o me, it feels more predictable to crop the center of the image, as that will most likely be the area of focus (and is usually what we do with images when we want to modify their aspect ratios in most contents, in the web and otherwise).

But subjectivity aside, I wouldn't change the familiar behaviour that users already know, unless we are certain the new approach leads to an improved user experience, so I'd leave a middle crop for now.

Is it worth moving this discussion to a separate issue perhaps?

mauromsl (changes requested):
further comment inline regarding change to default behaviour

docs/md/images-in-reader-interface.md
joemull: What we really need is a screen where users choose the crop area, like most image upload interfaces, so I don't think it's worth having a long discussion about this. I've changed it back to middle for now.

ajrbyers: > What we really need is a screen where users choose the crop area, like most image upload interfaces, so I don't think it's worth having a long discussion about this. I've changed it back to middle for now.

I think its worth opening an issue for this. It would be awesome to have an interstitial page for image uploads where we can deal with stuff like this.

src/submission/models.py
```diff
@@ -2561,6 +2562,34 @@ def iso639_1_lang_code(self):
...
+                    "identifier_type": "id",
+                    "identifier": self.pk,
+                    "file_id": self.large_image_file.pk,
+                },
+            )
+        elif self.issue and self.issue.large_image:
+            return self.issue.large_image.url
+        elif self.journal.default_large_image:
+            return self.journal.default_large_image.url
+        elif self.press.default_carousel_image:
```
ajrbyers: `Article` doesn't have a `Press` attribute.

PR #5027, opened 2025-10-31
Alt Text Management

### LINKED ISSUE #3938: image missing a text alternative [issue covers]
Issues are represented with a cover image which includes details on the contents as well as the volume  and date.  The volume and date information is included in the text below, but not the rest of the information.

The text alternative needs to convey the equivalent information - so for each of these the contents should also be listed in text / text alternative.

Image

https://www.iastatedigitalpress.com/archivalissues/issues/

### LINKED ISSUE #3943: image missing a text alternative [page header]
Alt text needed for journal logo images.

Image

https://www.iastatedigitalpress.com/archivalissues/issues/

https://www.iastatedigitalpress.com/farmreports/issue/97/info/?page=3

Image

ajrbyers:
Closes https://github.com/openlibhums/janeway/issues/3943
Closes https://github.com/openlibhums/janeway/issues/3938

## Alt Text

This PR introduces a new pattern for HTMX functionality in Janeway. There's a new `partial_views.py` module specifically for HTMX endpoints, separate from regular views. Partial templates live in `/appname/partials/`. This keeps HTMX-specific code isolated rather than mixing it into existing views and templates, hopefully this will make it easier to find and maintain as we add more interactive features.

Throughout the admin - article image management, journal settings, issue pages - there's now an "Edit alt text" button. Click it and you get an HTMX powered modal that lets you add or update alt text without page reloads. The modal itself is a reusable base template that other features can extend.

Note: This PR adds a new builtin templatetag (see `janeway_global_settings.py` and if this has been copied to `settings.py` this will need to be updated.

## Image Uploads

Journal images are now uploaded and removed individually without a full page save. Each image field on the Journal Images settings page has its own upload form and Remove Image button, powered by HTMX. A preview appears alongside the upload control once an image is set, and the alt text editor becomes available at that point too. Feedback on save and remove is shown via a toast notification. The old single-form submission for the images group has been removed.


StephDriver (changes requested):
## General reaction 🥳 💃 🍹 

But some small specific niggles:

### 1. security wrapper not working (as per discussion).

## Notes from the UI `manager/settings/images`
### 2. some images that show up in the journal image settings, don't have edit alt text options, 
 but there is no explanation as to why - so this will confuse users. In this example, the `default large image` has  a button, but the `press image override` does not.  The favicon and default profile image similarly have no alt-text button on my install.

Image

### 3. The alt text is saved on the "save alt text" button in the popup, but there is also a 'submit' button at the end of the page.   

Users may think they've saved changes they haven't - File changes are only actioned on submit, making it possible here to select a file, edit alt text and save it, then click away from teh page without submitting (the submit button may be off screen at bottom of page) and only have saved the alt text, while the file has not actually been added. 

Users may think they have undone changes that have actually been saved - the alt text edit popup once saved is saved, users are not warned that there is no way to undo a mistake here and they are accustomed to being able to click back off a page to cancel their edits. 

### 4. As page links to the documentation, that documentation either needs to be updated to include alt-text instructions, or there needs to be a second link to additional instruction on this new alt-text functionality.

Image


## Other

### 5. Profile image missing 'edit alt text' option
`/profile/` one can upload a profile image, but no edit alt text button.

### 6. Issues page cannot load

Image

```django
{% get_alt_text obj=issue file_path=issue.cover_image.url as alt_text %} 
```

### 7. Likewise collections
> Invalid block tag on line 20: 'get_alt_text', expected 'empty' or 'endfor'. Did you forget to register or load this tag?
```django
alt="{% get_alt_text obj=collection file_path=collection.hero_image_url default=collection.issue_title|striptags %}"
```

### 8. likewise homepage

... this continued I can't find a front of house page that might use one of these alt-texts that will load.

I don't think I can really look at much more right now, until the errors are sorted out. 


One final thought - I can't work out where on the Press manager this alt-text fits. Is this work covering press?

StephDriver (approved):
1. My concerns about there being two different save behaviours on the same page remain unaddressed. I realise this is a side-effect of adding in HTMX, but it is confusing and that needs mitigating, maybe some help text?

2. My concern remains about inability to add alt-text to user profile on the user profile edit page.  I realise we generate funcitonal alt-text for profile images in the form "profile picture of <person>" so perhaps this is future work - but I think the limitation needs to be clear somewhere.

3. I would like the early "WIP" commit messages updated to something more meaningful.

Other than that - as far as I can see it all works well.

PR #5029, opened 2025-11-04
Fix a bug that caused a recursion error when deleting authors

### LINKED ISSUE #5028: Cannot delete authors (recursion)
## Problem

Deleting a FrozenAuthor triggers a recursion between two signals, one on the FrozenAuthor and one on the Article-Account M2M.

This was not discovered earlier because one of the signals returns early in development, in order not to trigger a deprecation warning.

Removing the early return allows you to see the bug in development.

## Solution

- [ ] Remove the deprecation warning, as it is not essential
- [ ] Remove the early return, since it causes bugs to show up only in production
- [ ] Change the frozen author deletion signal so that it does not call the other signal if none of the objects it cares about are found

joemull:
Fixes #5028.

### DIFF
--- a/src/submission/models.py
+++ b/src/submission/models.py
@@ -3303,19 +3303,13 @@ class SubmissionConfiguration(models.Model):
 @receiver(pre_delete, sender=FrozenAuthor)
 def remove_author_from_article(sender, instance, **kwargs):
     """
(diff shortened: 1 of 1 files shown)


mauromsl:
merged as hotfix

Issue #5033, opened 2025-11-05
Reviewer account activation issues


S-Haime:
**Describe the bug**
There have been reports of editors trying to add new reviewers and being unable to do so because they cannot find them. Specifics have included:

* "I am facing [a] problem [...], that is when I try to select reviewers for a second round of peer-review, I cannot find the names of the reviewers who assessed the paper in the first round. Could you please kindly advise on this?" (https://janeway.freshdesk.com/a/tickets/11043) - OLH, one-click review enabled here. Unable to reproduce on Playtest thusfar...

* "Recently we have had two user accounts go inactive during the process of them serving as reviewers - so they sent in their reviews, but when we went to contact them to look at revisions, their user accounts could not be found. We could go into the user database and reactivate their account, but this seems like a bug that should be dealt with. The two users were [...] - not sure if the logs would explain why they got deactivated."
https://janeway.freshdesk.com/a/tickets/11071 - OLH, one-click review enabled, unable to reproduce so far.

** Testing **
***Tested on Playtest (docstest) - one-click peer review on***
Tested with two fake accounts not in DB, both were activated after filling in the pop-up and clicking **Add reviewer** but before the assignment event fired.

<img width="1639" height="402" alt="Image" src="https://github.com/user-attachments/assets/b10092a9-4a51-4775-a969-ec8c0b7552f5" />

<img width="949" height="780" alt="Image" src="https://github.com/user-attachments/assets/db5734f1-11be-44d8-b686-93f7b2aa10b1" />

All other test accounts listed above activated and remained active during further peer review rounds (after their own was completed) and after a revision was completed. They could be found on the assignment screen.

If an account was not activated (imported) it could not be found, regardless of whether it had the reviewer role or not.

If an unactivated account is (attempted to be) added as a reviewer, it will remain inactive. If it did not yet have the reviewer role, this role will be added.
<img width="945" height="476" alt="Image" src="https://github.com/user-attachments/assets/d18f5096-4546-4bf3-991f-0003871e3601" />

<img width="951" height="462" alt="Image" src="https://github.com/user-attachments/assets/c9c7026c-3077-4fde-b6dc-a3072a15c3a0" />

***Tested on Playtest (docstest) - one-click peer review off***
When trying to assign an existing, inactive user expected behaviour follows. See: https://github.com/openlibhums/janeway/issues/4186

When adding a new user who is marked as active and assigned to review, the account is indeed active and remains so after review completion.

[...]

PR #5041, opened 2025-11-14
Fix an issue where editors wouldn't be able to select reviewers with inactive accounts to review a paper, despite using one click access. 

mauromsl:
Hotfix for  #5033 

Added support for the following stories:

> As an editor, I can select users with inactive accounts as review candidates, as long as they have the reviewer role.

> As an editor, when I fill in the form to add a reviewer to a paper and it matches an existing inactive account, the account will be granted the reviewer role (thus displaying in the list of review candidates) and will be preselected as a reviewer in the assignment screen.

> As an editor, if I attempt to assign an inactive reviewer to a paper when one click access is not enabled, I will see a clear warning indicating the reviewer won't be able to log in and recommending that I enable One click access

> As a press manager, the default option for one click access will now be on (matching the new installation default). Journals who have opted to turn the feature on/off explicitely, remain unchanged.

image

### DIFF
--- a/src/core/migrations/0110_one_click_review_on_by_default.py
+++ b/src/core/migrations/0110_one_click_review_on_by_default.py
@@ -8,7 +8,7 @@ def update_default_setting(apps, schema_editor):
     SettingValue = apps.get_model("core", "SettingValue")
     setting_values = SettingValue.objects.filter(
         setting__name="enable_one_click_access",
-        journal__isnull=False,
+        journal__isnull=True,
     )
     setting_values.update(value="on")
 


src/core/migrations/0110_one_click_review_on_by_default.py
```diff
@@ -0,0 +1,26 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def update_default_setting(apps, schema_editor):
+    SettingValue = apps.get_model("core", "SettingValue")
+    setting_values = SettingValue.objects.filter(
+        setting__name="enable_one_click_access",
+        journal__isnull=False,
```
ajrbyers: For the default it would be `journal__isnull=True` right?

mauromsl: good catch!

ajrbyers (approved):

PR #5043, opened 2025-11-18
Document how to test ORCID login locally

joemull:


ajrbyers (approved):

docs/md/orcid.md
```diff
@@ -0,0 +1,58 @@
...
+1. Create an API client on orcid.org.
+
+2. Enter each production domain for your Janeway instance as a redirect URL on orcid.org. If Janeway is set to domain mode, each journal will need its own redirect URI.
+
+3. Enable ORCID and copy the keys from your client into the Django settings file (`src/core/settings.py`).
+
+```py
+ENABLE_ORCID = True
+ORCID_CLIENT_SECRET = "your client secret"
+ORCID_CLIENT_ID = "your client ID"
```
StephDriver: just a comment - but this made me wonder what syntax we have for showing things like 'your client secret' - one of the things I struggle with when looking at examples online is which part to replace - e.g. 
`"your client secret"` or `your client secret` - much worse when it isn't obvious if it should be a string or not or snake/kebab case used within quotes etc.  

I'm not saying this example here is unclear, more that it prompted me to think it would be good to have a standard syntax for replacable values.

docs/md/orcid.md
```diff
@@ -0,0 +1,58 @@
...
+
+Note that the URLs for the ORCID API are fixed, so you can leave the defaults in `src/core/janeway_global_settings.py`.
+
+## Testing ORCID login manually
+
+You can test the ORCID login flow locally, by creating a test client, adjusting your hosts file, and resetting the Janeway domain for the site you want to test.
+
+The following steps were written for Ubuntu, and are meant for testing the press site.
+
+1. Create a test client on orcid.org and enter a normal domain of your choosing (it won’t be accessed) with `http` and the Janeway port you use for development: `http://www.openlibhums.org:8000/`
```
StephDriver: would be good to have an example for completeness or is that the example?  If so call it an example?

docs/md/orcid.md
```diff
@@ -0,0 +1,58 @@
...
+
+2. Edit your hosts file to put in the domain from the redirect URI that orcid.org knows about, in place of the `localhost` domain:
+
+```sh
+sudo vim /etc/hosts
+```
+
+```
+127.0.0.1 www.openlibhums.org
+```
```
StephDriver: again, for clarity, label examples as examples.

docs/md/orcid.md
```diff
@@ -0,0 +1,58 @@
...
+
+This will let orcid.org redirect your browser to the domain it knows about when authentication is finished, while allowing your browser to load the normal Janeway development server.
+
+3. Restart the networking service for the change to take effect.
+
+```
+sudo systemctl restart NetworkManager
+```
```
StephDriver: This looks like a command to be used as is, rather than an example, and for clarity this needs to be distinguished from the examples used above.

docs/md/orcid.md
```diff
@@ -0,0 +1,58 @@
...
+
+4. Edit your Janeway site domain to the same domain, keeping the port number.
+
+```
+python src/manage.py alter_domains press
+Altering domain for Test A, current domain: localhost:8000
+Enter the new domain you wish to set: www.openlibhums.org:8000
+Altering domain record...... [Ok]
```
StephDriver: It is unclear here which parts are you typing, and which are prompts.

docs/md/orcid.md
joemull: Yes, agreed. For now I will follow this: https://developers.google.com/style/placeholders

joemull:
OK @StephDriver, I've given it a once-over for clarity with the help of your comments.

StephDriver (approved):

PR #5054, opened 2025-11-24
Editors can hide some author affiliation fields

### LINKED ISSUE #5044: Improvements to author affiliations for large journals
## User stories

Some editors want only minimal information to appear for authors affiliations. For example, Glossa often has multiple authors on each article, and a high volume, so it is too much work to fix up the job titles, departments, and affiliation dates. And they require the use of ORCID, so it is best to let users follow ORCID links to get more up-to-date and detailed information on the author's ORCID profile. In some cases the ROR ID and org name are the only desired pieces of affiliation metadata.

Other editors want more information, perhaps because they do not use ORCID as much, so the job titles, departments, and dates should be available to them.

## Proposed solution

- [ ] Create settings to allow editors to turn off some of the fields on the affiliation form: `title`, `department`, `start`, and `end`. Modify the templates and form so that the input fields do not show up when the setting is turned off, and nothing is output in the display templates (even if there is content).
- [ ] Create a block for help text at the bottom of the editor's **Edit author details** page, which shows which of these settings are currently on, and gives links to change them.
- [ ] Leave the affiliations for user accounts unchanged by these settings.

joemull:
This closes #5044

It also documents settings, with a guideline for naming that we ~are due to discuss~ discussed at the design meeting on 2 Dec. ~Please don't merge until we have affirmed the guideline.~ It's been updated to reflect the discussion.

### DIFF
--- a/src/submission/forms.py
+++ b/src/submission/forms.py
@@ -7,6 +7,7 @@ import re
 import warnings
 
 from django import forms
+from django.db.models import Q
 from django.utils.translation import gettext, gettext_lazy as _
 
 from submission import models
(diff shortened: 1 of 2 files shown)


docs/md/settings.md
ajrbyers, line 1: This is great.

src/submission/forms.py
```diff
@@ -546,13 +546,23 @@ class Meta:
...
+            if not self.journal.get_setting("metadata", "author_department"):
+                self.fields.pop("department")
+            if not self.journal.get_setting("metadata", "author_affiliation_dates"):
+                self.fields.pop("start")
+                self.fields.pop("end")
 
     def clean(self):
         cleaned_data = super().clean()
         if not self.instance:
+            # Todo: Does this ever run? Q is undefined.
```
ajrbyers: Todo here, check and update?

joemull: OK, that's fixed

PR #5057, opened 2025-11-27
Authors can be linked to matching accounts more easily

### LINKED ISSUE #5053: Unable to link author to account in some cases
## Problem

If you try to add an author by searching for their email address, Janeway checks two things:

1) Is there an account with this email already?
2) Is there an author record on this article with this email already?

Recently, an author record was created with the same email as an existing account, but the frozen author and account were not linked. My guess is that the author record was created with a non-matching email address, and then it was edited to one that did match an account.

The journal editor did not know how to resolve the issue. We had to manually delete the old author record and populate a new one.

## Proposed solutions

I can see a few options:

a. Detect when an unlinked email match exists between a frozen author record and an account, and provide a button in these cases to link them. When linking, do not snapshot the account fields and links, because this could overwrite unique data, just link the account via the `author` field.

b. If a frozen author email is edited in such a way that it matches a Janeway account, raise a validation error. Provide the user with a message explaining that there is an account with this email, and that the author record needs to be deleted before searching for the user by their email address.

joemull:
Fixes #5053.

Screenshot from 2025-11-27 16-25-12
Screenshot from 2025-11-27 16-25-23

This also fixes a bug where if editors go to edit author details for an author that *does* have an account, they do not see the account email pulled through, they just see an empty field for frozen_email, which is not what we want.

### DIFF
--- a/src/submission/views.py
+++ b/src/submission/views.py
@@ -1024,21 +1024,32 @@ def order_authors(request, article_id):
     return HttpResponse("Thanks")
 
 
-@login_required
+@require_POST
 @user_can_edit_article
 def link_author_to_account(request, article_id, author_id):
     next_url = request.GET.get("next", "")
+
     article = get_object_or_404(models.Article, pk=article_id, journal=request.journal)
     author = get_object_or_404(models.FrozenAuthor, pk=author_id, article=article)
-    account = get_object_or_404(core_models.Account, email__iexact=author.email)
-    author.author = account
-    author.save()
-    messages.add_message(
-        request,
-        messages.SUCCESS,
-        "%(author_name)s (%(email)s) is now linked to a user account."
-        % {"author_name": author.full_name(), "email": author.email},
+    account = get_object_or_404(
+        core_models.Account,
+        email__iexact=author.email,
+        accountrole__role__slug="author",
     )
+
+    author_account_form = forms.FrozenAuthorAccountForm(
+        {"author": account.pk},
+        instance=author,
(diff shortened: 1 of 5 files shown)


src/submission/views.py
ajrbyers, line 1: Nice and DRY.

src/submission/views.py
```diff
@@ -1033,6 +1024,34 @@ def order_authors(request, article_id):
     return HttpResponse("Thanks")
 
 
+@login_required
+@user_can_edit_article
+def link_author_to_account(request, article_id, author_id):
```
ajrbyers, line 1029: Since this links the author and saves changes, it should really be restricted to POST. Using GET here risks accidental state changes via caching or prefetching. There will also be no CSRF protection.

src/templates/admin/elements/current_authors_inner.html
```diff
@@ -36,6 +36,18 @@ <h2>{% trans "Current authors" %}</h2>
                   {% if article.correspondence_author and article.correspondence_author == author.author %}
                     {% include "admin/submission/submit_correspondence_author.html" with article=article %}
                   {% endif %}
+                  {% if unlinked_account %}
```
ajrbyers, line 39: See comment on view above, this section should be a form POST rather than a link.

ajrbyers (changes requested):
I've added two comments around the view for linking an account with the author. As its altering data it should require a POST action.

PR #5064, opened 2025-12-05
fix mat log overlaping nav

### LINKED ISSUE #4052: Header overwriting logo at some screenwidths
**Describe the bug**

https://traj.openlibhums.org/
image

**To Reproduce**
problem occurs with windows at pixel widths of 994 to 1423.

**Expected behavior**

this is for smaller widths:
image

and this for larger widths:
image

but the point of crossover is too low - the change is occurring between 993px and 994px, but should not be until approx 1423px.

 - same on both Chrome and Safari (on MacOS)

StephDriver:
closes #4052

### DIFF
--- a/src/themes/material/assets/mat.css
+++ b/src/themes/material/assets/mat.css
@@ -1184,3 +1184,49 @@ p + .btn-background {
   .nav-wrapper a:focus {
     display: block;
   }
+
+/* navbar overrides */
+:root {
+  --nav-height-px: 120px;
+  --nav-border-style: 4px solid;
+}
+
+@media only screen and (min-width: 993px) {
+  .header-overrides {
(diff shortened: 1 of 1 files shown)


StephDriver:
I'm closing this PR. 

- The changes can't be made universally without clashing with custom css.
- There is no easy way to override them when they clash within the current structure without resorting to javascript.
- Where clashes occur, it is primarily because custom css has already been used to fix this issue in a way that is specific to that journal, and is not compatible with this general fix. 

On discussion within the team, we decided that:

Material as a theme is near it's end of life, to long term solutions are not required.  Therefore for now, we will provide this fix as a basic template for use in custom css as and when needed.  But not alter the theme itself.

PR #5065, opened 2025-12-05
Fixed an issue when a duplicate organization is found (legacy model)

mauromsl:
Instead of creating additional organizations, we use a pre-exiting one

### DIFF
--- a/src/core/models.py
+++ b/src/core/models.py
@@ -2915,26 +2915,28 @@ class Organization(models.Model):
                 # Or maybe a primary affiliation has already been
                 # entered without a ROR for this
                 # account / frozen author / preprint author?
+                # If there is no `institution`, this method is being used to update
+                # the department or country in isolation, so we want the primary
+                # affiliation's org regardless of what its custom label is.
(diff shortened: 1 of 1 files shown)


joemull:
> During an import, we saw a `MultipleObjectsReturned` exception being raised here where the change is made. When creating a new object in the exception handler, with an `IntegrityError` against a unique constraint. relying on the pre-existing db record solved the issue

Hmm, which model was the unique constraint on? The only unique constraint on `Organization` is related to ror_id, but it can be an empty string, so that should not be an issue here.

The problem with relying on a pre-existing organization for a custom-named org is that the linked `OrganizationName` will be editable by end users. I could edit my affiliation and end up changing yours too. So, orgs that have custom names should not be used across accounts.

joemull (changes requested):
Going ahead and submitting a review to show the comment above as a requested change.

mauromsl:
I cannot reproduce this any longer, so I'm going to assume it was due to an unexpected state in a prod DB

Issue #5105, opened 2026-01-23
WCAG 1.3.3 ✅ (Supports): Sensory Characteristics


StephDriver:
## WCAG 1.3.3: Sensory Characteristics

**Level:** A
**Principle:** Perceivable
**Result:**  ✅ Supports

While OLH only partially supports it, any users affected can choose to use accessibility mode which does support it, therefore overall, Janeway supports this criterion.

- Material - ✅  Supports
- Clean - ✅ Supports
- A11y Mode - ✅  Supports
- OLH - 🟠  Partially Supports

### Description
This issue tracks the audit and implementation of WCAG 2.2 Success Criterion 1.3.3: Sensory Characteristics.

### Resources
- [WCAG 2.2 Success Criterion 1.3.3](https://www.w3.org/WAI/WCAG22/Understanding/sensory-characteristics.html)
>Goal
Instructions are understandable by more people.
What to do
Describe controls by name, not just by appearance or location.
Why it's important
People who are blind or have low vision need non-visual instructions.

### Audit Checklist
- [x] Review previous Audit Result
- [x] Review work identified in previous audit and list open issues.
- [x] Check for any other relevant open issues.
- [x] Retest open issues, are these issues still present?
- [x] Test for any new issues that may have been introduced.
- [x] Determine compliance status, add this to issue name.
- [x] Add any new issues to the Accessibility project for triage.
- [x] Replace previous audit in milestone with this update.

### Previous audit
> ❌ Sensory Characteristics
We do not rely on sound, size or orientation to convey information. We are in the process of upgrading our Captcha to make it more accessible. This audit found that we have used icons and typographical marks to convey information without providing non-visual alternatives. Colour is being used in the Material theme to show invalid user input.

#4126

 ### Progress since last audit

- openlibhums/janeway#3945
- openlibhums/janeway#4210
- openlibhums/janeway#4230
- openlibhums/janeway#4058
- openlibhums/janeway#4215
- openlibhums/janeway#4200
- openlibhums/janeway#4227
- openlibhums/janeway#3936
- openlibhums/janeway#4268 

 ### Audit Notes

Janeway does not to use sound, size or orientation to convey information. Janeway does not use sounds. Orientation is only handled in as much as the page reflows to match the screen-size with both portrait and landscape use anticipated.  The appearance of elements is handled through CSS and may be different for instance through custom styling, so cannot be used within the platform to describe elements.

[...]

PR #5106, opened 2026-01-23
a11y: update credit link markup for screen-readers

StephDriver:
single instance of non-compliance with WCAG 1.3.3 (#5105) for the OLH theme. 

the audit for this SC identified that the icon had no text-equivalent. Then while fixing, I also found the external link wasn't marked up so added that fix which would otherwise have been found on a later audit when looking at navigation. 

While fixing this, I also found that CrediT wasn't displaying on the OLH article page at all, due to a logic error. I have fixed that in the second commit here.

### DIFF
--- a/src/themes/OLH/templates/journal/article.html
+++ b/src/themes/OLH/templates/journal/article.html
@@ -480,13 +480,7 @@
 
                         {% if journal_settings.general.use_credit and credit_records_exist %}
                             <div class="section hide-for-small-only">
-                                <h2>{% trans "CRediT Roles" %} 
-                                    <a href="https://www.niso.org/publications/z39104-2022-credit" target="_blank">
-                                        <span class="sr-only">{% trans "Read more about Credit." %} </span>
-                                        <i aria-hidden="true" class="fa fa-info"></i>
-                                        {% include "elements/icons/link_external.html" %} 
-                                    </a>
-                                </h2>
+                                <h2 id="credit-roles-title">{% trans "CRediT Roles" %} </h2>
                                 <ul>
                                     {% for author, credit_records in article.authors_and_credits.items %}
                                         {% if credit_records.exists %}
@@ -501,6 +495,13 @@
                                         {% endif %}
                                     {% endfor %}
                                 </ul>
+                                <a href="https://www.niso.org/publications/z39104-2022-credit" 
+                                    target="_blank"
+                                    aria-describedby="credit-roles-title"
+                                    >
+                                    {% trans "Read More" %}
+                                    {% include "elements/icons/link_external.html" %} 
+                                </a>
                             </div>
                         {% endif %}
 


src/themes/OLH/templates/journal/article.html
```diff
@@ -478,22 +478,34 @@ <h2>{% trans "File Checksums" %} (MD5)</h2>
...
-                            <ul>
-                              {% for author, credit_records in article.authors_and_credits.items %}
-                                <li>
-                                  {{ author.full_name }}:
-                                  <span class="credit">
-                                    {% for record in credit_records %}
-                                      {{ record }}{% if not forloop.last %}, {% endif %}
+                        {% if journal_settings.general.use_credit %}
+                            {% for author_check in article.frozen_authors.all %}
+                                {% if author_check.credits.exists %}
```
joemull: This creates two levels of for-loops, causing the section to display multiple times, once for each author:

Image

I'd probably fix it instead by creating a new context variable called `credit_records_exist` in the view, which queries credit records for a any relationship to the article, like this:

```py
credit_records_exist = models.CreditRecord.objects.filter(frozen_author__article=article).exists()
```

Then use that instead of the broken piece of logic:

```html
{% if journal_settings.general.use_credit and credit_records_exist %}
```

src/themes/OLH/templates/journal/article.html
```diff
@@ -478,22 +478,34 @@ <h2>{% trans "File Checksums" %} (MD5)</h2>
...
+                                    <div class="section hide-for-small-only">
+                                    <h2>{% trans "CRediT Roles" %} 
+                                        <a href="https://www.niso.org/publications/z39104-2022-credit" target="_blank">
+                                            <span class="sr-only">{% trans "Read more about Credit." %} </span>
+                                            <i aria-hidden="true" class="fa fa-info"></i>
+                                            {% include "elements/icons/link_external.html" %}
```
joemull: This feels a bit too icon-y, and the two icons right next to each other are visually distracting, since they are not positioned the same way (I understand why, but nevertheless).

Maybe we can just get rid of the `i` and replace it with the phrase "read more", with the whole link in parentheses? This would make the external link icon work better.

StephDriver: I'd much prefer that!

src/themes/OLH/templates/journal/article.html
StephDriver: Logic updated as suggested.

src/themes/OLH/templates/journal/article.html
StephDriver: And Read more link.  It looked intrusive under the heading, so I put it at the end intead.

image

StephDriver:
Note to self - update is now working on OLH, but credit no longer displaying on Clean or Material, can't find it on Master either, need more investigation as to what is going wrong.  As I only started that bugfix because it was working on C & M and not on O.

PR #5110, opened 2026-01-26
2610 orcid tokens

everreau:
Addresses: https://github.com/openlibhums/janeway/issues/2610

When orcid is enabled:

- Save orcid auth token and expiration when orcid is authenticated
- Allow users to add orcid via authentication
- When orcid is removed by admin or user revoke token
- Notify users when token is not valid in profile and edit user views
- Don't allow users to manually edit orcids.
- Allow co-authors to request orcids
- Add is_orcid_valid flag to FrozenAuthors
- Add authentication element to orcid crossref xml


alainna:
So great, @everreau! I've suggested some changes in line with ORCID's [brand guidelines](https://info.orcid.org/brand-guidelines/)

This is only for journals at the moment, is that correct? Could the co-author use the ORCID login option to connect their iD to their account rather than connecting the iD in their profile? What happens if the co-author is a frozen author without an account, or if the account isn't active?

ajrbyers:
> @alainna. All of the login features also work for preprints. I don't see any reference at all to orcids in the preprint author interface. I think we (cdl) would send whatever orcid was in the author's account to crossref even though it's not visible and it appears in the preprint front end. The profile and the account are the same as far as I know. The interface on main is a bit different than what we're currently running. Maybe we should setup a time so you can take a look at what I have. I hadn't really thought about what would happen if a user connects their orcid after the frozen authors were already made. I'll take a look at that.

On the `FrozenAuthor` object `orcid` is a property that will either return `FrozenAuthor.frozen_orcid` atrribute and if there is none check if the linked account has an `orcid` attribute and return that instead.

src/core/tests/test_app.py
```diff
@@ -582,3 +583,80 @@ def setUp(self):
...
+
+    @override_settings(ORCID_URL="https://sandbox.orcid.org/oauth/authorize")
+    def test_profile_orcid_unverified(self):
+        self.admin_user.orcid = "0000-0000-0000-0000"
+        self.admin_user.save()
+        self.client.force_login(self.admin_user)
+        response = self.client.get(reverse("core_edit_profile"))
+        self.assertContains(response, "ORCID iD could not be validated.")
+        self.assertContains(response, "Connect your ORCID")
+        self.assertContains(response, "https://sandbox.orcid.org/0000-0000-0000-0000")
```
joemull, line 615: Does this test (or any of the ones below) rely on a network connection? We'll need to put mocks into any test that relies on a network connection.

src/core/migrations/0110_account_orcid_token_account_orcid_token_expiration_and_more.py
```diff
@@ -0,0 +1,32 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("core", "0109_salutation_name_20250707_1420"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="account",
+            name="orcid_token",
+            field=models.CharField(blank=True, max_length=40, null=True),
```
joemull: Character fields should not have `null=True`

src/core/migrations/0110_account_orcid_token_account_orcid_token_expiration_and_more.py
```diff
@@ -0,0 +1,32 @@
...
+            model_name="orcidtoken",
+            name="access_token",
+            field=models.CharField(blank=True, max_length=40, null=True),
+        ),
+        migrations.AddField(
+            model_name="orcidtoken",
+            name="access_token_expiration",
+            field=models.DateTimeField(blank=True, null=True),
+        ),
+    ]
```
joemull: It's not immediately clear to me why we'd need to add fields to the account rather than using the orcidtoken object--what's the thinking here?

src/core/tests/test_app.py
```diff
@@ -582,3 +583,80 @@ def setUp(self):
...
+
+    @override_settings(ENABLE_ORCID=False)
+    def test_profile_orcid_disabled(self):
+        self.client.force_login(self.admin_user)
+        response = self.client.get(reverse("core_edit_profile"))
+        self.assertContains(
+            response, '<input type="text" name="orcid" maxlength="40" id="id_orcid">'
+        )
+
+    def test_profile_orcid_enabled_no_orcid(self):
```
joemull, line 597: Maybe add an `override_settings` decorator on this and other tests that expect `ENABLE_ORCID=True`, in case someone has it off and runs tests.

src/core/models.py
```diff
@@ -485,6 +486,9 @@ class Account(AbstractBaseUser, PermissionsMixin):
     orcid = models.CharField(
         max_length=40, null=True, blank=True, verbose_name=_("ORCiD")
     )
+    orcid_token = models.CharField(max_length=40, null=True, blank=True)
```
joemull: Charfields should not have `null=True`

src/core/models.py
```diff
@@ -959,6 +969,8 @@ class OrcidToken(models.Model):
     expiry = models.DateTimeField(
         default=generate_expiry_date, verbose_name=_("Expires on")
     )
+    access_token = models.CharField(max_length=40, null=True, blank=True)
```
joemull: Same as the other one--charfields should not have `null=True`

src/core/views.py
```diff
@@ -282,6 +288,24 @@ def user_login_orcid(request):
...
+        request.user.orcid = orcid_id
+        request.user.orcid_token = access_token
+        request.user.orcid_expiration = expiration
+        request.user.save()
+        messages.add_message(
+            request,
+            messages.SUCCESS,
+            _("Your ORCID has been connected to your account."),
+        )
+        return redirect(logic.reverse_with_next("core_edit_profile", next_url))
```
joemull: Haven't been able to test this yet but I'm actually having trouble working out what's happening just by reading it, especially the case where they're not authenticated and they get sent to a different login screen. Can you add a comment above with context, like the other if / else branches have?

src/core/views.py
```diff
@@ -434,6 +458,9 @@ def register(request, orcid_token=None):
         if form.is_valid():
             if token_obj:
                 new_user = form.save()
+                new_user.orcid_token = token_obj.access_token
+                new_user.orcid_expiration = token_obj.access_token_expiration
+                new_user.save()
```
joemull, line 553: This makes me think of the same question I had above: why we would need more fields on the account. It doesn't introduce a situation where there could be two sources of truth, does it?

src/core/models.py
```diff
@@ -485,6 +486,9 @@ class Account(AbstractBaseUser, PermissionsMixin):
     orcid = models.CharField(
         max_length=40, null=True, blank=True, verbose_name=_("ORCiD")
     )
+    orcid_token = models.CharField(max_length=40, null=True, blank=True)
+    orcid_token_expiration = models.DateTimeField(null=True, blank=True)
+    date_orcid_requested = models.DateTimeField(blank=True, null=True)
```
joemull, line 484: It's not immediately clear to me why we'd need to add fields to the account rather than using the orcidtoken object, maybe in combination with some property methods that make it easy to get the needed info from the account object in templates--what's the thinking here?

src/repository/urls.py
```diff
@@ -68,6 +68,12 @@
         views.repository_delete_author,
         name="repository_delete_author",
     ),
+    re_path(
+        r"^manager/(?P<preprint_id>\d+)/author/(?P<account_id>\d+)/request_orcid/(?P<redirect_path>[-\w]+)/$",
+        views.repository_request_orcid,
+        name="repository_request_orcid",
+    ),
```
joemull: We've got some ambiguity in object names between "author" and "account" in other parts of Janeway, esp. around of the `FrozenAuthor` object on journals. So a small request here: change `/author/` to `/account/` in the URL, so it's clear what object the subsequent id refers to when rendered?

Also, rather than having an extra parameter for redirect path, would you consider using `url_with_return` in templates, and then picking up the `next_url` in the view? We just standardized this pattern last year and all the login views use it, so it would be good to be consistent.

src/repository/views.py
```diff
@@ -1466,6 +1467,28 @@ def preprints_author_order(request, preprint_id):
...
+def repository_request_orcid(request, preprint_id, account_id, redirect_path):
+    user = get_object_or_404(
+            core_models.Account,
+            pk=account_id,
+    )
+    core_logic.send_orcid_request(request, user)
+    messages.add_message(
+        request,
+        messages.SUCCESS,
+        f"Successfully requested ORCID from {user.full_name()}",
```
joemull: "ORCID iD"?

src/repository/views.py
```diff
@@ -1466,6 +1467,28 @@ def preprints_author_order(request, preprint_id):
 
     return HttpResponse("Complete")
 
+@login_required
+def repository_request_orcid(request, preprint_id, account_id, redirect_path):
+    user = get_object_or_404(
+            core_models.Account,
+            pk=account_id,
+    )
```
joemull: I wonder about security here. Could any user who knows how to form the URL run this function against any other account id?

We might need to introduce a check that makes sure the target user is a co-author of theirs. Maybe consider also using a POST request.

src/templates/admin/elements/accounts/orcid_field.html
```diff
@@ -0,0 +1,53 @@
+{% load orcid %}
+{% load static %}
+
+<div>
+  <label for="{{ form.orcid.id_for_label }}">
+    <strong>
+      {{ form.orcid.label }}
+    </strong>
+  </label>
```
joemull: The input is hidden, so I think this won't link up properly as accessible HTML.

src/templates/admin/elements/accounts/orcid_field.html
```diff
@@ -0,0 +1,53 @@
...
+  {% endfor %}
+  {% if form.orcid.value %}
+    <input type="hidden" name="{{ form.orcid.name }}" value="{{ form.orcid.value }}"/>
+    <p>
+        <a href="{{ form.instance.get_orcid_url }}">
+            <img alt="ORCID logo" src="{% static 'common/img/icons/orcid.png' %}">&nbsp;{{ form.instance.get_orcid_url }}
+        </a>
+    </p>
+    {% if form.instance.is_orcid_token_valid %}
+        {% if form.instance == request.user or request.user.is_admin %}
```
joemull: I think we want to check `request.user.is_staff` instead. The `is_admin` flag just relates to the Django admin pages.

src/templates/admin/elements/accounts/orcid_field.html
```diff
@@ -0,0 +1,53 @@
...
+  {% endif %}
+  {% if not form.orcid.value or not form.instance.is_orcid_token_valid %}
+    {% if form.instance == request.user %}
+        <a href="{% url_with_next_and_orcid_action 'core_login_orcid' action='add_profile_orcid' %}"
+            class="button expanded orcid-button">
+            Connect your ORCID iD
+        </a>
+    {% else %}
+        {% if form.instance.date_orcid_requested %}
+            <p>Request sent: {{ form.instance.date_orcid_requested|date:"m/d/Y" }}</p>
```
joemull: This date format will make the date unclear to users outside the States.

src/templates/admin/elements/repository/orcid.html
```diff
@@ -0,0 +1,10 @@
+{% if author.account.orcid %}
+    {{ author.account.orcid }}
+{% else %}
+    {% if author.account.date_orcid_requested %}
+        <p>Request sent: {{ author.account.date_orcid_requested|date:"m/d/Y" }}</p>
```
joemull: This date format will make the date unclear to users outside the States.

src/templates/admin/repository/submit/authors.html
```diff
@@ -75,6 +75,7 @@ <h2>{% trans 'Authors' %}</h2>
                                 <tr>
                                     <th>{% trans 'Name' %}</th>
                                     <th>{% trans 'Email' %}</th>
+                                    <th>{% trans 'ORCID' %}</th>
```
joemull: "ORCID iD"?

src/templates/admin/repository/submit/review.html
```diff
@@ -55,12 +55,16 @@ <h2>Authors</h2>
               <th>Email Address</th>
               <th>First Name</th>
               <th>Last Name</th>
+              <th>ORCID</th>
```
joemull: "ORCID iD"?

src/templates/admin/repository/article.html
```diff
@@ -227,6 +227,7 @@ <h2>Authors</h2>
                         <tr>
                             <th>{% trans 'Name' %}</th>
                             <th>{% trans 'Email' %}</th>
+                            <th>{% trans 'ORCID' %}</th>
```
joemull: "ORCID iD"?

src/utils/install/journal_defaults.json
```diff
@@ -1669,6 +1669,44 @@
...
+        },
+        "setting": {
+            "description": "Email sent when user requests co-authors add ORCID iD.",
+            "is_translatable": true,
+            "name": "orcid_request",
+            "pretty_name": "ORCID iD Request",
+            "type": "rich-text"
+        },
+        "value": {
+            "default": "<p>Dear {{ user.full_name }}<br/><br/>Your co-author has requested your ORCID iD. You can add it by through your profile {{ user_profile_url }}.<br/></p>"
```
joemull: Typo: "by through"

src/utils/install/journal_defaults.json
```diff
@@ -1669,6 +1669,44 @@
...
+        },
+        "editable_by": [
+            "editor",
+            "journal-manager"
+        ]
+    },
```
joemull, line 1690: It looks like this separate template isn't needed, since the email text is the same and the link is the same.  Only the subject differs. Is there another reason for it?

src/utils/orcid.py
```diff
@@ -48,12 +49,44 @@ def retrieve_tokens(authorization_code, site):
...
+    else:
+        expiration_date = None
+
+    return access_token, expiration_date, orcid_id
+
+
+def is_token_valid(orcid_id, token):
+    api_client = OrcidAPI(
+        settings.ORCID_CLIENT_ID, settings.ORCID_CLIENT_SECRET, sandbox=True
+    )
```
joemull, line 77: The sandbox parameter needs to be able to change in production. Maybe it should be based on whether `sandbox.orcid.org` is present in the URL in settings.py?

src/utils/orcid.py
```diff
@@ -67,7 +100,9 @@ def build_redirect_uri(site):
 def get_orcid_record(orcid):
     try:
         logger.info("Retrieving ORCID profile for %s", orcid)
-        api_client = OrcidAPI(settings.ORCID_CLIENT_ID, settings.ORCID_CLIENT_SECRET)
+        api_client = OrcidAPI(
+            settings.ORCID_CLIENT_ID, settings.ORCID_CLIENT_SECRET, sandbox=True
```
joemull: Same comment about changing `sandbox` in production.

joemull (changes requested):
This is hugely appreciated--thank you for putting it in.

I'm having trouble testing it out (an issue with OLH's ORCID sandbox that will take a few days to resolve), but I've gone ahead and given you some comments from reading the code. Hope they're clear but please do ping me if anything does not make sense.

src/utils/install/journal_defaults.json
everreau: @alainna suggested having different text for active and inactive users.  I asked her for text but she hasn't sent any yet. I'll check with her today.

src/utils/orcid.py
everreau: ugh, yes, this was actually mistakenly commited but I can actually fix it.

src/core/tests/test_app.py
everreau: None of the tests rely on network connection.  It's setting the orcid url purely for display purposes.  I noticed that ruff is failing now but it was passing at some point and the tests ran successfully.

src/core/views.py
everreau: I'm not sure If I understand what you are suggesting so I might be going down the wrong track. I don't think it introduces two sources of truth because I was treating the OrcidToken table as a temporary holding place for initiated but incomplete transaction information (for example, if a user starts the login process but doesn't have an account or if the user simply starts from the registration entry point). I suppose we could complete the transaction by linking the OrcidToken to the account but it would still be a little wonky.  Because I think we'd still be keeping the orcid field in Account (in the case that users are running with orcid disabled) and the OrcidToken itself has it's own expiration since it's intended to support passing information between views within a transaction (perhaps the names need to be clarified).

joemull: OK, I will take a closer look on the next review and try to work out what I mean and what possible solutions I can see.

src/utils/install/journal_defaults.json
alainna: Apologies @joemull @everreau: just back in office. Re: template: 

@joemull could you confirm whether this situation is true? If an author is added to a Janeway journal or repository article, then an account is created for them and set as "inactive". In order for the user to connect an authenticated ORCID iD to their Janeway account/author profile, they must first activate their Janeway account, then go to their profile to connect their ORCID iD. 

Is that correct? 

A second question: If an active user tries to sign into Janeway with ORCID, but there's no iD in their account, they will be redirected back to the Janeway login page and asked to sign in -- is that correct? 

(Apologies, our ORCID integration is currently undergoing issues; i may try to get a local copy running to test tomorrow.)

joemull: @alainna this is a complicated thing to answer, because the changes to journals and repositories are getting out of sync. Let's discuss on Discord.

joemull: But to try to answer just for the repository side:

> If an author is added to a Janeway journal or repository article, then an account is created for them and set as "inactive". In order for the user to connect an authenticated ORCID iD to their Janeway account/author profile, they must first activate their Janeway account, then go to their profile to connect their ORCID iD.

I believe that's correct for the behavior before this PR. However, we should consider whether we can provide these users with a link that will do it all--authenticate with ORCID and activate their Janeway account.

> If an active user tries to sign into Janeway with ORCID, but there's no iD in their account, they will be redirected back to the Janeway login page and asked to sign in -- is that correct?

No, if there is no ORCID for them in Janeway, Janeway will attempt to match on email address. If the email address in their Janeway account matches one of the email addresses on their public ORCID profile, it will add the ORCID and log them in automatically. However, if there is no match by email, Janeway will say there is no matching account and redirect them to the **registration** page. This could potentially result in duplicate accounts, or an error when the user types in their email and it says "an account with that email already exists". So I think this PR should introduce a third option, where Janeway is able to identify the user attached to an ORCID authentication token, and input the ORCID and activate the account.

src/core/tests/test_app.py
joemull: This is using a network connection, because `is_orcid_token_valid` calls the ORCID API using `requests`. I turned my internet off and got the error below. So I think we'll just want to mock it where the Janeway logic that is being tested ends, and the ORCID API code begins, in this and subsequent tests that use the API.

```
  File "/home/joemull/git/janepr/src/core/tests/test_app.py", line 563, in setUp                                                                                                                                    
    self.article_one = helpers.create_article(                                                                                                                                                                      
  File "/home/joemull/git/janepr/src/utils/testing/helpers.py", line 300, in create_article                                                                                                                         
    author.snapshot_as_author(article)                                                                                                                                                                              
  File "/home/joemull/git/janepr/src/core/models.py", line 893, in snapshot_as_author
    frozen_dict["is_frozen_orcid_valid"] = self.is_orcid_token_valid()
  File "/home/joemull/git/janepr/src/core/models.py", line 963, in is_orcid_token_valid
    return is_token_valid(self.orcid, self.orcid_token)
  File "/home/joemull/git/janepr/src/utils/orcid.py", line 76, in is_token_valid
    r = api_client._get_public_info(
  File "/home/joemull/git/janepr/.venv/lib/python3.10/site-packages/orcid/orcid.py", line 410, in _get_public_info
    return requests.get(request_url, headers=headers, 
  File "/home/joemull/git/janepr/.venv/lib/python3.10/site-packages/requests/api.py", line 73, in get
    return request("get", url, params=params, **kwargs)
  File "/home/joemull/git/janepr/.venv/lib/python3.10/site-packages/requests/api.py", line 59, in request
    return session.request(method=method, url=url, **kwargs)
  File "/home/joemull/git/janepr/.venv/lib/python3.10/site-packages/requests/sessions.py", line 589, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/joemull/git/janepr/.venv/lib/python3.10/site-packages/requests/sessions.py", line 703, in send
    r = adapter.send(request, **kwargs)
  File "/home/joemull/git/janepr/.venv/lib/python3.10/site-packages/requests/adapters.py", line 700, in send
    raise ConnectionError(e, request=request)
requests.exceptions.ConnectionError: HTTPSConnectionPool(host='pub.orcid.org', port=443): Max retries exceeded with url: /v2.0/0004-5678-9012-345X/record (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7c9bc79e8190>: Failed to establish a new connection: [Errno -3] Temporary failure in name resolution'))
```

src/core/models.py
```diff
@@ -485,6 +486,9 @@ class Account(AbstractBaseUser, PermissionsMixin):
     orcid = models.CharField(
         max_length=40, null=True, blank=True, verbose_name=_("ORCiD")
     )
+    orcid_token = models.CharField(max_length=40, blank=True, default="")
+    orcid_token_expiration = models.DateTimeField(null=True, blank=True)
+    date_orcid_requested = models.DateTimeField(blank=True, null=True)
```
joemull, line 484: The admin view for the account is a bit unique in that we have to manually add the fields we want to see.  I think it would be helpful to see these new ones in the admin when troubleshooting login issues for users, so could we add them in the list in `AccountAdmin.fieldsets`?

src/core/views.py
```diff
@@ -274,14 +299,83 @@ def user_login_orcid(request):
...
+            # this case is very unlikely but since this view
+            # doesn't require a login check to ensure they are
+            # already logged in, if not just redirect to
+            # non-orcid login
+            messages.add_message(
+                request,
+                messages.WARNING,
+                _("You must be logged in to connect an ORCID iD to your account."),
+            )
+            return redirect(logic.reverse_with_next("core_login", next_url))
```
joemull: OK, with my proposed change to the email template, this might catch a fair number of users. So if you like that change, let's make sure users come back here after they've done the non-ORCID login. Something like this:

```py
next_url = request.get_full_path()
redirect(logic.reverse_with_next("core_login", next_url))
```

src/core/tests/test_views.py
```diff
@@ -488,6 +499,69 @@ def test_action_register_redirects_with_next(self, retrieve_tokens):
...
+        response = self.client.get(
+            "/login/orcid/",
+            get_data,
+            follow=True,
+            SERVER_NAME=self.journal_one.domain,
+        )
+        self.assertIn(
+            f"/register/step/orcid/",
+            response.redirect_chain[0][0],
+        )
```
joemull: If I'm reading this right, the redirect to registration would happen here even if the account was active, because neither the ORCID nor email matches. Should we add some of those in here so it's clear the test is just testing active / inactive status?

src/core/tests/test_views.py
```diff
@@ -488,6 +499,69 @@ def test_action_register_redirects_with_next(self, retrieve_tokens):
...
```
joemull, line 538: I'm not sure what this test is testing. Is it about duplicate ORCIDs, or the account being inactive? I think it basically does what I was asking for, for the previous test, but it also has a bit in the name about duplicates, so I'm not sure.

src/core/logic.py
```diff
@@ -148,6 +148,42 @@ def send_confirmation_link(request, new_user):
...
+
+    user.date_orcid_requested = timezone.now()
+    user.save()
+
+    if user.is_active:
+        template = "orcid_request"
+        subject = "subject_orcid_request"
+    else:
+        template = "orcid_activate_request"
+        subject = "subject_orcid_activate_request"
```
joemull: I think we just want one of these now, since there is just one template, right?

src/core/logic.py
```diff
@@ -148,6 +148,42 @@ def send_confirmation_link(request, new_user):
...
+        publication_name = request.journal.name
+    elif request.repository:
+        publication_name = request.repository.name
+    else:
+        publication_name = request.press.name
+    context = {
+        "user": user,
+        "user_profile_url": request.site_type.site_url(
+            reverse("core_edit_profile"),
+        ),
```
joemull, line 170: Have you considered making the link straight to the verification step? I think this might be clearer for users who do not remember what the profile page looks like. If they are taken to the profile page, they have to scroll down to the social media and accounts section and notice the button to connect their ORCID, but if they are taken straight to the redirect cycle, they'll be prompted at each step what to do.

We have a helper function in this file that might be useful:
```py
reverse_with_query("core_edit_profile", { "action": "add_profile_orcid" })
```

src/core/models.py
```diff
@@ -948,6 +956,12 @@ def hypothesis_username(self):
         )[:30]
         return username.lower()
 
+    def get_orcid_url(self):
+        return f"{settings.ORCID_URL.replace('oauth/authorize', '')}{self.orcid}"
```
joemull: I wonder what are the benefits and drawbacks of this versus the logic in `orcid_uri` for the frozen author (and now the preprint author)? I'm thinking of a few things:

- The ORCID IDs in Janeway databases have sometimes been input as URLs (obviously something to fix when we make the field unique), so `get_orcid_url` would potentially result in repeating the host name and domain parts.
- It might be useful in some contexts to surface the `sandbox.orcid.org` domain, but this is not accounted for by `orcid_uri` because it hard-codes the domain.

Maybe it's time to combine the logic and have one method for `Account`, `FrozenAuthor`, and `PreprintAuthor`?

joemull (changes requested):
This is getting really close to being ready, and it's great that you were able to loop in the frozen author piece. I took a careful read through the logic and tested it out, and it's mostly working great for me. I just have a few requests around tests, the admin view, and maintainability, and one suggestion for usability--in the verification request, taking the user straight to the verification step rather than to their profile.

src/core/views.py
everreau: you're right this is better.

src/core/tests/test_views.py
everreau: I deleted this test.  I also don't recall exactly what I was trying to test but I think it might have merged with the other test as I worked through the login logic.

src/core/models.py
everreau: Good idea.  I added a utility method that combines both sets of logic.

everreau:
@joemull I think I've addressed all your comments.  Let me know if I've missed something.

src/core/views.py
joemull: This line hasn't changed since my comment above--I think `next_url` is still not set to the **current** view here. You need to add this line before the return statement, to get the path of the current request, so the user comes back here after logging in:

```py
next_url = request.get_full_path()
```

src/utils/orcid.py
```diff
@@ -168,3 +210,11 @@ def decode_state(encoded_state):
...
     if not COMPILED_ORCID_REGEX.match(orcid):
         raise ValidationError(f"{orcid} is not a valid ORCID")
+
+
+def normalized_orcid(orcid):
+    url = urlparse(settings.ORCID_URL)
+    result = COMPILED_ORCID_REGEX.search(orcid)
+    if result:
+        return f"{url.scheme}://{url.netloc}/{result.group(0)}"
+    return ""
```
joemull, line 220: Super!

src/core/logic.py
joemull: Thanks for making this change. I'd just request two small further changes for readability: 

1. Can we get away with just one level of reversing here, and rely on logic in the view to kick the user to the login screen?

   In other words, here we'd just send them to "core_login_orcid" with action="add_profile_orcid", and then, in that view, if they're not logged in, we'd get the current path, save it in next_url (see other comment), and kick them out to the login screen, expecting that Janeway will send them back to "core_login_orcid". OK with you?
2. Let's change the name of this template variable from "user_profile_url" to "orcid_verification_link" or something like that so it's clear in the template what it is.

joemull (changes requested):
Looking really good. Just missing one bit of logic around the redirection after non-ORCID login, and I had two nitpicks about readability.

src/core/views.py
everreau: ah, sorry, should be fixed now.  I think I missed this and mixed it up with a different comment.

src/core/logic.py
everreau: I do think the flow of logging in to janeway first and then orcid is a bit better for users but it likely won't affect many users since most will already be logged in to janeway so I'm fine with it.

everreau:
changes complete.

src/core/logic.py
joemull: @everreau OH! Now I understand the intent of having two levels of reversed URLs. If you want to change it back and add a comment explaining why, then I'd be happy with that. Let me know and then I'll move this along.

joemull:
@everreau OK super. See my reply here on the double redirects: https://github.com/openlibhums/janeway/pull/5110/changes#r3050292149

src/core/logic.py
everreau: 👍

joemull (approved):

src/utils/orcid.py
```diff
@@ -48,12 +50,49 @@ def retrieve_tokens(authorization_code, site):
...
+    if not token:
+        return False
+    is_sandbox = "sandbox" in settings.ORCID_URL
+    api_client = OrcidAPI(
+        settings.ORCID_CLIENT_ID, settings.ORCID_CLIENT_SECRET, sandbox=is_sandbox
+    )
+    r = api_client._get_public_info(
+        orcid_id, "record", token, None, "application/orcid+json"
+    )
+    return r.status_code == 200
```
mauromsl, line 82: There are extensive calls to this function, especially via the `is_orcid_token_valid`, sometimes during template rendering, sometimes during backend views (even through account loops). The request is non trivial, as it is fetching a public profile only to check if we get a 200 error.

I think we need to:
 - Use a HEAD request rather than GET
 - Consider handling token not valid (401 status code according to the docs) differently from other errors.
 - If we have a token and an expiry date recorded in our database, consider running this request only when strictly necessary (i.e. when directly interacting with ORCID API) and not while doing internal processing.
 - It sounds like it would be safe to consider caching a user's orcid token validity status at least for a single request cycle.

src/core/views.py
mauromsl, line 171: This function has become exceedingly long. Can we consider a refactor where shared logic can be moved to its own function and then write separate thinner views for branches such as `elif action == "add_profile_orcid":` (should probably be a view of its own).

It would be easier to maintain and, importantly, easier to unit test behaviour such as the scenario were we match multiple accounts with one ORCID

mauromsl (changes requested):
Thanks @everreau

I've gone through the PR and proposed some changes around two areas:

 - ORCID token validation
 - Refactor of the `user_login_orcid` view, for better maintainability.

I'm handing this over to Andy before passing it back to you, so you can get all the required feedback from us before making further changes

src/utils/orcid.py
everreau: @mauromsl thanks for the feedback.  I consulted with Alainna about proper orcid implementation and she agreed with your points.  I've implemented 2-4. As for using a HEAD request rather than GET I don't see how that can be done using the python orcid library (I did not introduce this). I could generate this call directly if you think it's worth it but it seems a bit clunky.

mauromsl:
@ajrbyers can you have a look when you are back?

ajrbyers:
> @ajrbyers can you have a look when you are back?

on it

src/core/views.py
```diff
@@ -283,6 +332,62 @@ def user_login_orcid(request):
...
+            messages.SUCCESS,
+            _("Your ORCID iD has been connected to your account."),
+        )
+    # return to profile page
+    return redirect(logic.reverse_with_next("core_edit_profile", next_url))
+
+
+@login_required
+@require_POST
+def request_orcid(request, account_id):
```
ajrbyers, line 378: As it currently stands any authenticated user who knows another users ID could spam them with requests.

src/core/forms/forms.py
```diff
@@ -278,6 +279,12 @@ class Meta:
             "enable_public_profile": YesNoRadio,
         }
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        if django_settings.ENABLE_ORCID:
+            self.fields["orcid"].widget = forms.HiddenInput()
```
ajrbyers: This will allow a user to alter their ORCID if they edit the underlying field, no?

ajrbyers (changes requested):
Generally, great, I spotted two issues.

everreau:
@ajrbyers @mauromsl I believe that this PR is ready to go.  Do you want me to rebase.  I tried to awhile ago but I started to get some unrelated test failures though maybe it was due to changes needed to default configuration.

mauromsl:
> @ajrbyers @mauromsl I believe that this PR is ready to go. Do you want me to rebase. I tried to awhile ago but I started to get some unrelated test failures though maybe it was due to changes needed to default configuration.

Yes please, at least to overcome the conflicts that have arisen since. We can then handle any test failures that may appear

PR #5113, opened 2026-01-28
Feature to allow the submission url to be overridden for custom pages.

RoseReatherford:
This feature allows for engineers to make `HIJACK_SUBMISSION_ENABLED` if they desire to make a truly custom submission experience for users.


mauromsl (commented):
Hi @RoseReatherford. Do you have an example on how HIJACK_SUBMISSION_ENABLED is used?

If done via 3rd party app/plugin, you should be able to modify any entries in Django's `URLResolver`, without requiring the custom setting in the `janeway` project, in the ready call for your app (as long as it loads after the default apps in `INSTALLED_APPS`

It can be retrieved with `from django.urls import get_resolver`, and can be altered from your app's `AppConfig.ready()`

If we were to look into overriding URLs from core, I think we'd favour a solution that was easier to generalize.

PR #5184, opened 2026-02-17
articles filter select and clear all buttons

### LINKED ISSUE #2097: "Select all" option on articles page
When a journal has many sections, it is difficult for users to filter the results down to a specific set of article sections

A button for (un)toogling all the sections in the articles and search pages

None

**Additional context**
Need to be added to all the built-in themes

StephDriver:
closes #2097

### DIFF
--- a/src/themes/OLH/templates/elements/journal/article_list_filters.html
+++ b/src/themes/OLH/templates/elements/journal/article_list_filters.html
@@ -7,9 +7,10 @@
         <h2 id="filter-title-{{ filter_context }}">{% trans 'Filter' %}</h2>
         <form method="GET" id="{{ facet_form.id }}-{{ filter_context }}">
             {{ facet_form|foundation }}
-            <button action="" class="button" type="submit">{% trans 'Apply' %}</button>
             <button type="button" class="button secondary" id="select-all-btn-{{ filter_context }}">{% trans 'Select all' %}</button>
             <button type="button" class="button secondary" id="clear-all-btn-{{ filter_context }}">{% trans 'Clear all' %}</button>
+            <br/>
+            <button action="" class="button" type="submit">{% trans 'Apply' %}</button>
         </form>
 
         <script src="{% static 'common/js/form-toggle-all.js' %}"></script>
--- a/src/themes/clean/templates/elements/journal/article_list_filters.html
+++ b/src/themes/clean/templates/elements/journal/article_list_filters.html
@@ -9,9 +9,10 @@
                 <h2 id="filter-title">{% trans 'Filter' %}</h2>
                 <form method="GET" id={{ facet_form.id }}>
                     {% bootstrap_form facet_form %}
-                    <button action="" class="btn btn-primary" type="submit">{% trans 'Apply' %}</button>
                     <button type="button" class="btn btn-secondary" id="select-all-btn">{% trans 'Select all' %}</button>
                     <button type="button" class="btn btn-secondary" id="clear-all-btn">{% trans 'Clear all' %}</button>
+                    <hr aria-hidden="true" />
+                    <button action="" class="btn btn-primary" type="submit">{% trans 'Apply' %}</button>
                 </form>
                 <script src="{% static 'common/js/form-toggle-all.js' %}"></script>
                 <script>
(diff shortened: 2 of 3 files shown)


StephDriver:
the problem was that OLH theme initialise the filter twice, once for mobile once for desktop, whereas teh other themes initalise it once. I modified the OLH article_list template to incldue a context for the filter (desktop, mobile) and used that to differentiate.

src/themes/clean/templates/elements/journal/article_list_filters.html
```diff
@@ -9,8 +10,13 @@ <h2 id="filter-title">{% trans 'Filter' %}</h2>
                 <form method="GET" id={{ facet_form.id }}>
                     {% bootstrap_form facet_form %}
                     <button action="" class="btn btn-primary" type="submit">{% trans 'Apply' %}</button>
-                    <button type="button" class="btn btn-secondary" onclick="window.location.href='{{ request.path }}'">{% trans 'Clear all' %}</button>
+                    <button type="button" class="btn btn-secondary" id="select-all-btn">{% trans 'Select all' %}</button>
+                    <button type="button" class="btn btn-secondary" id="clear-all-btn">{% trans 'Clear all' %}</button>
```
joemull, line 13: Any reason not to use something like `clear-all-btn-{{ filter_context|default:"unified" }}` here for readability?

src/themes/OLH/templates/elements/journal/article_list_filters.html
```diff
@@ -1,13 +1,20 @@
...
             {{ facet_form|foundation }}
             <button action="" class="button" type="submit">{% trans 'Apply' %}</button>
-            <button class="button warning" name="clear_all">{% trans 'Clear all' %}</button>
+            <button type="button" class="button secondary" id="select-all-btn{% if filter_context %}-{{ filter_context }}{% endif %}">{% trans 'Select all' %}</button>
+            <button type="button" class="button secondary" id="clear-all-btn{% if filter_context %}-{{ filter_context }}{% endif %}">{% trans 'Clear all' %}</button>
         </form>
+
+        <script src="{% static 'common/js/form-toggle-all.js' %}"></script>
+        <script>
+            initFormToggleButtons('select-all-btn{% if filter_context %}-{{ filter_context }}{% endif %}', 'clear-all-btn{% if filter_context %}-{{ filter_context }}{% endif %}', '{% if filter_context %}{{ facet_form.id }}-{{ filter_context }}{% else %}{{ facet_form.id }}{% endif %}');
```
joemull: As one long line, this is very hard to read. Can we not put each of these in their own JS variable above this line, and use the more succinct `|default` template filter to avoid the template if / else?

src/themes/clean/templates/elements/journal/article_list_filters.html
joemull, line 13: This changes the behavior of the button. It just clears the form fields, rather than reloading the page with those form inputs applied to the results.

Is that what you intended? If so, consider putting the clear and select buttons above the apply button. I think this would make it clearer that clicking **Apply** is expected every time as a final step, even after you use the select or clear buttons.

StephDriver: I don't understand the benefit. Could you elaborate @joemull ?

src/themes/OLH/templates/elements/journal/article_list_filters.html
StephDriver: Looking at this again, I realised the logic was not needed at all, as there is only one template that includes this, and it passes in the filter_context, so it will always be there. 

I have removed the logic, and changed the spacing of the script call to improve readability.

src/themes/clean/templates/elements/journal/article_list_filters.html
StephDriver: yes, I intended this because the page-load part was an accessibliity issue -the page reload was a change in context.  see [WCAG Technique H32](https://www.w3.org/WAI/WCAG22/Techniques/html/H32).  It may be that a users wants to clear the filter from the articles, it may also be that the just want to clear the selections prior to making new selections.   Having an apply button is more predictable. 

I like the idea of putting the apply button below, will do!

joemull: The latest changes obviated the need for this.

src/themes/OLH/templates/elements/journal/article_list_filters.html
joemull: Great!

PR #5192, opened 2026-02-25
Adds option to gather citation metrics using OpenAlex's API

ajrbyers:
Janeway currently pulls forward citation links from Crossref. This PR adds a parallel workflow using OpenAlex, which has broader coverage and a friendlier REST API.                                                                                                                                              
                                                                                                                                                                                                                                                                                                                    
  Model changes:                                                                                                                                                                                                                                                                                                    
  - New openalex identifier type to store OpenAlex Work IDs against articles
  - source field on ArticleLink/BookLink so you can tell whether a citation came from Crossref or OpenAlex
  - other object type for works that aren't articles or books (preprints, dissertations, etc.)

  New commands:
  - fetch_openalex_ids — looks up the OpenAlex Work ID for each published article that has a DOI, stores it as an Identifier
  - fetch_openalex_citations — for each article with an OpenAlex ID, fetches all citing works and saves them as ArticleLink/BookLink records. Uses cursor pagination and routes by work type.

  Both commands accept --mailto (polite pool), --journal_code, --article_id, and --delay.

  fetch_forward_links updated to set source="crossref" on its records so existing Crossref citations aren't overwritten by OpenAlex runs.


ajrbyers:
Ruff is complaining here that some files untouched by this PR would be formatted...

Issue #3726, opened 2023-09-20
Updates for Isolinear Plugin


ajrbyers:
**Is your feature request related to a problem? Please describe.**
The [Isolinear Plugin](https://github.com/BirkbeckCTP/isolinear) will work to combine the repository and journal workflows. There are some changes that are required to the core in order for this to be effective.

**Describe the solution you'd like**
- [ ] Add new hook to the Unassigned action buttons list
- [ ] Add new hook to the Review action buttons list
- [ ] Add a new setting to control displaying a linked article <> preprint on the article page pre-publication
    - [ ] When this setting is enable we should embed the preprint PDF on the article page
- [ ] Update the repository to run in headless mode so that there is not front end

ajrbyers:
Closed a duplicate of this: https://github.com/BirkbeckCTP/janeway/issues/3669

ajrbyers:
- [ ] Allow an XML download for the Preprint with just the stub before the article is published

PR #5194, opened 2026-02-26
Merge of Iowa and Isolinear features into master

ajrbyers:
Closes https://github.com/openlibhums/janeway/issues/3726

### DIFF
--- a/src/core/test_settings.py
+++ b/src/core/test_settings.py
@@ -1,9 +1,0 @@
-from core.settings import *
-
-DATABASES = {
-    "default": {
-        "ENGINE": "django.db.backends.sqlite3",
-        "NAME": ":memory:",
-    }
-}
-print("TEST SETTINGS LOADED")


src/themes/OLH/templates/repository/elements/hierarchy_list.html
```diff
@@ -0,0 +1,27 @@
+{% load i18n %}
+
+{% for item in hierarchy %}
+<li>
+  <div class="ou-tree-item {% if rou and item.unit == rou %}selected{% endif %}">
+    <svg class="ou-tree-item-icon" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round">
```
ajrbyers: Move to media.

src/api/permissions.py
```diff
@@ -29,3 +32,47 @@ def has_permission(self, request, view):
...
+            return True
+
+
+class IsPreprintOwner(permissions.BasePermission):
+    message = 'You must be the owner of this preprint to edit it.'
+
+    def has_permission(self, request, view):
+        # grant access to non-create/update requests
+        if request.method not in ['PUT', 'PATCH', 'DELETE']:
+            return True
```
mauromsl: It feels strange that a `GET` request checking for preprint owner of a resource returns True for all users.

I'd either rename this to `canEditPreprint` or preserve `isPreprintOwner` but move the `GET` check to the view that wants to allow it.

src/core/janeway_global_settings.py
```diff
@@ -721,3 +721,7 @@ def __len__(self):
 # Note that the provided theme CSS expects a default crop size of (1500, 648)
 # and may not work properly with a different size.
 DEFAULT_CROP_SIZE = (1500, 648)
+
+# This setting should only be enabled on Dev or where CORS is properly
+# configured to stop misuse of this endpoint.
```
mauromsl: should we add it to `dev_settings.py` then?

src/api/urls.py
```diff
@@ -11,11 +13,40 @@
...
+    router.register(
+        r'account/register',
+        views.RegisterAccount,
+        'register_account',
+    )
+    router.register(
+        r'account/activate',
+        views.ActivateAccount,
+        'activate_account',
+    )
```
mauromsl, line 58: Are these also only meant to be available when `API_ENABLE_ACCOUNT_SEARCH` is on?

in that case, should it just be `API_ENABLE_ACCOUNT_ENDPOINTS`?

src/core/templatetags/hooks.py
```diff
@@ -11,15 +11,20 @@
...
             hook_module = import_module(hook.get("module"))
             function = getattr(hook_module, hook.get("function"))
-            html = html + function(context, *args, **kwargs)
+            hook_output = function(context, *args, **kwargs)
 
-        return mark_safe(html)
+            if hook_output:
+                html = html + hook_output
+        if html:
+            return mark_safe(html)
```
mauromsl: If no exception is thrown, the default value will be ignored:
```
  if html:
      return mark_safe(html)
  else:
      return mark_safe(default_value)
```

src/core/plugin_loader.py
```diff
@@ -95,12 +95,7 @@ def validate_plugin_version(plugin_settings):
     current_version = version.parse(janeway_version.base_version)
     valid = current_version >= wants_version
 
-    if not valid:
-        raise ImproperlyConfigured(
-            "Plugin {} not  compatibile with current install: {} < {}".format(
-                plugin_settings.PLUGIN_NAME, current_version, wants_version
-            )
-        )
```
mauromsl, line 103: I imagine this is here for dev purposes. Perhaps we need a `if not valid and not settings.DEBUG`

src/identifiers/admin.py
```diff
@@ -97,8 +108,12 @@ class CrossrefDepositAdmin(admin.ModelAdmin):
     list_select_related = True
 
     def _journal(self, obj):
-        if obj and obj.crossrefstatus_set.first():
-            return obj.crossrefstatus_set.first().identifier.article.journal
+        fist_status = obj.crossrefstatus_set.first()
+
+        if obj and fist_status and fist_status.identifier.article:
```
mauromsl: The first part of this change seems wrong,  the check for `if obj` should happen on the first access of `obj.crossrefstatus_set.first` above. Otherwise it may raise an `AttributeError: Nonetype object has no attribute {}` at that point (which I guess it is why it was there in thefirst place.

src/identifiers/preprints.py
```diff
@@ -0,0 +1,119 @@
...
+    except (requests.exceptions.ConnectionError, requests.exceptions.Timeout) as e:
+        status = 'Error depositing. Could not connect to Crossref ({0}). Error: {1}'.format(
+            depositor.get_endpoint(verb='deposit'),
+            e,
+        )
+        crossref_deposit.result_text = status
+        crossref_deposit.save()
+        logger.error(status)
+        return status, e
+    if response.status_code == 200:
```
mauromsl: minor nitpick, do we care if the code is `200` or one of the other variants of `response.ok`? If we do, we need to handle them. If we don't, then just check against `response.ok`

src/identifiers/urls.py
```diff
@@ -9,46 +9,56 @@
 urlpatterns = [
     re_path(r"^pingback$", views.pingback, name="crossref_pingback"),
     re_path(
-        r"^(?P<article_id>\d+)/$", views.article_identifiers, name="article_identifiers"
+        r"^(?P<content_type>article|preprint)/(?P<object_id>\d+)/$",
+        views.identifiers,
+        name="identifiers",
     ),
```
mauromsl, line 15: do we need to support the previous URL for a release, so that pingbacks still work for deposits in flight? probably worth adding this one to the list of backwards-compatible URLs below

src/templates/admin/repository/submit/start.html
```diff
@@ -2,123 +2,124 @@
...
+          margin-left: 2rem;
+      }
+
+      .ou-depth-3 {
+          margin-left: 3rem;
+      }
+
+      .ou-depth-4 {
+          margin-left: 4rem;
+      }
```
mauromsl: we can make this a bit more robust by setting  the digit as a css varibale:

```css
.ou-depth {
  margin-left: calc(var(--ou-depth, 0) * 1rem);
}
```

Then the HTML render would be like:

```html
<label class="custom-radio ou-depth" style="--ou-depth: {{ depth }};">
```

PR #5212, opened 2026-03-05
Handle possibility of missing date_complete

### LINKED ISSUE #5206: Decision status can fail to load due to missing date
## Problem
I came across this while doing support. If a review is missing a `date_complete`, it can cause the decision status to choke on this line with an attribute error:

https://github.com/openlibhums/janeway/blob/4a00de15ca0bd0e0cc8ab245b15a3dcca5d9c841/src/review/models.py#L401-L403

## Version
4a00de15ca0bd0e0cc8ab245b15a3dcca5d9c841

## Proposed solution

Allow for an unknown date:

```diff
-            return f"Withdrawn {self.date_complete.date()}"
+            if self.date_complete:
+                date = self.date_complete.date()
+            else:
+                date = '[unknown date]'
+            return f"Withdrawn {date}"
```

joemull:
Fixes #5206.

### DIFF
--- a/src/review/models.py
+++ b/src/review/models.py
@@ -400,7 +400,11 @@ class ReviewAssignment(models.Model):
 
     def request_decision_status(self):
         if self.decision == RD.DECISION_WITHDRAWN.value:
-            return f"Withdrawn {self.date_complete.date()}"
+            if self.date_complete:
+                date = self.date_complete.date()
+            else:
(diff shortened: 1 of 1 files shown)


mauromsl (approved):
I believe #5221 was the missing piece leading to this, but since the field is still nullable, there is no harm here

PR #5235, opened 2026-03-19
Fixes a bug that was stopping all CREDIT roles from displaying

ajrbyers:
RadioSelect does not add a blank entry to the choices list so we don't need to slice the list.


joemull (changes requested):
Thanks for catching this. I approve of the change, but I think we should do the PR against 1.8 so people get it faster. OK with you?

ajrbyers:
Closed for https://github.com/openlibhums/janeway/pull/5284

PR #5239, opened 2026-03-24
docs: Add and document a working Lando setup for local Janeway development

kopfduenger:
## Summary

This PR updates and documents a tested Lando-based local development workflow for Janeway.

## What changed

- fix the tracked `.lando.yml` so a fresh clone has the required shared Lando configuration
- add a dedicated Lando guide under `docs/source/dev/lando.rst`
- link that guide from the developer installation docs
- remove redundant parallel Lando explanation from `docs/source/dev/installation.rst`
- link the Lando guide from `README.md`
- switch local Lando development defaults in `src/core/dev_settings.py` to:
  - `DEFAULT_HOST = "http://janeway.lndo.site"`
  - `URL_CONFIG = "domain"`
- revise the new and touched docs toward plainer, more direct wording by removing “just/simple”-style phrasing where it was avoidable

## Why

The previous Lando guidance was outdated for the current repository state. This PR documents a tested workflow and covers practical issues contributors may hit, including:

- current proxy/domain-mode behavior
- startup noise on first boot
- Docker/Lando version expectations
- `arm64` vs `amd64` package differences
- `hosts` file fallback when `*.lndo.site` does not resolve
- post-install domain alignment when installer defaults save `localhost`

It also reduces duplication in the developer docs by making the dedicated Lando page the main source of truth for this workflow.

## Notes for reviewers

- `.lando.local.yml` remains intentionally local and untracked
- local editable dependency checkouts under `src/` are not part of this PR
- the new Lando guide is integrated into the existing Sphinx / Read the Docs structure
- the wording changes aim to keep the setup instructions direct and easier to follow for new contributors

### DIFF
--- a/.lando.yml
+++ b/.lando.yml
@@ -39,7 +39,9 @@ services:
     overrides:
       environment:
         DB_HOST: "db"
-        PYTHONPATH: "/app/src"
+        # Include vendored editable package roots so Django app labels resolve
+        # to the actual package instead of a namespace package shell.
+        PYTHONPATH: "/app/src:/app/src/foundationform"
         JANEWAY_SETTINGS_MODULE: "core.dev_settings"
       labels:
         # Janeway runs its Django development server on port 8000, but Lando's


kopfduenger:
I replaced the earlier settings-file workaround with a Lando-specific fix.

This update adds `/app/src/foundationform` to `PYTHONPATH` in `.lando.yml`, so the vendored `foundationform` package resolves correctly in the local Lando environment. That means we no longer need to touch `src/core/janeway_global_settings.py` for this issue.

I verified this with:
- `lando rebuild -y`
- `import foundationform` resolving to the expected vendored package path
- `lando manage check` passing successfully

PR #5247, opened 2026-03-27
Warn editor of missing or incorrect review URL in notifications

ajrbyers:
A confirmation modal will now pop if the review url does not match the expected url.

image


src/review/views.py
```diff
@@ -400,18 +400,23 @@ def assignment_notification(request, article_id, editor_id):
             email_context=email_context,
             request=request,
         )
-        if form.is_valid():
+        skip = request.POST.get("skip")
+        form.is_valid()
+        if skip or form.is_valid():
```
joemull: Is this a drive-by bugfix, or are you trying to use a confirmable form here? Or did you mean to change the review assignment view?

If it's a bugfix, you don't need form.is_valid() in line 404.

src/review/tests/test_views.py
```diff
@@ -0,0 +1,1146 @@
...
+                "cc": "",
+                "bcc": "",
+                "subject": "You have been assigned",
+                "body": "You have been assigned as an editor.",
+                "send": "send",
+            },
+            SERVER_NAME=self.journal_one.domain,
+        )
+        self.assertEqual(len(mail.outbox), 1)
+        self.assertIn(self.editor.email, mail.outbox[0].to)
```
joemull, line 1147: The new tests making sure editor assignments are great, but you haven't made any tests for the changes to the review assignment view, and there aren't any pre-existing ones for this view, only for the corresponding repository view. Do you think it's worth adding a few?

src/review/forms.py
```diff
@@ -329,6 +329,34 @@ def check_for_potential_errors(self):
...
+        if not self.expected_url:
+            return []
+        body = self.cleaned_data.get("body", "")
+        if self.expected_url not in body:
+            return [
+                _(
+                    "The email body does not contain the expected reviewer URL "
+                    "(%(url)s). Do not copy this template from elsewhere, as it "
+                    "must include the link to the review request."
+                )
```
joemull, line 355: I'd suggest changing this to make it a bit easier to understand, like this:

"The email body does not contain the expected reviewer URL "
"(%(url)s), so the reviewer will not be able to complete the request. "
"/nThis can happen if the body was copy-pasted into "
" the text box from elsewhere."

joemull (changes requested):
I test it out and it works fine for me. A few changes requested.

src/review/views.py
ajrbyers: Yeah it was a drive-by I picked up whilst working on this. form.is_valid() must be called before form.as_dataclass() to populate cleaned_data, skip short circuits the call so I've shifted it to a variable.

src/review/tests/test_views.py
ajrbyers: I've added some tests for views.

src/review/forms.py
ajrbyers: Sure this is fine.

joemull (approved):

PR #5249, opened 2026-03-30
Errata backend

gamboz:
Add the possibility of linking an erratum to an article and register this relation with Crossref.

See also [discussion 5238](https://github.com/orgs/openlibhums/discussions/5238)

- add Genealogy model to link Articles
- add crossmark/updates section to crossref xml deposit for errata

ATM, errata are so rare that no front-end or manager interface is changed here, but an "admin" is provided. 

Please note the new dependency.


gamboz:
@mauromsl we have been notified of a small incoherence in the XML deposit specs (see https://github.com/orgs/openlibhums/discussions/5238#discussioncomment-16599730 ).

I've moved this MR to "draft".

I'll complete it after we publish our first erratum and ping again.

src/templates/common/identifiers/crossref_article.xml
```diff
@@ -54,6 +54,31 @@
                 <item_number item_number_type="article_number">{{ article.object.pk }}</item_number>
             </publisher_item>
 
+            {% if article.erratum_of %}
+            <crossmark>
+              <crossmark_policy>{{ article.object.journal|setting:'crossref_prefix' }}/not-used</crossmark_policy>
```
gamboz, line 59: :warning: according to XML specs, when recording an `<update>`, then a `<crossmark_policy>` DOI should also be provided, but, according to crossref support, that element is not used and any DOI would do. They suggested to use the article DOI, but I'm hardcoding a `10.11111/no-used` fake DOI (I feel it's less confusing...)

gamboz:
@mauromsl we talked with @ajrbyers on Friday and after a couple of minor fixes I'm setting this as ready for review.

Note that there might be some issues with migrations that (I think) are not related to these changes.\
Please let me know if you want me to point this to some other branch.

src/submission/models.py
```diff
@@ -2630,6 +2631,26 @@ def best_large_image_alt_text(self):
...
+        """
+        Return the "parent" article for which this article is an erratum.
+
+        This is intended to be used in
+        templates/common/identifiers/crossref_article.xml
+        """
+        if self.section.name != "Erratum":
+            return None
+        if not self.ancestors.exists():
+            return None
```
mauromsl: While it makes sense on the context of this PR, section.name is a customizable (and translatable) entry. Since we are adding a new model to register the relationships between articles, it would make sense to codify this (and other relationships) as part of that model, rather than relying on the indirection of the section name.

src/submission/models.py
```diff
@@ -3400,6 +3421,28 @@ def handle_defaults(self, article):
...
+    )
+    children = SortedManyToManyField(
+        Article,
+        related_name="ancestors",
+    )
+
+    def __str__(self):
+        return f"Genealogy: {self.parent} has {self.children.count()} kids"
+
+
```
mauromsl: I think the community are asking for a few more relationship types such as addendum, correction and so on. We also have a model on the hydra plugin for registering relationships such as translations.

Would it work for your use case if we were to port [the LinkedArticle model from Hydra plugin](https://github.com/openlibhums/Hydra/blob/33c8eb8082b6f11d8bc883f432a31461ac8a3786/models.py#L53) instead?

It behaves as a linking table, so there is one less join required when querying both sides of the relationship.

requirements.txt
```diff
@@ -59,6 +59,7 @@ python-magic==0.4.27
 pytz==2024.1
 requests==2.32.4
 six==1.16.0
+django-sortedm2m~=3.1
```
mauromsl: If we go with the approach I proposed about registering a model that behaves as a linking table, we could also use [our M2MOrderedThroughField](M2MOrderedThroughField). It would avoid the additional dependency and also maintain consistency with other ordered many to many relationships in the codebase.

mauromsl (changes requested):
Thank you @gamboz great work. I've proposed two main changes inline:

- Avoid using the section names as the mechanism for serializing the relationship type of "errata"
- Use a linker table with a `through` model relationship, ideally repurposing the model that already lives in the hydra plugin.

src/submission/models.py
```diff
@@ -2617,6 +2618,84 @@ def best_large_image_alt_text(self):
...
+            )
+        if direction == "ancestors":
+            # self is the "to-article"; return the "from-articles"
+            return Article.objects.filter(
+                linked_from__to_article=self, linked_from__relationship=link_type
+            )
+
+        raise ValueError(
+            f"Unknown relationship direction '{direction}' requested for {self.id}"
+        )
```
gamboz: I wrote these methods as helpers for the theme templates, but before seeing that there was the [sidebar_article_links](https://github.com/openlibhums/Hydra/blob/33c8eb8082b6f11d8bc883f432a31461ac8a3786/hooks.py#L16) hook in hydra.

ATM, these are not used and probably not necessary.

Let me know if you prefer I drop them 🙂 

Please also see commit [3dc9f69](https://github.com/openlibhums/Hydra/pull/12/changes/3dc9f69a0b2ef88d78516e035da70964b4359855) in Hydra

src/templates/common/identifiers/crossref_article.xml
gamboz: @mauromsl please feel free to resolve this thread if you have no objections

src/submission/models.py
gamboz: should be addressed by https://github.com/openlibhums/Hydra/pull/12

src/submission/models.py
gamboz: should be addressed by https://github.com/openlibhums/Hydra/pull/12

requirements.txt
gamboz: should be addressed by https://github.com/openlibhums/Hydra/pull/12

gamboz:
@mauromsl something funny happened to this MR: I don't see my recent commits and the changes are still the old ones...

Please ignore this MR until I understand what's going on :cry:

gamboz:
> @mauromsl something funny happened to this MR: I don't see my recent commits and the changes are still the old ones...
> 
> Please ignore this MR until I understand what's going on 😢

Situation back to normal... (was github unhappy with me? I will worry about this on the beach all next week :beach_umbrella: :partying_face: )

gamboz:
> ...something funny happened to this MR: I don't see my recent commits and the changes are still the old ones...

It happened again and I think it was due to me pushing the correct changes directly to github/sissamedialab, but also having a stale branch on our internal forge; the branches from the internal forge are automatically mirrored every so often to github/sissamedialab, and the stale branch kept overwriting the good one ( :disappointed: stupid me).

It should be fixed now. ATM I see this:

two changes only

@mauromsl I apologize for the confusion :pray: 

Please can you review this PR again and also https://github.com/openlibhums/Hydra/pull/12 ?

PR #5250, opened 2026-03-31
Editors can now skip copyeditor acknowledgement emails

### LINKED ISSUE #1027: Inconsistent behaviour/cannot skip sending message to copyeditor
**Describe the bug**
Logged in as an editor with a privileged account. I want to accept the completed Copyediting task. When I click on the 'Accept Copyedit' button on the left, a window appears saying: 
'You can write a note to the copyeditor thanking them for their time.' In fact, there is no 'can' about it, as I would appear to have no choice, as the usual 'skip' the email button is missing. There appears to be no way to break out of this other than to click on the 'send' button.
Then the copyedit appears to be accepted as I get "Editor accepted the Copyedit' in the action log:

**Janeway version**
https://demojournal.janeway.systems/copyediting/article/613/#

**To Reproduce**
As above.

**Expected behavior**
Get the option to skip this email requirement which is the case in similar situations (or change the text from 'can' to 'must' etc.)

**Screenshots**
Screenshot (150)

Screenshot (151)

Screenshot (152)

## Front-end Issues

Windows 19 with Chrome.

**Additional context**

ajrbyers:
Closes #1027 

- Editors can now skip sending this message
- It also now uses the email data class and has its own view/template to be consistent with others
- Added some tests to ensure skipping skips 😄


joemull (changes requested):
Looks pretty good. I just noticed a couple things:

- The HTML heading structure you use here is not accessible, but I assume you knew this and were copy-pasting a different existing template, with a plan to change all these later?
- I tested it out and noticed that the view for re-opening the copyedit still uses the modal. In scope or out of scope?
  Image
- Needs ruff formatting to pass CI

ajrbyers:
1. Yes its just using the standard template.
2. It probably is but I've added it
3. Ruff should pass now

joemull:
Assigning this back to @mauromsl since I won't have time to do it before planned leave.

joemull:
Taking this back since no one else has reviewed it during my leave.

joemull (approved):

joemull:
@ajrbyers assigning back to you to resolve conflicts.

PR #5263, opened 2026-04-13
Fix featured articles homepage hook

kopfduenger:


kopfduenger:
This PR fixes the Featured Articles homepage element so it no longer depends on the display name `"Featured Articles"` to decide whether the hook should run.

The previous implementation checked the element name, which made the feature fragile: if a journal renamed the homepage block, the featured articles context was no longer added and the block could stop rendering its articles. The change uses the stable `configure_url` (`featured_articles_setup`) instead, and also exposes that identifier in the plugin hook registration.

I added coverage for both the hook-level behavior and the journal homepage rendering case where the element has been renamed, so the intended behavior should be protected from regressions.

PR #5264, opened 2026-04-13
Redirect authors from submitted article edit links

kopfduenger:


kopfduenger:
This PR improves the user experience when authors follow an edit/submission link for an article that has already been submitted.

Previously, the `article_is_not_submitted` decorator returned a 404 once an article had a `date_submitted`, even for the article owner. That made stale dashboard links or bookmarked submission links feel like broken access. The new behavior redirects staff users and the article owner to the article dashboard instead, with an informational message explaining that the article has already been submitted.

The article dashboard now also gives authors a clearer status summary: current stage, submitted date, section, keywords, abstract, a live article link when published, and a note when no author action is currently required. I added tests for the redirect and the dashboard summary/published-link behavior.

PR #5265, opened 2026-04-13
Suggest journal keywords during submission

kopfduenger:


kopfduenger:
This PR adds journal-level keyword suggestions to the submission metadata keyword field.

The goal is to make the existing journal disciplines/keywords more useful during submission: when authors or editors enter article keywords, the Tag-it field now receives `request.journal.keywords` as `availableTags` in both the author submission metadata step and the admin metadata edit view. This keeps free keyword entry intact while surfacing the journal’s configured vocabulary as suggestions.

The implementation also escapes the rendered keyword values for JavaScript output, and I added tests for both submission and metadata-edit views to confirm that journal keywords appear as suggestions.

ajrbyers:
Thanks for this submission. I can see it would work well for a small set of keywords, however, we'd need this solution to work well for n+1 keywords. As an example the OLH along has 56,000 keywords in its database.

kopfduenger:
Good point, thanks. The current implementation inlines the journal keyword list into the page, which is fine for a small controlled discipline list but would not scale for large keyword sets such as OLH’s.

I’ll rework this to use server-side autocomplete instead: the form will request matching keyword suggestions only after the user starts typing, and the backend will return a limited result set. That should avoid loading all keywords into the page while preserving the suggestion behaviour.

ajrbyers:
> Good point, thanks. The current implementation inlines the journal keyword list into the page, which is fine for a small controlled discipline list but would not scale for large keyword sets such as OLH’s.
> 
> I’ll rework this to use server-side autocomplete instead: the form will request matching keyword suggestions only after the user starts typing, and the backend will return a limited result set. That should avoid loading all keywords into the page while preserving the suggestion behaviour.

We're standardising our approach to reactive interfaces using HTMX, not sure if that's something you've come across before.

kopfduenger:
Thanks, that makes sense. I’m familiar with the general idea of HTMX, though I haven’t used it much in Janeway yet.
I’ll take a look at the existing Janeway HTMX patterns and rework this so keyword suggestions are fetched server-side with a limited result set, rather than inlining all keywords in the page.

kopfduenger:
I had a quick look through the current codebase and couldn’t find existing HTMX usage or established `hx-*` patterns yet, so I’m guessing this is a direction you’re standardising on rather than something already widely present in the codebase? Is there a preferred Janeway HTMX pattern or branch I should follow? Otherwise I can propose a small HTMX-based server-side autocomplete with a limited result set instead of inlining all keywords into the page.

ajrbyers:
You are correct. We have some PRs in the works that we're using HTMX for, I can pull some out this afternoon and send them to you.

kopfduenger:
Great, thank you.

kopfduenger:
Following up on our earlier discussion in #5265, I put together a small clean patch for the HTMX/autocomplete approach suggested there.

It stays on top of the existing `submission.Keyword` model, is rebased onto current `master`, and I tested the affected workflow paths locally. I’ve opened it as a small follow-up PR in case it’s useful.

PR: https://github.com/openlibhums/janeway/pull/5300

PR #5284, opened 2026-04-22
Fix display issue for conceptualisation role

ajrbyers:
RadioSelect does not add a blank entry to the choices list so we don't need to slice the list.


ajrbyers:
Assigned to @mauromsl as its been approved on https://github.com/openlibhums/janeway/pull/5235 just changed to v1.8.

PR #5286, opened 2026-04-23
accessibility page text

### LINKED ISSUE #4852: 1.9 Accessibility information text
- #4694 will add a place for accessibility information.
- This will then need to be updated with each release.

So first, the information for 1.9 needs to be drafted!

- [ ] link to A11y project
- [ ] link to A11y roadmap
- [ ] link to contact page
- [ ] prominent link to Sitemap
- [ ] link to Janeway site for more info (? vpat on Janeway site?)

As part of this, may need to reorganise the A11y project board, so that say ''known bugs" are clearly listed in a single view, like how releases are.

Need to consider how this is translated.  If we are making a statement about Janeway accessibility in English, do we need to be clear that a translation by someone else  and give access to the original?

StephDriver:
closes #4852

### DIFF
--- a/src/journal/views.py
+++ b/src/journal/views.py
@@ -4,6 +4,7 @@ __license__ = "AGPL v3"
 __maintainer__ = "Birkbeck Centre for Technology and Publishing"
 
 import json
+import os
 import re
 
 from django.conf import settings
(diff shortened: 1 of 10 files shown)


StephDriver:
split original issue 4852 into three sub issues, as not all parts can be done yet - so that VPAT work can be reviewed and merged rather than wait around for a11y mode.

PR #5293, opened 2026-04-29
Enhance plugin list page layout and list them in alphabetical order

ajrbyers:
- Makes the Plugin list page look better
- Lists all plugins in alpha order


mauromsl (approved):
awesome!

PR #5296, opened 2026-04-30
Remove circular import

gamboz:
`journal.forms` imports `journal.logic` which in turn imports `journal.forms`.

`logic` is not used in `journal.forms` code, so it's a redundant import in any case.

This has not triggered any error just due to a lucky combination of factors, but changes in WJS project triggers this error 🙂

(thanks @yakky for noticing)

### DIFF
--- a/src/journal/forms.py
+++ b/src/journal/forms.py
@@ -13,7 +13,7 @@ from tinymce.widgets import TinyMCE
 
 from core import models as core_models
 from core.forms import FullSettingEmailForm
-from journal import models as journal_models, logic
+from journal import models as journal_models
 from utils.forms import CaptchaForm
 
 SEARCH_SORT_OPTIONS = [
--- a/src/review/tests.py
+++ b/src/review/tests.py
@@ -296,7 +296,7 @@ class ReviewTests(TestCase):
             article=article1,
             is_complete=False,
             review_round=round,
-            reviewer=self.regular_user,
+            reviewer=self.third_reviewer,
         )
         assignment.decision = None
         assignment.save()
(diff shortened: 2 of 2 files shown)


ajrbyers (approved):

ajrbyers:
@gamboz did tests pass locally for you?

gamboz:
> @gamboz did tests pass locally for you?

Didn't try: I was counting on the ci/cd, but didn't expect any issue...

Let me check...

gamboz:
> > @gamboz did tests pass locally for you?
> Let me check...

I have the same error locally,\
but I have the same error locally also for branch r-v1.8x \
but not for r-v1.7x


thinking out loud...

AFAICT, the two tests
- `test_review_assignment_form_valid`
- `test_review_assignment_form_bad_reviewer`

differ only because 
- the Account used as reviewer in the first test (`self.second_reviewer`) has the "reviewer" role for the article's journal,
- while the account used in the second test (`self.regular_user`) does not have that role.

in the form, this is reflected in the queryset from which to select a reviewer.

That queryset is built by `review.logic.get_reviewer_candidates()` and `review.logic.get_reviewers()`.

My suspicion is that, during the tests setup, we create some `ReviewAssigment`s using `regular_user` as reviewer, and this "percolates" into the queryset of the candidates.

Let me add a [commit](df9705c052d8660ea576a6d3121a12773f492a6e) to test... and yes if I use another user with the reviewer role wherever the regular-user was used as reviewer, then 
- `test_review_assignment_form_bad_reviewer` is ok  :partying_face: 
- but `test_review_assignment_form_valid` fails :cry: 

@ajrbyers IIC, the failing test has nothing to do with this MR. \
If it's ok with you, I'd like to submit this change for merge anyway (without the second commit), and address this issue in another MR.


[off topic] should I assign this MR to you?

gamboz:
@ajrbyers FYI
I've just rebased this and the issue with the tests seems resolved.

PR #5300, opened 2026-05-05
Improve journal keyword autocomplete and local static serving

kopfduenger:
This is a follow-up to the earlier keyword-entry discussion in #5265, using the HTMX/autocomplete direction suggested there.

What it changes:
- adds journal keyword suggestions during submission
- loads keyword suggestions remotely in the submission/admin metadata UI
- keeps local media/static serving working without enabling full debug mode

Notes:
- suggestions are sourced from the existing `submission.Keyword` records already used by journals and articles, rather than introducing a separate keyword store
- the branch is cleanly based on `upstream/master`

Validation:
- `lando manage check`
- `lando manage test submission.tests.test_workflow security.test_security`

Manual checks:
- journal settings keyword suggestions
- submission article info keyword suggestions
- article metadata keyword suggestions

### DIFF
--- a/src/core/include_urls.py
+++ b/src/core/include_urls.py
@@ -153,6 +153,11 @@ urlpatterns = [
         name="core_edit_settings_group",
     ),
     re_path(
+        r"^manager/settings/journal/keywords/suggest/$",
+        core_views.journal_keyword_suggestions,
+        name="core_journal_keyword_suggestions",
+    ),
+    re_path(
         r"^manager/settings/(?P<plugin>[-\w.:]+)/(?P<setting_group_name>[-\w.]+)/(?P<journal>\d+)/$",
         core_views.edit_plugin_settings_groups,
         name="core_edit_plugin_settings_groups",
--- a/src/core/views.py
+++ b/src/core/views.py
@@ -20,14 +20,14 @@ from django.urls import NoReverseMatch, reverse, reverse_lazy
 from django.shortcuts import render, get_object_or_404, redirect, Http404
 from django.utils import timezone
 from django.utils.decorators import method_decorator
-from django.http import HttpResponse, QueryDict
+from django.http import HttpResponse, QueryDict, JsonResponse
 from django.contrib.messages.views import SuccessMessageMixin
 from django.contrib.sessions.models import Session
 from django.core.validators import validate_email
 from django.core.exceptions import ValidationError
 from django.db import IntegrityError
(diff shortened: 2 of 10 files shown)


src/core/dev_settings.py
```diff
@@ -2,9 +2,10 @@
 # You should change this key before you go live!
 import os
 
-DEBUG = True
+DEBUG = False
```
ajrbyers: DEBUG should always be TRUE on dev_settings.py

src/core/dev_settings.py
```diff
@@ -2,9 +2,10 @@
...
 FORCE_BUILTIN_XSL = False
 SECRET_KEY = "uxprsdhk^gzd-r=_287byolxn)$k6tsd8_cepl^s^tms2w1qrv"
+SERVE_MEDIA_LOCALLY = True
```
ajrbyers: Django should never be used to serve media in a production environment so this needs to be removed.

src/core/dev_settings.py
```diff
@@ -38,6 +34,10 @@ def show_toolbar(request):
     "SHOW_TOOLBAR_CALLBACK": show_toolbar,
 }
 
+if DEBUG:
```
ajrbyers: These settings are dev only so this change seems superfluous.

src/core/urls.py
```diff
@@ -26,13 +26,21 @@
 ]
 
 try:
-    if settings.DEBUG or settings.IN_TEST_RUNNER:
```
ajrbyers, line 29: This needs to be removed, we dont want to support django serving media in any context other than development.

ajrbyers (changes requested):
Before I get furhter into this review there are some strange changes that seem not to relate directly to the actual changes needed. dev_settings is never used in a prod environment so there is no need to set DEBUG to False. We also dont want to support serving of media content from Django, its very slow compared to NGINX or Apache which are designed for this task. In dev when debug is set media will be served so there is no need for a setting for serving local media.

Please take a look at these and then send this PR back so we can check the feature.

kopfduenger:
Thanks, and sorry about the unrelated changes. You were right — those changes did not belong in this PR.

I’ve removed the unrelated `dev_settings.py` / `urls.py` static/media-serving changes and force-pushed the branch. I also re-checked the branch against current `upstream/master`, and the diff is now limited to the keyword/autocomplete-related files only.

PR #5301, opened 2026-05-05
Fix missing alt_text template tag loads

kopfduenger:
This is a small standalone fix for templates that use `get_alt_text` without loading the corresponding template tag.

What it changes:
- loads `alt_text` in templates that already use `get_alt_text`

Notes:
- this issue is independent of the keyword/autocomplete work
- keeping it separate avoids mixing unrelated changes into PR #5300

### DIFF
--- a/src/core/dev_settings.py
+++ b/src/core/dev_settings.py
@@ -2,9 +2,10 @@
 # You should change this key before you go live!
 import os
 
-DEBUG = True
+DEBUG = False
 FORCE_BUILTIN_XSL = False
 SECRET_KEY = "uxprsdhk^gzd-r=_287byolxn)$k6tsd8_cepl^s^tms2w1qrv"
+SERVE_MEDIA_LOCALLY = True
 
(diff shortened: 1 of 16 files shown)


ajrbyers:
it looks like you've got the alt text commit at the tip of the branch made for the keywords changes. It'll need to be put on a branch of its own, though this tag is loaded in janeway_global_settings.TEMPLATES so its possible you are over-writing this setting which causes the error.

ajrbyers (changes requested):
See comment

kopfduenger:
Thanks, that makes sense.

I checked this again, and you’re right on both counts:
- the `alt_text` commit ended up on top of the earlier keyword branch instead of a clean branch of its own
- `get_alt_text` is already available via `janeway_global_settings.TEMPLATES` builtins, so this is very likely caused by my local settings/template override rather than a real upstream issue

I’m going to drop this PR rather than keep pushing a fix for what looks like a local setup problem.

PR #5307, opened 2026-05-11
fix olh mobile nav dropdown

### LINKED ISSUE #5123: OLH Mobile Nav unusable
Working off Master at f23d166d8

Image

OLH mobile navigation dropdown overlaps the main nav.

StephDriver:
closes #5123

### DIFF
--- a/src/themes/OLH/assets/scss/app.scss
+++ b/src/themes/OLH/assets/scss/app.scss
@@ -1359,7 +1359,13 @@ button.list-group-item-danger.active:focus {
   float:right;
 }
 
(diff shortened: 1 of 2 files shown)


StephDriver:
Note:

while working on this, I investigated why the issue was on Journal nav but not on press nav.  I found there was a typo in the press (from 5 years ago, but missed in subsequent updates) for the classnames `.classname` instead of `classname` so the class was never being called.  The two are now consistent.

mauromsl (commented):
good catch!

PR #5310, opened 2026-05-14
fix: save tinymce edits to proofing task

StephDriver:
fixes bug reported in ticket 12047. 

edits to the tinymce when managing a proofing task were not saving.  The date changes were saving.


StephDriver:
Explanation.

The jqte was interfering with the tinymce. But jqte no longer in use on this template, so removed it.

Both were trying to manage the text area, the one visible to the user was (correctly) tinymce, but the one sending the data to the form was the (invisible) jqte.  So the changes the user made were not being sent.

ajrbyers (approved):

mauromsl (approved):

PR #5312, opened 2026-05-15
sitemap a11y improvements

### LINKED ISSUE #5170: Improve Front of House Sitemaps
Testing for WCAG 2.4.5 Multiple Ways (#5150) identified several bugs in the sitemaps as well as readability concerns, such that they could not be relied upon for multiple ways compliance. 

The sitemaps need to improve so that every page on the front of house is both findable and reachable by humans via the sitemaps.  Each page must also link back to the sitemap on which it appears. 

Requirements:
- Every page must appear on a sitemap and only once on each sitemap.
- Links on the sitemap should be contextual - and the same text should not be used for different destinations.
- Each page that appears on a sitemap, must have a link to that sitemap
- The structure of the sitemaps should be human navigable so that someone can find the page they want. 

Problems identified during audit:

- only the press footer links to sitemaps.
- The H1 of each sitemap does not identify the level it is on, e.g. Press, Journal, Issue.  Instead the name of hte press or journal is used, this creates confusion when press and journal have the same name, e.g. OLH.
- articles are listed by issue, and each issue is  separate sitemap, this means someone who wants to reach an article via the sitemaps has to know its issue.  Consider a sitemap listing all articles with issues as headings.
- The journal level appears to pick up some press level pages, but not others, such that for example, the OLHJ sitemap includes two "start submission" links, one to the Press and one to the journal submit\start pages.
- there is no order to the static links - consider alphabetical ordering
- the news items appear alongside other static pages, rather than with any heading identifying them as news items - consider these could be a sub-sitemap of the journal sitemap, so that all news items are together. Consider whether these may be best shown in date order. 

### From BRF:

- [ ] News items should have their own sub site map for each press and journal
- [ ] Footer link should take reader to sitemap the page appears on, issues to the journal sitemap
- [ ] Ensure links are de-duplicated

StephDriver:
closes openlibhums/janeway#5170


StephDriver:
## Related Fixes

While working on this, I found the following bugs and elements that were confusing to users and fixed them along the way.
### cms/index.html
- for clarity: added 'in navigation' column to page table, that makes clear if the custom page is being used in the navigation. This is important because only pages used in the navigation are picked up by the sitemaps. And the navigation can refer to press pages or journal pages, but only the custom journal pages will show up in the journal cms/index.html, so there can be confusion if the custom nav includes a link to a page of hte same name, but on the press not the journal. 

note: this follows the explanation given on the cms/nav.html page where it describes how to enter navigation links
> In most cases, this should be the the relative URL path to the page. The relative path is formed from 1) the journal code (acronym or abbreviation) if your journal homepage URL ends with the journal code, 2) the word “site”, and 3) whatever you put into the Link field for the corresponding page. For example, to link to a custom page you created, if the journal homepage URL is “example.com/abc”, and you put “research-integrity” in the Link field for the page, then the Link field for the nav item should be “abc/site/research-integrity”. For top-level nav items that should not also appear as sub-nav items (under themselves), leave the Link field empty. For external links, it should be an absolute URL.

- fix: the page table includes the relative link, but this did not include the journal code prefix, so the links given were as if on the press, rather than the link needed in the nav to reach that page. 
- TODO fix: the action buttons in the Navigation pane were all over the place. I aligned them.
- TODO clarity: the action buttons in the Navigation pane included "enable" in green and "disable" in yellow - which I initially misunderstood as green items being 'enabled' and yellow items 'disabled' - when the button refers to the action you **can** do, rather than the current state.  I have changed this to make it clearer which items are active and which not (and updated the language). 



## Key Features.

###  Front of house footer links to Sitemaps
These use a template_tag to go to the most relevant sitemap.   Most of this is intuitive:

- article -> issue sitemap from which the article is linked.  Any article not in an issue links to a virtual 'no issue' issue.
- issue -> journal index from which the issue sitemap is linked.  This is the primary issue. Collections are not regarded as issues and articles are only listed for their primary issue.
- static page, e.g. Accessibility -> Pages sitemap from which it is linked **except Home Page, see below**
- news item -> News sitemap from which it is linked.
- repository subject -> while preprints could have multiple subjects, in keeping with the issue:article 1:1 relationship, they are listed on their first subject.

But some less so...

- home page -> the siteindex for the press/journal/repo NOT the pages sitemap. Yes it is a page, but if you're on the home page, then the sitemap for the whole of that object is the most relevant.

### Which Static Pages?
#### Always-present 
- Home,
- Accessibility, 
- Log in, 
- Register,
- Privacy Policy (the url is a setting, so must be added that way). 

These appear regardless of any nav flag.  for example the 'home' page can be disabled in the custom navigation, but unlike other disabled pages, would always exist!

#### Built-in nav pages (journal-only)

mirrors each nav_* flag on the Journal model: 
- nav_contact → Contact, 
- nav_articles → Articles,
-  nav_issues → Issues, 
- nav_news → News (gated on has_news, see below), 
- nav_sub → Submissions, nav_start → Start Submission (also checks disable_journal_submission), 
- nav_review → Become a Reviewer. 

Plus journal settings: 
- enable_editorial_display → Editorial Team (with multi-page support), 
- keyword_list_page → Keywords.


#### CMS custom nav pages

NavigationItems with /site/ links (flat nav and subnav treated identically), plus non-CMS custom nav items.
`/site/page` links to press site, 
`code/site/page` links to journal site
Journals may include press links in their custom nav, even if they have pages of same name in the CMS.

#### News Gate
News suppression post-processing still applies to both press and journal to handle any custom nav items pointing to /news/ when no items exist.

### Handling of CMS pages

CMS pages are only included if they are in the Navigation. This is because we want to use the Navigation + Index pages (e.g. Issues) with the Sitemap to demonstrate compliance with WCAG Multiple Ways .  So if a CMS page isn't in the navigation, then it isn't in use.  Note this is how some installs are using CMS pages in order to have draft or backup pages by not putting them in the navigation. 

To make this clearer in the manager, I have included an additional column next to each CMS page, that shows where it is in the navigation (or that it isn't there). 

CMS pages that have no content, will 404, so these are also filtered out by the sitemap.

The CMS page name is not used in the sitemap as the label for the link, instead the label in the navigation is used, this is for consistency as it is the label presented to users.

### Handling of Nav links

External links are ignored, as they are not for the site.

For journals and repositories, press links are ignored, as they will be on the press site map

Orphaned Nav pages are also ignored, this is where a page was added as a sub-navigation, then the parent navigation item was deleted, but the sub-nav itself was not deleted.  These do not show up in the navigation, but will show up in a database query, so are specifically filtered out.

### Ensuring Unique Contextual LInks

For WCAG compliance, it is important that link text is unique on the page. We handle this in two stages.  First each URL should only appear once on a page (deduplication) and second where different URLs have the same text label, we disambiguate them by adding `[different text]` as a suffix.  There are several different types of disambiguation in the code, to handle press and journals having the same name, issues of same name, articles of same name, news items of same name etc.  Ultimately, the end goal is that all links on the page have unique text.

### Validation
As per [Sitemaps.org Protocol](https://www.sitemaps.org/protocol.html). This required a change in the structure of the sitemaps from what had been agreed during Backlog refinement.  We cannot have a mix of links and child sitemaps, so the static links have been moved into their own child sitemap. 

### Testing
Added *a [redacted] tonne of tests* - to guard against duplicates or missing items in the sitemaps in future.


## Docs
- updated `sitemaps-and-robots.md` developer doc.
- TODO update the userfacing docs to explain which pages will show up
- TODO create issue to re-audit for WCAG multiple ways compliance

StephDriver:
# Sitemap accessibility refactor

This PR addresses the WCAG 2.4.5 (Multiple Ways) compliance gaps surfaced by the openlibhums/janeway#5150 audit and replaces the legacy sitemap with a properly-nested siteindex hierarchy plus stylesheet.

## Hierarchy

The press sitemap is now a true sitemapindex. Each level contains only `<sitemap>` child references — never inline `<url>` entries — because the sitemaps.org protocol defines two separate root elements, `<sitemapindex>` (children: `<sitemap>`) and `<urlset>` (children: `<url>`), and neither schema permits the other's children. See the [sitemaps.org protocol](https://www.sitemaps.org/protocol.html) and the [`siteindex.xsd`](https://www.sitemaps.org/schemas/sitemap/0.9/siteindex.xsd) / [`sitemap.xsd`](https://www.sitemaps.org/schemas/sitemap/0.9/sitemap.xsd) schemas. Pages and canonical links live on dedicated `pages_sitemap.xml` sub-sitemaps at each level.

```
Press Index (siteindex)
├── Press pages (sitemap)
├── Press news (sitemap)               ← only when active news items exist
├── Journal Index (1..n, siteindex)
│   ├── Journal pages (sitemap)
│   ├── Journal news (sitemap)         ← only when active news items exist
│   ├── Issue sitemap (1..n)           ← only non-empty regular issues
│   └── Not in any issue (sitemap)     ← only when orphan published articles exist
└── Repository Index (0..n, siteindex)
    ├── Repository pages (sitemap)
    ├── Subject sitemap (1..n)         ← only non-empty subjects
    └── Not in any subject (sitemap)   ← only when orphan published preprints exist
```

Empty sub-sitemaps are skipped: no file written, no parent entry emitted.

## CMS Page as source of truth

`_published_cms_pages(owner)` returns every `cms.Page` scoped to the owner (`content_type` + `object_id`) with `is_draft=False`. Pages whose `content` is empty after HTML/whitespace stripping are excluded, since the view does not effectively serve them.

The pages sub-sitemap (`build_pages_sitemap_context`) is the union of:

- **Canonical links** — Home and Accessibility always; the rest gated by their existing `nav_*` flags or journal/press settings.
- **Published CMS pages** — `display_name` as the label, `edited` as `lastmod`.

The two lists are merged with a URL-keyed dedup. On collision the CMS page wins both its `display_name` (as label) and its `edited` date (as lastmod), since the CMS page is what the URL actually serves — its values describe the real content. The canonical's label only applies when no CMS page exists at that URL. Final ordering is case-insensitive alphabetical by label.

### Canonical lists

Cell = the gate for that page on that site type; **Always** = unconditional; **—** = not a page on that site type.

| Page | Press | Journal | Repository |
|---|---|---|---|
| Home | Always | Always | Always |
| Accessibility | Always | Always | Always |
| Log in | Always | Always | Always |
| Register | Always | Always | Always |
| Contact | Always | `nav_contact` | — |
| Privacy Policy | Always¹ | Always¹ | — |
| News | `active_news_items.exists()` | `active_news_items.exists()` | — |
| Journals | `publishes_journals && !disable_journals` | — | — |
| Conferences | `publishes_conferences` | — | — |
| Articles | — | `nav_articles` | — |
| Issues | — | `nav_issues` | — |
| Submissions | — | `nav_sub` | — |
| Start Submission | — | `nav_start && !disable_journal_submission` | — |
| Become a Reviewer | — | `nav_review` | — |
| Editorial Team | — | `enable_editorial_display` setting | — |
| Keyword List | — | `keyword_list_page` setting | — |
| About | — | — | Always |
| List | — | — | Always |

¹ Privacy Policy uses the `privacy_policy_url` setting if set, otherwise `/site/privacy/`, so it is always present.

## `lastmod` sources

Each entry type derives its `<lastmod>` from a different field. The guiding principle is to use a **publication date** wherever one exists, never a "last modified" date: database modification timestamps update on any save, so a migration touching that table could produce a new `lastmod` without the page actually changing. The Accessibility link is the exception — it has no model row, so its `lastmod` is the mtime of `src/a11y/conformance_data.json`, the file that drives what the page renders.

| Entry type | `lastmod` source | Field type | Stable? |
|---|---|---|---|
| Article (issue / not-in-any-issue sitemaps) | `article.date_published` | publication date | :white_check_mark: |
| Preprint (subject / not-in-any-subject sitemaps) | `preprint.date_published` | publication date | :white_check_mark: |
| News item | `item.posted` | `default=timezone.now` (set once at creation) | :white_check_mark: |
| Accessibility link | mtime of `src/a11y/conformance_data.json` | file mtime | :white_check_mark: |
| Canonical / static links (Home, Contact, Log in, Register, …) | `None` — `<lastmod>` omitted | — | n/a |
| CMS page | `page.edited` | `auto_now` | :orange_circle: |

### CMS page `lastmod` is brittle

`cms.Page` carries no creation or publication date — `edited` (`auto_now`) is the only timestamp it has, so it is used as the CMS page `lastmod`. For a CMS page this is *usually* the right value (it genuinely tracks content edits), but it shares the `auto_now` weakness: any `Page.save()`, and any **migration that touches the `cms_page` table**, rewrites `edited` to the migration's run time across every page at once. If a future migration causes a mass `lastmod` jump on CMS pages, this is why; the accepted trade-off is to keep `edited` rather than add a dedicated stable date field to `Page`.

## Accessibility

- Each rendered sitemap emits separate `<janeway:page_title>` and `<janeway:h1>` elements; the XSLT renders them into `<title>` and `<h1>` respectively. They are kept separate in preparation for anticipated a11y work to standardise page-title formulae across the codebase, so the document `<title>` and the visible `<h1>` can diverge later without reworking the sitemap templates.
- `<janeway:sitemap_level>` is emitted per context; the XSLT "About Janeway sitemaps" hierarchy explanation marks the current level with `aria-current="true"` and a visible `(current)` annotation, for every level (`press`, `press-pages`, `press-news`, `journal`, `journal-pages`, `journal-news`, `issue`, `not-in-any-issue`, `repository`, `repository-pages`, `subject`, `not-in-any-subject`):

The "About Janeway sitemaps" hierarchy panel rendered on a journal index page: a nested bullet list (Press Index → Press pages, Press news, Journal Index (current) → Journal pages, Journal news, Issues, Not in any issue; Repository Index → Repository pages, Subjects, Not in any subject), with the current level shown in bold and annotated "(current)".
- `<janeway:group>` on each child sitemap entry routes the press index through the XSLT's grouped-rendering path (h3 headings for Pages / News / Journals / Repositories / Issues / Subjects) rather than the flat-list fallback.

### Name-clash disambiguation (WCAG 2.4.4)

Where two different entities share a name and co-occur as links on one page, disambiguation suffixes (`[press]`, `[journal]`, `[repository]`) distinguish them; sub-sitemaps that reference only one entity use the plain `.name`. The per-page breakdown of which clash set is checked and which labels are suffixed is in [the Name-clash disambiguation table](#name-clash-disambiguation-wcag-244-1) under the Expected output matrix.

The Pages and News child labels at the index level use the generic site type, not the owner name: `"Press Pages"` / `"Press News"`, `"Journal Pages"` / `"Journal News"`, `"Repository Pages"`. These are already unique within their index (no journal/repo child shares them), so there is no need to repeat the owner name in the label.

### Content-label disambiguation

`_disambiguate_labels_by_date` runs on news items, article entries, preprint entries, and pages-sitemap entries. For clashing labels it appends progressively finer suffixes — `[year]` → `[Mon YYYY]` → `[DD Mon YYYY]` — until unique, falling back to `[#1]`, `[#2]` numbered ascending by date so existing items keep their numbers as new ones are added. When entries have no `date` (e.g. a canonical clashing with a CMS page), it falls straight to the numbered suffix.

### Unique, contextual link text

The target is **no two links on one page with the same accessible name pointing to different destinations**, so that when a screen-reader user pulls up the page's list of links every entry is unique and self-describing.

`_disambiguate_labels_by_date` only de-duplicates *within a single list*. It does not, on its own, cover a label that collides with a link in a *different* section of the page — for example, on a journal news sub-sitemap the parent back-link is the journal name (e.g. "Journal of Pink Sky") and a news item could share that exact title. To cover the cross-section case, each list is given ARIA context: its links are associated (via `aria-labelledby`) with their section heading, so the accessible name a screen reader announces is the link text *plus* its context (e.g. "Journal of Pink Sky — News item" vs "Journal of Pink Sky — higher-level sitemap"). The visible text is preserved inside the accessible name, satisfying WCAG 2.5.3 (Label in Name).

## Footers

Every front-of-house page now has a visible "Sitemap" link in the theme footer (clean, OLH, material). `page_sitemap_url` resolves the most specific sitemap for the current page:

| Current page | Footer "Sitemap" link target |
|---|---|
| Article | Its canonical issue sub-sitemap (`primary_issue` if set, else the first issue); the no-issue sub-sitemap if it has no regular issue |
| Preprint | Its first-alphabetical subject sub-sitemap; the no-subject sub-sitemap if it has none |
| News item | The news sub-sitemap for its owner |
| Home page | The relevant index (press / journal / repository) |
| Any other page | The pages sub-sitemap for its owner |

The article and preprint rows pick the **same single canonical sub-sitemap** the page is listed in (see "Article routing" / "Preprint routing"), so the footer link always points to a sitemap that actually contains the page.

## URL routes

New (and renamed) named routes:

- `press_news_sitemap`, `journal_news_sitemap` — `/news_sitemap.xml`
- `press_pages_sitemap`, `journal_pages_sitemap`, `repository_pages_sitemap` — `/pages_sitemap.xml`
- `journal_no_issue_sitemap` — `/issue/no_issue_sitemap.xml`
- `repository_no_subject_sitemap` — `/subject/no_subject_sitemap.xml`

`press_views.sitemap` keeps its existing dispatching to journal/repository views; new shims `pages_sitemap` and `news_sitemap` follow the same dispatch pattern.

## Management command

`generate_sitemaps` orchestrates the writers in dependency order. For each level it pre-checks the gates so empty sub-sitemaps are not written and parent indexes don't reference them:

```
press → write_press_sitemap, write_pages_sitemap, write_news_sitemap (if news exists)
journal → write_journal_sitemap, write_pages_sitemap, write_news_sitemap (if news exists),
          write_issue_sitemap per non-empty regular issue,
          write_not_in_any_issue_sitemap (if orphan articles exist)
repository → write_repository_sitemap, write_pages_sitemap,
             write_subject_sitemap per non-empty subject,
             write_not_in_any_subject_sitemap (if orphan preprints exist)
```

`--site_type` and `--codes` filters work as before.

### Journal visibility

The command writes a journal sitemap for every **`is_remote=False`** journal in the filter set. The two visibility flags are handled differently:

- **`is_remote=True`** — the journal is *not hosted in Janeway*, so it has no local pages to describe. No files are written and it is not referenced anywhere.
- **`hide_from_press=True`** — the journal still has a live, hosted site, so its sitemap **is** generated. It is simply not linked from the press index. Its siteindex is therefore a top-level sitemap: `build_journal_index_context` emits **no `<janeway:higher_sitemap>`** and instead a `<janeway:note>` ("This journal is hidden from the press, …"), which the XSLT renders as a short *Note* section in place of the "Higher-level sitemap" link. With no press reference on the page, no `[journal]`/`[press]` name-clash suffix is applied.

`is_archived` is **not** a visibility flag (it is journal metadata meaning "no longer publishing"); archived journals are still browsable and are included normally. Conferences (`is_conference=True`) are hosted Janeway sites and are listed in the press index alongside journals.

## What's deliberately not here

`_canonical_preprints_for_subject` is kept (single-canonical-subject restriction for multi-tagged preprints) — see "Preprint routing" above for why.

`extend_to_journals=True` on a press CMS page still renders that page on journals at request time, but for sitemap purposes each pages sub-sitemap lists only its own-site CMS pages.

## Tests and docs

Dedicated sitemap test classes covering context builders, pages-link ordering and de-duplication, news ordering, the `page_sitemap_url` template tag, plain-label/XML-escaping, and schema validation against `sitemap.xsd` / `siteindex.xsd`. 

## Expected output matrix

A condition-by-condition breakdown of what the sitemap generator produces, intended as a checklist for manual testing and for designing automated tests.

### Which files are written

Generated under `src/files/sitemaps/`. `<code>` is `journal.code` or `repository.short_name`.

| Path | Level | Written when |
|---|---|---|
| `sitemap.xml` | press index | A `Press` row exists (i.e. always, on any real install). |
| `pages_sitemap.xml` | press | Always — Home and Accessibility are unconditional canonicals. |
| `news_sitemap.xml` | press | `press.active_news_items.exists()`. |
| `<code>/sitemap.xml` | journal index | Journal is in the run's filter set **and** `is_remote=False` — pages canonicals (Home + Accessibility) guarantee at least one page. Hidden (`hide_from_press=True`) journals are still written, but with no higher-level link (see "Journal visibility"). |
| `<code>/pages_sitemap.xml` | journal | Same gate as journal index. |
| `<code>/news_sitemap.xml` | journal | `journal.active_news_items.exists()`. |
| `<code>/<issue_id>_sitemap.xml` | issue | `issue.issue_type.code == 'issue'` **and** `_canonical_articles_for_issue(issue).exists()` (articles whose canonical issue is this one). |
| `<code>/no_issue_sitemap.xml` | journal | `_articles_not_in_any_regular_issue(journal).exists()`. |
| `<code>/sitemap.xml` | repo index | `live=True` repo, **and** at least one subject with published preprints **or** at least one orphan preprint. Repos with no published preprints get no files at all. |
| `<code>/pages_sitemap.xml` | repo | Same gate as repo index. |
| `<code>/<subject_id>_sitemap.xml` | subject | `_canonical_preprints_for_subject(subject).exists()` (preprints whose canonical subject is this one). |
| `<code>/no_subject_sitemap.xml` | repo | `_preprints_without_subject(repo).exists()`. |

### Press index — child sitemap entries

Listed in the press `sitemap.xml`. Order: pages → news → journals (sorted) → repositories (sorted).

| Child | Appears when |
|---|---|
| Pages | Always (label: `"Press Pages"`). |
| News | `press.active_news_items.exists()` (label: `"Press News"`). |
| Each journal | `hide_from_press=False && !is_remote`. |
| Each repository | `live=True`. |

### Journal index — child sitemap entries

Order: pages → news → issues (date desc) → not-in-any-issue.

| Child | Appears when |
|---|---|
| Pages | Always (label: `"Journal Pages"`). |
| News | `journal.active_news_items.exists()` (label: `"Journal News"`). |
| Each regular issue | `issue_type.code == 'issue'` **and** `_canonical_articles_for_issue(issue).exists()`. |
| Not in any issue | `_articles_not_in_any_regular_issue(journal).exists()`. |

### Repository index — child sitemap entries

Order: pages → subjects (alphabetical) → not-in-any-subject.

| Child | Appears when |
|---|---|
| Pages | Always (label: `"Repository Pages"`). |
| Each subject | `subject.published_preprints().exists()`. |
| Not in any subject | `_preprints_without_subject(repo).exists()`. |

### Pages sitemap — canonical link presence

Canonical links are always added before CMS pages; the merged list is deduped by URL, disambiguated by date when labels clash, and sorted case-insensitive by label.

#### Press

| Canonical | Gate |
|---|---|
| Home | Always. |
| Accessibility | Always (lastmod = mtime of `src/a11y/conformance_data.json`). |
| Journals | `publishes_journals && !disable_journals`. |
| Conferences | `publishes_conferences`. |
| News | `active_news_items.exists()`. |
| Contact | Always. |
| Log in, Register | Always. |
| Privacy Policy | `privacy_policy_url` setting if set, otherwise `/site/privacy/`. |

#### Journal

| Canonical | Gate |
|---|---|
| Home | Always. |
| Accessibility | Always. |
| Contact | `nav_contact`. |
| Articles | `nav_articles`. |
| Issues | `nav_issues`. |
| News | `active_news_items.exists()`. `nav_news` only controls the default nav link; the `/news/` URL is always served, so it's listed whenever active news exists. |
| Submissions | `nav_sub`. |
| Start Submission | `nav_start && !disable_journal_submission` setting. |
| Become a Reviewer | `nav_review`. |
| Editorial Team | `enable_editorial_display` setting (label from `editorial_group_page_name`). |
| Keyword List | `keyword_list_page` setting. |
| Log in, Register | Always. |
| Privacy Policy | `privacy_policy_url` setting if set, otherwise `/site/privacy/`. |

#### Repository

| Canonical | Gate |
|---|---|
| Home, Accessibility, About, List, Log in, Register | All always (no setting/`nav_*` gating). |

### CMS page inclusion (Press & Journal pages sitemaps)

| CMS Page state | Included? |
|---|---|
| `is_draft=True` | No. |
| `is_draft=False`, `content` null / empty / whitespace-only / HTML wrapper with no text (`<p></p>`, `<p>&nbsp;</p>`) | No. |
| `is_draft=False`, content has visible text | Yes. URL: `{owner.site_url}/site/{name}/`; label: `display_name`; lastmod: `edited`. |
| Press CMS page with `extend_to_journals=True` | Press pages sitemap only — *not* the journal's. The page still renders on journals at request time. |
| Press CMS page on a journal pages sitemap | Never — content_type scoping prevents leakage. |

### Article routing (which sub-sitemap an article appears in)

`published` = `stage=STAGE_PUBLISHED && date_published<=now()`. "Regular" issue means `issue_type.code='issue'`; collections (and any other custom issue types) are not regular.

Each published article appears in **exactly one** issue sub-sitemap — its *canonical* regular issue — matching the preprint/subject model and sitemap.org's "each URL in one place" guidance. The canonical regular issue is `primary_issue` if that is a regular issue, otherwise the first regular issue (`issues_list` order); `primary_issue` is an optional field, so the first-issue fallback covers the common case. The footer "Sitemap" link points to this same canonical issue.

| Article state | Sub-sitemap |
|---|---|
| Published, in one regular issue | That issue's `<issue_id>_sitemap.xml`. |
| Published, in two or more regular issues | Only the canonical (`primary_issue`, else first) regular issue's sitemap. |
| Published, in a regular issue AND a collection | The canonical regular issue's sitemap. **Not** in `no_issue_sitemap.xml`. |
| Published, only in collection(s) | `no_issue_sitemap.xml`. |
| Published, in no issues at all | `no_issue_sitemap.xml`. |
| Unpublished / future-dated | Nowhere. |

Journal-index inclusion uses the same canonicalised count: an issue whose articles are all canonicalised elsewhere is omitted from the index (and not written), so no empty issue sub-sitemaps appear.

### Preprint routing (subject sub-sitemaps)

Each preprint appears in exactly one subject sub-sitemap — whichever of its subjects sorts first alphabetically by name (`_canonical_preprints_for_subject`). This matches sitemap.org's "each URL in one place" guidance and the alphabetical-first choice the `page_sitemap_url` template tag already makes for the footer link.

| Preprint state | Sub-sitemap |
|---|---|
| Published, one subject | That subject's `<subject_id>_sitemap.xml`. |
| Published, multiple subjects | Only the alphabetically-first subject's sitemap. |
| Published, no subjects | `no_subject_sitemap.xml`. |
| Unpublished | Nowhere. |

Repository-index inclusion uses the same canonicalised count: a subject whose preprints are all canonicalised elsewhere is omitted from the index (and not written), so no empty sub-sitemaps appear.

### News item routing & ordering

| Item state | Sub-sitemap |
|---|---|
| Active (`start_display<=now` **and** `end_display>=now` or null), `content_type=press` | Press `news_sitemap.xml`. |
| Active (**same** `start_display`/`end_display` gate), `content_type=journal` | That journal's `news_sitemap.xml`. |
| `start_display` in future | Nowhere (until the date arrives). |
| `end_display` in past | Nowhere. |

Press and journal news use the **same** `ActiveNewsItemManager` (`Press.active_news_items` / `Journal.active_news_items`), so the `start_display`/`end_display` window is applied identically; only the owning `content_type` differs. Items are ordered newest first (`-posted`).

### Name-clash disambiguation (WCAG 2.4.4)

Suffix `[press]` / `[journal]` / `[repository]` is added **only on pages where two different entities sharing the same name actually co-occur**:

| Page | Clash set | Suffix applied to |
|---|---|---|
| Press index | press.name + all journal names + live repo names | The press's own H1/title + any clashing journal/repo child label. |
| Journal index | journal.name vs press.name | Journal H1/title + parent (press) back-link, when they collide. |
| Repository index | repo.name vs press.name | Repo H1/title + parent (press) back-link, when they collide. |
| Pages / news / issue / subject sub-sitemaps | n/a — only one entity appears | No suffix. |

### Content-label disambiguation (date-based)

`_disambiguate_labels_by_date` runs on news, article, preprint, and pages-sitemap entries. For labels colliding case-insensitively:

| Strategy (tried in order) | Trigger |
|---|---|
| `[YYYY]` | Year alone makes the group unique. |
| `[Mon YYYY]` | Year wasn't enough. |
| `[DD Mon YYYY]` | Month-and-year wasn't enough. |
| `[#1]`, `[#2]`, … (ascending by date) | Even date-time isn't enough, **or** at least one clashing entry has `date=None` (e.g. a canonical link clashing with a CMS page that share a label/URL). |

## Schema validation

I have run this on a test copy of the olh database, for a very large test case. — `urlset` files (`pages`, `news`, `issue`, `subject`) against `sitemap.xsd`, and `sitemapindex` files (press/journal/repository indexes) against `siteindex.xsd`:

```
# from src/files/sitemaps/
xmllint --noout --schema ../../utils/schema/sitemap.xsd.xml pages_sitemap.xml news_sitemap.xml */pages_sitemap.xml */news_sitemap.xml */*_sitemap.xml
xmllint --noout --schema ../../utils/schema/siteindex.xsd.xml sitemap.xml */sitemap.xml
```

**Result:** ✅  all xmllint checks pass.

## General Validation
Manual test, with a random sample have all footer links behaving as above. Further I git-tracked the sitemaps generated across the work, so they have demonstrated a consistent generation (this was what caught the lastmod bug).

StephDriver:
TODO:

- [x] create sitemap doc from above. (separate from robots into dedicated sitemap file).
- [x] go through the tests line by line, checking against the logic of above (above has been gone through and matches the intention).  Consider any test gaps
- [x] Consider whether user-docs are needed for sitemaps DECISION - No. not a stand alone anyway.
- note that once merged, review openlibhums/janeway#5150 as this update was identified as all that was needed to bring us into compliance with WCAG 2.4.5. And then reaudit and update conformance.

StephDriver:
Work Done:

- updated the CMS pages to have drafts (with previews) and tidied up the CMS interface a bit.   Then we can use CMS published pages as source of truth for additional pages for sitemaps.
- changed utils/tests.py into a package and separated out the sitemap tests.
- rewrote `generate_sitemaps` from ground up, to use a different structure that would validate with sitemaps schema, and satisfy WCAG requirements for link deduplication and disambiguation.
- updated the progress bars while running the generator to show number of sitemaps per object.
- created a new template_tag for linking to sitemap from each front of house footer.
- wrote a new set of tests for this new `generate_sitemaps`.
- split the original `sitemaps-and-robots.md` and then updated the sitemaps part, with tables to explain which pages are included in sitemaps and when (lots and lots of logic to cover).  I have reconciled these back to the tests.
- manual tests of new sitemaps alongside automated tests
  - loading them up and moving around the structure.
  - git tracked the sitemaps output, so that I could confirm they were consistently generated.
  - checked back and forth from the new sitemap footer links.


Notes:
1. a11y-mode work will also need to be updated to include this footer link too. see openlibhums/clarity#7
2. Original docs recommended setting chron to run `generate_sitemap` every 30 min.  Is this still wise? This is a lot of processing.  Do we need to consider some kind of incremental update command. 
3. There's a lot of logic. I have been over it line by line several times until my eyes blurred - good luck with it. 
4. This moved beyond what was described in the original issues because we needed a source of truth for which CMS pages to include, and having explored using the Nav, it was clear the better solution was to introduce draft/publish flags for CMS.  Further, the structure had to be amended to pass the validation, as we could only have either URLs or site-indexes listed on any one sitemap, we couldn't mix them.

StephDriver:
rebased onto r-v1.9.x

ajrbyers:
@StephDriver passed this back as tests are failing here.

StephDriver:
@ajrbyers which tests?  All of mine are passing.

```bash
(venv)  stephdriver@Keeper-of-Keys r-1_9 % git status
On branch b-5170_a11y_sitemaps
Your branch is up to date with 'origin/b-5170_a11y_sitemaps'.

nothing to commit, working tree clean
(venv)  stephdriver@Keeper-of-Keys r-1_9 % python src/manage.py test utils
Found 332 test(s).
Creating test database for alias 'default'...
System check identified no issues (18 silenced).
..........................................................................................................................................Processing  12345 ... status accepted
.Processing  12345 ... status delivered
.Processing  12345 ... status failed
................................................................................................................................................................................................
----------------------------------------------------------------------
Ran 332 tests in 32.587s

OK
Destroying test database for alias 'default'...
```

and those processing lines are on the branch before my changes.

```bash
stephdriver@Keeper-of-Keys r-1_9 % git status
On branch r-v1.9.x
Your branch is up to date with 'origin/r-v1.9.x'.

nothing to commit, working tree clean
stephdriver@Keeper-of-Keys r-1_9 % jvenv
(venv)  stephdriver@Keeper-of-Keys r-1_9 % python src/manage.py test utils
Found 197 test(s).
Creating test database for alias 'default'...
System check identified no issues (18 silenced).
Processing  12345 ... status accepted
.Processing  12345 ... status delivered
.Processing  12345 ... status failed
...................................................................................................................................................................................................
----------------------------------------------------------------------
Ran 197 tests in 23.142s

OK
Destroying test database for alias 'default'...

```

ajrbyers:
> @ajrbyers which tests? All of mine are passing.
> 
> ```shell
> (venv)  stephdriver@Keeper-of-Keys r-1_9 % git status
> On branch b-5170_a11y_sitemaps
> Your branch is up to date with 'origin/b-5170_a11y_sitemaps'.
> 
> nothing to commit, working tree clean
> (venv)  stephdriver@Keeper-of-Keys r-1_9 % python src/manage.py test utils
> Found 332 test(s).
> Creating test database for alias 'default'...
> System check identified no issues (18 silenced).
> ..........................................................................................................................................Processing  12345 ... status accepted
> .Processing  12345 ... status delivered
> .Processing  12345 ... status failed
> ................................................................................................................................................................................................
> ----------------------------------------------------------------------
> Ran 332 tests in 32.587s
> 
> OK
> Destroying test database for alias 'default'...
> ```
> 
> and those processing lines are on the branch before my changes.
> 
> ```shell
> stephdriver@Keeper-of-Keys r-1_9 % git status
> On branch r-v1.9.x
> Your branch is up to date with 'origin/r-v1.9.x'.
> 
> nothing to commit, working tree clean
> stephdriver@Keeper-of-Keys r-1_9 % jvenv
> (venv)  stephdriver@Keeper-of-Keys r-1_9 % python src/manage.py test utils
> Found 197 test(s).
> Creating test database for alias 'default'...
> System check identified no issues (18 silenced).
> Processing  12345 ... status accepted
> .Processing  12345 ... status delivered
> .Processing  12345 ... status failed
> ...................................................................................................................................................................................................
> ----------------------------------------------------------------------
> Ran 197 tests in 23.142s
> 
> OK
> Destroying test database for alias 'default'...
> ```

Well never make an assumption. Its actually ruff formatting see: https://github.com/openlibhums/janeway/actions/runs/27752395486/job/82105825907?pr=5312

StephDriver:
> > @ajrbyers which tests? All of mine are passing.
> > ```shell
...
> Well never make an assumption. Its actually ruff formatting see: https://github.com/openlibhums/janeway/actions/runs/27752395486/job/82105825907?pr=5312

ooh, that's interesting, because I was running a formatter with a pre-commit hook - which won't let me commit until it passes, so that means my formatter is out of sync with the one testing here. I'll have to dig into it more on Monday as to what and why.

StephDriver:
I haven't worked out why my existing hook wasn't firing for that file, when it fired for others - the most likely explanation is that it the difference crept in during a rebase.  But the two missing empty lines have now been duly added.

StephDriver:
Unpicking the test failures.

- sorted out the migrations
- updated the ruff formatting and added new guard as it was testing disc versions not staged versions pre commit. #5360
- went through all my new tests, updated them - found an error in my code relating to how draft access codes where handled (it allowed reuse) - fixed that, and updated tests around this new behaviour.
- tested locally with python 3.11 (as that is the one that keeps failing here) - I usually work in python 3.12, and I couldn't reproduce the errors here with that.  Can't actually reproduce the problem with either.

Failures that are on the base branch too
- ERROR: test_pdf_to_text (src.core.tests.test_files.TestFilesHandler.test_pdf_to_text)
- FAIL: test_search_includes_article (src.journal.tests.test_journal_frontend.TestJournalSite.test_search_includes_article)
- FAIL: test_section_editor_cannot_see_pii_when_enabled 

I **cannot** reproduce the test errors from github locally. They seem to come down to `submission_article.primary_issue_id` which I refer to here, but I didn't create. So my best guess is some kind of missing migration - but that's only affecting GH and not local?

src/cms/models.py
```diff
@@ -70,6 +73,15 @@ class Page(models.Model):
...
         return "{0} - {1}".format(self.content_type, self.display_name)
 
+    def save(self, *args, **kwargs):
+        if self.is_draft and not self.preview_token:
+            self.preview_token = uuid4()
+        elif self.pk:
+            existing = Page.objects.get(pk=self.pk)
+            if self.is_draft and not existing.is_draft:
+                self.preview_token = uuid4()
+        super().save(*args, **kwargs)
```
mauromsl: Instead of doing these checks (which won't run when an object is created without calling `.save()`explicitely), is there a reason to not use an [uuid field](https://docs.djangoproject.com/en/4.2/ref/models/fields/#uuidfield)with a default like the one provided in the docs as an example:
```python
preview_token = models.UUIDField(default=uuid.uuid4, editable=False)
```

If the reason is that we want to cycle them, I question why we need to do that instead of having a token for the lifetime of the page

src/cms/views.py
```diff
@@ -63,6 +64,20 @@ def index(request):
...
+        page_id = request.POST.get("toggle_draft")
+        page = get_object_or_404(
+            models.Page,
+            pk=page_id,
+            content_type=request.model_content_type,
+            object_id=request.site_type.pk,
+        )
+        page.is_draft = not page.is_draft
+        if not page.is_draft:
+            page.preview_token = str(uuid4())
```
mauromsl: if we provide a token to all pages, we do not need to do this check whenever a page is toggled. Also, toggling the page as draft multiple times would reset the token, which I'm not sure we want. Finally, the same token generation logic lives inside the `.save()` call below.

src/cms/views.py
```diff
@@ -63,6 +64,20 @@ def index(request):
...
 
+    if request.POST and "toggle_draft" in request.POST:
+        page_id = request.POST.get("toggle_draft")
+        page = get_object_or_404(
+            models.Page,
+            pk=page_id,
+            content_type=request.model_content_type,
+            object_id=request.site_type.pk,
+        )
+        page.is_draft = not page.is_draft
```
mauromsl: This nifty toggle makes sense in the context of a single thread of execution, but it is unreliable in terms of concurrency. e.g if two users use the toggle button around the same time (or an accidental double click), the results can be unpredictible. Having explicit on/off buttons is more reliably

src/cms/views.py
```diff
@@ -114,6 +129,16 @@ def view_page(request, page_name):
         object_id=request.site_type.pk,
     )
 
+    if page.is_draft:
+        access_code = request.GET.get("access_code")
+        if not access_code or access_code != str(page.preview_token):
```
mauromsl: I'm not sure why we are casting preview_token to `str`. It will always be either a str or `None`, in which case the latter would be converted into `"None"

src/core/templatetags/sitemap_tags.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def _subject_sitemap_url(repo, subject):
+    path = reverse("repository_sitemap", kwargs={"subject_id": subject.pk})
+    return repo.site_url(path=path)
+
+
+def _no_subject_sitemap_url(repo):
+    path = reverse("repository_no_subject_sitemap")
+    return repo.site_url(path=path)
```
mauromsl: Do these need to be pseudo-private?

I'm trying to anticipate issues like the below use of 

`from utils.logic import _canonical_issue`

I've noticed that LLM tools have  a tendency to make in-module functions and constant private by default, so might be worth considering them first

docs/md/sitemaps.md
```diff
@@ -0,0 +1,263 @@
...
+
+Each published article appears in **exactly one** issue sub-sitemap — its *canonical* regular issue — matching the preprint/subject model and sitemap.org's "each URL in one place" guidance. The canonical regular issue is `primary_issue` if that is a regular issue, otherwise the first regular issue (`issues_list` order); `primary_issue` is an optional field, so the first-issue fallback covers the common case. The footer "Sitemap" link points to this same canonical issue.
+
+| Article state | Sub-sitemap |
+|---|---|
+| Published, in one regular issue | That issue's `<issue_id>_sitemap.xml`. |
+| Published, in two or more regular issues | Only the canonical (`primary_issue`, else first) regular issue's sitemap. |
+| Published, in a regular issue AND a collection | The canonical regular issue's sitemap. **Not** in `no_issue_sitemap.xml`. |
+| Published, only in collection(s) | `no_issue_sitemap.xml`. |
+| Published, in no issues at all | `no_issue_sitemap.xml`. |
```
mauromsl, line 235: I'm not sure I understand what `no_issue_sitemap.xml` means or how it is intended to be used, does it mean "not primary" issue? If an article is published in a collection, we put it there bundled with articles without an issue? surely there should be a sitemap for a collection the well

src/cms/migrations/0027_alter_page_is_draft_default_true.py
```diff
@@ -0,0 +1,22 @@
...
+            model_name="historicalpage",
+            name="is_draft",
+            field=models.BooleanField(default=True),
+        ),
+        migrations.AlterField(
+            model_name="page",
+            name="is_draft",
+            field=models.BooleanField(default=True),
+        ),
+    ]
```
mauromsl: can we squash these migrations before we merge to 1.9? I think we started making the field `default=False` and then flipped to `default=True` at some point during development. This transition is unnecessary for the release.

Additionally, please consider having `default=False` in the model, as the opposite introduces a minor breaking change (any other code generating pages will need to be updated to cope with default behaviour being creating a draft)

src/templates/common/issue_sitemap.xml
```diff
@@ -5,23 +5,20 @@
...
   </janeway:higher_sitemap>
-  {% for article in issue.get_sorted_articles %}
+  {% for article in article_entries %}
     <url>
       <loc>{{ article.url }}</loc>
-      <lastmod>{{ article.fast_last_modified_date.isoformat }}</lastmod>
-      <changefreq>monthly</changefreq>
-      <janeway:loc_label>{{ article.title|striptags }}</janeway:loc_label>
+      <lastmod>{{ article.lastmod }}</lastmod>
+      <changefreq>never</changefreq>
```
mauromsl: issues and articles can change, especially when publishing a rolling release. I think the previous value of `monthly` exists to ensure indexers will come back and continue to have the content updated, which we want to preserve.

Consider these for the other pages such as news, if they are editable in Janeway, we shouldn't use `<changefreq>never</changefreq>`

src/utils/logic.py
```diff
@@ -205,67 +210,793 @@ def get_log_entries(object):
...
+        articles_qs = _articles_not_in_any_regular_issue(journal).order_by("title")
+        page_title = f"Sitemap - Not in any issue, {journal.name}"
+        sitemap_level = "not-in-any-issue"
+
+    article_entries = []
+    for article in articles_qs:
+        # Use the publication date, not last_modified: an article's
+        # last_modified (and that of its related galleys/files/issues) is
+        # bumped by unrelated events — migrations, cache-on-render saves —
+        # which would churn the sitemap lastmod without the page changing.
```
mauromsl: The last modified date in the old sitemap system considered all the modifications up and downstream, such as changes to sections, authors, keywords, issues etc.

I think sending a false positive here is better here than having a false negative and having outdated indexed metadata potentially forever. This is particularly important for important changes such as retractions or COPE related changes

src/utils/logic.py
```diff
@@ -205,67 +210,793 @@ def get_log_entries(object):
...
+    """
+    from collections import defaultdict
+
+    groups = defaultdict(list)
+    for i, entry in enumerate(entries):
+        groups[entry["title"].lower()].append((i, entry))
+
+    for _lower_title, group in groups.items():
+        if len(group) == 1:
+            continue
```
mauromsl, line 306: If I'm reading this correctly, we are going to loop 3 * n over the entire content catalogue when generating sitemaps. Considering how long this can take, I'd recommend trying to do make these changes within a single loop.

src/utils/logic.py
```diff
@@ -205,67 +210,793 @@ def get_log_entries(object):
...
+    """Disambiguate clashing content labels using progressively finer date granularity.
+
+    entries: list of dicts, each containing at least 'title' (str) and 'date'
+    (date or datetime).  Modifies 'title' in-place for clashing groups.
+
+    Tries [year], [Mon YYYY], [DD Mon YYYY] in turn; if still not unique, falls
+    back to sequential [#1], [#2], ... assigned by ascending date (so oldest = #1
+    and existing items keep their numbers as new ones are added).
+    """
+    from collections import defaultdict
```
mauromsl: import statement shouldn't be inside the function

src/utils/logic.py
```diff
@@ -205,67 +210,793 @@ def get_log_entries(object):
...
+    """Journal siteindex context: pages → news → issues (date desc) → not-in-any-issue."""
+    press = journal.press
+    if journal.hide_from_press:
+        # A journal hidden from the press is not linked from the press index,
+        # so its siteindex is a top-level sitemap with no higher level. With no
+        # press reference on the page there is no name clash to disambiguate.
+        # _plain_label keeps it a plain str so the template escapes any "&".
+        journal_label = _plain_label(journal.name)
+    else:
+        clash_names = _local_clash_names(journal.name, press.name)
```
mauromsl, line 860: where is this used?

src/utils/logic.py
```diff
@@ -205,67 +210,793 @@ def get_log_entries(object):
...
+    return {name for name, count in counts.items() if count > 1}
+
+
+def _local_clash_names(*names):
+    """Names that appear more than once in the given list.
+
+    Used at sub-press levels (journal, repo, sub-sitemaps) where only a small
+    fixed set of entity names appears on a single page.
+    """
+    from collections import Counter
```
mauromsl: move import statement to the top

src/utils/logic.py
```diff
@@ -205,67 +210,793 @@ def get_log_entries(object):
...
+
+
+def _build_clash_names(press):
+    """Names shared by more than one entity in the press's full child set.
+
+    Used at the press index level to decide when to disambiguate the press's
+    own name and the names of its journals/repositories with [press] / [journal]
+    / [repository] suffixes (WCAG 2.4.4 same-link-text).
+    """
+    from collections import Counter
```
mauromsl: Move import statement to the top

src/utils/logic.py
```diff
@@ -205,67 +210,793 @@ def get_log_entries(object):
...
+
+def build_press_index_context(press):
+    """Press siteindex context: pages → news → journals → repositories."""
+    journals = press.journals_az.filter(
+        hide_from_press=False,
+        is_remote=False,
+    )
+    repos = repo_models.Repository.objects.filter(live=True)
+    clash_names = _build_clash_names(press)
+    press_label = _suffixed_name(press.name, clash_names, "[press]")
```
mauromsl, line 801: where is `press_label` used?

src/utils/logic.py
```diff
@@ -205,67 +210,793 @@ def get_log_entries(object):
...
+
+
+def build_press_index_context(press):
+    """Press siteindex context: pages → news → journals → repositories."""
+    journals = press.journals_az.filter(
+        hide_from_press=False,
+        is_remote=False,
+    )
+    repos = repo_models.Repository.objects.filter(live=True)
+    clash_names = _build_clash_names(press)
```
mauromsl, line 800: Where is `clash_names` used

src/utils/logic.py
```diff
@@ -205,67 +210,793 @@ def get_log_entries(object):
...
+    }
+
+
+def build_repo_index_context(repo):
+    """Repository siteindex context: pages → subjects → not-in-any-subject."""
+    press = repo.press
+    clash_names = _local_clash_names(repo.name, press.name)
+    repo_label = _suffixed_name(repo.name, clash_names, "[repository]")
+    press_label = _suffixed_name(press.name, clash_names, "[press]")
+    child_sitemaps = [
```
mauromsl, line 932: I don't think we are using these here either

src/core/include_urls.py
```diff
@@ -440,16 +440,53 @@
...
+    re_path(
+        r"^pages_sitemap.xml$",
+        press_views.pages_sitemap,
+        name="journal_pages_sitemap",
+    ),
+    re_path(
+        r"^pages_sitemap.xml$",
+        press_views.pages_sitemap,
+        name="repository_pages_sitemap",
+    ),
```
mauromsl: These URLS have identical regex matches, Django will only ever match one of them. Why do we need the other identical patterns but with different names?

src/core/templatetags/sitemap_tags.py
```diff
@@ -0,0 +1,136 @@
...
+        {% page_sitemap_url request.resolver_match.url_name article %}
+        {% page_sitemap_url request.resolver_match.url_name preprint %}
+    """
+    request = context.get("request")
+    if request is None:
+        return ""
+
+    # Fall back to the page's primary object in context so the article/preprint
+    # branches work on every theme, even footers that don't pass an object.
+    if obj is None:
```
mauromsl: Worth being extra defensive here, as the footer link could techincally pass in a falsey value (i.e. `''`). Worth checking `if obj:` instead

mauromsl (changes requested):
Mostly comments regarding cleanup of migrations, unused code and inline import statements.

The more important comment is around the change to the `lastmod` attribute on article sitemaps. Do you agree that giving a false and potentially outdated `lastmod` date could be problematic?

Finally a note around performance: I've noted we run a large number of loops during title / date processing of issue items and preprints, including N+1 queries. For publications that run larger items (big repositories or conference proceedings) Would it be worth breaking these sitemaps into smaller ones? My concerns:

 - Memory consumption: Since we have to hold all the model instances in memory in order to disambiguate dates/titles, memory requirements for some instances might be very high. Optmise by memoizing only the title and date strings we are worried about
 - Compute time: Each of the disambiguation / processing done per sitemap, comes with its own loop. Instead build the dictionary of memoized strings on the caller and pass it to the children to only run a single loop and do checks in constant time.
 - (Probably not an issue but worth mentioning) There is a 10MB file size limit  and 50K link limit per sitemap on [sitemaps.org recommendations](https://www.sitemaps.org/protocol.html).

I can help providing prod-like data to test these if you need them @StephDriver

src/cms/models.py
StephDriver: @mauromsl 

My take on this was that each time the page becomes a draft it needs a different token because the token grants anonymous access to that draft.  A page could exist for years. Staff could change. Allowing old links (i.e old staff) to access that page during new drafts felt like it could be a problem and not what  a user would expect.  I realise 99.9999% of the time it likely wouldn't matter.  But it seemed good practice to me to keep it unique per drafting process for a page, rather than unique per page.

src/cms/views.py
StephDriver: very good catch.  fixed: 252ff7f

src/cms/views.py
StephDriver: probably a hangover from my java need to type everything.  fix 4fa09df

src/core/templatetags/sitemap_tags.py
StephDriver: this is my own preference really - I like to have clear public/private scoping so that I don't call something I wasn't expecting or rather, I get an error if I try - so I just got into the habit even here where it doesn't have same meaning.  But that's probably more java thinking from me too.  It helps me keep clear in my head what I intended.  But happy to revert them.  51de04b

docs/md/sitemaps.md
StephDriver: No issue is the bucket any articles that haven't already been put on a sitemap fall into so that we can logically demonstrate that all of them will be accounted for no matter what in regards to the WCAG multiple ways SC.  

This was my way of solving the problem that an article needed to appear only once, but could be in multiple issues/collections - and having set them out by issue, it wouldn't make sense to have a collection with only "some" of that collection's articles, the ones that weren't under issues as well.   I couldn't find a neat solution as to a 1 to 1 grouping that would cover all articles.   So in the end my thinking was that having a bucket for 'all else' will meet the requirements, and this could be improved on later if we find a better way to group articles in the sitemaps.

src/templates/common/issue_sitemap.xml
StephDriver: 884b9e2

src/cms/migrations/0027_alter_page_is_draft_default_true.py
StephDriver: ok, I this has taken some thought but I think I get it - you want the behaviour of defaulting to draft to be in the code rather than in the database?

7725a0b

src/utils/logic.py
StephDriver: I take the point.  I changed it as there were likely to be fewer false negatives than false positives, but I hadn't considered that the impact of false-positives is negligible. 84951f3a0

src/utils/logic.py
StephDriver: good catch.  91e014cbf

src/utils/logic.py
StephDriver: 04a6a3f

src/utils/logic.py
StephDriver: 04a6a3f

src/utils/logic.py
StephDriver: 04a6a3f

src/utils/logic.py
StephDriver: yes, they are used in the if journal.hide_from_press logic  lower down in the function and it's not very readable.  The concepts are:

- journals which are hide_from_press should not have an upper link to the press in the sitemap
- As we output at the bottom a structure showing the journals are sub-sitemaps of press, there needs to be a note to explain this is an exception.
- if the press name isn't there, then any name clash (eg. OLH the publisher, OLH the Journal) does not apply

 I've updated for clarity.   f688ae4

src/utils/logic.py
StephDriver: [see above](https://github.com/openlibhums/janeway/pull/5312#discussion_r3602354073)

src/utils/logic.py
StephDriver: [see above](https://github.com/openlibhums/janeway/pull/5312#discussion_r3602354073)

src/utils/logic.py
StephDriver: [see above](https://github.com/openlibhums/janeway/pull/5312#discussion_r3602354073)

src/core/include_urls.py
StephDriver: good catch. I don't know how I did that. 40a7a7f

src/core/templatetags/sitemap_tags.py
StephDriver: f473078

StephDriver:
@mauromsl  when testing these changes I found that the repo A11y page was falling back to the press page, but that's because we don't have repo data yet, and haven't built a repo version. So it shouldn't be there at all and the sitemaps shouldn't be listing it.  So I've added another commit to remove that link from Repo footers and sitemaps.

I also found the material repository had no footer link to the sitemap, so I added that too.

StephDriver:
@mauromsl there's now a branch adding the a11y page to repos, so I have removed the commit where I excluded them here from the sitemap for not exisiting.  And I've rebased the conflicts.

StephDriver:
following rebase, tests are failing, so now working on those.

rebasing b-5170_a11y_sitemaps onto an updated r-v1.9.x 
  NameError: name 'journal_models' is not defined in src/utils/logic.py, 

  1. On the original b-5170_a11y_sitemaps branch, src/utils/logic.py carried the module-level `import from journal import models as journal_models`, inherited from the point the branch diverged from master.
  2. none of this branch's commits modify the import line itself 
  3. Independently, master commit 4ad0c8f45 ("feat #3970) remove that line, and the removal was correct in the context of master at that time.
  4. When rebasing, this branch hadn't change the import, but it no longer existed on the base branch, so it didn't show as a conflict, and simply disappeared from this branch.

And there's a deprecation warning on `press.journals_az` which sitemap generator used to get an alphabetical list of journals.  The replacement doesn't necessarily provide journals in alphabetical order, so I've moved the ordering into the sitemap logic itself.

src/cms/models.py
mauromsl: I understand, the only issue being that `.save()` is not a reliable way of doing this. There are multiple scenarios where the ORM does not invoke `.save`, especially when doing bulk operations. If we want to reliably refresh these, we'd need to add a `signal`. To me it seems more trouble than its worth. It also means that the smallest of tweaks to a draft would require issuing new links to it to everyone else, since the link doesn't point to a snapshot of the draft, it will get dropped and stop working altogether

src/cms/views.py
```diff
@@ -4,6 +4,7 @@
 __maintainer__ = "Birkbeck Centre for Technology and Publishing"
 
 import os
+from uuid import uuid4
```
mauromsl: I don't think we use this here anymore

mauromsl (changes requested):
Thanks Steph, a couple more pointers

StephDriver:
@mauromsl 

**remove the last CMS migration if you have it locally on your database as I've edited it!**

- I removed the superfluous import
- I changed the CMS draft UUID to persist and updated tests
- this also meant the migration needed to change to generate UUID for exisiting published CMS items. I squashed this with the existing migration.
- I removed the GUI warning that access to page from preview link would be lost on publish, as this is no longer the case.
- I rebased the whole lot again - so it can be merged.

🤞

mauromsl (approved):
we got there!

PR #5314, opened 2026-05-19
Adds a11y mode options

### LINKED ISSUE #4947: Create "Clarity" styling for accessibility mode for front of house
After Clean has been set up as an accessibility mode for OLH and Material #4946 

Next we need to improve the default 'Clean' theme to set of colours and fonts optimised for accessibility.  These are normally set at Journal level as part of branding, but the accessibility mode should prioritise accessibility over branding - so good colour contrasts, readability of font etc.

ajrbyers:
- Adds a new a11y mode option that can be enabled per journal
- When enabled an a11y button is added after skip to content and in the nav (desktop and mobile)
- Closes https://github.com/openlibhums/janeway/issues/4947


ajrbyers:
Assigned to @StephDriver for initial review.

StephDriver:
Hi Andy, having a few problems getting this running.

1. from Press home page, I can enable a11y mode, but the new theme that loads looks like clean (could be clarity, harder to tell on home page) which then also has an option to 'enable' which takes me back. So either it doesn't play nice with dev environment `?theme=` and is skipping to the wrong theme, or the text isn't changing to show I'm in the mode.

2. Journal pages don't seem to have a11y mode in the nav at all.

ajrbyers:
> Hi Andy, having a few problems getting this running.
> 
>     1. from Press home page, I can enable a11y mode, but the new theme that loads looks like clean (could be clarity, harder to tell on home page) which then also has an option to 'enable' which takes me back. So either it doesn't play nice with dev environment `?theme=` and is skipping to the wrong theme, or the text isn't changing to show I'm in the mode.
> 
>     2. Journal pages don't seem to have a11y mode in the nav at all.

Make sure you have clarity installed.

Double check you are not over-writing the CONTEXT_PROCESSORS variable in your local settings (note that `"core.context_processors.accessibility_mode"` is added to janeway_global_settings.py)

In debug mode theme= alwyas takes precedent so dont use it, switch the themes manually from General

Clarity will be merged into core for 1.9 so a11y mode currently assumes you have it 😄

StephDriver:
Notes on how to get this working:

1. Clone Clarity to `src/themes`
2. `"core.context_processors.accessibility_mode"`  into `settings.py`
3. Migrate
4. `python src/manage.py load_default_settings`

StephDriver:
### Bug - unable to disable mode in some circumstances.
To reproduce:
- enable accessibilty mdoe
- log in
- try to disable accessibility mode - it won't.
note - if you log in and then enable accessibility mode you're fine, but if you enable it before logging in the bug occurs.
- then if you log out, accessibility mode is disabled, despite it being enabled in the session before you logged in.

### General Notes:
- some of the changes made on master are not here, so it is harder to tell whether an issue is with a11y mode or it hasn't be included yet. For example, the cross references back to the text from footnotes and bibliography do not seem to be here regardless of state of a11y mode.
- This has removed the Accessibility link from the footer - is that deliberate?
- the window size at which a11y mode is in narrow screen is not the same as OLH, so you can move from narrow screen a11y to main layout OLH. 
- I note `paper` is hardcoded in as the palette - is this expected to change? 
- I'm unconvinced that 'disable' is a good term to use for moving between a11y and standard modes.
- I'm unconvinced that `skip-to-main` and `enable accessibility mode` should share space, with only one visible at a time especially as they take up different vertical space such that the page shifts between them, I think they should be side by side with skip-to-main remaining visible when accessibility mode appears.
- Why the different styles for `skip to main` and `enable accessibility mode` ?
image
image


### A11y work done on other themes that is not here
- keyboard trap on enlarged tables. To reproduce, navigate with a keyboard, enlarge a table in an article, then try to navigate to the close button (you can't). 
- keyboard 'something wrong' when following an in-page link (e.g. from text citation to the bibliography). It lands in the right place in the bibliography, but the next `tab` takes you to the `skip-to-main` at the top of the page instead of the backlink.
- Dyslexia mode needs to exclude the icons specifically,
image

StephDriver (changes requested):
see notes above - there is one bug that must be fixed, the others are feedback for your consideration.

ajrbyers:
Some changes to clarity here: https://github.com/openlibhums/clarity/pull/6

src/core/views.py
```diff
@@ -2622,6 +2623,68 @@ def set_session_timezone(request):
...
+    ):
+        return redirect(next_url)
+
+    referer = request.META.get("HTTP_REFERER")
+    if referer and url_has_allowed_host_and_scheme(
+        url=referer,
+        allowed_hosts={host},
+        require_https=request.is_secure(),
+    ):
+        return redirect(referer)
```
mauromsl: Can we add a message here? Something like `"Failed to redirect to disallowed url:%s ", referrer `

mauromsl (approved):
Looks good, just one minor thing to add, otherwise approved

StephDriver (commented):
### action not status
Andy I like the on/off change in language and appreciate the bug fix, but it is now unclear that this is an action rather than a status.  

Image

### Side by side hidden buttons
This has been fixed with the changes to clarity for the journal and press, but still appears collapsed on the repository.

Image

Image

### Footer A11y link
One further thought, if we are having an accessibility link in the nav, do we also want it in the footer?

### Repositories

Repository gets the a11y-mode link next to skip to main (and it works), but nothing in the nav.

[btw, the repo a11y mode looks fab]

### Log out A11y mode changes

- with A11y mode ON or OFF
- log into account that already has a11y-mode on
- log out of that account
- A11y mode is turned off  -- expected it to still be on, as was on for user account that was logged in.  Doesn't return to whatever it was before log in either, always turns off.

- Turn a11y mode on
- log into account that doesn't have a11y mode on previously, and it stays on ✅ 
- log out, and a11y mode turns off

I think we need to be very clear when A11y mode turns on, stays on, turns off etc on actions other than manual switch on/off.  These are my expectations.

#### Principles
  1. **Anonymous default is OFF.** 
  2. **On login**, if the user explicitly toggled during the anonymous session,
     that choice wins and updates the account. If they did *not* touch it, the
     account preference is applied. This is key so that users who have switched on a11y mode to use the site don't lose that on log in.
  3. **The mode is sticky across logout** — logging out never changes the current
     mode. The account simply retains its last saved value.   The mode could have been changed while logged in, so this is not necessarily the same as the anonymous value.


#### Login Mode Changes
 | Test | Anonymous session | Account before | Expected mode after login | Expected account after login | Rule |
  |---|-------------------|----------------|---------------------------|------------------------------|------|
  | ✅  | Untouched | OFF | **OFF** | OFF | Nothing to apply |
  | ✅ | Untouched | ON  | **ON**  | ON  | Account preference applied (rule 2) |
  | ✅ | Explicit ON  | OFF | **ON**  | **ON** ← updated | Explicit action wins + writes back (rule 2) |
  | ✅  | Explicit ON  | ON  | **ON**  | ON  | Consistent |
  | ✅  | Explicit OFF | OFF | **OFF** | OFF | Consistent |
  | :x: | Explicit OFF | ON  | **OFF** | **OFF** ← updated | Explicit OFF wins + writes back (rule 2) |

#### Logout Mode Changes

This should be simpler, no need to consider an explicit on/off, just need to persist whatever the state was in the account.

 | Test | Mode at logout | Account | Expected mode after logout | Expected account after logout | Rule |
|--|--|--|--|--|--|
  | :x: | ON  | ON  | **ON**  | ON  | Stays on (rule 3) |
  | ✅  | OFF | OFF | **OFF** | OFF | Stays off (rule 3) |

StephDriver (changes requested):
See comments above I think there are some changes still to make.

StephDriver (changes requested):
Almost there - so close!


image


versus

image


as keyboard user !=> screen reader user.  Could both have the toggle that shows the current state?

StephDriver (approved):
approved, but handed back to you for the conflicts.

PR #5315, opened 2026-05-19
Fix inline validation errors on password update form

### LINKED ISSUE #4642: Password update: Users need informative and persistent messaging if password update fails due to not meeting requirements
### Issue 

Users who try to update their password with one that does not meet Janeway's requirements are confused: 
1. no instruction as to password requirements
2. confusing error messages display in separate floaty boxes at the top right of the screen (not next to password field) and which disappear after a few seconds 
3. no indication that the password has not been updated 

**Janeway version**
1.5.6, 1.7.4 

**To Reproduce**

1. Update account password with a password that doesn't meet the requirements (e.g. too short, no uppercase characters) 
Image

2. See page refresh with a confusing error message floating box at the top right of the corner that does not explain that the password has not been updated and disappears in seconds: 
Image

**Expected behavior**

1. Password requirements should be displayed in the password update box 
Image
2. Error message should be persistent, clear, and display in or near the password update box, similar to the error message on the registration form when a non-compliant password is input, e.g. ``Password not updated: <reasons>``
Image

ajrbyers:
Closes #4642

### DIFF
--- a/src/core/forms/forms.py
+++ b/src/core/forms/forms.py
@@ -277,6 +277,60 @@ class RegistrationForm(forms.ModelForm, CaptchaForm):
         return user
 
 
+class PasswordChangeForm(forms.Form):
+    """
+    A form for changing the password of an already-authenticated user.
+
+    Validates the current password, confirms the two new-password fields
+    match, and runs the press password policy check so that all failures
+    are surfaced as inline form errors rather than disappearing toast
+    messages.
+    """
+
+    old_password = forms.CharField(
+        label=_("Current password"),
+        widget=forms.PasswordInput,
+    )
+    new_password_one = forms.CharField(
+        label=_("New password"),
+        widget=forms.PasswordInput,
+    )
+    new_password_two = forms.CharField(
+        label=_("Confirm new password"),
+        widget=forms.PasswordInput,
+    )
+
+    def __init__(self, *args, user=None, request=None, **kwargs):
+        super().__init__(*args, **kwargs)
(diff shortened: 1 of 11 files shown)


src/core/forms/forms.py
```diff
@@ -246,6 +246,60 @@ def save(self, commit=True):
...
+            raise ValidationError(_("Current password is incorrect."))
+        return value
+
+    def clean(self):
+        cleaned = super().clean()
+        new_one = cleaned.get("new_password_one")
+        new_two = cleaned.get("new_password_two")
+
+        if new_one and new_two and new_one != new_two:
+            self.add_error("new_password_two", _("Passwords do not match."))
```
mauromsl, line 320: All we've done here is change the punctuation, it is worth updating the key on PO files

mauromsl (changes requested):
one minor tidbit

src/core/forms/forms.py
```diff
@@ -246,6 +246,60 @@ def save(self, commit=True):
...
+
+        if new_one and self.user and self.request:
+            problems = self.user.password_policy_check(self.request, new_one)
+            for problem in problems:
+                self.add_error("new_password_one", _("Password not updated: ") + str(problem))
+
+        return cleaned
```
mauromsl, line 329: we are adding the errors on the form inline, which is great, but we lost the `messages` / toastr prompt. Is it worth also displaying one?

mauromsl (approved):

PR #5331, opened 2026-06-05
Redirect unslashed RSS feed URLs instead of serving catch-all

### LINKED ISSUE #3411: Preprints: RSS feed URL returns error if lacks trailing slash
**Describe the bug**
When accessing a preprints repository RSS feed, an empty file returns if the URL lacks a final trailing slash. This doesn't occur with journal feed URLs as far as I know. 

**Janeway version**
1.4.4 RC6

**To Reproduce**
1. Go to https://dev.eartharxiv.org/rss/preprints
2. View the file; see no preprints data 
3. Compare: Go to https://dev.eartharxiv.org/rss/preprints/
4. View the file; see preprints data as expected

**Expected behavior**
``/rss/preprints`` is functionally equivalent to ``/rss/preprints/``

**Additional context**
I only noticed this when I accessed the URL directly from my browser. My normal feed reader automatically corrected the issue. However, the Janeway journals feeds tested at Michigan and Iowa displayed the same data regardless of the trailing slash.

ajrbyers:
Closes #3411

### DIFF
--- a/src/rss/tests.py
+++ b/src/rss/tests.py
@@ -1,0 +1,57 @@
+__copyright__ = "Copyright 2026 Birkbeck, University of London"
+__author__ = "Andy Byers"
+__license__ = "AGPL v3"
+__maintainer__ = "Birkbeck Centre for Technology and Publishing"
+
+from django.test import TestCase, override_settings
+from django.urls.base import clear_script_prefix
+from django.utils import timezone
+
+from repository import models as repository_models
+from utils.testing import helpers
+
+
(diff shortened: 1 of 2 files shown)


StephDriver (approved):
Reproduced the bug that showed different content and confirmed this fixed it. 

Diffed the output to check `/preprints` and `/preprints/` was the same after the fix, and that they were the same as `/preprints/` before the fix. 

Ran the new tests, all pass.

PR #5340, opened 2026-06-09
a11y reading options bar

### LINKED ISSUE #5337: Extend dyslexia mode to more reading text format options
The ability to format text is a key access adjustment for many. Some are more comfortable with dark-modes. Others need a particular combination of colours or contrast ratio in order to read text. 

WCAG 2.2 AA requires us not to prevent users overriding our text format. Janeway already provided a dyslexia mode on the OLH theme, and in 1.9 work this was extended to all themes.  This particularly looks at the readability of **long** text (e.g. articles) rather than the readability of the interface in general.  Some people may navigate the site with a screen-reader but prefer to be able to read text visually, or vice versa - some may also do both at different times.  As a publishing platform, breaking down access barriers for reading the text of publications is very important.

This issue introduces text formatting functionality. It builds upon existing work already done for 1.9, including the new Accessibility Mode and the extension of dyslexia mode to all themes.  It will not be comprehensive, but will provide basic functionality for colour and font choices as well as bring the existing dyslexia options into a single text format UI which can then be extended in the future.

The formats will link to a target class for the parent element.  So this becomes a generic text-format UI that can be dropped into pages other than `article` and given other targets.  By only targeting some elements (e.g. article text) it also provides a focused reading option for these longer texts, and clear separation between them and side-bar content.

StephDriver:
closes #5337 

openlibhums/clarity#10 ports these changes to clarity.


StephDriver:
This is effectively a 'focused reading' styling
- [x] basic functionality
  - [x] article pages
  - [x] news items
  - [x] cms pages
  - [x] preprints
  - [x] accessibility information page
- [x] make persistent (as per a11y-mode as a model)
- [x] theme by theme visual styling tweaks
- [x] screen reader testing and updates


Extras:
- preprint OLH theme bug that didn't display preprint html (but material version did) - one line fix included.

StephDriver:
### Limitations / Future Development.

- Initially I introduced options for custom colours, but making the colour pickers accessible to both keyboard and screen-reader users across all three themes became a major piece of work which is out of scope for this issue.  So I have left the code side of the customisation for use in future, but it is just not wired up to the templates.  Note - the custom colour pickers work fine by mouse. see 6a8ab05e4132f4598633a61e7cb6d34abe55d872

- This has been created as a generic feature, which *could* be added to any page, however the targets on each page need to be set individually. So for this first version I have focused on areas of high text content (article, preprint, CMS, news item).  

- Material reading bar is not fully screen-reader accessible. But it is as accessible as the Material Nav - the limitation is Material itself. The state of items in the reading bar is announced, but the drop-downs do not have a `menu popup collapsed` or `menu popup expanded` state announced to user.  This is the same as with the Material Nav.

- Text size differences between themes - especially noticeable on mobile, the default text size is different on each theme, which means having customised it on say a Press running Clean and then going to a Journal on OLH, the customisations carry across, but the text looks a very different size.  I'm classing this as out of scope, because the majority of users will not move between themes, it is easy enough to then adjust and this issue is intended to setup the basic functionality.

StephDriver:
## Screenshots

### Non-article pages
image

### Article pages

Also include the draw-attention animation toggle, cite, email and print options, depending on settings. And material doesn't have the cite modals.

image

### Pre-set colours and fonts
I've created a doc to describe how to amend the fonts and colours, they are kept in `src/core/text_formats.py` for easy editing. Also see [note about custom colours above](https://github.com/openlibhums/janeway/pull/5340#issuecomment-4720001481).

image

image

### Toggles
I've copied the Italics toggle from the previous dyslexia mode, and added one for `dark mode` which swaps the foreground and background of the current text colours, and one for the `draw attention highlighting` which is a way to suppress the animation on using an internal page link that briefly draws attention by highlighting the landing element.

docs/md/reading-options-bar.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Extending the reading text-format options
+
+### Adding a new font
+
+If the font is a web-safe stack (e.g. system serif/sans), only step 3 is required. 
+
+1. Put the font files in `src/static/common/fonts/<family>/` (`.woff2`/`.woff`
+   preferred; `.otf`/`.ttf` work too). Include the font's licence file.
+2. Add an `@font-face` block to `src/static/common/css/text_readability.css`:
```
ajrbyers, line 33: Does this file exist?

src/core/logic.py
```diff
@@ -849,6 +840,188 @@ def accessibility_mode_active(request):
...
+    }
+
+
+# --- Reading options preferences -------------------------------------------
+
+# Allowed fonts, schemes and size bounds derive from the text_format registry,
+# the single source of truth shared with the bar template and the JS. Server-side
+# validation means a stale or tampered stored value can never lodge an
+# unresolvable preference (the same guarantee the JS rollback gives live).
+_HEX_COLOUR_RE = re.compile(r"^#[0-9a-fA-F]{6}$")
```
ajrbyers, line 944: No need to redefine this here as we can get it from the already imported `text_format`

src/static/common/js/text_readability.js
```diff
@@ -1,63 +1,687 @@
...
+    });
+  });
+}
+
+  // Reflect state back onto controls.
+function syncControls() {
+  // Keep the menu-button labels in step with the selection. 
+  setLabelText('.tf-font-select', (FONTS[state.font] || FONTS['default']).label);
+  setLabelText('.tf-scheme-select', (COLOURS[state.scheme] || COLOURS['default']).label);
+  document.querySelectorAll('.tf-custom-colours').forEach(function (block) {
```
ajrbyers, line 413: Duplicated code here.

src/static/common/js/text_readability.js
```diff
@@ -1,63 +1,687 @@
...
 
-// Track cumulative resize value
-var cumulativeResize = 0;
-var minResize = -3;
-var maxResize = 6;
+// Initialize on page load
+if (document.readyState === 'loading') {
+  document.addEventListener('DOMContentLoaded', initialise);
+} else {
+  initialise();
```
ajrbyers, line 10: Do we need to eager call initialise here if we're calling it on demand below?

src/templates/common/elements/journal/reading_options_bar.html
```diff
@@ -0,0 +1,189 @@
+{% load i18n %}
+{% load hooks %}
+{% load static %}
+{% comment %}
+    Front of House self contained drop in.
+    Include only once per page as would otherwise load resources twice.
+    add marker class `.text-format-region` to define the target parent(s).
+{% endcomment %}
+<section aria-label="{% trans 'Reading options' %}" class="invisible-landmark reading-options-landmark">
+<div id="tf-bar" data-sticky-container class="row reading-options-bar{% if text_format_preferences.hideReadingBar %} tf-bar-hidden{% endif %}">
```
ajrbyers, line 10: data-sticky is on two divs here.

ajrbyers (commented):
I've added a couple of inline comments but generally:

- The bar should remain in its original place as with OLH on all four themes, it feels awkward at the top of the page

StephDriver:
> The bar should remain in its original place as with OLH on all four themes, it feels awkward at the top of the page

@ajrbyers  I disagree. I tried placing the bar where it was on OLH originally - but it means this was then confusing as it cannot be universally added to non-article pages and the majority do not have banner images.  Placing it in different places on different pages felt clunky and meant different CSS was needed in different places.  Placing it at the top of the page is the only 'universal' position I could find that worked across other templates, e.g. news and CMS pages.   And then I made it sticky so once you scroll down past the banner image it will appear at top of window anyway.

ajrbyers:
> > The bar should remain in its original place as with OLH on all four themes, it feels awkward at the top of the page
> 
> @ajrbyers I disagree. I tried placing the bar where it was on OLH originally - but it means this was then confusing as it cannot be universally added to non-article pages and the majority do not have banner images. Placing it in different places on different pages felt clunky and meant different CSS was needed in different places. Placing it at the top of the page is the only 'universal' position I could find that worked across other templates, e.g. news and CMS pages. And then I made it sticky so once you scroll down past the banner image it will appear at top of window anyway.

Ah right I was thinking about it specifically from the article POV. I'll take another look this morning.

StephDriver:
@ajrbyers I've added a commit which sets the default to hide the reading options bar. It is then found in the accessibility menu. I hope that is a good compromise to get this out the door.

ajrbyers (approved):

PR #5357, opened 2026-06-20
Document manual UI testing best practices

joemull:


docs/md/manual-ui-testing.md
```diff
@@ -0,0 +1,38 @@
...
+## Using multiple accounts and roles
+
+When testing a user interface manually, you often need to be logged in with several different accounts that have different roles.
+
+You can use Firefox container tabs to switch quickly between roles.
+
+## Testing on a phone or other device at home
+
+When you are on secure local network, you can run the Janeway local server in such a way that you can test it on other devices.
+
```
StephDriver, line 14: I think this section might need subsections for different OS or something very clear that these are linux instructions.  For example, when working on macOS, I can use the machine name `http://keeper-of-keys.local` as a substitute for "localhost" instead of having to find the IP or set a domain alias.  e.g. `http://keeper-of-keys.local:8000/olh/article/id/10801/?theme=clean`  I still have to use the additional argument on running the server, as here, but the rest no.

docs/md/manual-ui-testing.md
```diff
@@ -0,0 +1,38 @@
...
+1. Run your development server with an additional argument to specify an IP of four zeroes, and the same port as normal: `python src/manage.py 0.0.0.0:8000`.
```
StephDriver: `python src/manage.py 0.0.0.0:8000` -> `python src/manage.py runserver 0.0.0.0:8000`

docs/md/manual-ui-testing.md
```diff
@@ -0,0 +1,38 @@
...
+## Switching between themes
+
+On a development installation, you can quickly switch to a different front-office theme by adding a query parameter.
+
+For example, adding `?theme=material` to the URL will load the page in the Material theme.
+
+Caveats:
+
+- This only works when `DEBUG` is turned on.
+- The Hourglass press theme (used by the Open Library of Humanities) does not support this feature.
```
StephDriver, line 40: This also affects the new A11y mode - currently we cannot use the`?theme=` when testing anything with a11y mode as it interferes with how that swaps. I'm not sure if this is a temporary until clarity is merged here.

docs/md/manual-ui-testing.md
```diff
@@ -0,0 +1,38 @@
...
+This guide offers tips for testing the Janeway user interface manually during development.
+
+## Using multiple accounts and roles
+
+When testing a user interface manually, you often need to be logged in with several different accounts that have different roles.
+
+You can use Firefox container tabs to switch quickly between roles.
+
+## Testing on a phone or other device at home
+
```
StephDriver, line 12: Maybe a line about how janeway has different wide and narrow screen layouts so testing needs to cover different widths of window.  But somehow distinguish this from implying narrow = test on mobile.

StephDriver (changes requested):
this looks like a good starting point - and I'm suspecting it is a starting point for this PR (no issue to tell me what the aim is here). 

I have added a handful of direct comments.   Overall, good - sure, let's start here. 

If you are looking to add more now, I think the it might be worth mentioning the `build_assets` and how to set the domain as those are the two I use most often when UI testing. So easy to forget to build the assets - and so wonder why your change didn't work or to end up with every attempt to reach a page redirecting to example.org if the domain isn't set - especially if you open in a different port to normal to say compare the UI on master alongside your branch.

docs/md/manual-ui-testing.md
StephDriver: see also [devops - screen-reader testing](https://github.com/openlibhums/devops/blob/main/workflows/screenreader_testing.md) which is basic notes on how I set up on mac for screenreader testing on iphone. Key point there was that you need to turn off the debug tool bar or it interferes with any (and all) a11y tests as it's not very accessible and covers other content.

joemull: OK, that's good to know. I don't think I can add it myself since I don't have this setup, but ideally you or another Mac user can add it to this doc in the future.

What I have done is put these instructions under a Linux h3 to make clear they all apply to this setup--with an implied Mac set of instructions to come later.

docs/md/manual-ui-testing.md
joemull: I've rephrased it to omit mentioning phones.

joemull:
> I think the it might be worth mentioning the `build_assets` and how to set the domain as those are the two I use most often when UI testing. So easy to forget to build the assets - and so wonder why your change didn't work or to end up with every attempt to reach a page redirecting to example.org if the domain isn't set - especially if you open in a different port to normal to say compare the UI on master alongside your branch.

Please do add things you find helpful once this is merged! I have been approaching these docs-in-progress in a very streamlined manner--only adding sections and pages when they are fresh in my mind because I've just dealt with them while making or fixing something, so I am going to refrain from aiming for completeness at this stage, or I would never emerge from the writing cave 😄

StephDriver (approved):

PR #5361, opened 2026-06-22
ruff check staged version of file

### LINKED ISSUE #5360: ruff pre-commit git-hook targets disk not staged files
The codebase uses a git-hook to check python files have Black formatting prior to commits.  But this check passes when formatting changes exist, but are not staged for that commit. 

- stage changes
- attempt a commit
- commit rejected as not in format
- `ruff format .`
- attempt commit
- commit accepted.
- formatting changes NOT committed, as were never staged, but hook passes them as on disc they are formatted.

While the output advises the developer to re-stage, there is currently no guard to ensure that, with the commit passing even if this step is omitted.

StephDriver:
closes #5360

### DIFF
--- a/.git-hooks/pre-commit
+++ b/.git-hooks/pre-commit
@@ -12,12 +12,44 @@ if [ -z "$STAGED_PYTHON_FILES" ]; then
 fi
 
 echo "Checking formatting for staged Python files..."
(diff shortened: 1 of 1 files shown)


mauromsl (approved):
good catch!

PR #5377, opened 2026-07-01
fix: #5348 escape literal % in search SQL before Article.objects.raw()

### LINKED ISSUE #5348: Search with pesky terms result in server error
**Describe the bug**

When processing a search, we build the final query by rendering the ORM query (stringify_queryset()/mogrify()), then hand-wrapping it (Article.objects.raw(...)) (see [here](https://github.com/openlibhums/janeway/blob/819fe88a8781c2b275bc884d3d74dd8dac668899/src/submission/models.py#L894))
- `mogrify()` escapes the term, leaving literal % in the text
- `.raw()` does placeholder-parsing that text; IIC, with params =() (empty but not None), 

**Janeway version**

1.8

**To Reproduce**

Try this search:

(please don't click too often as these links go to real journals)

- https://www.comicsgrid.com/search/?article_search=50%25%2520chance%2520to%2520double%2520or%2520half%2520amount
- https://pynchonnotes.openlibhums.org/search/?article_search=50%25%2520chance%2520to%2520double%2520or%2520half%2520amount

**Expected behavior**

The user sees a 500 server error

The system gives
- if using psycopg2 → IndexError
- if using psycopg3 → ProgrammingError

**Additional context**

Adding the following right after `inner_sql = self.stringify_queryset(queryset)` [here](https://github.com/openlibhums/janeway/blob/819fe88a8781c2b275bc884d3d74dd8dac668899/src/submission/models.py#L894) should fix it
```python
        # mogrify() already inlined the parameters into inner_sql, but .raw()
        # makes psycopg3 re-parse the string for placeholders. A literal '%'
        # coming from the search term (e.g. "50%") then looks like a bad
        # placeholder ("%2", ...). Double them so psycopg3 collapses them back.
        inner_sql = inner_sql.replace("%", "%%")
```

thisismyurl:
Hey folks,

Thanks to whoever triaged #5348 — the diagnosis in the issue is spot on and saved me the hunt.

`postgres_search()` builds its final ordered query by wrapping `stringify_queryset()` in `Article.objects.raw()`. The catch is that `stringify_queryset()` runs the params through `cursor.mogrify()`, so what comes back is fully-interpolated SQL — every value already baked in as a string literal. When a search term contains a `%` (someone searching for "50%"), that `%` is now sitting in the SQL as a literal. Then `Article.objects.raw()` hands the string to the cursor with `params=()` — Django's default is an empty tuple, not `None` — so psycopg runs `%`-formatting over it again, hits the stray `%`, and raises `TypeError: not enough arguments for format string`. Any search for a term with a `%` in it 500s.

The fix doubles the literal `%` to `%%` in the mogrified SQL before it goes into `raw()`, so they survive that second formatting pass unchanged. Because `stringify_queryset()` has already interpolated everything, there are no real placeholders left to protect — every `%` in that string is literal, so doubling them all is safe.

I added a regression test that searches for a `%`-bearing term and asserts the query executes rather than raising (it guards the Postgres path, since the `raw()` query only runs there). I confirmed the mechanism directly too — a mogrified string with a literal `%` raises under `%`-formatting with empty params, and doubling it clears the error.

One thing I wasn't sure about: I branched this off `master`, but I noticed recent bug fixes going to `r-v1.9.x` — happy to retarget or add a backport if that's the flow you'd prefer.

closes #5348

(full disclosure: AI helped me identify the issue and verify my work)

### DIFF
--- a/src/submission/models.py
+++ b/src/submission/models.py
@@ -893,6 +893,14 @@ class ArticleSearchManager(BaseSearchManagerMixin):
         # distinct fields to match order_by fields
         inner_sql = self.stringify_queryset(queryset)
 
+        # stringify_queryset() returns SQL with every parameter already
+        # interpolated via cursor.mogrify(), so any '%' left in the string is a
+        # literal (e.g. from a search term like "50%" or a LIKE pattern). Article
+        # .objects.raw() defaults params to () rather than None, so it re-runs
+        # %-formatting on this SQL and chokes on those literal '%'. Double them so
+        # they survive that pass unchanged. (#5348)
+        inner_sql = inner_sql.replace("%", "%%")
+
         if "relevance" in sort:
             # Relevance is not a field but an annotation
             return Article.objects.raw(
--- a/src/submission/tests/test_article_search.py
+++ b/src/submission/tests/test_article_search.py
@@ -183,3 +183,39 @@ class ArticleSearchTests(TransactionTestCase):
         result = [a for a in queryset]
 
         self.assertEqual(result, [article])
+
+    @override_settings(ENABLE_FULL_TEXT_SEARCH=True)
+    def test_article_search_term_containing_percent(self):
+        """A search term containing '%' must not raise (#5348).
+
+        stringify_queryset() mogrifies the term into the raw SQL, so a literal
+        '%' from the term reaches Article.objects.raw(), which re-runs
+        %-formatting on the string and previously raised. Evaluating the search
+        should now complete instead of erroring.
+        """
+        from django.db import connection
+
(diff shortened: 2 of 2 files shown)


mauromsl (changes requested):
Thanks @thisismyurl

We are actually preparing a patch release v1.8.1 with some minor fixes.

Would you mind changing the base of your branch and PR to `r-v1.8.x`?

thisismyurl:
Done — rebased onto r-v1.8.x and switched the PR base to match. Same single commit, clean rebase.

mauromsl (approved):

PR #5382, opened 2026-07-07
add alt text for css pseudo elements

### LINKED ISSUE #5381: add alt-text to css pseudo elements
we have several inline lists that have pseudoelements e.g.

```css
.bar-sep li:not(:last-child)::after {
    content: " | ";
}
```

we should ensure these (and any others) have appropriate alt text where needed.  So something like:

```css
.bar-sep li:not(:last-child)::after {
    content: " | " / "";
}
```

StephDriver:
closes #5381

### DIFF
--- a/src/static/admin/css/app.css
+++ b/src/static/admin/css/app.css
@@ -5158,7 +5158,7 @@ ul.actions li a {
 
 /* Create the caret/arrow with a unicode, and style it */
 .caret::before {
-  content: "\25B6";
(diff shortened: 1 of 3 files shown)


StephDriver:
Note - when I went through the codebase, most of the use of css pseudo elements were to add text or punctuation which was being used as that punctuation, so only three changes were needed, all of which were decorative rather than semantic.

mauromsl (approved):

PR #5396, opened 2026-07-17
A11y options info onto the A11y page

### LINKED ISSUE #5289: Include A11y Mode info on A11y page
Information about Accessibility Mode and recommendations for when to use it, need to be added to the A11y information page.  

This work cannot be done until the Accessibility mode is present.

StephDriver:
closes #5289

### DIFF
--- a/src/templates/common/elements/a11y/a11y_options.html
+++ b/src/templates/common/elements/a11y/a11y_options.html
@@ -1,5 +1,7 @@
 {% load i18n %}
-<h2>{% trans 'Reading Options' %}</h2>
+<h2>{% trans 'Accessibility Options' %}</h2>
+
+<h3>{% trans 'Text' %}</h3>
     <p>
         {% blocktrans %}
             The Reading Options bar lets you customise how text is displayed. It is available on text-heavy pages
@@ -8,7 +10,7 @@
         {% endblocktrans %}
     </p>
 
-<h2>{% trans 'Accessibility Mode' %}</h2>
+<h3>{% trans 'Accessibility Mode' %}</h3>
     <p>
         {% blocktrans %}
             Accessibility Mode is a user-enabled theme built from the ground up to be accessible. The current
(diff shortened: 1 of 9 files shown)


src/templates/common/elements/a11y/a11y_options.html
```diff
@@ -0,0 +1,34 @@
+{% load i18n %}
+<h2>{% trans 'Accessibility Options' %}</h2>
+<p>
+The accessibility options may be accessed from the Accessibility Menu <i class="fa fa-universal-access" aria-hidden="true"></i> at the top of each page.
```
ajrbyers: This section is not marked for translation like the others.

src/templates/common/elements/a11y/a11y_options.html
```diff
@@ -0,0 +1,34 @@
...
+<h3>{% trans 'Text' %}</h3>
+    <p>
+        {% blocktrans %}
+            The Reading Options bar lets you customise how text is displayed. It is available on text-heavy pages
+            and targets large areas of text, such as journal articles, rather than surrounding page elements,
+            so the reading area stays focused.
+        {% endblocktrans %}
+    </p>
+
+<h3>{% trans 'Accessibility Mode' %}</h3>
```
ajrbyers, line 19: a11y mode can be enabled/disabled so we should either describe that or guard this paragraph using the setting.

ajrbyers (changes requested):
Just two comments. Generally awesome.

src/templates/common/elements/a11y/a11y_options.html
StephDriver: @ajrbyers 
As in at press level for the whole press? per journal? 

some of our overall compliance is based on a11y mode being the alternative for any user on a theme which isn't accessible, so if staff can disable it, then are we informing end users it is missing? or are we silently allowing them to think Janeway itself is not Accessible?

ajrbyers: Its per journal. It will be enabled on all OLH journals but we cant say that for all folks using Janeway.

StephDriver: as per discussion elsewhere, leaving this as we will be making a11y mode a fixed feature.

PR #5399, opened 2026-07-21
Enhance keyboard accessibility for clean table modal

### LINKED ISSUE #5387: Some keyboard navigation updates have been lost
this is looking at r-1.9.x. 

1. keyboard trap, cannot dismiss enlarged table modal.
- only affects Clean and Clarity. 
- Clean was working in PR #4901.  So compare with that to find what has been lost. 
- To reproduce. Navigate an article which has tables using the keyboard. Open a modal, try to dismiss the modal by tabbing to the close button - you can't!   Compare with Material and OLH where it works fine. 

2. on navigating an in page link, on tabbing the keyboard focus returns to the top of the page, rather than the landing element. 
- affects OLH, untested on other themes.
- was working when introduced in PR #4789

ajrbyers:
Closes #5387

### DIFF
--- a/src/journal/tests/test_journal_frontend.py
+++ b/src/journal/tests/test_journal_frontend.py
@@ -1,3 +1,8 @@
+import os
+import shutil
+
+from django.conf import settings
+from django.core.files.uploadedfile import SimpleUploadedFile
 from django.test import TestCase, override_settings
 from django.urls import reverse
 from django.utils import timezone
@@ -8,6 +13,7 @@ from journal.tests.utils import make_test_journal
 from press.models import Press
 from utils.testing import helpers
 from utils import setting_handler
+from core import files as core_files
 from core import models as core_models
 from submission import models as submission_models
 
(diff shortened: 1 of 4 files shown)


StephDriver (commented):
## Confirmed & approved:

1. clean keyboard trap where having opened a modal it could not be dismissed has been resolved.

2. olh page links now keep focus at landing point rather than sending it to top of page.

## Questions
1. Did the scope of this include the Figure modals? there's something not quite right, as if having tabbed to the close button, one tabs again, it is possible to end up back on the main page, but with it still in shadow because the modal hasn't closed.  I think when open the modal should keep focus trapped to the modal until it is dismissed, certainly it shouldn't return to the main page without dismissing the modal. 

The table modals keep the focus within the modal until dismissed, the figure modals do not.

2. Is there a related clarity PR? The bug affected clean and clarity.

StephDriver (changes requested):

ajrbyers:
Closed by https://github.com/openlibhums/janeway/issues/5387

PR #5405, opened 2026-07-23
feat: #3755 make the submitted file version 1 automatically

### LINKED ISSUE #3755: Repository/Preprints: The first submission should automatically be version 1
As a Repository Manager, I am annoyed that I *must* select the one PDF awaiting as "version" in order to start my review of the preprint. There is only PDF, so clearly it should be the first version! 

The file/PDF submitted by the author should automatically be interpreted by Janeway as Version 1. 

There may be cases in which the author accidentally uploads multiple PDFs due to the lack of information in the user interface. (ref: https://github.com/BirkbeckCTP/janeway/issues/2810) In this case, the most recent file should be flagged as Version 1 and the Manager pointed toward that file.

As a user submitting a preprint, I'd like my initial submission to be flagged as v1 of that preprint 

**Additional context**
This could be combined with the Summary Page at the end of preprint submission. On that page, the submitting author should see the files (PDFs) they have submitted and have the option to designate one as the primary file -- the one which will be Version 1: https://github.com/BirkbeckCTP/janeway/issues/3478

Jayashanker-Padishala:
Closes #3755

## What this does

When an author completes preprint submission (`Preprint.submit_preprint()`), the submission file now automatically becomes **PreprintVersion 1**. Repository managers no longer have to select the one pending file as a "version" in the manager view before they can start reviewing.

## Design notes

- The version is created from `preprint.submission_file`, which the submission file-upload view overwrites on every upload — so it always points at the author's **most recent** file. That covers the accidental multiple-upload case discussed in the issue (ref BirkbeckCTP/janeway#2810): the latest file is the one flagged as version 1.
- Guarded by `has_version()`: if a version already exists nothing is created, so the change is idempotent and cannot duplicate versions.
- If there is no submission file (edge case — the files step normally blocks completion without one), behavior is unchanged: no version is created.
- The manager's existing "make version" control is untouched and still works for subsequent versions.

## Tests

Three new tests in `repository/tests/test_models.py`:
- submitting creates version 1 pointing at the submission file
- submitting when a version already exists does not create a duplicate
- submitting without a submission file creates no version

`DB_VENDOR=sqlite python src/manage.py test repository` — 47/47 pass locally. `ruff format --check` clean on both changed files.

### DIFF
--- a/src/repository/models.py
+++ b/src/repository/models.py
@@ -1043,6 +1043,12 @@ class Preprint(models.Model):
         self.current_step = 5
         self.save()
 
+        # The most recently uploaded submission file automatically becomes
+        # version 1, so managers do not have to select it manually before
+        # starting their review.
+        if self.submission_file and not self.has_version():
+            self.make_new_version(self.submission_file)
+
     def subject_editors(self):
         editors = []
         for subject in self.subject.all():
--- a/src/repository/tests/test_models.py
+++ b/src/repository/tests/test_models.py
@@ -135,6 +135,23 @@ class TestModels(TestCase):
                 article_two,
             )
 
+    def test_submit_preprint_creates_first_version(self):
+        self.preprint_one.submit_preprint()
+        self.assertEqual(self.preprint_one.preprintversion_set.count(), 1)
+        version = self.preprint_one.current_version
+        self.assertEqual(version.version, 1)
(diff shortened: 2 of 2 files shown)


ajrbyers:
@alainna might be quite excited by this PR.

ajrbyers (approved):

PR #5420, opened 2026-08-04
Merge the Clarity theme into Janeway's core

ajrbyers:
This PR merges Clarity into the core of Janeway and adds a handler for the palette setting.


src/core/janeway_global_settings.py
```diff
@@ -623,6 +623,20 @@ def __len__(self):
...
+    "clarity": {
+        "clarity_palette": [
+            "evergreen",
+            "ocean",
+            "cardinal",
+            "paper",
+        ],
+    },
+}
+
```
mauromsl: These would be easier to handle if we used the `Setting`model.

I propose we keep the concept of theme related settings and we use a pattern similar to what we do for plugins. instead of `plugin:<plugin_name>` make the group `theme:<theme_name>`

We get the benefit of managing these via Janeway alone and being able to set default values.

Themes are not yet janeway/django applications and they lack an "entrypoint" so for now they will need to exist in the JSON document that contains all the default settings, or consider baking their management `clarity/build_assets.py though I'm not sure the latter is a good approach so I'd prefer just adding them to the JSON document for now and explore giving themes a load procedure later on

src/core/logic.py
```diff
@@ -809,6 +802,47 @@ def get_theme_list():
...
+    theme_settings = [
+        {
+            "name": "journal_base_theme",
+            "object": setting_handler.get_setting(
+                "general", "journal_base_theme", journal
+            ),
+            "choices": [[each, each] for each in settings.CORE_THEMES],
+            "themes": non_core_themes,
+        },
+    ]
```
mauromsl, line 846: If we turn the theme settings into `core.Setting` objects, this needs updating. The encapsulation here works in our favour :)

mauromsl (changes requested):
Looking good, just one proposal around settings

mauromsl (approved):

PR #5423, opened 2026-08-04
fix: reject keywords over the model max length with a form error

### LINKED ISSUE #5418: Jounal & Repository: keywords that go over charlimit cause submisison to fail, should get an error message instead
**Describe the bug**
When a submitting author to a Janeway Journal or Repository inputs a single keyword that exceeds the character limit, their submission fully fails and returns a server error. The author is given no feedback about the cause of the error nor how to resolve the issue. 

**Janeway version**
1.8

**To Reproduce**
1. Log into a Janeway Journal or Repository and start a new submissoin. 
2. Fill in the required fields, then input an extremely long keyword, eg ``Keywords:electronic waste (e-waste); transboundary e-waste trade; greenhouse gas emissions; climate change; Basel Convention; extended producer responsibility (EPR); life-cycle assessment (LCA); circular economy.``
Image

3. Note that janeway automatically transforms that string into a single keyword 
Image

4. Submit to proceed to next step; note error which gives no helpful information to the user: 
Image

5. Check reporting/logs and note error:  ``DataError: value too long for type character varying(200)``

**Expected behavior**
Authors are going to make mistakes. If they input a keyword that exceeds the limit, they should be returned to the same submission page with an error message, eg: ``A keyword cannot exceed 200 characters. Please input your keywords one at a time, hitting Enter after each keyword.``

ajrbyers:
Closes #5418.

### DIFF
--- a/src/utils/forms.py
+++ b/src/utils/forms.py
@@ -75,6 +75,21 @@ class KeywordModelForm(ModelForm):
             field = self.fields["keywords"]
             field.initial = ",".join(current_keywords)
 
+    def clean_keywords(self):
+        posted_keywords = self.cleaned_data.get("keywords", "")
+        max_length = submission_models.Keyword._meta.get_field("word").max_length
+        for keyword in posted_keywords.split(","):
+            if len(keyword) > max_length:
+                raise ValidationError(
+                    _(
+                        "A keyword cannot exceed %(max_length)s characters. "
+                        "Please enter keywords one at a time, pressing Enter "
+                        "after each keyword."
+                    ),
+                    params={"max_length": max_length},
+                )
+        return posted_keywords
+
     def save(self, commit=True, *args, **kwargs):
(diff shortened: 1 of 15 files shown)


src/utils/forms.py
```diff
@@ -75,6 +75,21 @@ def __init__(self, *args, **kwargs):
...
+            if len(keyword) > max_length:
+                raise ValidationError(
+                    _(
+                        "A keyword cannot exceed %(max_length)s characters. "
+                        "Please enter keywords one at a time, pressing Enter "
+                        "after each keyword."
+                    ),
+                    params={"max_length": max_length},
+                )
+        return posted_keywords
```
mauromsl, line 96: Is it worth adding something to tagit to have front end validation before the form is submitted?

mauromsl (commented):
optional comment, otherwise good to merge

mauromsl (approved):

PR #5437, opened 2026-08-14
Improve access denial messages for task assignments

### LINKED ISSUE #2970: Permission denied when opening an assignment link
**Describe the bug**
Revierwers, copyeditors, typesetters, and authors sometimes see "Permission denied" when they try to follow a task link sent to them via email. This most often happens when they are already logged into Janeway with the wrong account. The task has been assigned to a different account or email, so their current account does not have the correct permissions.

While I can understand why this happens, the real problem is that the user is given no information to understand why this is happening and how to resolve the problem.

**Janeway version**
v1.4.1

**To Reproduce**
1. Log in to Janeway.
2. Follow a link for a task not assigned to the account you're logged in with, such as a Review Assignment Request link (with one-step reviewer access turned off)
3. See "Permission denied"

**Expected behavior**
Janeway should recognize the problem and provide the user with the information to understand the error and fix it, in a secure way.

**Additional context**
This may be difficult to achieve in a secure way, but it is definitely possible and worth doing given how often this problem occurs.

ajrbyers:
Closes #2970

Adds options for decorators to pass messages to deny_access. Most security decorators now determine a reason for the user being denied access in such a way that they do not leak whether the assignment exists or not when that particular user isnt the owner of the task.

### DIFF
--- a/src/security/decorators.py
+++ b/src/security/decorators.py
@@ -466,17 +466,18 @@ def article_author_required(func):
     @base_check_required
     def wrapper(request, *args, **kwargs):
         article_id = kwargs["article_id"]
-        article = models.Article.get_article(request.journal, "id", article_id)
+        article = get_object_or_404(
+            models.Article,
+            pk=article_id,
+            journal=request.journal,
+        )
 
-        # A missing or cross-journal id is answered like an article the
-        # reader did not write, so ids cannot be enumerated.
-        if not article or not article.user_is_author(request.user):
+        if request.user.is_author(request) and article.user_is_author(request.user):
+            return func(request, *args, **kwargs)
+        elif not article.user_is_author(request.user):
             deny_access(request, article_not_yours(request))
-
-        if not request.user.is_author(request):
+        else:
             deny_access(request, required_roles=["author"])
-
-        return func(request, *args, **kwargs)
 
     return wrapper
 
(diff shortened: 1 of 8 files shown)


src/security/decorators.py
```diff
@@ -431,10 +468,15 @@ def wrapper(request, *args, **kwargs):
...
-            deny_access(request)
+        # A missing or cross-journal id is answered like an article the
+        # reader did not write, so ids cannot be enumerated.
+        if not article or not article.user_is_author(request.user):
+            deny_access(request, article_not_yours(request))
+
+        if not request.user.is_author(request):
+            deny_access(request, required_roles=["author"])
+
+        return func(request, *args, **kwargs)
```
mauromsl: Should the first clasuse (`if not article`) raise a 404 instead? the other two clauses could still be the same as they were before. Also the comment added there doesn't make a whole lot of sense to me

src/security/decorators.py
```diff
@@ -483,18 +525,31 @@ def copyeditor_for_copyedit_required(func):
...
+        # assignment, another journal's, and a missing one cannot be
+        # told apart.
+        copyedit = copyediting_models.CopyeditAssignment.objects.filter(
+            pk=copyedit_id,
+            copyeditor=request.user,
+            article__journal=request.journal,
+        ).first()
+
+        if not copyedit:
+            deny_access(request, copyedit_not_assigned(request))
```
mauromsl: similarly to the above, we are swapping a 404 for a 403 here, which doesn't seem right.

src/security/decorators.py
```diff
@@ -483,18 +525,31 @@ def copyeditor_for_copyedit_required(func):
...
-            return func(request, *args, **kwargs)
-        else:
-            deny_access(request)
+        # Staff keep access: the views here do not narrow the assignment
+        # to its copyeditor.
+        if request.user.is_staff:
+            if copyediting_models.CopyeditAssignment.objects.filter(
+                pk=copyedit_id,
+            ).exists():
+                return func(request, *args, **kwargs)
```
mauromsl: what does this change aim to achieve?

src/security/decorators.py
```diff
@@ -483,18 +525,31 @@ def copyeditor_for_copyedit_required(func):
...
     def wrapper(request, *args, **kwargs):
         copyedit_id = kwargs["copyedit_id"]
-        copyedit = get_object_or_404(
-            copyediting_models.CopyeditAssignment, pk=copyedit_id
-        )
 
-        if (
-            request.user == copyedit.copyeditor
-            and request.user.is_copyeditor(request)
-            or request.user.is_staff
```
mauromsl, line 493: I believe all we need in this PR is to avoid the shortcircuit on the `AND`
```
            request.user == copyedit.copyeditor
            and (
              request.user.is_copyeditor(request)
              or request.user.is_staff
            )
```

mauromsl (changes requested):
I like the new error messages and will improve useability when facing these 403 screens.

Added a couple minor comments inline regarding the changes to the security checks. In general I prefer the default return on these to be `deny_access` and the conditional logic to be the allow flag, somehow it gives me higher confidence even though all you've done is apply de morgan's law, so they should be equivalent :D

mauromsl (approved):

PR #5443, opened 2026-08-17
fix: #5255 prevent shortcode conflict with reserved system routes

### LINKED ISSUE #5255: Journal codes can conflict with reserved system routes
**Describe the bug**
Where a journal code matches a reserved system url route, the site middleware routes the request to the journal homepage instead of the intended system view. For example, if a journal has the code `cms`, navigating to `press.com/cms/` loads the journal homepage rather than the CMS manager.

This issue is specific to journals using path mode and does not affect domain mode directly.

**Janeway version**
All versions.

**To Reproduce**
1. Create a journal with a code that matches a reserved system url route (e.g. `cms`, `manager`)
2. Navigate to `[press-url]/[code]/`
3. Observe that the journal homepage loads instead of the expected system view

**Suggested fix 1: Blocklist reserved URL prefixes**

Add validation on journal code creation and editing that checks the proposed code against reserved URL prefixes derived from the URL conf. If a conflict is found, the form should reject the code with a clear error message.

**Suggested fix 2: Namespace all back-office URLs**

Move all system routes under a common prefix (e.g. `press.com/*/cms/`) that is unlikely to ever be a valid journal code. This reduces the number of potential conflicts to one.

BARF notes: We are going to implement fix 1 and reconsider how we build site url paths separately

StephDriver:
closes #5255


StephDriver:
Notes:
1. this doesn't only apply to journals, but to repos.
2. on creation journal and repo codes need to be cross checked as if two have the same name it would also cause a conflict
3. other than these codes, the reserved routes would be part of the codebase and so can be a static list with a unittest to prevent drift, rather than needing to be dynamically built each time they  are needed for validation.
4. Baking in accessibility alongside this - WCAG 3.3.1, 3.3.3, 3.3.6 around errors.

StephDriver:
The first commit here sets the respository shortcode to be unique.  I have included a safeguard to warn if non-unique values are found on migration. Ordinarily this would be a breaking change, however in this particular case I cannot see how duplicate values could have gone unnoticed.  They would have caused problems accessing the different repos. So all this is doing is bringing the repository shortcode into line with teh journal shortcode which already has `'is_unique`.

This makes the validation against reserved routes simpler. 

@ajrbyers are you happy with this approach?

StephDriver:
pausing this work as 1.9 testing is more urgent.  This is in a state that someone else could probably pick it up if need be. I think the solution is mostly there, but it needs full testing.

work to do:
- review failing tests.  Some are also failing on master.
- manually test

PR #5448, opened 2026-08-18
Add repository settings for managing preprint identifiers

ajrbyers:
Closes https://github.com/openlibhums/janeway/issues/5450


src/security/decorators.py
```diff
@@ -1575,6 +1575,38 @@ def inner(request, *args, **kwargs):
...
+        elif (
+            request.user.is_production(request)
+            or request.user.is_editor(request)
+            or request.user.is_staff
+        ):
+            return func(request, *args, **kwargs)
+
+        deny_access(request)
+
+    return wrapper
```
mauromsl, line 1610: The decorator doesn't seem to honour its name, it is not checking `repository.identifier_management` and we are not calling `get_object_by_content_type`. Following the nested calls, I don't see where this happens either. Shouldn't the check happen here to be more explicit?

Also, if we are adding a toggle, is it worth adding a similar setting for the journal?

src/repository/tests/test_views.py
```diff
@@ -593,3 +593,116 @@ def test_submit_with_id_redirects_when_draft_has_no_type(self):
...
+    def test_identifiers_view_accessible_when_setting_on(self):
+        self.client.force_login(self.repo_manager)
+        response = self.client.get(
+            reverse(
+                "identifiers",
+                kwargs={"content_type": "preprint", "object_id": self.preprint.pk},
+            ),
+            SERVER_NAME=self.server_name,
+        )
+        self.assertEqual(response.status_code, 200)
```
mauromsl, line 673: I'm not sure if this is working because the user is a repo manager or because of the new setting

mauromsl (changes requested):
A change suggestion on the security decorator to either fix the behaviour of the new setting or make it more explicit

mauromsl (approved):

PR #5473, opened 2026-08-25
fix articles page sort, filter, search

### LINKED ISSUE #5464: Articles "sort" not working
1.9 Articles page, OLH theme.

"sort articles by" does nothing.

Image

Works in Material, Clean and Clarity. only an OLH issue.

### LINKED ISSUE #5471: filter date fields not providing labels to screen readers
1.9 rc 1

Filter date fields on article page have labels but these are not being read aloud by voice over when testing with screen reader. 

Affects all but Material, which has a modal popup.

Image

StephDriver:
closes #5464
closes #5471

### DIFF
--- a/src/themes/clarity/templates/elements/journal/article_list_filters.html
+++ b/src/themes/clarity/templates/elements/journal/article_list_filters.html
@@ -1,6 +1,7 @@
 {% load classname %}
 {% load bootstrap4 %}
 {% load static %}
+{% load field_aria %}
 
 {% if facet_form.fields %}
     <div class="card">
@@ -8,7 +9,18 @@
             <section aria-labelledby="filter-title">
                 <h2 id="filter-title">{% trans 'Filter' %}</h2>
                 <form method="GET" id={{ facet_form.id }}>
-                    {% bootstrap_form facet_form %}
+                    {% for field in facet_form.visible_fields %}
+                        {% if field.field.widget.input_type == "date" %}
+                            {% with described_by_id=field.auto_id|add:"-label" %}
+                                <div class="form-group">
+                                    <label id="{{ described_by_id }}" for="{{ field.auto_id }}">{{ field.label }}</label>
+                                    {% field_aria field describedby=described_by_id css_class="form-control" %}
+                                </div>
+                            {% endwith %}
+                        {% else %}
+                            {% bootstrap_field field %}
+                        {% endif %}
+                    {% endfor %}
                     <button type="button" class="btn btn-secondary" id="select-all-btn">{% trans 'Select all' %}</button>
                     <button type="button" class="btn btn-secondary" id="clear-all-btn">{% trans 'Clear all' %}</button>
                     <hr aria-hidden="true" />
--- a/src/themes/clarity/templates/elements/journal/article_list_filters.html
+++ b/src/themes/clarity/templates/elements/journal/article_list_filters.html
@@ -12,7 +12,7 @@
                     <button type="button" class="btn btn-secondary" id="select-all-btn">{% trans 'Select all' %}</button>
                     <button type="button" class="btn btn-secondary" id="clear-all-btn">{% trans 'Clear all' %}</button>
                     <hr aria-hidden="true" />
-                    <button action="" class="btn btn-primary" type="submit">{% trans 'Apply' %}</button>
+                    <button action="" class="btn btn-primary" type="submit" aria-label="{% trans 'Apply filter' %}">{% trans 'Filter' %}</button>
                 </form>
                 <script src="{% static 'common/js/form-toggle-all.js' %}"></script>
                 <script>
--- a/src/themes/clean/templates/elements/journal/article_list_filters.html
+++ b/src/themes/clean/templates/elements/journal/article_list_filters.html
@@ -1,6 +1,7 @@
 {% load classname %}
 {% load bootstrap4 %}
 {% load static %}
+{% load field_aria %}
 
 {% if facet_form.fields %}
     <div class="card">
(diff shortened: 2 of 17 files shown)


StephDriver:
These changes are all about the `articles` page.  
- fix a bug in OLH that meant the sort wasn't working. 
- improve accessibility so that page is compatible with keyboard and screen reader
- improve accessibility (readability) by changing button labels to `sort`, `filter` and `search` instead of a mix of apply and those terms. Nice and clear which one applies which.
- changes css so that the `clear all` and `apply all` buttons in the filter are quieter than the `filter` itself. 
- tweak css so that the elements within the `sort` block are neatly aligned (as previously in Clarity, but not great in other themes). 

Added two template tags
- one is for including aria in form fields - this was needed to make the date fields screen reader compatible, but is likely to be useful elsewhere. A template tag is needed because you need to pass in the relevant ID to the relevant field, and hte id may only be known through loops of the calling template, i.e. not static.
- other is for adding ids that aren't duplicates to elements that otherwise would be duplicated due to the way they are generated (e.g. in a loop).   IDs being needed for `describedby` and `labelledby` but also needing to be unique on the page for valid HTML.

src/themes/clean/templates/elements/journal/article_list_filters.html
```diff
@@ -1,18 +1,30 @@
...
+                            {% endwith %}
+                        {% else %}
+                            {% bootstrap_field field %}
+                        {% endif %}
+                    {% endfor %}
                     <button type="button" class="btn btn-secondary" id="select-all-btn">{% trans 'Select all' %}</button>
                     <button type="button" class="btn btn-secondary" id="clear-all-btn">{% trans 'Clear all' %}</button>
                     <hr aria-hidden="true" />
-                    <button action="" class="btn btn-primary" type="submit">{% trans 'Apply' %}</button>
+                    <button action="" class="btn btn-primary" type="submit" aria-label="{% trans 'Apply filter' %}">{% trans 'Filter' %}</button>
```
joemull: Couldn't we just change the label to "Apply filter" if you think that's clearer? Keep it simple

src/themes/material/assets/mat.css
```diff
@@ -1694,3 +1700,33 @@ nav .a11y-toggle-host .a11y-toggle-indicator > .fa {
...
+  background-color: transparent;
+  color: #343434;
+  text-transform: uppercase;
+  cursor: pointer;
+}
+
+.btn-outline:hover,
+.btn-outline:focus {
+  background-color: rgba(0, 0, 0, 0.05);
+}
```
joemull, line 1732: I thought we weren't changing this theme any more? Visual changes to the Material theme will interfere with people's custom styling (they often use Material for custom styling) and increase the volume of support queries during the 1.x LTS term

src/themes/clarity/templates/elements/journal/article_list_filters.html
```diff
@@ -1,18 +1,30 @@
...
             <section aria-labelledby="filter-title">
                 <h2 id="filter-title">{% trans 'Filter' %}</h2>
                 <form method="GET" id={{ facet_form.id }}>
-                    {% bootstrap_form facet_form %}
+                    {% for field in facet_form.visible_fields %}
+                        {% if field.field.widget.input_type == "date" %}
+                            {% with described_by_id=field.auto_id|add:"-label" %}
+                                <div class="form-group">
+                                    <label id="{{ described_by_id }}" for="{{ field.auto_id }}">{{ field.label }}</label>
+                                    {% field_aria field describedby=described_by_id css_class="form-control" %}
```
joemull, line 17: Why is aria needed here? Do we know why VO does not use the standard relationship established by `id`  and `for` to create the expected behavior?

joemull (changes requested):
Mostly looks good. Two questions

src/themes/material/assets/mat.css
StephDriver: During the previous 1.9 work I got the focus-visible working across Material. for some reason this part was either missed, or the change lost during a rebase, so this felt like it was simply fixing something already supposed to be complete.

src/themes/clarity/templates/elements/journal/article_list_filters.html
StephDriver: Yes, this came from actual testing with VO. And then some digging into why it wasn't working, trying of different solutions until I found one that did provide the right information.

src/themes/material/assets/mat.css
joemull: Sorry, I'm not talking about focus-visible styles, I'm talking about the secondary quieter button styles. The visual change there will be something that stands out on the page, won't it? Is it strictly necessary? With this change we'd be effectively introducing a new style of button for a theme that we want to reduce the maintenance burden for, and it might have the opposite effect.

src/themes/clarity/templates/elements/journal/article_list_filters.html
joemull: Testing it only establishes *that* it is happening, but have you been able to discover *why* it does that? Asking why is relevant because manually overriding the relationship between elements with ARIA in this way goes against well-accepted guidance and good practice about how to use HTML and ARIA for accessibility. At the least, it is worth uncovering why so that we can be sure we are not working against a VO feature. At most, we can document the reason in a code comment so that when reading this code in the future, we will not think it is a mistake and remove it.

joemull (changes requested):
See comments above (not attached to this review but separate--apologies)

src/themes/material/assets/mat.css
StephDriver: I mostly did that because it looked pretty confusing to me without that change. But I take the point, it's not strictly necessary, more something that was quick while I was there on a page that is unlikely to get more attention any time soon.  Having go the rest of it working,  the two minutes that required seemed a good use of time to reduce confusion (and the other themes were like that, so not unprecidented. It felt like fixing an oversight.

But I take your point. And I certainly don't think it's worth debating so I've remove that change.

src/themes/clarity/templates/elements/journal/article_list_filters.html
StephDriver: This is from my browser history when I was investigating this problem.
- https://stackoverflow.com/questions/62882886/safari-voiceover-issue-when-reading-input-field
- https://stackoverflow.com/questions/35215298/safari-osx-voiceover-not-reading-aria-label-for-input
- https://bugzilla.mozilla.org/show_bug.cgi?id=1611428
- https://dev.to/mfranzke/voiceover-bug-on-ios-safari-blocks-date-time-related-inputs-especially-in-react-4f61
- https://github.com/w3c/aria-practices/issues/669
- https://bugs.webkit.org/show_bug.cgi?id=294649
- https://hassellinclusion.com/blog/input-type-date-ready-for-use/


Collectively from these I concluded:

1. My test results were not unique - the root cause had been a known problem for some time.
2. Aria could not be relied upon for Date fields.
3. a work around to make the field usable with screen readers that didn't rely on the standard "lable for" OR a custom date input field was needed. The workaround was the smaller code-change.

ajrbyers (approved):

Issue #5505, opened 2026-09-06
JATS article titles contain escaped HTML tags rather than safe JATS tags


joemull:
## Problem

Up to now, in JATS, we have not marked article titles as safe, stripped them, or transformed them, just sent them with Django's automatic escaping.

https://github.com/openlibhums/janeway/blob/2c986c0a88581ac9977413e56e20789746956734/src/templates/common/encoding/article_jats_1_2.xml#L36

This means markup inside them may not be properly parsed by downstream systems, and we cannot complete #5495 because we can't mark MathML markup as safe until we transform other HTML in the title into JATS elements.

These are the HTML elements we allow in article titles: 

https://github.com/openlibhums/janeway/blob/2c986c0a88581ac9977413e56e20789746956734/src/utils/const.py#L25-L26

These are the [JATS XML elements](https://jats.nlm.nih.gov/publishing/tag-library/1.2/element/article-title.html) we'd want to support based on the above list: bold, italic, sub, sup

## Possible solutions

- Write an XSL script that does this, like we did for the abstract.

PR #5506, opened 2026-09-06
Support for LaTeX mathematics in titles and abstracts

### LINKED ISSUE #5495: Support for mathematics markup in article titles and abstracts
## Problem

Articles in mathematics typically use tex syntax in their titles and abstracts to communicate topics more succinctly and precisely than they can in English. 

We already parse math markup when it comes through as part of a JATS XML galley, but we do not yet support titles and abstracts. Here is an example from UTMJ, which just joined Janeway:

https://voljournals.utk.edu/utmj/article/pubid/vol2026-iss2-2/

Image

For examples of articles that render with equations in titles and abstracts, look at maths articles in arxiv: https://arxiv.org/html/2608.17267v1

Note: without this functionality, maths and many engineering journals are kept from using Janeway effectively, so it is really a make-or-break feature for some fields.

## Investigation

- [x] Research what downstream metadata recipients expect when it comes to maths
  - Crossref expects [name-spaced MathML in JATS](https://www.crossref.org/documentation/schema-library/markup-guide-metadata-segments/mathml/).
  - Datacite [expects no HTML](https://datacite-metadata-schema.readthedocs.io/en/4.7/appendices/appendix-1/descriptionType/#abstract) in abstracts and [titles](https://datacite-metadata-schema.readthedocs.io/en/4.7/properties/title/#title)

### Accessibility caveats
- The available solutions are not going to provide accessible MathML (i.e. Content MathML or the more recent browser standard MathML Core), only presentational MathML, which is what we already request from JATS vendors and what MathJax uses as a base. For more see https://doi.org/10.48550/arXiv.2605.16562 and https://github.com/roniemartinez/latex2mathml/discussions/289.

## Proposed solution A

- [ ] Upgrade MathJax to version 4 (we currently use version 2 via a CDN)
- [ ] Configure MathJax to pick up tex syntax in titles and abstracts on the public article page
- [ ] Add this functionality to back office pages where titles and abstracts can be edited. This will be key so that users can anticipate how things will render, and they need the metadata to be functional during production. In other words, import and configure MathJax to work in tandem with TinyMCE

## Proposed solution B - MVP

- [ ] Use the Python library [latex2mathml](https://pypi.org/project/latex2mathml/) to convert math in titles and abstracts to mathml before they are displayed on the article page, output in generated JATS stubs (for Crossref and OAI), displayed via the JSON API, or sent to DataCite in JSON. Create a Django template filter / serializer to make this behavior explicit
- [ ] If needed, adjust the MathJax CDN in the article page to do the same thing with inline MathML in titles and abstracts as it already does with the MathML output by XSLT for rendered JATS

joemull:
Closes #5495.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -36,7 +36,7 @@ django-ipware==7.0.1
 more-itertools==10.1.0
 jsmin==3.0.1
 kitchen==1.2.4
-latex2mathml==3.81.0
+latex2mathml==3.78.1
 libsass==0.23.0
 lxml==6.0.0
 markdown
--- a/requirements.txt
+++ b/requirements.txt
@@ -36,6 +36,7 @@ django-ipware==7.0.1
 more-itertools==10.1.0
 jsmin==3.0.1
 kitchen==1.2.4
+latex2mathml==3.81.0
 libsass==0.23.0
 lxml==6.0.0
 markdown
--- a/src/core/templatetags/latex_mathml.py
+++ b/src/core/templatetags/latex_mathml.py
@@ -1,0 +1,107 @@
+import re
+from xml.etree import ElementTree as ET
+
+import latex2mathml.converter
+import latex2mathml.exceptions
+
+from django import template
+
+register = template.Library()
+
+BLOCK_LATEX_RES = [
+    # \[...\]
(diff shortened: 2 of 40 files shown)


joemull:
Blocked by #5505.

PR #5508, opened 2026-09-08
document orcid config troubleshooting

### LINKED ISSUE #5507: update ORCID config to include troubleshooting
> invalid_request: Missing parameter: redirect_uri — "<journal URL>" does not match registered client "<client id>"

This error has been reported a few times. It is indicative of an incorrect URL entered into ORCID during configuration.  

We had some [troubleshooting notes (private link)](https://github.com/openlibhums/devops/commit/08b77a49bd55f81c6bd2a2e946648c101f666b4b), and want to combine these with the public doc on ORCID config as that means only one place to look, and they could be useful for others managing their own installs. 

| Type | Redirect URI |
|------|-------------|
| Press (and everything within it)| `https://<press-domain>/` |
| Individual Journal — domain config | `https://<journal-domain>/` |
| Individual Journal — no press and on path config | `https://<press-domain>/<journal_code>/login/orcid/` |

StephDriver:
closes #5507

### DIFF
--- a/docs/md/orcid.md
+++ b/docs/md/orcid.md
@@ -8,7 +8,13 @@ It can be set up in a few steps.
 
 1. Create an API client on orcid.org.
 
-2. Enter each production domain for your Janeway instance as a redirect URL on orcid.org. If Janeway is set to domain mode, each journal will need its own redirect URI.
+2. Enter each production domain for your Janeway instance as a redirect URI on orcid.org. If Janeway is set to domain mode, each journal will need its own redirect URI.  
+
+| Type | Redirect URI |
+|------|-------------|
+| Press (and everything within it)| `https://<press-domain>/` |
+| Individual Journal — domain config | `https://<journal-domain>/` |
+| Individual Journal — no press and on path config | `https://<press-domain>/<journal_code>/login/orcid/` |
 
 3. Enable ORCID and copy the keys from your client into the Django settings file (`src/core/settings.py`). Replace _`CLIENT_SECRET`_ and _`CLIENT_ID`_ with the values from orcid.org.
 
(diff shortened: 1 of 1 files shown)


docs/md/orcid.md
```diff
@@ -8,7 +8,13 @@ It can be set up in a few steps.
...
 1. Create an API client on orcid.org.
 
-2. Enter each production domain for your Janeway instance as a redirect URL on orcid.org. If Janeway is set to domain mode, each journal will need its own redirect URI.
+2. Enter each production domain for your Janeway instance as a redirect URI on orcid.org. If Janeway is set to domain mode, each journal will need its own redirect URI.  
+
+| Type | Redirect URI |
+|------|-------------|
+| Press (and everything within it)| `https://<press-domain>/` |
+| Individual Journal — domain config | `https://<journal-domain>/` |
+| Individual Journal — no press and on path config | `https://<press-domain>/<journal_code>/login/orcid/` |
```
joemull: Are we sure about this? Can we identify an example?

joemull (changes requested):
Thanks for updating this. It looks good except for the individual journal on path mode redirect URI, which still confuses me and I'd want to see an example to make sure it is correct.

docs/md/orcid.md
StephDriver: @ajrbyers could you confirm table is correct please?  You were my source for the information.

ajrbyers: Both are valid but the best approach is just to use the apex domain for path mode see: 
image

docs/md/orcid.md
```diff
@@ -8,7 +8,13 @@ It can be set up in a few steps.
...
 1. Create an API client on orcid.org.
 
-2. Enter each production domain for your Janeway instance as a redirect URL on orcid.org. If Janeway is set to domain mode, each journal will need its own redirect URI.
+2. Enter each production domain for your Janeway instance as a redirect URI on orcid.org. If Janeway is set to domain mode, each journal will need its own redirect URI.  
+
+| Type | Redirect URI |
+|------|-------------|
+| Press (covers everything within it that is on path config)| `https://<press-domain>/` |
+| Individual Journal — domain config (required regardless of press config) | `https://<journal-domain>/` |
+| Individual Journal — no press and on path config | `https://<press-domain>/<journal_code>/` |
```
joemull, line 17: Based on my understanding of Andy's comment, on line 17, the pattern is just `https://<press-domain>`. 

This means it's exactly the same as the other two lines, in that what you want is the site domain (whether journal or press).

So, can we simplify this section by removing this table? The text in step 2 makes it clear that "If Janeway is set to domain mode, each journal will need its own redirect URI."

docs/md/orcid.md
```diff
@@ -84,3 +90,24 @@ The following steps were written for Ubuntu, and are meant for testing the press
...
+
+### Determine the correct URI
+
+The URI to register depends on whether you are configuring a press or a journal, and for journals, whether it uses domain or path configuration.
+
+**To identify journal configuration type:** open the journal in a browser and look at the URL.
+- Own domain (e.g. `https://myjournal.org/`) → **domain configuration**
+- Press domain with journal code (e.g. `https://press.org/myjournal/`) → **path configuration**
+
+See [Setting up ORCID login](#setting-up-orcid-login) for the format of the redirect URI.
```
joemull, line 108: Can we simplify this as above? That is, you don't need the journal code in path mode.

joemull (changes requested):
This can be simplified further, based on my current understanding. All that matters is that you should just put site domains as redirect URIs. In path mode this is the press only, and in domain mode this also includes each journal domain.

PR #5517, opened 2026-09-15
Enhance discussion system with new features and fixes

ajrbyers:


src/discussion/views.py
```diff
@@ -1,105 +1,34 @@
...
-    """
-    modal = None
-
     if object_type == "article":
-        object_to_get = get_object_or_404(
+        obj = get_object_or_404(
             submission_models.Article,
             pk=object_id,
             journal=request.journal,
         )
```
mauromsl, line 20: Now that the view is just decorated with `@login_required` I think any logged in user would be able to pass in an arbitrary id and snoop in on the threads page for any article/preprint. I see the `threads_list_partial` view does filter the list of threads before passing it to the template render, as does the partial for loading the active thread via HTMX, so at least I don't think we'd be leaking the threads themselves.

src/discussion/partial_views.py
```diff
@@ -0,0 +1,534 @@
...
+        thread=thread,
+        participant=user,
+        added_by=request.user,
+        request=request,
+    )
+    thread.create_system_post(
+        actor=request.user,
+        body=f"{request.user.full_name()} added {user.full_name()} to the discussion",
+    )
+    return HttpResponse(status=204)
```
mauromsl, line 305: the m2m manager is idempotent, so adding the same user twice will have no effect, but the code below isn't, so I think the event would fire again if you add the same user multiple times (e.g. clicking the button multiple times on the browser). This would lead to multiple emails being sent out AFAIC. I think its worth checking if the user is already a participant before this block runs and send a django message like "User is already a participant"

mauromsl (changes requested):
This works surprisingly well, only spotted a couple quirks that might be worth addressing.

ajrbyers:
> This works surprisingly well, only spotted a couple quirks that might be worth addressing.

Thanks for the vote of confidence xD

